PR #96, opened 2015-01-28
@orta => Adds optional delay to stubs

### LINKED ISSUE #95: Delay in stubs
It would be sweet if there were a way to randomly or deterministically add a short delay to stub request returns to make it a little more "real".

ashfurrow:
Fixes #95.

### DIFF
--- a/Demo/DemoTests/MoyaProviderSpec.swift
+++ b/Demo/DemoTests/MoyaProviderSpec.swift
@@ -56,7 +56,29 @@ class MoyaProviderSpec: QuickSpec {
                         expect(endpoint1).to(equal(endpoint2))
                     }
                 })
-                
+
+                it("delays execution when appropriate") {
+                    let closure = { (target: GitHub) -> (Moya.StubbedBehavior) in
+                        return .Delayed(seconds: 2)
+                    }
+
+                    let provider = MoyaProvider(endpointsClosure: endpointsClosure, stubResponses: true, stubBehavior: closure)
+
+                    let startDate = NSDate()
+                    var endDate: NSDate?
+                    let target: GitHub = .Zen
+                    waitUntil(timeout: 3){ done in
+                        provider.request(target, completion: { (data, statusCode, response, error) in
+                            endDate = NSDate()
+                            done()
+                        })
+                    }
+
+                    expect{
+                        return endDate?.timeIntervalSinceDate(startDate)
+                    }.to( beGreaterThanOrEqualTo(NSTimeInterval(2)) )
+                }
+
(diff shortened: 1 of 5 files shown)


ashfurrow:
Went with this implementation (using an optional) instead of a default value (like in the endpoints resolver closure) because the optional behaviour is so trivial, it's replaced with only a single coalescing operator. Thought a default value would be overkill.

orta:
implementation looks good

ashfurrow:
Changed my mind about the "overkill" thing – this is more expressible. Adds tests, updates Quick+Nimble.

ashfurrow:
@orta Free to merge whenevs.

Demo/DemoTests/MoyaProviderSpec.swift
```diff
@@ -56,7 +56,29 @@ class MoyaProviderSpec: QuickSpec {
                         expect(endpoint1).to(equal(endpoint2))
                     }
                 })
-                
+
+                it("delays execution when appropriate") {
```
orta: nice test

ashfurrow: Think I might move to `waitUntil` for all necessarily asynchronous tests, instead of `toEventually`.

Demo/DemoTests/MoyaProviderSpec.swift
```diff
@@ -56,7 +56,29 @@ class MoyaProviderSpec: QuickSpec {
...
+                    let target: GitHub = .Zen
+                    waitUntil(timeout: 3){ done in
+                        provider.request(target, completion: { (data, statusCode, response, error) in
+                            endDate = NSDate()
+                            done()
+                        })
+                    }
+
+                    expect{
+                        return endDate?.timeIntervalSinceDate(startDate)
```
kylef: I'm surprised expect doesn't use @autoclosure

ashfurrow: It has a version that does that, too.

ashfurrow: I just went a bid overboard. Was originally planning a more complex test. Might be worth correcting.

kylef: Ah, I haven't used Quick/Nimble yet

Demo/DemoTests/MoyaProviderSpec.swift
orta: makes sense here, wouldn't want this in apps though :👍

PR #98, opened 2015-02-04
Fixed code samples to reflect the real apis

raphaelmor:
a few '!' were missing and the order of the parameters in Endpoint constructors were wrong

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ might look like.
 ```swift
 private extension String {
     var URLEscapedString: String {
-        return self.stringByAddingPercentEncodingWithAllowedCharacters(NSCharacterSet.URLHostAllowedCharacterSet())!
+        return self.stringByAddingPercentEncodingWithAllowedCharacters(NSCharacterSet.URLHostAllowedCharacterSet())
     }
 }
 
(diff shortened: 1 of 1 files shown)


raphaelmor:
I still have an issue though with :
```swift
provider.request(.Zen, completion: { (data, error) in
            if let data = data {
                // do something with the data
            }
        })
```
It complains about `Missing argument for parameter 'method' in call`

README.md
```diff
@@ -115,11 +115,11 @@ Next, we'll set up the endpoints for use with our API.
 
 ```swift
 public func url(route: MoyaTarget) -> String {
-    return route.baseURL.URLByAppendingPathComponent(route.path).absoluteString
+    return route.baseURL.URLByAppendingPathComponent(route.path).absoluteString!
 }
 
 let endpointsClosure = { (target: GitHub, method: Moya.Method, parameters: [String: AnyObject]) -> Endpoint<GitHub> in
-    return Endpoint<GitHub>(URL: url(target), method: method, parameters: parameters, sampleResponse: .Success(200, target.sampleData))
+    return Endpoint<GitHub>(URL: url(target),sampleResponse: .Success(200, target.sampleData), method: method, parameters: parameters )
```
ashfurrow: Very strange! The `Endpoint` initializer [has default values](https://github.com/ashfurrow/Moya/blob/f9ccd82cfe6135faeaa0f35247e00524beff4873/Endpoint.swift#L28) for those two parameters – was this not working for you?

raphaelmor: for this particular line in the commit it is just an issue with the order of the parameters

ashfurrow:
That's strange, it shouldn't complain at all. If you look in the unit tests, you'll find code like the following that works fine: 

```swift
let target: GitHub = .Zen
provider.request(target, completion: { (data, statusCode, response, error) in
    if let data = data {
        // ...
    }
})
```

It's very strange that you're seeing this error, and I'd really appreciate your help in tracking down what's going on. The first thing I'd like you to do is make sure that you're on the latest version of Moya (since we've added a few parameters to the completion handler – something also not added to the readme yet oops). Next, if you're still seeing the error, could I ask you to upload a sample project somewhere? One that demonstrates the error – that way we can troubleshoot what's going on and get a fix for everyone. 

Thanks again for bringing this to my attention, and I'm sorry I couldn't answer earlier – I was at a conference all weekend. 

Thanks! :beers:

ashfurrow:
Sorry this took so long – thanks again for your help! :bamboo:

PR #100, opened 2015-02-10
[WIP] Adds 'Lazy' type, so sampleData can be requested later

colinta:
I'm having trouble getting the specs to run... also, if you think this PR is worthwhile, I would like to add some specs around this specific feature.  Let me know what you think!

Example:

```swift
let endpointsClosure = { (target: GitHub, method: Moya.Method, parameters: [String: AnyObject]) -> Endpoint<GitHub> in
    return Endpoint<GitHub>(URL: url(target), method: method, parameters: parameters,
        sampleResponse: .Lazy({ return .Success(200, target.sampleDataFromSomeExpensiveOperation) })
    )
}
```

### DIFF
--- a/Endpoint.swift
+++ b/Endpoint.swift
@@ -11,17 +11,8 @@ import Alamofire
 
 /// Used for stubbing responses.
 public enum EndpointSampleResponse {
-    case Success(Int, NSData)
-    case Error(Int?, NSError?, NSData?)
-    case Lazy(()-> EndpointSampleResponse)
-
-    func evaluate() -> EndpointSampleResponse {
-        switch self {
-        case Success, Error: return self
-        case Lazy(let lazy):
-            return lazy().evaluate()
-        }
-    }
+    case Success(@autoclosure () -> (Int, NSData))
+    case Error(@autoclosure () -> (Int?, NSError?, NSData?))
 }
 
 
--- a/Endpoint.swift
+++ b/Endpoint.swift
@@ -11,8 +11,8 @@ import Alamofire
 
 /// Used for stubbing responses.
 public enum EndpointSampleResponse {
-    case Success(@autoclosure () -> (Int, NSData))
-    case Error(@autoclosure () -> (Int?, NSError?, NSData?))
+    case Success(@autoclosure () -> Int, @autoclosure () -> NSData)
+    case Error(@autoclosure () -> Int?, @autoclosure () -> NSError?, @autoclosure () -> NSData?)
 }
 
 
--- a/Endpoint.swift
+++ b/Endpoint.swift
@@ -11,8 +11,17 @@ import Alamofire
 
 /// Used for stubbing responses.
 public enum EndpointSampleResponse {
-    case Success(@autoclosure () -> Int, @autoclosure () -> NSData)
-    case Error(@autoclosure () -> Int?, @autoclosure () -> NSError?, @autoclosure () -> NSData?)
+    case Success(Int, NSData)
+    case Error(Int?, NSError?, NSData?)
+    case Closure(@autoclosure () -> EndpointSampleResponse)
+
+    func evaluate() -> EndpointSampleResponse {
+        switch self {
+        case Success, Error: return self
+        case Closure(let closure):
+            return closure().evaluate()
+        }
+    }
 }
 
 
--- a/Endpoint.swift
+++ b/Endpoint.swift
@@ -13,7 +13,17 @@ import Alamofire
 public enum EndpointSampleResponse {
     case Success(Int, NSData)
     case Error(Int?, NSError?, NSData?)
+    case Lazy(()-> EndpointSampleResponse)
+
+    func evaluate() -> EndpointSampleResponse {
+        switch self {
+        case Success, Error: return self
+        case Lazy(let lazy):
+            return lazy().evaluate()
+        }
+    }
 }
+
 
 /// Class for reifying a target of the T enum unto a concrete Endpoint
 public class Endpoint<T> {
--- a/Moya.swift
+++ b/Moya.swift
@@ -126,13 +126,19 @@ public class MoyaProvider<T: MoyaTarget> {
             let behavior = stubBehavior(token)
 
             let stub: () -> () = {
-                switch endpoint.sampleResponse.evaluate() {
-                case .Success(let statusCode, let data):
+                var statusCode : Int
+                var data : NSData
+                var optStatusCode : Int?
+                var optData : NSData?
+                var optError : NSError?
+
+                switch endpoint.sampleResponse {
+                case .Success(let successClosure):
+                    (statusCode, data) = successClosure()
                     completion(data: data, statusCode: statusCode, response:nil, error: nil)
-                case .Error(let statusCode, let error, let data):
-                    completion(data: data, statusCode: statusCode, response:nil, error: error)
-                default:
-                    Void()  // evaluate will always return Success or Error, but exhaustive switches and all that...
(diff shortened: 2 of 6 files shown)


colinta:
I took an initial stab at this using `@autocomplete` on `sampleResponse:` (and called in the `if stubbedResponses` code), but that had an unfortunate side effect that `sampleResponse: .Success(..)` was not autodetected to be of type `EndpointSampleResponse` (e.g. it had to be `EndpointSampleResponse.Success`).  I figured backwards compatibility was preferred, and this syntax is explicit, which is a win.

ashfurrow:
@colinta Hey! Sorry I haven't reviewed this yet. I just moved from Amsterdam to New York and some things have fallen between the cracks. I'll get to it ASAP – thanks for contributing!

Endpoint.swift
```diff
@@ -13,8 +13,18 @@ import Alamofire
 public enum EndpointSampleResponse {
     case Success(Int, NSData)
     case Error(Int?, NSError?, NSData?)
+    case Lazy(()-> EndpointSampleResponse)
+
+    func evaluate() -> EndpointSampleResponse {
+        switch self {
+        case Success, Error: return self
```
ashfurrow: Nice – didn't know you could use commas to delimit a list of case statements. I've been using `fallthrough` – this should clean up my code a bunch!

Endpoint.swift
```diff
@@ -13,8 +13,18 @@ import Alamofire
...
+        case Lazy(let lazy):
+            return lazy().evaluate()
```
ashfurrow: I like this, but the variable name didn't make it immediately obvious how this was working. What do you think about renaming `lazy` to `closure` or something similar?

Moya.swift
```diff
@@ -126,11 +126,13 @@ public class MoyaProvider<T: MoyaTarget> {
             let behavior = stubBehavior(token)
 
             let stub: () -> () = {
-                switch endpoint.sampleResponse {
+                switch endpoint.sampleResponse.evaluate() {
                 case .Success(let statusCode, let data):
                     completion(data: data, statusCode: statusCode, response:nil, error: nil)
                 case .Error(let statusCode, let error, let data):
                     completion(data: data, statusCode: statusCode, response:nil, error: error)
+                default:
```
ashfurrow: Hmm. Maybe we can rethink this a bit. I'm really OK with dropping support for all non-lazy sample responses. Maybe instead of adding to the enum, we can replace the `sampleResponse` property on `Endpoint` with a closure that returns a sample response. Thoughts?

ashfurrow:
Looks great! Tests would be :+1: Let me know if I can help getting them running on your machine.

colinta:
I like the "always lazy" idea.  For that to be backwards compatible we'd need to use an autoclosure, right?

```swift
public init(URL: String, sampleResponse: EndpointSampleResponse, method: ...
  =>
public init(URL: String, sampleResponse: @autoclosure () -> EndpointSampleResponse, method: ...
```

I attempted this and there were so many repercussions that I went with this more straightforward way instead.  I'll take another stab at it.

I'm down with `s/Lazy/Closure/` if we go that route.

colinta:
Btw when I run `make` or `make test` I get this error:

    ⌦ The sandbox is not in sync with the Podfile.lock. Run 'pod install' or update your CocoaPods installation.

colinta:
Ahh, right, OK I just reimplemented the autoclosure style locally, and here's the error that turned me off to the idea:

```
var endpoint = Endpoint<MyAPI>(URL: url(target),
  sampleResponse: .Success(200, target.sampleData),
  method: method,
  parameters: parameters)
🚫 '(@autoclosure () -> EndpointSampleResponse).Type' does not have a member named 'Success'
```

Right, so since this method expects a closure, it wouldn't support `.Success`.  This would be a breaking change.

If `(()->EndpointSampleResponse).Type` could be extended to include `.Success` and `.Failure` then we'd have a way of having this feature *with* backwards compatibility.

colinta:
(O_O)

```swift
public enum EndpointSampleResponse {
    case Success(@autoclosure () -> (Int, NSData))
    case Error(@autoclosure () -> (Int?, NSError?, NSData?))
}
```

colinta:
This is still not ideal.  With this autoclosure trick, the `.Success / .Failure` values now look like this - they have to be wrapped in a closure:

`.Success((200, target.sampleData))`

ashfurrow:
Hmm. To be honest, I'm *totally* cool with breaking changes, so don't worry about it. We're versioning the library so people can update if/when they want to. 

I don't know if `@autoclosure` is going to work going forward. Swift 1.2 changes the semantics surrounding this construct such that these types of closures cant be stored for later – they must be executed immediately (as far as I understand it). I think moving to a pure closure would be OK.

ashfurrow:
Oh, and `bundle exec pod install` or just `pod install` should fix the error with testing.

colinta:
Is there a `Podfile` I'm missing?

    > bundle install --path vendor
    ...
    Your bundle is complete!
    It was installed into ./vendor
    > bundle exec pod install
    [!] No `Podfile' found in the project directory.

ashfurrow:
What folder are you running it from?

colinta:
03:39 PM ~/Code/Swift/Moya ‹ master ›
    > ls
    CHANGELOG.md             Gemfile                  Makefile                 Moya.swift               docs
    Demo                     Gemfile.lock             Moya+ReactiveCocoa.swift RACSignal+Moya.swift     vendor
    Endpoint.swift           LICENSE                  Moya.podspec             README.md                web

colinta:
(root project folder, same as Makefile)

ashfurrow:
Try it in the `Demo` directory, where the Podfile is.

colinta:
Success!  I'll work on the specs.

Regardless, I'm unsure what style you like the most at this point:

1. new type, `.Closure()`, which returns an `EndpointSampleResponse `
2. autoclosure in `init`, which requires changing `.Success => EndpointSampleResponse.Success` because the type can't be detected
3. autoclosure in the enum initializer, which is backwards compatible but might not work in 1.2

I still like option 1 (the original patch) because if `@autoclosure` is a moving target, I don't feel much better about using it in `Endpoint.init` than using it in the enums.

Sounds like you're leaning towards option 2 - it just requires the sample response type to have the explicit `EndpointSampleResponse.` prefix.  Not *as* pretty, but not cumbersome.  And :+1: on the simple use of `@autoclosure` there.

ashfurrow:
I think the first option, adding a new case to the enum, would be awesome. Avoiding `@autoclosure` seems to be a good idea for now. Let me know if I can help!

colinta:
OK, can do - and yeah, I agree, I think it's the most straightforward.
I'll push up that version soon with specs.

Are you hanging out with Eloy these days?  Tell him Colin says hi! :-D

ashfurrow:
@alloy? Only on Slack :) I've left Amsterdam, sadly.

alloy:
👋😆

colinta:
@ashfurrow OK, I was working on the specs, and I *accidentally* (i swear) did this:

`.Closure(.Success(200, ...`

But then I was like, WAIT a second... `case Closure(@autoclosure () -> EndpointSampleResponse)` would make this a very easy opt-in feature!  It supports the syntax I accidentally typed, and as far as I can tell, this syntax will still be supported in 1.2.

I thought it looked great, so I'm pushing it up for review here.  I can still revert to the non-@autoclosure code, e.g. `Closure({ return .Success(200, ...`

ashfurrow:
Looks awesome!

PR #111, opened 2015-04-17
Show network activity indicator when the request is performed

avgx:


ashfurrow:
This is a great idea! I think I'd prefer to have a closure as a parameter to the MoyaProvider initializer, or a variable property, that is called to begin/end. That way developers using something like [this](https://cocoapods.org/pods/BOTNetworkActivityIndicator) can provider their own implementations. Thoughts?

avgx:
As for me closure will be great

ashfurrow:
Hey @avgx! Is there anything I can do to help on this?

avgx:
failed to add closure due to poor swift experience.
added property.
used with this code:
```
class MyReactiveMoyaProvider<T where T: MoyaTarget> : ReactiveMoyaProvider<T> {

    override init(endpointsClosure: MoyaEndpointsClosure, endpointResolver: MoyaEndpointResolution = MoyaProvider.DefaultEnpointResolution(), stubResponses: Bool  = false, stubBehavior: MoyaStubbedBehavior = MoyaProvider.DefaultStubBehavior) {
        super.init(endpointsClosure: endpointsClosure, endpointResolver: endpointResolver, stubResponses: stubResponses, stubBehavior: stubBehavior)
    }

    override var hasNetworkActivity: Bool {
        get {
            return self.hasNetworkActivity
        }
        set(newValue) {
            println("MyReactiveMoyaProvider.hasNetworkActivity \(newValue)")
            UIApplication.sharedApplication().networkActivityIndicatorVisible = newValue
        }
    }
}
```

ashfurrow:
Cool, OK. The PR isn't _quite_ accurate, but I like the idea. Basically, you could have two network calls: A and B. A is slow, but B is fast. First we start A, then start B. B finishes quickly, but A is still running. However, since B finished, we set our `hasNetworkActivity` to `false`. 

I'm really interested in this – I'll work on it today and let you know.

ashfurrow:
So I've got my idea up here: https://github.com/ashfurrow/Moya/pull/112 It still needs tests, but does that make sense?

ashfurrow:
Cool, I'm going to close this @avgx – if I can help in any other way, or if the solution in #112 isn't sufficient for your needs, please let me know!

avgx:
Great! That is almost exactly what I wish to express, but failed to make it compilable!

PR #112, opened 2015-04-22
@orta => Adds Network Activity Thingies

ashfurrow:
So we want users to be able to hide/show the network activity indicator without actually doing that work (in case they're [this](https://cocoapods.org/pods/BOTNetworkActivityIndicator) or something similar to manage that for them). 

This adds an optional closure parameter to the provider class initializer that is passed either `.Began` or `.Ended` whenever a network call begins or ends. 

Also: 

- Splits up tests.
- Removes separate integration test target.
- Replaces integration tests (which actually would hit the network) with OHHTTPStubs. Now we're testing both the stubbing _and_ live behaviour of Moya.


orta:
Looks good.

orta:
Mergable on green. And we should announce on @ArtsyOpenSource

PR #117, opened 2015-05-13
Adding iOS dynamic framework with project

MichaelMcGuire:
This pull request does the following:
* Adds a new Moya project at the root with an iOS Dynamic Framework target
    * The project/target are only used for building with Carthage.  The tests and demo are not included.
* Adds a `Cartfile` and `Cartfile.resolved` pointing to ReactiveCocoa and Alamofire as dependencies.
* Updates the `.gitignore` to ignore the `Carthage\Checkouts` and `Carthage\Builds` directory.  

I spent a considerable amount of time trying to add the dynamic framework target to the Demo project, to avoid requiring a separate project.  Ultimately, I think it is doable, but would require the use of submodules for dependencies.  I tried going the route of using a Cartfile, but because the Cartfile wasn't in the root directory, Carthage was failing to take note of the dependencies when I tried including Moya via Carthage.

I've tested via my own fork and I was able to add and build Moya via Carthage.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,5 +22,9 @@ DerivedData
 # CocoaPods
 Pods
 
+# Carthage
+Carthage/Checkouts
+Carthage/Builds
+
 # AppCode
 .idea/
--- a/Cartfile
+++ b/Cartfile
@@ -1,0 +1,2 @@
+github "ReactiveCocoa/ReactiveCocoa" "v3.0-beta.4"
+github "Alamofire/Alamofire" ~> 1.2.2
--- a/Cartfile.resolved
+++ b/Cartfile.resolved
@@ -1,0 +1,4 @@
+github "Alamofire/Alamofire" "1.2.2"
+github "robrix/Box" "1.2.2"
+github "antitypical/Result" "0.4.2"
+github "ReactiveCocoa/ReactiveCocoa" "1c8b18751c4af1f3c6003077ff57bed80e184c3a"
--- a/Moya.xcodeproj/project.pbxproj
+++ b/Moya.xcodeproj/project.pbxproj
@@ -1,0 +1,331 @@
+// !$*UTF8*$!
+{
+	archiveVersion = 1;
+	classes = {
+	};
+	objectVersion = 46;
+	objects = {
+
+/* Begin PBXBuildFile section */
+		BF7424E21B0401C70086929D /* Moya.h in Headers */ = {isa = PBXBuildFile; fileRef = BF7424E11B0401C70086929D /* Moya.h */; settings = {ATTRIBUTES = (Public, ); }; };
+		BF7424FC1B04025C0086929D /* Endpoint.swift in Sources */ = {isa = PBXBuildFile; fileRef = BF7424F81B04025C0086929D /* Endpoint.swift */; };
+		BF7424FD1B04025C0086929D /* Moya.swift in Sources */ = {isa = PBXBuildFile; fileRef = BF7424F91B04025C0086929D /* Moya.swift */; };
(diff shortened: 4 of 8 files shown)


ashfurrow:
Cool, thanks! 



Pretty sure this is failing on travis because of Swift 1.2. However, I would be more comfortable if we could add a command to the Makefile like `make test_carthage_build` or something to ensure that we can test Carthage's integration process and everything in the future. Does that sound OK?

orta:
Could switch to CircleCI

ashfurrow:
Also a good idea.

MichaelMcGuire:
Absolutely.  Not familiar with Travis or Circle, but I'll see about getting the Travis build functioning :+1:

ashfurrow:
Travis build won't work with Swift 1.2 because they're still on Xcode 6.2 :cry: 

The more and more I think about it, the more I'm not really comfortable with Carthage files in the main repo. I'm totally cool with people using Moya and Carthage, but I'm not personally going to use it and I'm afraid of what happens when we go to implement new Moya features. Contributors would have to make sure that Moya works in CocoaPods _and_ Carthage, which is doubling deployment effort for this repo :confused: In the worst case, we say we have Carthage support but it doesn't work.

I'm open to suggestions – if you comfortable with it, I'd love to let you handle Carthage and have the README contain instructions to use the Cartfile on your fork. Then you could pull in updates to your fork.

That's a lot to ask, though.

MichaelMcGuire:
After typing up my comment and doing a bit of research, I see now what you were getting at with Travis.

As for a separate fork, I wouldn't mind doing that for now.  A better solution is bound to come along to support both systems, but until then I wouldn't mind maintaining that.  If a better solution to including a dynamic framework target into the Demo project without using Carthage files for dependency management could be found, might you be more open?  If so, I can continue down that path when I get some time.

ashfurrow:
Hmm. Having the projects together would be nicer from a "fewer files in the repo" perspective, but it's really that I'm afraid that having Carthage supports is a promise to developers using Moya, and one that I'm not confident I can keep. 

But you're right – this isn't a problem that only we have. I'm hoping that the community can solve this problem at a higher level so library authors aren't in this awkward spot. Maybe [this](https://github.com/neonichu/Rome) could help? 

For now, I would be much obliged if you were to maintain a fork with Carthage and submit a new PR updating the README to point Carthage users to use your repo :bow: :heart:

zdavison:
Supporting Carthage is really minimal work, and considering its extreme simplicity in comparison to CocoaPods is something we'll probably see more of in future. Plus, having a dynamic framework target in your project is good practice now that they are available (and now that Cocoapods deployments use them also).

ashfurrow:
From what people working on this repo and others have experienced, adding Carthage is not as minimal as you might expect. My issue is the problem of misleading developers by saying "we support Carthage", when I don't use it and won't know if Moya supports it correctly.

MichaelMcGuire:
I didn't want to get into a battle over it :peace:.  At this point, I agree that if you're using the CocoaPods template for creating a pod with example/demo project, it is a pain to get Carthage support working if you have dependent libraries.  I am planning on submitting a separate PR as requested, just have gotten a bit busy! :smile_cat:

ashfurrow:
Not at all! We're all after the same goal: making our own jobs easier. Take your time – whenever you have a chance.

MichaelMcGuire:
@ashfurrow That was definitely not directed at you!  I thought language like "really minimal work" and "extreme simplicity in comparison" was getting into a CocoaPods vs Carthage battle :smile:, but perhaps I am just too sensitive.

neonichu:
Should be really easy with the approach from CocoaPods/pod-template#106 and ensuring that it continues to work can be done via CI. I'll make a PR once everything is in place inside both Carthage and CP.

PR #120, opened 2015-05-14
Adding cancelable behavior to MoyaProvider requests

### LINKED ISSUE #119: Cancelable Network requests
The current API does not provide a way to cancel a network request.  Alamofire's `Request` object exposes `NSURLSessionTasks`'s `cancel` method, but `MoyaProvider` does not keep a reference to the `Request` object to later cancel, nor does it return a token to the client that could be passed later for canceling.

Since supporting this would require a change to the API, I was wondering what your thoughts were on supporting this?  I wouldn't mind helping out if you wanted to have some discussion on how this would look.

MichaelMcGuire:
Resolves #119 

* Functionality will only work for delayed stubbed responses and actual network requests
* For stubbed responses, returns an NSError similar to what you get from a real network cancelation

Don't be shy about any feedback. Nitpick away! :bow:

### DIFF
--- a/Demo/DemoTests/MoyaProviderIntegrationTests.swift
+++ b/Demo/DemoTests/MoyaProviderIntegrationTests.swift
@@ -72,7 +72,7 @@ class MoyaProviderIntegrationTests: QuickSpec {
                         expect{message}.toEventually( equal(userMessage) )
                     }
                     
-                    it("returns an error when canceled") {
+                    it("returns an error when cancelled") {
                         var receivedError: NSError?
                         
                         let target: GitHub = .UserProfile("ashfurrow")
@@ -81,7 +81,7 @@ class MoyaProviderIntegrationTests: QuickSpec {
                         }
                         token.cancel()
                         
-                        expect(receivedError).toEventuallyNot(beNil())
+                        expect(receivedError).toEventuallyNot( beNil() )
                     }
                 }
 
--- a/Demo/DemoTests/MoyaProviderIntegrationTests.swift
+++ b/Demo/DemoTests/MoyaProviderIntegrationTests.swift
@@ -71,6 +71,18 @@ class MoyaProviderIntegrationTests: QuickSpec {
                         
                         expect{message}.toEventually( equal(userMessage) )
                     }
+                    
+                    it("returns an error when canceled") {
+                        var receivedError: NSError?
+                        
+                        let target: GitHub = .UserProfile("ashfurrow")
+                        let token = provider.request(target) { (data, statusCode, response, error) in
+                            receivedError = error
+                        }
+                        token.cancel()
+                        
+                        expect(receivedError).toEventuallyNot(beNil())
+                    }
                 }
 
                 describe("a provider with network activity closures") {
--- a/Demo/DemoTests/MoyaProviderSpec.swift
+++ b/Demo/DemoTests/MoyaProviderSpec.swift
@@ -256,6 +256,31 @@ class MoyaProviderSpec: QuickSpec {
                         expect(provider.inflightRequests.count).to(equal(0))
                     }
                 })
+                
+                describe("a reactive provider with delayed stubs") {
+                    var provider: ReactiveMoyaProvider<GitHub>!
+                    beforeEach {
+                        let closure = { (target: GitHub) -> (Moya.StubbedBehavior) in
+                            return .Delayed(seconds: 2)
+                        }
+                        
+                        provider = ReactiveMoyaProvider(endpointsClosure: endpointsClosure, stubResponses: true, stubBehavior: closure)
+                    }
+                    
+                    it("returns canceled error when delayed execution is canceled") {
+                        var receivedError: NSError?
+                        let target: GitHub = .Zen
+                        waitUntil(timeout: 3) { done in
+                            let disposable = provider.request(target).subscribeError { (error) -> Void in
+                                receivedError = error
+                                done()
+                            }
+                            disposable.dispose()
+                        }
+                        
+                        expect(receivedError).toNot(beNil())
+                    }
+                }
             }
             
             describe("with stubbed errors") {
--- a/Demo/DemoTests/MoyaProviderSpec.swift
+++ b/Demo/DemoTests/MoyaProviderSpec.swift
@@ -56,12 +56,12 @@ class MoyaProviderSpec: QuickSpec {
                         expect(endpoint1).to(equal(endpoint2))
                     }
                     
-                    it("returns an cancelable object when a request is made") {
+                    it("returns a cancellable object when a request is made") {
                         let target: GitHub = .UserProfile("ashfurrow")
                         
-                        let cancelable: Cancelable = provider.request(target) { (_, _, _, _) in }
+                        let cancellable: Cancellable = provider.request(target) { (_, _, _, _) in }
                         
-                        expect(cancelable).toNot(beNil())
+                        expect(cancellable).toNot(beNil())
 
                     }
                 })
@@ -138,7 +138,7 @@ class MoyaProviderSpec: QuickSpec {
                     }.to( beGreaterThanOrEqualTo(NSTimeInterval(2)) )
                 }
                 
-                it("returns canceled error when delayed execution is canceled") {
+                it("returns cancelled error when delayed execution is cancelled") {
                     let closure = { (target: GitHub) -> (Moya.StubbedBehavior) in
                         return .Delayed(seconds: 2)
                     }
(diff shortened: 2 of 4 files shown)


Demo/DemoTests/MoyaProviderIntegrationTests.swift
```diff
@@ -71,6 +71,18 @@ class MoyaProviderIntegrationTests: QuickSpec {
                         
                         expect{message}.toEventually( equal(userMessage) )
                     }
+                    
+                    it("returns an error when canceled") {
```
ashfurrow: *cancelled

Demo/DemoTests/MoyaProviderSpec.swift
```diff
@@ -55,6 +55,15 @@ class MoyaProviderSpec: QuickSpec {
                         let endpoint2 = provider.endpoint(target, method: Moya.DefaultMethod(), parameters: Moya.DefaultParameters())
                         expect(endpoint1).to(equal(endpoint2))
                     }
+                    
+                    it("returns an cancelable object when a request is made") {
```
ashfurrow: *cancellable

Demo/DemoTests/MoyaProviderSpec.swift
```diff
@@ -55,6 +55,15 @@ class MoyaProviderSpec: QuickSpec {
...
+                        let target: GitHub = .UserProfile("ashfurrow")
+                        
+                        let cancelable: Cancelable = provider.request(target) { (_, _, _, _) in }
+                        
+                        expect(cancelable).toNot(beNil())
```
ashfurrow: Awesome :+1:

Demo/DemoTests/MoyaProviderSpec.swift
```diff
@@ -121,12 +130,33 @@ class MoyaProviderSpec: QuickSpec {
...
                         }
+                        return
                     }
 
                     expect{
                         return endDate?.timeIntervalSinceDate(startDate)
                     }.to( beGreaterThanOrEqualTo(NSTimeInterval(2)) )
                 }
+                
+                it("returns canceled error when delayed execution is canceled") {
```
ashfurrow: *cancelled

Demo/DemoTests/MoyaProviderSpec.swift
```diff
@@ -121,12 +130,33 @@ class MoyaProviderSpec: QuickSpec {
...
+                    let closure = { (target: GitHub) -> (Moya.StubbedBehavior) in
+                        return .Delayed(seconds: 2)
+                    }
+                    
+                    let provider = MoyaProvider(endpointsClosure: endpointsClosure, stubResponses: true, stubBehavior: closure)
+                    
+                    var receivedError: NSError?
+                    let target: GitHub = .Zen
+                    waitUntil(timeout: 3) { done in
```
ashfurrow: Since we're immediately cancelling, is the `timeout: 3` necessary? I think it uses one second as a default, which would be cool to stick with unless there's a reason not to.

Demo/DemoTests/MoyaProviderSpec.swift
```diff
@@ -226,6 +256,31 @@ class MoyaProviderSpec: QuickSpec {
...
+                    var provider: ReactiveMoyaProvider<GitHub>!
+                    beforeEach {
+                        let closure = { (target: GitHub) -> (Moya.StubbedBehavior) in
+                            return .Delayed(seconds: 2)
+                        }
+                        
+                        provider = ReactiveMoyaProvider(endpointsClosure: endpointsClosure, stubResponses: true, stubBehavior: closure)
+                    }
+                    
+                    it("returns canceled error when delayed execution is canceled") {
```
ashfurrow: Same

Demo/DemoTests/MoyaProviderSpec.swift
```diff
@@ -226,6 +256,31 @@ class MoyaProviderSpec: QuickSpec {
...
+                        var receivedError: NSError?
+                        let target: GitHub = .Zen
+                        waitUntil(timeout: 3) { done in
```
ashfurrow: Same too :)

Demo/DemoTests/MoyaProviderSpec.swift
```diff
@@ -226,6 +256,31 @@ class MoyaProviderSpec: QuickSpec {
...
+                            let disposable = provider.request(target).subscribeError { (error) -> Void in
+                                receivedError = error
+                                done()
+                            }
+                            disposable.dispose()
```
ashfurrow: Perfect.

Moya.swift
```diff
@@ -99,6 +99,25 @@ public protocol MoyaTarget : MoyaPath {
     var sampleData: NSData { get }
 }
 
+/// Protocol to define the opaque type returned from a request
+public protocol Cancelable {
+    func cancel()
+}
+
+/// Internal token that can be used to cancel requests
+struct CancelableToken : Cancelable {
```
ashfurrow: Cancellable I think?

Moya.swift
```diff
@@ -99,6 +99,25 @@ public protocol MoyaTarget : MoyaPath {
...
+    
+    let cancelAction: () -> ()
+    
+    init(cancelAction: () -> ()) {
```
ashfurrow: I think Swift's implicit initializers mean you don't need this at all.

Moya.swift
```diff
@@ -99,6 +99,25 @@ public protocol MoyaTarget : MoyaPath {
...
+        self.cancelAction = cancelAction
+    }
+    
+    func cancel() {
```
ashfurrow: Do we care about cancelling things more than once? Probably not. I'd rather not expose state anyway :)

Demo/DemoTests/MoyaProviderSpec.swift
```diff
@@ -226,6 +256,31 @@ class MoyaProviderSpec: QuickSpec {
...
+                        }
+                        
+                        expect(receivedError).toNot(beNil())
```
ashfurrow: So, this is tricky to test. The disposable will invoke its block when the subscription is ended, which is :+1: However, when you cal `disposable.dispose()`, that means that the `subscribeError` subscription is now disposed and will never be invoked. 

This is really really hard – the request is cancelled after its signal is not subscribed to anymore, which is behaviour _on top_ of the regular `MoyaProvider`. In practical use, no one using the RAC extension wants to know if a request was cancelled; it has only been cancelled because they no longer care about it. 

I think what we should test is that after disposing of the subscription, the `waitUntil` times out and `done()` is never invoked. Thoughts?

ashfurrow:
This looks amazing! :heart_eyes_cat: 

One small nitpicky note: The rest of the tests use spacing around the expectation, like `expect(thing).to( beNil() )` – it's something that @orta got me doing and I think it looks nicer. 

Otherwise, looks awesome. Very nice job with the tests :+1:

Moya.swift
MichaelMcGuire: Not exactly sure what you're referring to here?  Are you thinking of not having the struct at all and instead just returning a closure that could be executed?

MichaelMcGuire:
The single/double L thing made me laugh.  We had an epic battle on a project a few years ago on which spelling was correct.  Eventually, [American pride](http://grammarist.com/spelling/cancel/) won out, even though (as an American) I preferred the two L's.  Just got so used to coding it that way, that I do it naturally.  

The only thing not done was fixing the ReactiveProvider test.  You can see I tried doing what you suggested, which made sense, but waitUntil() fails the test if it times out.  :crying_cat_face: 

I didn't see anything else in Nimble that would allow for an easy way to wait a certain amount of time, but I could lift some of the code from waitUntil() to give us similar functionality without failing the test?  Any other ideas?

MichaelMcGuire:
Oh! Also, I added the spacing around the expectation in the integration tests (since it was the only one without it), but in the `MoyaProviderSpec` the tests (except for one) did not have the spacing.  If you'd like me to just add the spacing on all the tests, let me know.

ashfurrow:
Hey there!

Sorry, missed the update. I'll take a look in a bit – gonna merge the swift-1.2 branch into master and get back to you.

MichaelMcGuire:
Not a problem at all! From Twitter could tell you've been pretty busy.

ashfurrow:
Ha, yeah. Jet lag took me out of commission for the past two days, too. 

Why don't I try to submit a PR to your branch that does the ReactiveProvider tests? Then we can do that separately, then merge this one into `swift-1.2`, then `swift-1.2` into `master` then :tada:

PR #126, opened 2015-05-26
A lot of things

### LINKED ISSUE #60: Method Type should really be a part of the API
The header API for two requests on the same URL can be different when comparing POST vs GET for example.

### LINKED ISSUE #74: Make defaultParams part of MoyaPath protocol
From our use in [Eidolon](https://github.com/artsy/eidolon/), we almost always call `request` with the [`defaultParameters`](https://github.com/artsy/eidolon/blob/29cfd75f3a63ed2d320a87659953f7cd5071ecb4/Kiosk/App/Networking/ArtsyAPI.swift#L40-L114) we've defined on our target, since the target contains all the details necessary for the action, anyway. So why not codify that requirement? 

Thoughts anyone?

### LINKED ISSUE #123: Moya Documentation
Moya documentation either exists in the source code (ew) or in the README (which doesn't scale). A solution to this issue would ideally include:

- Restructuring existing README docs into `/docs` directory.
- Rewrite README to outline:
 - The problem Moya solves.
 - How Moya solves it, at a high-level.
 - Why Moya's solution is a good idea.
 - Installation instructions.
 - A link to the docs and to the new issue URL.
- Document the capabilities of Moya.
- Provide some example use cases of its interesting features.

ashfurrow:
Lots of changes:

- Documentation revamp. Expanded a bunch and added examples. 
- Parameters moved from `request()` functions to the target enums, instead. 
- HTTP methods moved from `request()` functions to target enums, instead. 
- The result of the last two is that there is only one `request()` method in the non-reactive `MoyaProvider` class, and one more in the `ReactiveProvider` class. 
- Related changes to the endpoint-resolving closure.
- `SampleResponse` enum now accepts a closure that returns sample data instead of the data itself.

Overall, I think this is a general simplification of the Moya framework. Lots of things that made sense when I started this didn't turn out being as useful as I had anticipated. This really supports a simple use: setup a target enum and create a provider. No messing around with `Endpoint`s at all anymore, unless you want to.

Once merged, I'll release this as a `1.0.0`, since it has breaking changes (noted in the changelog). 

Fixes #123 (documentation)
Fixes #74 (move parameters to target)
Fixes #60 (moves HTTP method to target)


docs/Providers.md
```diff
@@ -0,0 +1,72 @@
...
+Notice that we don't have to specify the generic type in the `MoyaProvider` 
+initializer anymore, since Swift will infer it from the type of our
+`endpointsClosure`. Neat!
+
+This `endpointsClosure` is about as simple as you can get. Check out the
+[Endpoints](Endpoints.md) documentation for more on _why_ you might want
+to do this.
+
+The next optional initializer parameter is `endpointResolver`, which resolves
+and `Endpoint` to an actual `NSURLrequest`. Again, check out the [Endpoints](Endpoints.md)
```
nanoxd: Spelling
```diff
+an
-and
```

```diff
+NSURLRequest
-NSURLrequest
```

docs/Providers.md
```diff
@@ -0,0 +1,72 @@
...
+to do this.
+
+The next optional initializer parameter is `endpointResolver`, which resolves
+and `Endpoint` to an actual `NSURLrequest`. Again, check out the [Endpoints](Endpoints.md) 
+documentation for how and why you'd do this. 
+
+The third optional parameter, `stubResponses`, is very straightforward: 
+should the provider access the API or should it return the stubbed data? 
+
+Another option is to provide a `stubBehavior` of either `.Immedaite` (the
```
nanoxd:
```diff
+ .Immediate
- .Immedaite
```

docs/Providers.md
```diff
@@ -0,0 +1,72 @@
...
+    }
+}
+```
+
+That's it! The `request()` method returns a `Cancellable`, which has
+only one public function, `cancel()`, which you can use to cancel the
+request. 
+
+Remember, *where* you put your target and the provider, are completely up 
+to you. You can check out [Artsy's implementation](https://github.com/Artsy/eidolon/Kiosk/AppNetworking/ArtsyAPI.swift)
```
nanoxd: URL is invalid. Should be `https://github.com/artsy/eidolon/blob/master/Kiosk/App/Networking/ArtsyAPI.swift`

docs/ReactiveExtensions.md
```diff
@@ -0,0 +1,55 @@
...
+  and errors if unsuccessful.
+- `mapJSON()` tries to map the response data to a JSON object and
+  errors if unsuccessful.
+- `mapString()` tries to map the response data to a string and
+  errors if unsuccessful.
+
+In the error cases, the error's `domain` is `MoyaErrorDomain`. The code
+is one of `MoyaErrorCode`'s `rawValue`s, where appropriate. Wherever
+possible, underlying errors are provided and the original response
+data is included in the `NSError`'s `userInfo' dictionary using the
```
nanoxd: Missing a backtick on `userInfo`

docs/ReactiveExtensions.md
```diff
@@ -0,0 +1,55 @@
+ReactiveCocoa
+=============
+
+Moya providers an optional `ReactiveMoyaProvider` subclass of
```
nanoxd:
```diff
+ Moya providers are an optional
- Moya providers an optional
```

docs/ReactiveExtensions.md
```diff
@@ -0,0 +1,55 @@
...
+`MoyaProvider` that does a few interesting things. Instead of
+calling the `request()` method and providing a callback closure
+to be executed when the request completes. Instead, we use
```
nanoxd: Too many "instead"s, I propose a change to 

>Instead of calling the `request()` method and providing a callback closure to be executed when the request completes, we emit `RACSignal`s.

docs/ReactiveExtensions.md
```diff
@@ -0,0 +1,55 @@
...
+
+1. The signal sends a value, a `MoyaResponse` instance.
+1. The signal completes.
+
+If the request produces an error (typically a NSURLSession error), 
+then it sends an error, instead. The error's `code` is the failing
+request's status code, if any, and the response data, if any.
+
+The `MoyaResponse` class contains a `statusCode`, some `data`,
+an a(n optional) `NSURLResponse`. You can use these values however
```
nanoxd:
```diff
+ and a(n optional)
- an a(n optional)
```

docs/ReactiveExtensions.md
```diff
@@ -0,0 +1,55 @@
...
+you like in `subscribeNext` or `map` calls. 
+
+To make things even awesomer, Moya provides some extensions to 
+`RACSignal` that make detail with `MoyaResponses` really easy.
```
nanoxd: `detail` -> `dealing`

docs/Targets.md
```diff
@@ -0,0 +1,117 @@
+Targets
+=======
+
+You use Moya to define a Swift `enum` that will is your *target*. Then, the rest
```
nanoxd: :confused: Not sure what you mean on this line.

nanoxd:
I added a few doc changes but overall it's a step in the right direction. Good job!

ashfurrow:
Thanks for the feedback! Yeah, my fingers were getting pretty tired >.<

PR #128, opened 2015-05-27
Remove redundancy in circle.yml and Makefile

pietbrauer:
I am not sure if the steps declared in the dependencies section of the `circle.yml` are actually needed, so I removed them and used the `Makefile` (which does kind of the same) instead.


ashfurrow:
Cooooool thanks!

PR #131, opened 2015-05-28
Update ReactiveExtensions.md

alexito4:
*I think it's the wrong wording but maybe I just don't understand it xD*

A minnor wording change.

PS: Kudos for the documentation 👏


ashfurrow:
Both are accurate (think "provider" as in `MoyaProvider` class), but I like your wording better. Thanks!

PR #133, opened 2015-06-06
Add RxSwift provider

### LINKED ISSUE #124: RxSwift Provider
[This](https://github.com/kzaher/RxSwift) looks very cool.

alcarvalho:
This is my initial attempt to add RxSwift support, as mentioned in #124 

Still missing the `inflightRequests` management like in the RAC provider. And the filters as well. I had to factor out the MoyaResponse so it can be a common core for the two reactive frameworks.

Feedback is very much appreciated! :smiley: 

Fixes #124.

### DIFF
--- a/Demo/Podfile.lock
+++ b/Demo/Podfile.lock
@@ -6,8 +6,13 @@ PODS:
   - Moya/Core (1.0.0):
     - Alamofire (~> 1.2.0)
   - Moya/Reactive (1.0.0):
-    - Moya/Core
+    - Moya/ReactiveCore
     - ReactiveCocoa (= 3.0-beta.6)
+  - Moya/ReactiveCore (1.0.0):
+    - Moya/Core
+  - Moya/RxSwift (1.0.0):
+    - Moya/ReactiveCore
+    - RxSwift (= 1.4)
   - Nimble (0.4.2)
   - OHHTTPStubs (4.0.2)
   - Quick (0.3.1)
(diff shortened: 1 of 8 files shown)


alcarvalho:
It seems like an unrelated thing failed. It's likely something with a real connection. All tests are passing on my machine.

MoyaResponse.swift
```diff
@@ -0,0 +1,30 @@
+//
+//  File.swift
```
orta: best outright kill all these comments :+1:

Demo/Podfile.lock
```diff
@@ -6,8 +6,13 @@ PODS:
   - Moya/Core (1.0.0):
     - Alamofire (~> 1.2.0)
   - Moya/Reactive (1.0.0):
-    - Moya/Core
+    - Moya/ReactiveCore
```
ashfurrow: This is very smart.

ashfurrow:
@alcarvalho This is fantastic work, thank you. 

The CI sometimes fails – I haven't figured out why yet. It's related to the delayed stubbing behaviour :\

Anyway, this is great. Could you follow @orta's suggestion and remove those comments from the new files? And could you add a CHANGELOG entry for 1.1? 

Thanks again! :tada:

MoyaResponse.swift
alcarvalho: Sure thing. Even remove all copyright notices?

ashfurrow: Actually, you know what? Just leave them. They're all in the other files, too, so I'd rather make them consistent in the same commit.

Demo/Podfile.lock
alcarvalho: Thank you! :smiley:

MoyaResponse.swift
alcarvalho: Ok! I'll just correct it, then...

ashfurrow: Opened an issue to remove them later: https://github.com/ashfurrow/Moya/issues/134

PR #157, opened 2015-06-24
Update RxSwift and other pods of Demo project

alcarvalho:
I started this to update RxSwift and to try to solve #153, but cocoapods does not behave as expected and it downgrades the version if we add `~>` to the version spec. Any thoughts?

This also fixes the Demo to not update `Quick` beyond 0.3.x. Because we don't support Swift 2.0 yet.


alcarvalho:
Also keeping the Pods folder in the project generates a lot of noise when we update something. :disappointed:

ashfurrow:
Yeah, the `~>` signifier is a bit weird. It basically uses [Semantic Versioning](http://semver.org) to see if a library can be updated. This doesn't work for Quick, etc, because they often have breaking changes before they hit 1.0. 

I think this PR makes sense – relaxing the requirement around RxSwift, which is already post-1.0, should be fine. Could you add an entry to the changelog? Thanks!

alcarvalho:
Do you think it's better to squash these commits? Or maybe separate the Pods folder mess from the rest of the updates?

ashfurrow:
I don't mind either way, but if I had to pick, there'd be one commit with changes to the podfile/podspec/changelog, and one commit with changes to the Pods/ directory.

alcarvalho:
I too prefer that. Done! ;)

alcarvalho:
It's ok. But atom removed some white spaces that I haven't noticed before from the changelog.

CHANGELOG.md
```diff
@@ -3,11 +3,12 @@
...
-- **Breaking change** Redefines types of `DefaultEndpointMapping` and `DefaultEnpointResolution` class functions on `MoyaProvider`. You no longer invoke these functions to return a closure, rather, you reference the functions themselves _as_ closures. 
+- **Breaking change** Redefines types of `DefaultEndpointMapping` and `DefaultEnpointResolution` class functions on `MoyaProvider`. You no longer invoke these functions to return a closure, rather, you reference the functions themselves _as_ closures.
 - **Breaking change** Renames `endpointsClosure` parameter and property of `MoyaProvider` to `endpointClosure`.
 - **Breaking change** Renames `ReactiveMoyaProvider` to `ReactiveCocoaMoyaProvider` for consistency.
-- Fixes problem that the `ReactiveMoyaProvider` initializer would not respect the stubbing behaviour it was passed. 
+- Fixes problem that the `ReactiveMoyaProvider` initializer would not respect the stubbing behaviour it was passed.
 - Adds official Carthage support – [@neonichu](http://github.com/neonichu)
+- Relaxes version dependency on RxSwift
```
ashfurrow: Add your github username! Give yourself credit!

alcarvalho: Ok! :laughing:

alcarvalho:
If you have no further comments, I will merge this PR. :)

ashfurrow:
:+1:

PR #160, opened 2015-07-02
Using the correct functions to send events to Observer.

alcarvalho:
I have implemented this when first learning how to use RxSwift. Now I found that the correct way to send events is by using `sendNext`, `sendError` etc

This PR fixes that.


ashfurrow:
Looks good to me, thanks!

PR #161, opened 2015-07-02
Change ReactiveCocoa subspec to ReactiveCocoa/Core

ejensen:
The ReactiveCocoa subspec only depends on `ReactiveCocoa/Core`. ReactiveCocoa's default subspec is the larger `ReactiveCocoa/UI`

### DIFF
--- a/Moya.podspec
+++ b/Moya.podspec
@@ -32,7 +32,7 @@ Pod::Spec.new do |s|
   s.subspec "ReactiveCocoa" do |ss|
     ss.source_files = "Moya/ReactiveCocoa/*.swift"
     ss.dependency "Moya/ReactiveCore"
-    ss.dependency "ReactiveCocoa", "3.0-beta.6"
+    ss.dependency "ReactiveCocoa/Core", "3.0-beta.6"
   end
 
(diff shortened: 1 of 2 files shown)


ashfurrow:
That's a really great point, thanks!

PR looks good – would you mind adding an entry to the changelog?

ejensen:
It appears the [ReactiveCocoa/Core](https://github.com/CocoaPods/Specs/blob/master/Specs/ReactiveCocoa/3.0-beta.6/ReactiveCocoa.podspec.json) subspec is broken when using frameworks. It's missing a header referenced in the bridging header.

ashfurrow:
Interesting! I'll take a look later. It could be just that ReactiveCocoa was not intended to be used standalone, or that I didn't anticipate this need when I made the podspec.

ashfurrow:
Yeah, looks like ReactiveCocoa's `Core` subspec is not intended to be used standalone. CocoaPods bundles in whatever UI stuff you need depending on your platform. Could be that this is a use case the ReactiveCocoa developers _did_ intend, but they don't contribute to the podspec. 

Thanks again for the contribution!

PR #162, opened 2015-07-03
[WIP] Swift 2.0

esttorhe:
# Fix #156 

This is still a WIP but wanted to create the `PR` anyways.
I tried tackling the `ReactiveCocoa` task but it proved harder than expected due to its `LlamaKit` dependency and it being stuck on `1.2` and no work to migrate to `2.0` is being made.

I went down the :rabbit: hole and this is a major rework I think.
So far `Moya` compiles on the `Demo` project with the current `Podfile` that's leaving `ReactiveCocoa` and `RxSwift` out of it (only core is being integrated).

This is what's done so far.

## Done:
- Currently using Quick and Nimble `swift-2.0` branches
- AlamoFire `swift-2.0` branch as well
- Moya Core is compiling
- Minimal Swift 2.0 migration (not reworked; only made it compile)
- @garethknowles migrated the code for `RxSwift`

## TODO:
- [ ] Migrate `ReactiveCocoa` to `swift2` branch
  - [ ] Generate a `podspec` for `swift2` branch first
    - ~~Migrate `LlamaKit` to `Swift2` :scream:~~ <-- `LlamaKit` is no longer a dependency
- [x] ~~Migrate `RxSwfit`~~

PR #167, opened 2015-07-09
Fixes for swift 2.0 in Xcode 7.3

garethknowles:
I have looked at getting Moya (specifically for RxSwift) working with swift 2.0.
Using the new Xcode 7.3 beta and by totally ignoring ReactiveCocoa, I think I've managed to get Moya somewhat working.

It seems to be working well in my own project but I haven't got the Demo project and its tests working. I also had problems with ReactiveCocoa due to Carthage setup which I'm not very familiar with.

The main concern for me is the return defer in Moya+RxSwift.swift due to the changes which won't allow defer's to return.  I don't really see the need for a defer anyway so have simply replaced with a return closure.

Nonetheless, hopefully this might help with the migration over to 2.0

### DIFF
--- a/Moya.podspec
+++ b/Moya.podspec
@@ -32,12 +32,12 @@ Pod::Spec.new do |s|
   s.subspec "ReactiveCocoa" do |ss|
     ss.source_files = "Moya/ReactiveCocoa/*.swift"
     ss.dependency "Moya/ReactiveCore"
-    ss.dependency "ReactiveCocoa", "3.0-beta.6"
+    ss.dependency "ReactiveCocoa"
   end
 
   s.subspec "RxSwift" do |ss|
     ss.source_files = "Moya/RxSwift/*.swift"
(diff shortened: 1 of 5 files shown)


ashfurrow:
Hey, thanks so much! This is a great step forward. @esttorhe has done some work in #162 – maybe you two could work together?

Until RxSwift renames `defer`, if it does, you just need to surround it in backticks like `` `defer ` ``.

esttorhe:
Hi, I rebased this branch into my PR and fixed the defer issue.

--
Esteban Torres
(+506)8813-0934
Twitter: @esttorhe




On Sat, Jul 11, 2015 at 8:07 AM -0700, "Ash Furrow" <notifications@github.com> wrote:












Hey, thanks so much! This is a great step forward. @esttorhe has done some work in #162 – maybe you two could work together?



Until RxSwift renames defer, if it does, you just need to surround it in backticks like `defer `.

Issue #149, opened 2015-06-17
Release 2.0


ashfurrow:
Is this ready for release? It's a 2.0 because of semantic versioning (we have breaking changes). The remaining issues are all long-term goals and most should be doable without breaking things. 

Anyone have feels about this?

alcarvalho:
One thing we could fix before the release are the versions to which we depend on CocoaPods to support sane updates on them like mentioned on #153. I have also checked that if we update all the pods of the Demo project we fetch the version 0.4.0 of Quick that won't compile with Swift 1.2.

alcarvalho:
I just found a problem with cocoapods versioning, though.

If we add the `~>` to ReactiveCocoa, when we run a `pod update` on the Demo project it downgrades to alpha version:

`Installing ReactiveCocoa 3.0.0-alpha.1 (was 3.0-beta.6)`

ashfurrow:
Hmm. So running `pod update` isn't really something that should be done frequently; I can't remember the last time I ran it. Running `pod install` instead looks at the Podfile.lock file to see what specific versions need to be used. Does that make sense?

alcarvalho:
Yeah, but the intent here was to update the pods, really. I don't know if that should be done on a library regularly. I did it also to test if everything still works with the latest versions of the libraries that we should support.

ashfurrow:
For sure, but Quick 0.4.0 uses Swift 2, as I understand it, which Moya does not yet.

alcarvalho:
Yeah, that's why I have restricted it to 0.3.x in here #157

ashfurrow:
OK, so this looks ready. Once #174 fixes CI, I'll release a 2.0 :tada:

ashfurrow:
Also pushed to @CocoaPods trunk: https://raw.githubusercontent.com/CocoaPods/Specs/27abd436816cd75b5fa76230c1c325d7810467dc/Specs/Moya/2.0.0/Moya.podspec.json

PR #172, opened 2015-08-03
Updates AlamoFire dependency to 1.3.0

colinta:
AlamoFire has added SSL Pinning, and a host of other features that I'm sure are just lovely, but that I don't have an immediate need for.

But I think SSL Pinning is the most compelling feature of this release.

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,2 +1,2 @@
-github "Alamofire/Alamofire" ~> 1.2.0
+github "Alamofire/Alamofire" ~> 1.3.0
 github "ReactiveCocoa/ReactiveCocoa" ~> 3.0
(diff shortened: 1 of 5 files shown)


colinta:
```
> make ci
..............................................

	 Executed 46 tests, with 0 failures (0 unexpected) in 3.557 (3.574) seconds
```

colinta:
Whaddup with those specs failing on CircleCI?  I haven't seen that on previous PRs.

colinta:
Oh, I see, looks like it's a known issue since #161.

rynbyjn:
Looks legit to me, would be sweet to get SSL Pinning in!

ashfurrow:
OK cool, I'll merge and we can fix CI later. 

I don't think `master` is in a stable state to release a build, so you'll have to use this in your Podfile:

```ruby
pod 'Moya', :head
```

colinta:
Do we need a minor version bump in Moya to use `:head` that way?

ashfurrow:
I don't believe so, `:head` fetches from the repo's `HEAD` iirc.

ashfurrow:
OK, so once #174 is done, I'll be release 2.0 in #149, so we should be :+1: soon.

PR #186, opened 2015-08-18
Rename DefaultEnpointResolution to *Endpoint*

mauricerkelly:
While reviewing some of the Moya documentation, I noticed that there was a class method on `MoyaProvider` called `DefaultEnpointResolution()`. I’ve assumed this to be a typo, and have updated it be named `DefaultEndpointResolution()`.

Sorry if the naming was intentional, or this PR is viewed as being overly pedantic!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+- Add backward-compatible call from `DefaultEnpointResolution` to `DefaultEndpointResolution` on `MoyaProvider` class. `DefaultEndpointResolution` is now used internally as the default resolver. `DefaultEnpointResolution` can be removed in a future major release.
+
 # 2.1.0
 
 - Add option to pass an `Alamofire.Manager` to `MoyaProvider` initializer
--- a/Demo/DemoTests/MoyaProviderSpec.swift
+++ b/Demo/DemoTests/MoyaProviderSpec.swift
@@ -319,7 +319,7 @@ class MoyaProviderSpec: QuickSpec {
                     }
 
                     class TestProvider<T: MoyaTarget>: ReactiveCocoaMoyaProvider<T> {
-                        override init(endpointClosure: MoyaEndpointsClosure = MoyaProvider.DefaultEndpointMapping, endpointResolver: MoyaEndpointResolution = MoyaProvider.DefaultEnpointResolution, stubBehavior: MoyaStubbedBehavior = MoyaProvider.NoStubbingBehavior, networkActivityClosure: Moya.NetworkActivityClosure? = nil, manager: Manager = Alamofire.Manager.sharedInstance) {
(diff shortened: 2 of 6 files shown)


mauricerkelly:
I probably should add a line to the changelog.md given that this changes the API - but I’ll leave it until I get the thumbs up/down.

ashfurrow:
Wow, nice catch! Thanks for the PR :cake: 

You mention the changelog – very good point. We  Typically, breaking changes require a major update to Moya (so, we'd need to release 3.0.0 after merging this). However, a major release for _just_ a typo seems extreme, especially when it's on an optional parameter. @orta I'd appreciate your thoughts on this.

orta:
Yeah, afraid this _is_ a major version break.  Two options: 

* Make a backwards API compatible function?
* Save the PR till you're ready for a v3

CHANGELOG.md
```diff
@@ -15,7 +15,7 @@
 - **Breaking change** Combines `MoyaPath` and `MoyaTarget` protocols.
 - **Breaking change** Renames `Moya/Reactive` subspec to `Moya/ReactiveCocoa`.
 - **Breaking change** Removes `stubResponses` from initializer; replaced with new stubbing behavior `.NoStubbing`. Added class methods to `MoyaProvider` to provide defaults, while allowing users to still change stubbing behaviour on a per-request basis.
-- **Breaking change** Redefines types of `DefaultEndpointMapping` and `DefaultEnpointResolution` class functions on `MoyaProvider`. You no longer invoke these functions to return a closure, rather, you reference the functions themselves _as_ closures.
```
ashfurrow: Since we're going to make this a change to `3.0.0`, we should keep this line the same so people still using 2.x can rely on the changelog.

mauricerkelly:
Hey guys, sorry for not coming back sooner. I like the idea of the backwards compatible function, but I want to make sure I have the same understanding as yous. I’m thinking something like this:

```swift
public class func DefaultEnpointResolution(endpoint: Endpoint<T>) -> NSURLRequest {
    return DefaultEndpointResolution(endpoint)
}

public class func DefaultEndpointResolution(endpoint: Endpoint<T>) -> NSURLRequest {
    return endpoint.urlRequest
}
```

ashfurrow:
That makes sense to me – we could add a new issue to remove it for 3.0.

orta:
Yep :+1:

mauricerkelly:
Okay, I’ve force pushed an update to this PR that has the following changes:

* Adds `DefaultEndpointResolution` as a method
* Updates `DefaultEnpointReolution` to call through to `DefaultEndpointResolution`
* Updates the `MoyaProvider#init` methods to use `DefaultEndpointResolution` as the default
* Adds an entry to the `CHANGELOG.md` file that explains the change - no version number with it, as I’m assuming you fine folks will add a version at the top when you decide what the next release will be.
* Fixes a comment typo from `metods` to `methods`

Let me know if there’s anything else that I’ve missed, or can improve upon.

ashfurrow:
Looks great, thanks!

mauricerkelly:
Thanks for accepting it - was a pleasure at this end :-)

orta:
good job

PR #202, opened 2015-08-27
Deprecate wrong-named function

skywinder:
Maurice Kelly add new function and fix typo in fde0eca3d63d79b615093db8c072673091c53375.
I believe we should left old-named method for backward compatability for a while, but it would be nice to deprecate wrong named method. Then we can delete it in next major update.


orta:
Yep!

Good thinking @skywinder

Issue #151, opened 2015-06-17
Support for HTTP Basic Auth?


davidbarsky:
This is likely entirely my fault for missing it, but — I looked in the Kiosk app, and Artsy seems to encode the API keys into the the URL itself, but the API I’m working against encodes the keys using HTTP Basic Auth. Is that a supported feature of Moya, or is that something that I should drop down into Alamofire for (or even open a PR for)?

Thanks.

davidbarsky:
(I should add — a lengthy answer isn’t necessary! just a yes or no would suffice.)

ashfurrow:
tl;dr Moya doesn't currently support this.

Hmm. Interesting question. Alamofire seems to [handle this with `NSURLCredential`](https://github.com/Alamofire/Alamofire/blob/922c1ca3580b7adc9235bede1e99580e02ad3e9f/Source/Request.swift#L76-L80). Is this how you've done HTTP auth in the past? 

Maya doesn't currently have support for this, but using the `sharedSession` `NSURLSession`, so you _should_ be able to set things there and have all HTTP requests use those credentials. Obviously that's not a great fix, though. 

I guess what this would really need is some kind of (optional) way to specify which `NSURLSession` to use: the default, or one configured for your endpoints alone. 

Others are likely more familiar with the network library underneath Alamofire. Anyone have suggestions on how to implement this?

davidbarsky:
Unless I misunderstand, HTTP Basic Auth is implemented by Alamofire as just a chained method after `.request`. 

Here’s my naive suggestion: pass the username/password to `MoyaProvider#sendRequest` as an optional parameter of type `NSURLCredential`, and if it’s not nil, call `.authenticate` after the `Alamofire.Manager.sharedInstance.request(request)` call.

Granted, that makes the method messy, hence my referring to it as “naive”.

ashfurrow:
That makes sense. But it also adds a lot of complexity around individual API calls. Developers now need to remember which endpoints need auth and which don't, and that need is spread across their entire codebase. 

I think we need one of two things:

- an optional part of the `MoyaTarget` protocol that specifies per-target credentials
- an optional closure in the `MoyaProvider` initializer that gets last-moment chance to specify credentials based on the target. If it returns credentials, they're added to the request in the provider. 

I like the second option better because most users don't need HTTP basic auth and it keeps our protocol leaner. What do you think?

davidbarsky:
[...]

Issue #144, opened 2015-06-14
Make it easier to subclass MoyaProvider


ashfurrow:
It uses default values for its initializers that need to be duplicated by subclasses. 

```swift
public init(endpointClosure: EndpointClosure = MoyaProvider.DefaultEndpointMapping,
    requestClosure: RequestClosure = MoyaProvider.DefaultRequestMapping,
    stubClosure: StubClosure = MoyaProvider.NeverStub,
    networkActivityClosure: Moya.NetworkActivityClosure? = nil,
    credentialClosure: CredentialClosure? = nil,
    manager: Manager = Alamofire.Manager.sharedInstance) {

    self.endpointClosure = endpointClosure
    self.requestClosure = requestClosure
    self.stubClosure = stubClosure
    self.networkActivityClosure = networkActivityClosure
    self.credentialClosure = credentialClosure
    self.manager = manager
}
```

Dunno how we could do this..

ashfurrow:
This is really tricky. Basically our providers are containers for data (their configs) and a `request` method. That's, like, the definition of a class. 

I suppose we could move `MoyaProvider` to be a protocol and put `request()` in a protocol extension? Or maybe turn Moya into a `struct` instead of a class. The real hassle is just updating initializers that don't do anything but call `super.init(...` so structs would remove that pain most of the time.

ashfurrow:
Alternatively, it seems a bit odd that subclasses _must_ override the initializer. 

```swift
override public init(endpointClosure: EndpointClosure = MoyaProvider.DefaultEndpointMapping,
    requestClosure: RequestClosure = MoyaProvider.DefaultRequestMapping,
    stubClosure: StubClosure = MoyaProvider.NeverStub,
    networkActivityClosure: Moya.NetworkActivityClosure? = nil,
    credentialClosure: CredentialClosure? = nil,
    manager: Manager = Alamofire.Manager.sharedInstance) {
    
        super.init(endpointClosure: endpointClosure, requestClosure: requestClosure, stubClosure: stubClosure, networkActivityClosure: networkActivityClosure, credentialClosure: credentialClosure, manager: manager)
}
```

All the subclasses ever do is call `super.init()`. How can we abstract this?

ashfurrow:
Maybe instead of subclassing, we can add extensions to the existing `MoyaProvider` type?

ashfurrow:
[...]

PR #205, opened 2015-08-31
Added authentication support

avf:
Hi,

I added HTTP Basic auth support (and other protocols that Alamofire supports, like NTLM) by adding an optional closure to the `MoyaProvider` initializer. This has been discussed previously and I've basically implemented what Ash suggested [here](https://github.com/Moya/Moya/issues/151#issuecomment-112961113).

I've only tested this with NTLM (yeah, I know...), so this should probably be tested a bit more with other authentication protocols before being merged. There also seems to be an issue with the current Alamofire version and NTLM, for more info [see here](https://github.com/Alamofire/Alamofire/issues/721).

### DIFF
--- a/Moya/Moya.swift
+++ b/Moya/Moya.swift
@@ -101,17 +101,20 @@ public class MoyaProvider<T: MoyaTarget> {
     /// Closure that resolves an Endpoint into an NSURLRequest.
     public typealias MoyaEndpointResolution = (endpoint: Endpoint<T>) -> (NSURLRequest)
     public typealias MoyaStubbedBehavior = ((T) -> (Moya.StubbedBehavior))
+    public typealias MoyaCredentialClosure = (T) -> (NSURLCredential?)
     
     public let endpointClosure: MoyaEndpointsClosure
     public let endpointResolver: MoyaEndpointResolution
+    public let credentialClosure: MoyaCredentialClosure?
     public let stubBehavior: MoyaStubbedBehavior
     public let networkActivityClosure: Moya.NetworkActivityClosure?
     public let manager: Manager
 
     /// Initializes a provider.
-    public init(endpointClosure: MoyaEndpointsClosure = MoyaProvider.DefaultEndpointMapping, endpointResolver: MoyaEndpointResolution = MoyaProvider.DefaultEndpointResolution, stubBehavior: MoyaStubbedBehavior = MoyaProvider.NoStubbingBehavior, networkActivityClosure: Moya.NetworkActivityClosure? = nil, manager: Manager = Alamofire.Manager.sharedInstance) {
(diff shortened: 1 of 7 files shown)


ashfurrow:
This is great work, thanks @avf :cake: This is exactly how I would've implemented it, though it does make https://github.com/Moya/Moya/issues/144 even more important to deal with. 

Just a few things – you'll need to pull in changes from the master branch so we can merge your pull request – happy to help with that. 

Some tests would be nice, but it can be challenging to write them. If you like, we can write them together! Also, could you add an entry to the changelog? Make sure to give yourself credit!

Again, fantastic work :boom:

Moya/Moya.swift
```diff
@@ -126,11 +129,12 @@ public class MoyaProvider<T: MoyaTarget> {
...
         let request = endpointResolver(endpoint: endpoint)
+        let credential = credentialClosure?(token)
         let stubBehavior = self.stubBehavior(token)
 
         switch stubBehavior {
         case .NoStubbing:
-            return sendRequest(request, completion: completion)
+            return sendRequest(request, credential:credential, completion: completion)
         default:
             return stubRequest(request, completion: completion, endpoint: endpoint, stubBehavior: stubBehavior)
```
ashfurrow: Do you think it could be valuable to use the credentials when stubbing a request, too? I can't _really_ see how, but wanted to see if it was something we might want to add.

avf: Hmm, good question, the only reason I can think of is for debugging purposes maybe, to see if the credentials are set correctly for each request? Haven't really used the stubbing yet, so I'm not sure.

avf:
Thanks for your kind words :-)

I've merged the master and you should be able to merge the pull request now.

About tests: I'm pretty busy right now, but maybe I'll give it a shot next week. I'm familiar with OHTTPStubs, but I haven't used Nimble before. I'll message you when I get started or if I need help, ok?

ashfurrow:
OK, sounds great! Feel free to reach out to ash@ashfurrow.com with questions. If I have some time this week, I'll send you some first steps that might help.

ashfurrow:
Hey! It's been a few weeks so I thought I'd check in. I never got around to writing some tests (busy weekend!) but this week I plan to get this merged. It'll take some work to resolve some conflicts – I'll take care of those – and I'll write a few tests. I'll post a new pull request and ask you to review it for us, if that's ok? Thanks again!

ashfurrow:
Hey! So the tests you added look lovely. I just realized, though, that the `manager` we recently added should provide, this, right? The Alamofire's `Manager` class seems to handle a lot of customization of `NSURLSession` and requests. 

Do you think HTTP auth is common enough that we should have a closure specifically for it? Or should we stick to the manager? I like the closure since it keeps things abstracted with the `MoyaTarget` protocol. But it also adds yet another `init()` parameter :sweat_smile: Any thoughts, @avf?

ashfurrow:
Thought about this more, and I really like you're implementation. It lets people auth on per-target instead of per-URL, which is the whole point of Moya. Great work, I'll get this merged ASAP.

ashfurrow:
OK, this is merged :tada: thanks again for your hard work, @avf. I've release 2.4.0 on CocoaPods trunk, and tagged a release on the repo for Carthage.

avf:
Hey, thanks for doing the merge! I should really turn on github notifications ;-)

PR #207, opened 2015-09-04
Update to MoyaCompletion closure for recent Alamofire changes

jarrodrobins:
I did a `carthage update` this morning and saw that Alamofire pushed a new commit to their `swift-2.0` branch which breaks Moya. The commit (https://github.com/Alamofire/Alamofire/commit/64a004472c16313d04671518ef03ef4b72293dbc) swaps out NSError for ErrorType. 

I've updated MoyaCompletion to now take `ErrorType` rather than `NSError` to be inline with Swift 2 changes.

I should note that I haven't updated the RxSwift nor ReactiveCocoa extensions simply because I wasn't sure how you would want to handle the following bit that's present in both Moya+RxSwift and Moya+ReactiveCocoa.

    if let statusCode = statusCode {
        observer.on(.Error(NSError(domain: error.domain, code: statusCode, userInfo: error.userInfo)))
    } else {
        observer.on(.Error(error))
    }

As you can see, this just replaces the statusCode if it's been given back by Alamofire. We can't suck out the domain and userInfo from the ErrorType unless it's cast to NSError. Xcode 7 beta 6 spits out a warning that this always succeeds so I was certainly averse to putting that particular code smell in. I suppose we could potentially make a new ErrorType within Moya which has the statusCode as an argument but I didn't know whether that'd be a bit of an overkill? Alternatively we could just simplify it and return the existing error. I haven't been using either of these extensions at this point so I figured I'd leave that to someone who might have a bit more expertise with them on how it should be handled.

### DIFF
--- a/Moya/Moya.swift
+++ b/Moya/Moya.swift
@@ -2,7 +2,7 @@ import Foundation
 import Alamofire
 
 /// Block to be executed when a request has completed.
-public typealias MoyaCompletion = (data: NSData?, statusCode: Int?, response: NSURLResponse?, error: NSError?) -> ()
+public typealias MoyaCompletion = (data: NSData?, statusCode: Int?, response: NSURLResponse?, error: ErrorType?) -> ()
 
 /// General-purpose class to store some enums and class funcs.
 public class Moya {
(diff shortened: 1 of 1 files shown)


ashfurrow:
Cool – looks great, thanks!

PR #215, opened 2015-09-13
Carthage Support

### LINKED ISSUE #154: Carthage error with RxSwift
i installed Moya with Carthage and i get a error in xcode when building because i don't have RxSwift, maybe to have a another branch or something for RxSwift? 

this is the error i get

    dyld: Library not loaded: @rpath/RxSwift.framework/RxSwift
    Referenced from: /Users/marvinnazari/Library/Developer/CoreSimulator/Devices/9A8009C8-48F2-47FD-86A8-7E6251DF5B60/data/Containers/Bundle/Application/CB3D1236-77FF-4664-8EEF-D94DD08CE55D/X.app/Frameworks/Moya.framework/Moya
    Reason: image not found

### LINKED ISSUE #184: Make releasing easier
Currently it takes a lot of steps:

1. Update podspec version.
1. In Demo directory, `pod update Moya`.
1. Commit.
1. Tag new version.
1. Push, including tags.
1. `pod trunk push Moya.podspec`.
1. Hope that current tag lints fine on trunk push.

We should automate this process, preferably using Fastlane.

### LINKED ISSUE #204: Carthage having issues with Pods-DemoTests-Moya
It seems it still builds Moya.framework with this podfile:
```
github "Alamofire/Alamofire" ~> 1.3.0
github "ReactiveCocoa/ReactiveCocoa" ~> 3.0
github "ashfurrow/Moya"
```

**But it is spitting out this error:**
`xcodebuild: error: Scheme Pods-DemoTests-Moya is not currently configured for the build action.`

### LINKED ISSUE #216: Demo: dyld: Library not loaded: @rpath/libswiftCoreLocation.dylib
I can't get demo running. I'm using Xcode 7.0. Here is error log from console:

````
dyld: Library not loaded: @rpath/libswiftCoreLocation.dylib
  Referenced from: /Users/hello/Library/Developer/Xcode/DerivedData/Demo-djsbbkgjbpaearfipsplncpfociz/Build/Products/Debug-iphonesimulator/Moya.framework/Moya
  Reason: image not found
````
````
$ cat Podfile
platform :ios, '8.0'

source 'https://github.com/CocoaPods/Specs.git'

use_frameworks!

# Swift indirect dependencies we need to force to Swift 2
pod 'Box', :head
pod 'Result', :head

pod 'ReactiveCocoa', :git => 'https://github.com/ashfurrow/ReactiveCocoa.git', :branch => 'swift2'
pod 'RxSwift', '~> 2.0.0-alpha'
pod 'Alamofire'

pod 'Moya/ReactiveCocoa', :path => "../"
pod 'Moya/RxSwift', :path => "../"
pod 'Moya', :path => "../"

target 'DemoTests' do

  pod 'Quick', :git => 'https://github.com/Quick/Quick'
  pod 'Nimble', :git => 'https://github.com/Quick/Nimble'
  pod 'OHHTTPStubs'

end
````

I'd be glad if someone could help. Thanks!

### PR DESCRIPTION
**Pull request by ashfurrow**
This pull requests adds Carthage support to Moya, including its two extensions, RxMoya and ReactiveMoya. It builds off of @justinmakaila's work in #169 to include Carthage support, but works using a different solution that @orta and I implemented this afternoon. Let me explain.

In #154, we discussed a solution that involved splitting up Moya into several repositories to generate separate `.framework` files for each Moya, ReactiveMoya, and RxMoya. The solution in this PR instead uses thin podspecs _solely_ for the ReactiveCocoa and RxSwift extensions. This lets the `Pods` Xcode project build separately-named targets, which compile into the frameworks that Carthage needs.

The benefits of this approach are that we keep all of our code in _one_ repository instead of having to synchronize version numbers, tags, pull requests, and so on between three or four repositories. This would be a significant challenge.

The downsides are that we need to keep versions synced between the three podspecs (I've written a `rake` task to automate this). Additionally, if the public APIs to RxSwift or ReactiveCocoa change, Carthage users will need to continue to use the versions required by the CocoaPods builds. I actually kind of like the idea of one Moya existing in parallel on two dependency managers.

## Work Done

- We switched from a `Makefile` to a `Rakefile`.
- Automated releases.
- Re-added `Pods/` directory for Carthage support.
- Duplicated the RxMoya and ReactiveMoya subspecs into their own standalone podspecs, and trimmed them down.
- Removes unnecessary `Box` dependency from the Demo's `Podfile`.
- Updated documentation.

## Next Steps

The [ReactiveMoya](https://github.com/Moya/ReactiveMoya) and [RxMoya](https://github.com/Moya/RxMoya) repositories are no longer needed and should be deleted after this pull request is merged. Any existing Carthage users will need to update their Cartfiles.

## Notes

`MoyaResponse.swift` is included in both Carthage-specific podspecs, but it only appears once in the `Pods` project's file pane. The file is a member of both targets, but only appears once. The ambiguity between the files referenced in both frameworks led me to split up the unit tests into distinct files (to avoid having both `MoyaResponse`s in scope). I also tidied them up a bit.

Fixes #154.
Fixes #184.
Fixes #204.
Fixes #216.

### CODE REVIEW

**Comment by justinmakaila** [comment]
How does this work for Cartfiles? What do carthage users do to use ReactiveMoya, RxMoya, and Moya, on a per requirement basis? Is that a concept anymore, or are we back to the single Moya.framework?



—Justin Makaila

**Comment by orta** [comment]
It generates all three frameworks when you do `github "Moya/Moya"` - as a user, you can choose which ones to integrate this way:

```
~/d/i/l/moya_cart ⏛  cat Cartfile
git "file:///Users/orta/dev/ios/libs/Moya" "internal_podspecs"

~/d/i/l/moya_cart ⏛  carthage update
*** Fetching Moya
*** Checking out Moya at "e6a00bff2b51838e4449afac952b3937d1cfa6de"
*** xcodebuild output can be found in /var/folders/mk/gxhr550d5mbcv5s9_f61x84h0000gn/T/carthage-xcodebuild.uPEpwO.log
*** Building scheme "Moya" in Pods.xcodeproj
*** Building scheme "ReactiveMoya" in Pods.xcodeproj
*** Building scheme "RxMoya" in Pods.xcodeproj

~/d/i/l/moya_cart ⏛  ls Carthage/Build/iOS
Moya.framework         ReactiveMoya.framework RxMoya.framework
```

**Comment by justinmakaila** [comment]
Before I started the Carthage refactor, that's initially what I did. I had three frameworks output, with all of their dependencies. The downside was that in every file where you wanted to use any component of `ReactiveMoya`, you had to import `Moya ` as well. 

Is this still the case? Or do `ReactiveMoya` and `RxMoya` have the `Moya` core embedded?

**Comment by orta** [comment]
Right now that's not the case, but I don't see any reason to not make that behaviour default @ashfurrow ?

**Comment by justinmakaila** [comment]
It shouldn't be the default, it was an annoyance, for which we initially compensated by creating the separate repositories. This allowed for Cartfile level specification of the flavor of Moya, and an output of a single Moya.franework.



—Justin Makaila

**Comment by orta** [comment]
So you want the behaviour where you would always do `@import Moya;` and it would work fine regardless of which framework you've imported? This is totally doable using [module_name](https://guides.cocoapods.org/syntax/podspec.html#module_name) in the other Podspecs. I *think* it would generate three differently named frameworks for drag & drop but they would all all work with `@import Moya;`. Can't guarantee that Xcode allows the generation of the three frameworks with the same module name, but can be tested.

**Comment by justinmakaila** [comment]
Yes, exactly. That way, if you change your flavor, you don't have to search for all the imports and update, you just update your usage of the API.



—Justin Makaila

**Comment by ashfurrow** [comment]
OK, I'll play around with the `module_name` for the specs and verify it works :+1:

**Comment by gfontenot** [comment]
fwiw, pulling this down and running `carthage build --no-skip-current` results in a failure (with Xcode 7):

```
❯ carthage build --no-skip-current
*** xcodebuild output can be found in /var/folders/ym/cjv2ctc130v3slskyn7zw_2r0000gn/T/carthage-xcodebuild.Kbexbi.log
*** Building scheme "RxMoya" in Demo.xcworkspace
** BUILD FAILED **


The following build commands failed:
        CompileSwift normal x86_64 /Users/gordon/Code/Moya/Moya/Moya/RxSwift/Moya+RxSwift.swift
        CompileSwift normal x86_64 /Users/gordon/Code/Moya/Moya/Moya/ReactiveCore/MoyaResponse.swift
        CompileSwiftSources normal x86_64 com.apple.xcode.tools.swift.compiler
(3 failures)
/Users/gordon/Code/Moya/Moya/Moya/RxSwift/Moya+RxSwift.swift:2:8: error: no such module 'Moya'
/Users/gordon/Code/Moya/Moya/Moya/RxSwift/Moya+RxSwift.swift:2:8: error: no such module 'Moya'
A shell task failed with exit code 65:
** BUILD FAILED **


The following build commands failed:
        CompileSwift normal x86_64 /Users/gordon/Code/Moya/Moya/Moya/RxSwift/Moya+RxSwift.swift
        CompileSwift normal x86_64 /Users/gordon/Code/Moya/Moya/Moya/ReactiveCore/MoyaResponse.swift
        CompileSwiftSources normal x86_64 com.apple.xcode.tools.swift.compiler
(3 failures)
```

**Comment by ashfurrow** [comment]
Hmm. Looks like it doesn't have the other Moya files. Makes sense since they're no longer subspecs. Thanks @gfontenot!

**Comment by ashfurrow** [comment]
@justinmakaila Looks like CocoaPods can't handle having frameworks with the same module name. 

```
[!] The 'Pods' target has frameworks with conflicting names: Moya.
```

This means that users will need to `import Moya`, `import RxMoya`, or `import ReactiveMoya` respectively. Switching flavours would mean a find-and-replace. Kind of a pain, but people don't often switch flavours, and it's not too much effort. It also fits better with the Carthage folks' philosophy. 

@gfontenot Running the command `carthage build --no-skip-current --platform iOS` (we don't support Mac yet) now produces the following results:

```
*** xcodebuild output can be found in /var/folders/c9/qfv9zg9n23v5yt50skqpmtg40000gn/T/carthage-xcodebuild.5IPQce.log
*** Building scheme "Moya" in Demo.xcworkspace
*** Building scheme "ReactiveMoya" in Demo.xcworkspace
*** Building scheme "RxMoya" in Demo.xcworkspace
```

This seems good to me :tada: 

There's a failure on CI – looks like Xcode is getting confused about something when building the unit tests. Investigating now.

ashfurrow:
:green_apple:

ashfurrow:
Instead of removing the no-longer-needed [ReactiveMoya](https://github.com/Moya/ReactiveMoya) repo, we should transfer its ownership to someone so that GitHub redirects will work long enough for people to update their configs. @justinmakaila would you like it?

ashfurrow:
I've fixed #216 on this branch – just rearranging the podfile a bit. I've run carthage locally and it seems to be :+1:

justinmakaila:
Sure, I'll take it



—Justin Makaila

ashfurrow:
Hmm. Failing on CI, succeeding locally with Xcode, and failing from `xcodebuild` but for a _different_ reason.

ashfurrow:
OK, thanks @justinmakaila!

ashfurrow:
Hmm, @justinmakaila I can't transfer it to you because you have a fork already. GitHub made me transfer it to me _first_ anyway, so the redirects will continue to work. Let me know if you still want it – you'd need to delete your fork first.

ashfurrow:
:tea: This is green again. I don't want to keep it open too much longer, or we're likely to get scope creep :sweat_smile: The Demo apps runs and the unit tests pass, and carthage builds fine – going to give some more time for review, then merge and release 2.2.0 :tada:

PR #218, opened 2015-09-14
@orta: Moves framework files

### LINKED ISSUE #195: Support ReactiveCocoa with Swift 2.0
Right now `Moya` compiles and works on its `swift-2.0` branch including `RxSwift` but `ReactiveCocoa` is lagging behind due to a non-official `podspec`.

We need to support `ReactiveCocoa` as a dependency working with `Swift 2.0`

ashfurrow:
Before: Moya was a dependency of the Carthage-only subspecs.
Now: Moya files are included within the Carthage-only subspecs.
Side-effects: `GitHub` enum must be duplicated throughout unit tests because there are now multiple `MoyaTarget` protocol lookups available, which confuse the Swift compiler. No preprocessor means we need to duplicate these. Will to explore options later, but just want to get this shipped for now. 

Incidentally fixes #195.


orta:
This is what I expected carthage behavior would be like based on this comment from [justin](https://github.com/Moya/Moya/pull/218/files#diff-04c6e90faac2675aa89e2176d2eec7d8R83) :+1:

orta:
good to go on green.

ashfurrow:
Coooooool, thanks.

PR #219, opened 2015-09-14
Added endpointByAddingParameterEncoding

AlexanderKaraberov:
I think this method can be useful to avoid boilerplate, because in my case I have to use different parameter encodings for some selected requests and I didn't find another method to quickly change encoding for a particular API request, except from directly calling init with all parameters again.

### DIFF
--- a/Moya/Endpoint.swift
+++ b/Moya/Endpoint.swift
@@ -55,6 +55,12 @@ public class Endpoint<T> {
 
         return Endpoint(URL: URL, sampleResponse: sampleResponse, method: method, parameters: parameters, parameterEncoding: parameterEncoding, httpHeaderFields: newHTTPHeaderFields)
     }
+    
+    /// Convenience method for creating a new Endpoint with the same properties as the receiver, but with another parameter encoding.
+    public func endpointByAddingParameterEncoding(parameterEncoding: Moya.ParameterEncoding) -> Endpoint<T> {
+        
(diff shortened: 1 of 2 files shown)


Moya/Endpoint.swift
```diff
@@ -55,6 +55,12 @@ public class Endpoint<T> {
 
         return Endpoint(URL: URL, sampleResponse: sampleResponse, method: method, parameters: parameters, parameterEncoding: parameterEncoding, httpHeaderFields: newHTTPHeaderFields)
     }
+    
+    /// Convenience method for creating a new Endpoint with the same properties as the receiver, but with another parameter encoding.
+    public func endpointByAddingParameterEncoding(parameterEncoding: Moya.ParameterEncoding) -> Endpoint<T> {
+        
+        return Endpoint(URL: URL, sampleResponse: sampleResponse, method: method, parameters: parameters, parameterEncoding: parameterEncoding, httpHeaderFields: httpHeaderFields)
```
ashfurrow: Looks great! What do you think about changing the parameter to `newParameterEncoding` so we can avoid confusing it with the instance variable?

ashfurrow:
Super! Don't forget to add a changelog entry with the update!

Moya/Endpoint.swift
AlexanderKaraberov: Yes, I think it will be reasonable.

ashfurrow:
Coooool this is up on CocoaPods trunk: https://raw.githubusercontent.com/CocoaPods/Specs/73d48221ff011b8eb62174b9761bca23c5ba78ca/Specs/Moya/2.2.2/Moya.podspec.json 

Thanks again!

Issue #118, opened 2015-05-14
RAC 3.0 support


MichaelMcGuire:
Were you planning on incorporating ReactiveCocoa extensions based on the 3.0 API?  I'm about to create my own extensions and wouldn't mind sending them along if you were interested.

Thanks for the awesome library!

ashfurrow:
:bow: A pull request with the extensions would be awesome! Let me know what I can do to help.

ashfurrow:
I've not looked too extensively into this, and would appreciate guidance or assistance.

Maybe we could add this new functionality to the _existing_ `ReactiveMoyaProvider` class? Or would we want a 3.0-only class? Since the two APIs are interchangeable, it might make sense to have only one class.

nanoxd:
I think it would be best to have it in one class. Perhaps it would be best to have three versions of the API:

1. ReactiveCocoa 2 (RACSignals)
2. ReactiveCocoa 3 Bridge ([RACSignal -> SignalProducer](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/swift-development/ReactiveCocoa/Swift/ObjectiveCBridging.swift))
3. ReactiveCocoa 3 (SignalProducer)

2 would provide an API for those already using `RACSignal` throughout their app and want to switch over piecemeal to RAC 3.

sebinsua:
Has anybody done any work on this yet?

rpowelll:
I'm starting a new project with Moya and RAC 3.0, so it's likely I'll end up writing these extensions as part of them. If I'm happy with how they turn out, I'd be happy to contribute them back.

ashfurrow:
@rpowell that would be lovely!

rpowelll:
@sammoore Yeah I've been working on a small project on and off over the past few weeks. I'll work on getting my extensions polished up and publish the project on the weekend if all goes as planned.

justinmakaila:
This is all taken care of in github.com/Moya/ReactiveMoya. Unfortunately, there's no CocoaPods support yet, only Carthage.

zbeckman:
What's the current/best configuration for use with Swift 2 (Xcode 7) and RAC? It looks like the Moya swift-2.0 branch, w/ RAC swift-2.0 branch, and a carthage pull of ReactiveMoya?

ashfurrow:
[...]

PR #224, opened 2015-09-19
Adds ReactiveCocoa 4 Support

petester42:
This should address #118. It is based on @justinmakaila's work for ReactiveCocoa 3 but with processing methods moved to an extension. I also added tests for the new `SignalProducer` code (kind of a duplication of the `RACSignal` tests).

Everything should be :green_apple:  and ready to go!

If anyone has input on how to make [this](https://github.com/Moya/Moya/compare/Moya:master...petester42:reactive-cocoa-4?expand=1#diff-6e6eda96ca4cf5e24ec7f03907dbbbcbR15) nicer and keep the compiler happy it would be great. The force cast should never fail but I don't really like it.


ashfurrow:
Awesome, thanks! First glance looks great, I'll do an in-depth review tomorrow. Maybe a generic constraint could avoid the forced cast?

Moya/ReactiveCocoa/Moya+ReactiveCocoa.swift
```diff
@@ -2,68 +2,76 @@ import Foundation
...
-                objc_sync_enter(weakSelf)
-                weakSelf.inflightRequests[endpoint] = signal
-                objc_sync_exit(weakSelf)
             }
-            
-            return signal
         }
     }
+    
+    public func request(token: T) -> RACSignal {
```
ashfurrow: Oh, this is smart. So this is _not_ a breaking change, then? Existing users won't need to change any code?

ashfurrow:
This looks _amazing_ – really well done both of you. I didn't see a changelog entry – could you make one?

If this is a breaking change, I'd suggest we merge into the 3.0-develop branch instead of master. If it's _not_ a breaking change, I'm comfortable merging this.

Moya/ReactiveCocoa/Moya+ReactiveCocoa.swift
petester42: It shouldn't be if the developer has their types set properly. For example if the developer just has something like this `let signal = provider.request(target)` then there will be an error. They would need to specify the type or chain it directly with a subscribe like so `provider.request(target).subscribeNext { _ in  }`. It's not really breaking but it might cause errors for some users.

ashfurrow: Hmm. I'd want to be sure that we're not breaking semantic versioning. Just to be safe, I'll try using your branch on [our app](https://github.com/artsy/eidolon). It uses it thoroughly, so if we don't run into issues, I don't expect others will either. I'll get back to you tomorrow :boat:

ashfurrow:
Sorry, didn't get to this today :( I will do it first thing tomorrow!

ashfurrow:
:grimacing: This'll need changes from master pulled in. Let me know if I can help!

petester42:
I think that should do it.

ashfurrow:
Hmm, still showing as conflicted here: 

screen shot 2015-09-21 at 10 55 06 pm

Can you double-check?

petester42:
There we go! xcodeproj conflicts are the worst.

ashfurrow:
Definitely, thanks :) I'll get back to you about this within a few hours.

ashfurrow:
OK, I'm not aware of anyone who uses the RAC 2 extensions for thoroughly than our app, and in migrating I only saw _one_ issue:

```swift
let request = Provider.sharedProvider.request(.FindExistingEmailRegistration(email: email))

return request.map { (response) -> NSNumber in
            let moyaResponse = response as! MoyaResponse
            return moyaResponse.statusCode != 404
        }
```

It said the type was ambiguous and was easily fixed. 

This feels OK to me, but I know @orta is the semver police here. We'll merge into master if we get the :+1: or merge into 3.0.0-develop otherwise.

ashfurrow:
OK, since _technically_ this is a a breaking change, and I don't want to give Moya users unexpected surprises when the `pod update`, I'm going to merge this into 3.0.0-develop instead. We'll be releasing 3.0 sometime this week, anyway, so you won't have to wait long to see this deployed :cake:

Thanks again for your awesome work!

ashfurrow:
OK, cool – this is merged in the develop branch, you can see your contributions to the history here: https://github.com/Moya/Moya/commits/3.0.0-develop

PR #225, opened 2015-09-19
Adds RxSwift Data Processing

petester42:
Why does ReactiveCocoa get to have all the fun. Here are the same data processing extensions but for RxSwift. Tested and all.

Enjoy!


Moya/ReactiveCore/MoyaError.swift
```diff
@@ -0,0 +1,11 @@
+import Foundation
```
ashfurrow: Smart.

ashfurrow:
Pierre-Marc, this is _fantastic work_, thank you very much for your contribution! Well tested, well thought-out. Only thing I can find missing is an entry in the `CHANGELOG.md` file. 

Good to merge this once the changelog is in – then we'll release a new version for CocoaPods/Carthage so people can start using this immediately.

PR #229, opened 2015-09-22
Removes Inflight Request tracking/deduplication.

ashfurrow:
This caused issues for developers not expecting Moya to deduplicate network requests, as well as made it a _pain_ to test and maintain. Ideally, if people even want these, we can further subclass our Rx providers and limit the scope of what they do _just_ to inflight tracking. I don't think anyone will miss it, anyway.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Next
 
+- Removes inflight request tracking for ReactiveCocoa and RxSwift providers. **Breaking Change**
+
 # 2.3.0
 
 - Adds data processing functions for use with `RxMoyaProvider`
--- a/Demo/DemoTests/MoyaProviderIntegrationTests.swift
+++ b/Demo/DemoTests/MoyaProviderIntegrationTests.swift
@@ -194,30 +194,6 @@ class MoyaProviderIntegrationTests: QuickSpec {
 
                         expect{message}.toEventually( equal(userMessage) )
                     }
-                    
-                    it("returns identical signals for inflight requests") {
-                        let target: GitHub = .Zen
-                        let signal1 = provider.request(target)
-                        let signal2 = provider.request(target)
-                        
-                        expect(provider.inflightRequests.count).to(equal(0))
-                        
-                        var receivedResponse: MoyaResponse!
-                        
-                        signal1.subscribeNext { (response) -> Void in
-                            receivedResponse = response as? MoyaResponse
-                            expect(provider.inflightRequests.count).to(equal(1))
-                        }
-                        
-                        signal2.subscribeNext { (response) -> Void in
-                            expect(receivedResponse).toNot(beNil())
-                            expect(receivedResponse).to(beIndenticalToResponse( response as! MoyaResponse) )
-                            expect(provider.inflightRequests.count).to(equal(1))
-                        }
-                        
-                        // Allow for network request to complete
-                        expect(provider.inflightRequests.count).toEventually( equal(0) )
-                    }
                 }
             }
         }
--- a/Demo/DemoTests/ReactiveCocoaMoyaProviderTests.swift
+++ b/Demo/DemoTests/ReactiveCocoaMoyaProviderTests.swift
@@ -53,31 +53,6 @@ class ReactiveCocoaMoyaProviderSpec: QuickSpec {
             expect(receivedResponse) == sampleResponse
         }
 
-        it("returns identical signals for inflight requests") {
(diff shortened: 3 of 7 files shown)


ashfurrow:
Eek this shouldn't be a PR into master...

AvdLee:
Passing by, as I was looking for this feature. I did miss it, as my app requests many of the same API endpoints at the same time. I was wondering why all these calls were fired.

Just for my understanding, what would ideally be the way to handle these duplicate calls and prevent them to be all fired? Should I create my own inflights and duplicate detection?

colinta:
My thoughts on this: Ideally this would be something that could be wrapped up in a plugin - it seems out of scope with "core Moya."

But in order to do this in a plugin, we would need to expand the plugin API to include an ability to delay/cancel requests. Not unreasonable, and a "MoyaRequestThrottler" could be really handy!

ashfurrow:
It may help to look in an older release to see how we used to do it. It might lend itself well to one solution or another.

AvdLee:
Ok, I gave it a shot, but my knowledge did not let me create a certain solution using plugins. I'm really looking into a way to fix this for all Moya users. Anyone here to give me a good shot in the right direction?

ashfurrow:
It's tricky – I'm not sure this can be done with a plugin, not without significantly changing what a plugin is and does. The best bet would probably be to subclass `MoyaProvider`, but that doesn't benefit users already using the `RxMoyaProvider` or `ReactiveMoyaProvider`. Not sure there's an easy answer to this, but I'd love to hear ideas!

AvdLee:
I was thinking about a simple boolean added to the initializer of MoyaProvider `trackInflights:Bool = false` to make it available for all provider types. This makes if easy to enable for those who need it.

Creating a dictionary <Endpoint, [Moya.Completion]> containing all completion blocks for that request and call them when that existing request is finished.

How about that?

ashfurrow:
That makes sense and is a good solution, but we've had trouble before with "just add an initializer parameter" :smile: It eventually led to https://github.com/Moya/Moya/issues/144 and the idea of Moya plugins in the first place. I'm personally fine with it – it's going to be a breaking change either way, so reconsidering plugins might be worth it.

AvdLee:
Ok, I gave it a shot: 

https://github.com/AvdLee/Moya/commit/a029491b3219716c91b32feb0172e0f74be67154

For a quick test I duplicated the zen request. Checked it in Charles, one request, three completions.

Curious about your feedback!

ashfurrow:
Looks cool! Want to add tests and submit a pull request?

AvdLee:
I would love to, but I'm unexperienced with Nimble and Quick. I'm having trouble to even execute the existing tests, but I'm happy to learn and see this working.

Is there anything to do before the tests works with a clean checkout?

ashfurrow:
Nope, it should be a matter of opening the workspace and hitting ⌘U. What errors are you getting?

AvdLee:
Got it working, needed to force update the project to version 4.0.1 of Nimble, as a clean checkout used version 3.0.0. Had the convert string to staticstring error.

This is the unit test I've got now, which is close, but not successful yet:

```swift
        it("returns 3 times but executes one request") {
            var networkActivityCallTimes = 0
            var requestCompletionTimes = 0
            let plugin = NetworkActivityPlugin { (change) -> () in
                if change == .Ended {
                    networkActivityCallTimes += 1
                }
            }
            
            let provider = MoyaProvider<GitHub>(plugins: [plugin], trackInflights: true)
            let target: GitHub = .Zen
            
            provider.request(target) { _ in
                requestCompletionTimes += 1
            }
            provider.request(target) { _ in
                requestCompletionTimes += 1
            }
            provider.request(target) { _ in
                requestCompletionTimes += 1
            }
            
            expect(requestCompletionTimes).toEventually(beIdenticalTo(3))
            expect(networkActivityCallTimes).toEventually(beIdenticalTo(1))
        }
```

It's probably a small thing, any suggestions here?

ashfurrow:
Tricky. Not sure relying on the plugins is the best way to test. You could use a custom endpoint closure or something similar. 2.4.1 was the last version of Moya to support this on all providers (it was even turned on by default!) and it was tested [like this](https://github.com/Moya/Moya/blob/2.4.1/Demo/DemoTests/MoyaProviderIntegrationTests.swift#L267) but wasn't super-great either.

AvdLee:
Ok, got it! https://github.com/Moya/Moya/pull/477

ashfurrow:
Cool, thanks!

PR #228, opened 2015-09-22
Renames many things

ashfurrow:
These are meant for the 3.0.0 development branch, since they break things. 

Basically this renames a tonne of things that either no longer make sense, or never did :sweat_smile: Updated the docs and tried to add comments where possible.

PR #230, opened 2015-09-22
Makes Endpoint Mapping Asynchronous

### LINKED ISSUE #209: Async endpoint resolution
Our `endpointResolver` currently returns an `NSURLRequest`, which is cool, but makes working with async APIs like [Heimdall](https://github.com/rheinfabrik/Heimdall.swift) very difficult (by the nature of OAuth, [these requests cannot be synchronous](https://github.com/rheinfabrik/Heimdall.swift/blob/f67abd632b6ccc4604ae8fa2b6eab9d6b7855310/Heimdall/Heimdall.swift#L193).

I'm thinking that the the endpointResolver is passed in a closure that accepts the NSURLRequest as an argument, and the resolver calls this closure when the request is finished. Suggestions?

ashfurrow:
(Relies on the #228 pull request – didn't want the eventual merge conflicts.)

This is just an idea to make the mapping from `Endpoint` to `NSURLRequest` asynchronous. This is useful if you need to refresh network credentials (asynchronously) before being able to generate the `NSURLRequest` with a proper signature. 

As discussed in #209, I'm not _totally_ happy with how the cancellation works here, since it relies on state. If anyone has suggestions for a lightweight way to synchronously return a token that can be used to cancel a request that hasn't yet been created, please chime in :notes: 

Fixes #209.

### DIFF
--- a/Moya/Moya.swift
+++ b/Moya/Moya.swift
@@ -87,10 +87,20 @@ public protocol Cancellable {
 
 /// Internal token that can be used to cancel requests
 struct CancellableToken: Cancellable {
-    let cancelAction: () -> ()
+    let cancelAction: () -> Void
 
     func cancel() {
         cancelAction()
+    }
+}
+
+struct CancellableWrapper: Cancellable {
+    var innerCancellable: CancellableToken? = nil
+
+    private var isCancelled = false
+
+    func cancel() {
+        innerCancellable?.cancel()
     }
 }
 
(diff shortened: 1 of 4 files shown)


Moya/Moya.swift
```diff
@@ -135,15 +145,25 @@ public class MoyaProvider<Target: MoyaTarget> {
...
-        case .Never:
-            return sendRequest(request, completion: completion)
-        default:
-            return stubRequest(request, completion: completion, endpoint: endpoint, stubBehavior: stubBehavior)
+        var cancellableToken = CancellableWrapper()
+
+        let performNetworking = { (request: NSURLRequest) in
+            if cancellableToken.isCancelled { return }
+
+            let stubBehavior = self.stubClosure(token)
```
ashfurrow: Should use `[weak self]` here.

ashfurrow:
OK, no feedback. Gonna merge this into the other feature branch, then resolve conflicts from there.

PR #231, opened 2015-09-22
Custom manager support

rickerbh:
There is an issue where if a custom Alamofire manager is set, it's not actually used for making the requests in Moya - the default Alamofire one is. This pull request features a test to exhibit this behaviour, and a fix to use the manager that's set on the MoyaProvider. Let me know if any further changes are required.


ashfurrow:
Huh, good catch! And a great test! Could you add an entry to the changelog?

ashfurrow:
Added it in, and merged to master. Thanks again!

Issue #220, opened 2015-09-14
Logging HTTP requests


AlexanderKaraberov:
I don't know if this is relevant, but Moya adds another level of abstraction on top of Alamofire which is on top of NSURLSession, so with all this debug logging of all the network stuff can be hard. I think may be we can implement something similar to incredibly useful https://github.com/AFNetworking/AFNetworkActivityLogger in such a way it fits Moya's architecture? IMO it will be very useful. E.g. add enableLogging flag to Endpoint with print() calls or something better?

ashfurrow:
Sounds like a good idea – doesn't sound too difficult, either. We could hook into the existing provider initialization, if that makes sense, or use NSNotifications. Thoughts?

AlexanderKaraberov:
If we add something like `print("\(applicationName): \(date) Request Headers:\(request.request!.allHTTPHeaderFields)")` inside `func request(token: T, completion: MoyaCompletion)` it will be too intrusive. Or you mean smth different in "hook into the existing provider initialization".  So `NSNotification` sounds like a good idea.

ashfurrow:
Aye, so this is what I was thinking:

A user could pass in an endpoint resolver or other callback we already have – the one they pass in would log stuff. So that's how it could work now. With a new library, we could call a function that curries back a function to pass in. 

```swift
Provider<TargetEnum>(endpointResolver: MoyaNetworkLogger.StandardLogger)
... or ...
Provider<TargetEnum>(endpointResolver: MoyaNetworkLogger.StandardLogger { target -> NSURLRequqest in
    return /* our custom request, which the logger will then log before passing onto Moya */
})
```

Make sense?

AlexanderKaraberov:
Great idea. We can go with this solution. What about logging a response?

ashfurrow:
Hmm, good question. There's no existing way to log these at all – we may have to add yet another callback to the provider :cry:

ashfurrow:
What if we added a property to the provider – could be constant or not – that conformed to a `Logger` protocol or something, and it gets callbacks about requests and responses. We could provide a basic logger, and people would be free to implement the protocol and provide their own. Thoughts?

AlexanderKaraberov:
Just to be sure I understood you correctly. We will have a `protocol Logger` with some methods. Also we would have some kind of trait/typeclass (`protocol extension`) with default implementations of the logging methods. And we will use it as some kind of delegate?

ashfurrow:
[...]

PR #235, opened 2015-09-30
Adds http response and requests logging capabilities 

AlexanderKaraberov:
This feature was discussed in the issue [220](https://github.com/Moya/Moya/issues/220). Current PR adds `Logger` protocol with three methods to log response, request and data, their default implementations via `MoyaLogger` and new `MoyaProvider` ivar called `logger`.  Also, I think we should add some info or instructions about logging to the Moya Documentation, although I don't know where to put it :disappointed: .

### DIFF
--- a/Logger.swift
+++ b/Logger.swift
@@ -1,8 +1,3 @@
-
-//  Created by Alexander Karaberov on 9/30/15.
-//
-//
-
 import Foundation
 
 
--- a/Logger.swift
+++ b/Logger.swift
@@ -1,0 +1,17 @@
+
+//  Created by Alexander Karaberov on 9/30/15.
+//
+//
+
+import Foundation
+
+
+public protocol Logger {
+    
+    func logNetworkRequest(request: NSURLRequest) -> Void
+    
+    func logNetworkResponse(response: NSHTTPURLResponse?) -> Void
+    
+    func logNetworkResponseData(data: NSData) -> Void
+    
+}
--- a/Moya/Moya.swift
+++ b/Moya/Moya.swift
@@ -109,15 +109,17 @@ public class MoyaProvider<T: MoyaTarget> {
     public let stubBehavior: MoyaStubbedBehavior
     public let networkActivityClosure: Moya.NetworkActivityClosure?
     public let manager: Manager
+    public let logger: Logger?
 
     /// Initializes a provider.
-    public init(endpointClosure: MoyaEndpointsClosure = MoyaProvider.DefaultEndpointMapping, endpointResolver: MoyaEndpointResolution = MoyaProvider.DefaultEndpointResolution, stubBehavior: MoyaStubbedBehavior = MoyaProvider.NoStubbingBehavior, credentialClosure: MoyaCredentialClosure? = nil, networkActivityClosure: Moya.NetworkActivityClosure? = nil, manager: Manager = Alamofire.Manager.sharedInstance) {
+    public init(endpointClosure: MoyaEndpointsClosure = MoyaProvider.DefaultEndpointMapping, endpointResolver: MoyaEndpointResolution = MoyaProvider.DefaultEndpointResolution, stubBehavior: MoyaStubbedBehavior = MoyaProvider.NoStubbingBehavior, credentialClosure: MoyaCredentialClosure? = nil, networkActivityClosure: Moya.NetworkActivityClosure? = nil, manager: Manager = Alamofire.Manager.sharedInstance, logger: Logger? = nil) {
         self.endpointClosure = endpointClosure
         self.endpointResolver = endpointResolver
         self.credentialClosure = credentialClosure
         self.stubBehavior = stubBehavior
         self.networkActivityClosure = networkActivityClosure
         self.manager = manager
+        self.logger = logger
     }
 
     /// Returns an Endpoint based on the token, method, and parameters by invoking the endpointsClosure.
@@ -175,6 +177,9 @@ private extension MoyaProvider {
         // We need to keep a reference to the closure without a reference to ourself.
         let networkActivityCallback = networkActivityClosure
 
+        //Log network request to the console
+        self.logger?.logNetworkRequest(request)
+        
         var request = manager.request(request)
         
         if let cred = credential {
@@ -185,10 +190,16 @@ private extension MoyaProvider {
 
                 networkActivityCallback?(change: .Ended)
 
+                //Log network response to the console
+                self.logger?.logNetworkResponse(response)
+            
                 // Alamofire always sends the data param as an NSData? type, but we'll
                 // add a check just in case something changes in the future.
                 let statusCode = response?.statusCode
                 if let data = data {
+                    //Log response data
+                    self.logger?.logNetworkResponseData(data)
+                    
                     completion(data: data, statusCode: statusCode, response: response, error: error)
                 } else {
                     completion(data: nil, statusCode: statusCode, response: response, error: error)
--- a/Moya/Moya.swift
+++ b/Moya/Moya.swift
@@ -177,7 +177,7 @@ private extension MoyaProvider {
         // We need to keep a reference to the closure without a reference to ourself.
         let networkActivityCallback = networkActivityClosure
 
-        //Log network request to the console
+        //Print network request 
         self.logger?.logNetworkRequest(request)
         
         var request = manager.request(request)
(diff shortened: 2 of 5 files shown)


ashfurrow:
This looks _great_! Really good work. I've got a few _really_ small pieces of feedback that I'll provide inline. Additionally:

- Can you add a note to the changelog?
- Documentation can go in the `/docs` directory. Add it to the `Providers.md` file (which describes the provider initializer).
- The CI is failing because we really need to address #144. Could you update the `MoyaProvider` subclasses for ReactiveCocoa and RxSwift?

Let me know if I can help with anything – thanks again! :cake:

Logger.swift
```diff
@@ -0,0 +1,12 @@
+import Foundation
+
+
+public protocol Logger {
+    
+    func logNetworkRequest(request: NSURLRequest) -> Void
```
ashfurrow: These probably don't need the explicit `-> Void`, do they? They could use some descriptive comments though.

Moya/Moya.swift
```diff
@@ -175,6 +177,9 @@ private extension MoyaProvider {
         // We need to keep a reference to the closure without a reference to ourself.
         let networkActivityCallback = networkActivityClosure
 
+        //Print network request
```
ashfurrow: Let's not use the word "print", since a logger could do _anything_. A better comment could be:

```swift
// Log network request.
```

(Note the space after the `//` and punctuation – it's _super_ not a big deal, but I'd like to keep it consistent. Goes for the other additions to the file, too.)

Moya/Moya.swift
```diff
@@ -175,6 +177,9 @@ private extension MoyaProvider {
         // We need to keep a reference to the closure without a reference to ourself.
         let networkActivityCallback = networkActivityClosure
 
+        //Print network request 
+        self.logger?.logNetworkRequest(request)
```
ashfurrow: Don't need `self.` in this scope.

MoyaLogger.swift
```diff
@@ -0,0 +1,67 @@
...
+    private var date: String {
+        
+        dateFormatter.dateFormat = dateFormatString
+        dateFormatter.locale = NSLocale(localeIdentifier: "en_US_POSIX")
+        return dateFormatter.stringFromDate(NSDate())
+    }
+    
+    public func logNetworkRequest(request: NSURLRequest) {
+        
+        let requestOutput = String(format: "%@: [%@] Request:  %@", loggerId, date, request.description)
```
ashfurrow: I really like your implementation here. One suggestion: where we are printing out multiple lines, does it make sense to include the same preamble on each line, when it's for the same request? What if we had something like:

```swift
let requestOutput = String(format: "%@: [%@] Request:  %@", loggerId, date, request.description)
print(requestOutput)
```

And then later, had something like:

```swift
let requestHeadersOutput = String(format: "        Request Headers:  %@", loggerId, date, headers)
print(requestHeadersOutput)
```

Does that make sense?

MoyaLogger.swift
```diff
@@ -0,0 +1,67 @@
...
+        print(requestOutput)
+        
+        if let headers = request.allHTTPHeaderFields {
+            let requestHeadersOutput = String(format: "%@ [%@] Request Headers:  %@", loggerId, date, headers)
+            print(requestHeadersOutput)
+        }
+        
+        if let bodyStream = request.HTTPBodyStream {
```
ashfurrow: We _could_ make this struct have a `verbose` property that defaults to `false` in its initializer, which could silence things like the headers and body stream, body stream, and response data. I don't know – I've never written a network logger before :sweat_smile: Thoughts?

ashfurrow:
OK, that's all my feedback. Again, terrific work – thanks!

aschuch:
Looking good. 

One concern that I have: The default logger uses multiple `print` statements to log a single request/response. I guess that multiple simultaneous requests/responses (on different threads) could corrupt the log output in a way that could be misleading to the user...

I would like to suggest to at least accumulate a the log output of each of the `logNetwork...` methods into a single `String` that gets printed at the very end of the method. This way, a single log is somewhat _atomic_.

ashfurrow:
Aye, that's a good idea @aschuch

AlexanderKaraberov:
@aschuch "I would like to suggest to at least accumulate a the log output" - very reasonable, thanks - will reimplement it to `print` everything  after we get either a response or error. Or may be we should use `NSLog`? Because `NSLog()` ensures that all strings passing through it are written in a thread safe manner to `stderr`. Or it will look not good in Swift code?

AlexanderKaraberov:
@ashfurrow  Thanks a lot for feedback! Yeah - there are lots of style errors, so I asked for review. Will fix it and add comments to `protocol` methods.

MoyaLogger.swift
AlexanderKaraberov: Seems reasonable. We can even add several flags (debug/verbose/info) or something like that.

ashfurrow: Yup! I don't think we need to go overboard – people are free to write their own – I'll leave it up to you.

MoyaLogger.swift
AlexanderKaraberov: @ashfurrow So, you want to add date/id preamble at the very beginning? And then just log response, headers, request, etc. after without it? 
It can be useful to get rid of noise in the log output.

ashfurrow:
NSLog is a good question. Honestly I don't know. Anyone have any resources? Blog posts? Surely Mike Ash has something to say on the matter...

Logger.swift
AlexanderKaraberov: @ashfurrow Agree. I think I can add some comments which will serve as a hint for user what he can implement it his own logger.

MoyaLogger.swift
ashfurrow: Yeah, well if we aggregate the log output anyway, as discussed, repeating the preamble in the same string seems superfluous. Again, up to you!

AlexanderKaraberov:
@ashfurrow `"Could you update the MoyaProvider subclasses for ReactiveCocoa and RxSwift?"`. Sorry I'm not sure I understood your comment...
You mean only to add `logger` parameter to initializer in Reactive/Rx Moya extensions? Or add some FRP friendly implementation of the Logger? :smile:

AlexanderKaraberov:
@aschuch `"Anyone have any resources? Blog posts?"` - yes, we need some authoritative opinion here about the thread safety. Also I'm not sure if it is good from the code style viewpoint...

ashfurrow:
Yes, exactly – just the parameters :smile_cat:

aschuch:
I think it is reasonable to accumulate the request log output, then `print` the whole thing. Don't see any reason to use NSLog here. Also, I think its fine to log the request and the response separately (i.e. don't wait for a response to print request + response together).

petester42:
As far as I can tell `print` doesn't write to the device log which `NSLog` does. It might be useful to debug things when you device is not connected. Since its extensible leaving it as just print is fine

ashfurrow:
Yeah, I think `print` is fine :smile: Don't sweat the small stuff, eh?

@AlexanderKaraberov Do you want this in before we release 3.0? It's a matter of adding this feature to an existing release, or forcing users to adopt 3.0 (with breaking changes) to use it.

ashfurrow:
Closing in favour of new work to be done on #238.

AlexanderKaraberov:
@ashfurrow Sorry for my slow response. So, I have to reimplement logger as a `Moya Plugin`?
Where can I find some guidelines how should I do this correctly?

AlexanderKaraberov:
@ashfurrow "Do you want this in before we release 3.0? It's a matter of adding this feature to an existing release, or forcing users to adopt 3.0 (with breaking changes) to use it." If I will understood new Moya Plugin system and reimplement logger I think we can include logger plugin in the 3.0 release.

`P.S` All the info about `Plugins` here: https://github.com/Moya/Moya/pull/238, right?

AlexanderKaraberov:
@ashfurrow Oh. Sorry Again, I'm sooo slow :smile:  (deleted previous comments),
As I found [here](https://github.com/Moya/Moya/pull/240). You have already merged it with `Plugin` reimplementation and code cleaning. Thanks for terrific clean-up work!
Are there smth. left for me in in the logger functionality? Or I can look for other features to implement :smiley:

ashfurrow:
I think the logger is well-covered, but we've got lots of open issues – feel free to dig in to any that look interesting to you!

PR #237, opened 2015-10-02
Plugins

aschuch:
I would like to propose a plugin system for Moya.
`MoyaProvider` can now be initialized with a set of `Plugin`s.  A `Plugin` receives callbacks to perform side effects wherever a request is sent or received.

For example, a plugin may be used to
 * log network requests
 * hide and show a network avtivity indicator
 * inject additional information into a request

This kind of tackles #144 as well, as the number of additional closures is now much lower and can be initialized with an arbitrary number of plugins

I have also added plugins for the existing `CredentialClosure` and `NetworkActivityClosure` to make it easy for everyone to adopt this change. Maybe we should also try to build a plugin from the Logger in #235?

Would love to get some feedback on the idea or how to make this even more generic for everyone to use (I'm not super happy with the subclassing approach). :heart:


aschuch:
Actually, this should be merged into develop.
Continued in #238.

PR #238, opened 2015-10-02
Plugins

### LINKED ISSUE #144: Make it easier to subclass MoyaProvider

aschuch:
I would like to propose a plugin system for Moya.
`MoyaProvider` can now be initialized with a set of `Plugin`s.  A `Plugin` receives callbacks to perform side effects wherever a request is sent or received.

For example, a plugin may be used to
 * log network requests
 * hide and show a network avtivity indicator
 * inject additional information into a request

This kind of tackles #144 as well, as the number of additional closures is now much lower and can be initialized with an arbitrary number of plugins

I have also added plugins for the existing `CredentialClosure` and `NetworkActivityClosure` to make it easy for everyone to adopt this change. Maybe we should also try to build a plugin from the Logger in #235?

Would love to get some feedback on the idea or how to make this even more generic for everyone to use (I'm not super happy with the subclassing approach). :heart: 

Fixes #144.

### DIFF
--- a/Moya/Moya.swift
+++ b/Moya/Moya.swift
@@ -202,9 +202,7 @@ private extension MoyaProvider {
         let plugins = self.plugins
         
         // Give plugins the chance to alter the outgoing request
-        for p in plugins {
-            request = p.willSendRequest(token, request: request)
-        }
+        plugins.forEach { request = $0.willSendRequest(token, request: request) }
         
         // Perform the actual request
         request.response { (_, response: NSHTTPURLResponse?, data: NSData?, error: ErrorType?) -> () in
--- a/Moya/Moya.swift
+++ b/Moya/Moya.swift
@@ -7,15 +7,6 @@ public class Moya {
     /// Closure to be executed when a request has completed.
     public typealias Completion = (data: NSData?, statusCode: Int?, response: NSURLResponse?, error: ErrorType?) -> ()
 
-    /// Network activity change notification type.
-    public enum NetworkActivityChangeType {
-        case Began, Ended
-    }
-
-    /// Network activity change notification closure typealias.
-    /// Explicitly outside the MoyaProvider type since it doesn't touch the Target generic.
-    public typealias NetworkActivityClosure = (change: NetworkActivityChangeType) -> ()
-
     /// Represents an HTTP method.
     public enum Method {
         case GET, POST, PUT, DELETE, OPTIONS, HEAD, PATCH, TRACE, CONNECT
@@ -117,30 +108,27 @@ public class MoyaProvider<Target: MoyaTarget> {
     /// Closure that decides if/how a request should be stubbed.
     public typealias StubClosure = Target -> Moya.StubBehavior
 
-    public typealias CredentialClosure = Target -> NSURLCredential?
-
     public let endpointClosure: EndpointClosure
     public let requestClosure: RequestClosure
     public let stubClosure: StubClosure
-    public let credentialClosure: CredentialClosure?
-
-    public let networkActivityClosure: Moya.NetworkActivityClosure?
     public let manager: Manager
+    
+    /// A list of plugins
+    /// e.g. for logging, network activity indicator or credentials
+    public let plugins: [Plugin<Target>]
 
     /// Initializes a provider.
     public init(endpointClosure: EndpointClosure = MoyaProvider.DefaultEndpointMapping,
         requestClosure: RequestClosure = MoyaProvider.DefaultRequestMapping,
         stubClosure: StubClosure = MoyaProvider.NeverStub,
-        networkActivityClosure: Moya.NetworkActivityClosure? = nil,
-        credentialClosure: CredentialClosure? = nil,
-        manager: Manager = Alamofire.Manager.sharedInstance) {
+        manager: Manager = Alamofire.Manager.sharedInstance,
+        plugins: [Plugin<Target>] = []) {
 
         self.endpointClosure = endpointClosure
         self.requestClosure = requestClosure
         self.stubClosure = stubClosure
-        self.networkActivityClosure = networkActivityClosure
-        self.credentialClosure = credentialClosure
         self.manager = manager
+        self.plugins = plugins
     }
 
     /// Returns an Endpoint based on the token, method, and parameters by invoking the endpointsClosure.
@@ -152,9 +140,6 @@ public class MoyaProvider<Target: MoyaTarget> {
     public func request(token: Target, completion: Moya.Completion) -> Cancellable {
         let endpoint = self.endpoint(token)
         let stubBehavior = self.stubClosure(token)
-
-        let credential = credentialClosure?(token)
-
         var cancellableToken = CancellableWrapper()
 
         let performNetworking = { (request: NSURLRequest) in
@@ -162,9 +147,9 @@ public class MoyaProvider<Target: MoyaTarget> {
 
             switch stubBehavior {
             case .Never:
-                cancellableToken.innerCancellable = self.sendRequest(request, credential: credential, completion: completion)
+                cancellableToken.innerCancellable = self.sendRequest(token, request: request, completion: completion)
             default:
-                cancellableToken.innerCancellable = self.stubRequest(request, completion: completion, endpoint: endpoint, stubBehavior: stubBehavior)
+                cancellableToken.innerCancellable = self.stubRequest(token, request: request, completion: completion, endpoint: endpoint, stubBehavior: stubBehavior)
             }
         }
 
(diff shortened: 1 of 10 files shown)


aloco:
I really like this approach! Since Moya get´s more and more awesomeness, there should be a consistent way in adding new features without having dozens of specific closures (like the already mentioned credentials, logger, network activity). 
It would also reduce the amount of breaking changes if new features were implemented as ```Plugin```s whenever possible.

Moya/Moya.swift
```diff
@@ -212,30 +197,22 @@ public extension MoyaProvider {
...
-        let networkActivityCallback = networkActivityClosure
-
+    func sendRequest(token: Target, request: NSURLRequest, completion: Moya.Completion) -> CancellableToken {
         var request = manager.request(request)
+        let plugins = self.plugins
         
-        if let cred = credential {
-            request = request.authenticate(usingCredential: cred)
+        // Give plugins the chance to alter the outgoing request
+        for p in plugins {
```
ashfurrow: Maybe something like...

```swift
plugins.forEach{ $0.willSendRequest(token, request: request) }
```

Does that make sense?

Moya/Plugins/CredentialsPlugin.swift
```diff
@@ -0,0 +1,31 @@
...
+        if let credentials = credentialsClosure(token) {
+            request.authenticate(usingCredential: credentials)
+        }
+        return request
+    }
+    
+    override func willSendStubbedRequest(token: Target, request: NSURLRequest) {
+        // Just call the closure here to make it possible to test the 
+        // credentials closure when stubbed requests are turned on
+        let _ = credentialsClosure(token)
```
ashfurrow: You can drop the `let` here.

Moya/Plugin.swift
```diff
@@ -0,0 +1,44 @@
...
+    // protocol Plugin {
+    //      typealias T: MoyaTarget
+    //      ...
+    // }
+    //
+    // let plugins = [Plugin]()
+    // 
+    // This does not work, because `plugins` is now unable to infer the actual type of the typealias `T`.
+    
+    func willSendRequest(token: Target, request: Alamofire.Request) -> Alamofire.Request {
```
ashfurrow: Why don't we overload this method with an identical one, except that it doesn't return a request? That way if the function doesn't need to modify the request, it can just omit `return` and let the type inference of Swift handle it?

ashfurrow:
Generally looks great! I was a bit wary of the idea of "plugins", but I really like the implementation. 

I have a few pieces of general feedback:

- I saw the note about avoiding protocols – let's see if we can think of something besides inheritance? Nothing comes to mind, though ...
- Should a plugin be able to distinguish between stubbed requests and real ones? 
- Maybe the plugin functions should accept the `manager` as a parameter? I dunno, probably unnecessary but it feels like something that _should_ happen. Thoughts?
- Could we call them something other than "plugins"? :stuck_out_tongue: 

Thanks again!

Moya/Moya.swift
aschuch: Huh, I put that in for debugging. Good catch!

Moya/Plugin.swift
aschuch: Good point. Alamofire.Request is passed by reference anyway, maybe we can get rid of the return anyway?

In case we want to go forward with the overloading, calling just  `plugins.forEach{ request = $0.willSendRequest(token, request: request) }` would no longer work, since it will always call the function with the return value.

Should we invoke both methods within the `forEach` closure?

Moya/Plugins/CredentialsPlugin.swift
aschuch: :tada:

aschuch:
Thanks for the feedback. I'll push your remarks asap.

* First of, I don't really like the subclassing approach either. I tried a few things with protocols but none of them really seemed to work out. I'd love to discuss alternatives though.
* I decided to give the `Plugin` separate calls for stubbed requests and real ones, because the parameters are a bit different. This could, however, be abstracted into a single method on the plugin that accepts a stubbed/not-stubbed parameter.
* Could easily add the `manager` to the plugin functions, not sure what the plugin would do with it though. Do you have any use cases that come to mind?
* I am not too fond on the term _plugin_ either. Maybe _mixin_? _sideeffect_? I just named it plugin, because everyone is sort of familiar of this term. 

:boom:

ashfurrow:
Ahhh, sorry I meant the `provider` parameter (through which, I guess, the `manager` would be accessible too).

I guess what I'm _really_ saying about the stubbed/not stubbed functions is that I don't think the plugin should even know if the request it's being called with is being stubbed or not. 

Nomenclature is hard, let's give it a think :)

Moya/Plugin.swift
ashfurrow: I hadn't thought of that, actually. Good point. If the request can be modified in-place, may as well get rid of the return value.

ashfurrow:
Also, I agree with using a logging plugin for this – it looks like we're on board to go forward with this solution, so let's I'm going to close #235. Happy to recreate the logger from there as a plugin, if I could help.

ashfurrow:
And don't forget about the changelog! :poodle:

esttorhe:
Wow :scream: 
Great idea and implementation @aschuch :bow: :ok_hand:

PiXeL16:
Great!

aschuch:
Ok, so I've removed the duplication of methods for stubbed/real requests and added the `provider` to the callbacks.

I have also put together a version [on a different branch](https://github.com/aschuch/Moya/commit/34c10bb4644868bc38a64cc84486549cfc556135) that shows how `Plugin` can be implemented in terms of a protocol. This requires a `PluginContainer` wrapper in order to keep a reference within `MoyaProvider` and satisfy the type system. 

I also tried to make  `PluginContainer` conform to `ArrayLiteralConvertible` to keep the API clean, but could not figure out a way to accept the type constraints.

```swift
extension PluginContainer: ArrayLiteralConvertible {
    init<P: Plugin where P.T == T>(arrayLiteral elements: P...) {
        willSendRequestClosures = elements.map { $0.willSendRequest }
        didReceiveResponseClosures = elements.map { $0.didReceiveResponse }
    }
}
```

aschuch:
Any opinions towards keeping the subclass vs. the protocol approach?

ashfurrow:
Leaning towards protocols, personally. Will have time to review shortly.

ashfurrow:
Tricky problem – I think the root of the issue is that we want the plugins to both be able to be used with any type of `MoyaTarget`, but refer to the specific type that's conforming to that target. 

What I mean is, I'm not sure we need to use a typealias'd protocol or a generic class. What about the following?

```swift
public protocol Plugin {

    func willSendRequest<Target: MoyaTarget>(request: Alamofire.Request, provider: MoyaProvider<Target>, token: Target) -> Alamofire.Request
    
    func didReceiveResponse<Target: MoyaTarget>(data: NSData?, statusCode: Int?, response: NSURLResponse?, error: ErrorType?, provider: MoyaProvider<Target>, token: Target)

}
```

Even if someone wanted a plugin to support a _specific_ target type, they should be able to do that:

```swift
class ExamplePlugin: Plugin {

    func willSendRequest<Target: MoyaTarget>(request: Alamofire.Request, provider: MoyaProvider<Target>, token: Target) -> Alamofire.Request {
        return request
    }

    func willSendRequest(request: Alamofire.Request, provider: MoyaProvider<HTTPBin>, token: HTTPBin) -> Alamofire.Request {
        return request
    }

    func didReceiveResponse<Target: MoyaTarget>(data: NSData?, statusCode: Int?, response: NSURLResponse?, error: ErrorType?, provider: MoyaProvider<Target>, token: Target) {

    }
}
```

@aschuch I'll send a PR over to your fork – looking forward to hearing what you think.

ashfurrow:
Hmm, wait that might not work. Anyway, I'll play with it.

ashfurrow:
Nah, this is silly. You have a working solution, let's merge this PR and keep the idea of moving to a non-inheritance approach for another release.

aschuch:
Awesome. Thanks for your input everyone. :heart:

Issue #236, opened 2015-10-01
Release 3.0.0


ashfurrow:
Once #228 is merged.

Unless there's anything showstopping we need in? I was hoping to get #144 in, but haven't got a plan on that yet :sweat:

ashfurrow:
Waiting (possibly) on #144.

aschuch:
I'd like to introduce #238 in 3.0.0 as well. :tada:

ashfurrow:
Agreed. 

--
Ash Furrow
iOS Developer, Author
http://ashfurrow.com

ashfurrow:
Just gonna tidy up the docs around plugins, and we should be good to go on this.

ashfurrow:
Fixed in https://github.com/Moya/Moya/commit/d61003dd3c5e4bea18d69622a9038daa0aae7064 and pushed to CocoaPods trunk: https://raw.githubusercontent.com/CocoaPods/Specs/9186a2cd0f90f5cc20fa56d8a01f93bdeb1480a0/Specs/Moya/3.0.0/Moya.podspec.json

PR #240, opened 2015-10-03
A few other renames, documentation updates

### LINKED ISSUE #200: Improve Endpoint Documentation
`Failure` is truly a network failure – the status code is optional, even. It's meant to simulate when the device cannot connect to the internet. The `Success` case is a "success" in that the API returns some data. it's agnostic to the idea that a 404 would be a failure – it succeeded because it returned data.

### LINKED ISSUE #220: Logging HTTP requests

ashfurrow:
This does a few things.

- All sample responses are now evaluated lazily, obviating the need for a `Closure` case in the sample response enum.
- The `EndpointSampleResponse` has been simplified. No one really understood what the difference between `Success` and `Error` was (looking at the unit tests, it doesn't look like _I_ even understood the difference).
- Adds documentation for authentication (HTTP and OAuth).
- Adds logging plugin based on @AlexanderKaraberov's work in #235).

Generally I really like how 3.0.0 has been coming along. Thanks to a lot of work from a lot of different contributors, Moya has gotten simpler and less complex, while also far more powerful. 

After this I thing we should go ahead and release 3.0.0 (see #236).

Fixes #200.
Fixes #220.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,10 +24,12 @@
 - Removes inflight request tracking for ReactiveCocoa and RxSwift providers. **Breaking Change**
 - Adds support for ReactiveCocoa 4 by moving `ReactiveCocoaMoyaProvider` to use `SignalProducer` instead of `RACSignal`
 - Renamed `EndpointSampleResponse` cases: **Breaking Change**
-  - `Success` to `NetworkResponse`
+  - `Success` to `NetworkResponse`, now contains `NSData` instead of `() -> NSData`.
   - `Error` to `NetworkError`
   - Additionally, `NetworkError` no longer has a status code or data associated with it. This represents an error from the underlying iOS network stack, like an inability to connect. See [#200](https://github.com/Moya/Moya/issues/200) for more details.
-- ReactiveCocoa provider no longer replaces errors that contain status codes (an unlikely situation) with its own errors. It passes all errors directly through. 
+  - Also additionally, removed `Closure` case (see below).
+- Changed `Endpoint` to use a `sampleResponseClosure` instead of a `sampleResponse`, making all sample responses lazily executed. **Breaking Change**
+- ReactiveCocoa provider no longer replaces errors that contain status codes (an unlikely situation) with its own errors. It passes all errors directly through.
 
 # 2.4.1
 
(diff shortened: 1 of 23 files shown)


ashfurrow:
Hey @aschuch – you mentioned we can do in-place modification of the `Request` objects, right? I'm thinking that may be better than having the plugin functions return their own request – less overhead. Thoughts?

aschuch:
Yup. Since `Alamofire.Request` is implemented as a class, we could simply modify the request object in place. There is no real need to return it, just illustrates the intent of the method and would be future proof in case `Alamofire.Request` might be changed to a struct or something.

I still think we could get rid of it now, can always change it back later...

Your implementation of the logger is pretty awesome as well :tada:

ashfurrow:
OK cool. @aschuch I've pushed a change for that and would like your go-ahead: https://github.com/Moya/Moya/commit/55d8ed9eef1be7011247174a54a3ce77cb37c0df

The logger was a lot easier after @AlexanderKaraberov did all the work already 😅

One more thing – I don't care for having to include Alamofire in the plugins. One of Moya's big selling points is that you don't have to care about Alamofire. We're leaking that up through our `manager` property and now the plugins. It's not a _huge_ deal, but I'd love to nip this in the bud. 

What if we created a new protocol in `Moya.swift` that included things in `Request` that you might want to do – authentication, for example – and then made `Request` conform to the protocol in an empty extension. The nice thing is since no one else uses the protocol, we can add more functions to it later if `Request` adds them, or if we forgot to add them. So expanding the protocol wouldn't be a breaking change. Thoughts?

ashfurrow:
Yeah, I like that a lot actually. Work is done in https://github.com/Moya/Moya/commit/d68c03e276a13f58c4ce3e185ca97042bcb5429b if anyone wants to discuss that specifically.

aschuch:
Looking good. 
I like the abstraction of the Request a lot. Much cleaner now.

AlexanderKaraberov:
"Your implementation of the logger is pretty awesome as well". Completely agree :+1: , my code was a rather poor implementation.
@ashfurrow  What are the next plans for 3.0+ release. I use `Moya` intensively in my pet and job Swift projects, so I'm interested about the future updates.

ashfurrow:
@AlexanderKaraberov Yeah, just released 3.0! We'll be continuing to fix bugs that are reported, and implement features that are requested as they come up. If you have any ideas or suggestions, just open an issue :smile_cat:

Issue #177, opened 2015-08-05
RxSwift example in README


ashfurrow:
To match the ReactiveCocoa one.

retsohuang:
I'm using RxSwift + Moya in my project, Is the sample code like this  would help?
```
      provider.request(.UserProfile("ashfurrow")).subscribe(next: {
          image = UIImage(data: $0)
        }, error: {
          println($0)
      })
```

ashfurrow:
Yeah, exactly! We'd need to unindent the code and add Swift code highlighting (see the existing README samples' markdown). :cake:

argentounce:
Can you help me setup a Provider with RxSwift? How a simple RxMoyaProvider should look? How the closure should look?

It would be good that the demo includes examples for both RAC and RX. 

Currently I got my way with the Artsy implementation but im having problems declaring the provider. I know there is a branch about Rx Docs, but this probably be some kinda Faqsy.

ashfurrow:
Hey @argentounce – good question!

The RxMoyaProvider is created using the _exact same_ method as a regular `MoyaProvider`. You just say:

```swift
let provider = RxMoyaProvider<MyAPI>()
```

And that's it! All the examples in the docs that show how to configure the provider apply identically to the RxSwift and RAC providers. Demos for those projects are a great idea :100:

argentounce:
@ashfurrow Thanks! working now :yum:

PR #243, opened 2015-10-07
Updates reactive extension documentation

petester42:
Updates the documentation to reflect the 3.0.0 changes to the reactive providers. Addresses #227 and also updates the documentation to ReactiveCocoa 4. Possibly both could ReactiveCocoa and RxSwift documentation could be combined since they are very similar but I left it separate since most likely a user will be looking for one or the other and not both.

It also updates the examples in the main readme and adds an RxSwift one (#177).

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,18 +115,47 @@ parameter encoding.
 
 For examples, see the [documentation](docs/).
 
-ReactiveCocoa Extensions
-------------------------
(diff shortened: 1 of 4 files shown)


orta:
This is great work!

ashfurrow:
Fab!

PR #247, opened 2015-10-07
Closes #245 Update Endpoints.md to fix endpointResolver typo

### LINKED ISSUE #245: Support for setting body of request without passing parameters
So I've dug into this project and maybe I missed it but is there a way to set the body of the request without sending parameters. I took a look at Artsy's app and they basically have a lot of boilerplate like this. 

```
  case CreateUser(let email, let password,let phone,let postCode, let name):
            return [
                "email": email, "password": password,
                "phone": phone, "name": name,
                "location": [ "postal_code": postCode ]
            ]
```

So what I'm assuming here. And perhaps I'm wrong is that the parameters are converted to 'x-www-form-urlencoded' if the request is a .POST? Does this happen automatically or am I missing something?

Perhaps there is a way to extend this do allow sending an object and passing some sort of request encoding closure to Moya. 

I don't think the current setup is bad for small objects but imagine a complex JSON object body with nested array objects.

gregpardo:
Also make wording a bit more clear by discussing the requestClosure first and then then how endpointClosure should not be used for the same purpose after.

### DIFF
--- a/docs/Endpoints.md
+++ b/docs/Endpoints.md
@@ -137,10 +137,7 @@ let requestClosure = { (endpoint: Endpoint<GitHub>, done: NSURLRequest -> Void) 
 provider = MoyaProvider<GitHub>(requestClosure: requestClosure)
 ```
 
-Note that the `endpointResolver` is *not* intended to be used for any sort of 
-application-level mapping. This closure is really about modifying properties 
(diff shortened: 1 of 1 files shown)


ashfurrow:
Looks great, thanks!

Issue #246, opened 2015-10-07
Carthage installs require explicit dependency listing


JRHeaton:
Hi there. I've noticed that this library is set up to resolve its own dependencies via CocoaPods, but shares the Pod project's framework targets to be built as shared so that Carthage can see them. So far so good.

The "issue" is that when installing Moya via Carthage, if you were to simply list Moya itself as a dependency, it will not fetch for example, Alamofire, which is a mandatory dependency of Moya.

Thus, our app that uses Moya must have the following entries in the Cartfile (since we also use ReactiveMoya):
```
github "Alamofire/Alamofire" ~> 2.0.2
github "Moya/Moya"	 ~> 2.4.1
github "ReactiveCocoa/ReactiveCocoa" "swift2"
```

I think this can _possibly_ be remedied by including a Cartfile for Moya in the root of the repository that lists the dependencies. It won't need to be used obviously during development of Moya, but will provide a way for Carthage to know which deps to download + build. At the very least, I feel like this should be detailed in the README so that people don't have to debug this.

gregpardo:
Erm realized I should have used ~> 2.0 on that instead... Grr

JRHeaton:
I've also realized another fault due to Moya's project dependencies coming from the podspec. We are attempting to install Moya via Carthage with yesterday's Xcode beta, and using the `swift2` branch of ReactiveCocoa. Moya builds its own version of RAC that comes from its pod dependency on it. Because of this, and since it's an older version of RAC than we actually depend on, Carthage fails to build Moya's RAC w/ the beta command line tools.

gregpardo:
I remember seeing talk about this in the Carthage issues tracker. If I remember correctly there is nothing like subspecs in Carthage so essentially ReactiveMoya and RxMoya need to be separated into their own repositories/frameworks I believe.

JRHeaton:
Just brainstorming possible solutions here.

RXMoya, ReactiveMoya, and Moya could each have their own (shared) framework target inside of a dedicated xcodeproj (allowing carthage to build them), and then your subspecs could be kept to point to the same files still for people installing through pods. The demo project could install the library via either method (or both, with two targets).

orta:
We've gone through all this before, and eventually found that this was the least-bad ways to support Carthage users without placing an unreasonable amount of extra work on people maintaining and improving Moya. We're not moving to separate repos and multiple xcodeprojects to keep up to date for a a file or two per sub-project.

The alamofire PR looks like a good idea

[...]

PR #248, opened 2015-10-07
#246 Add Cartfile and separate sections for cocoapods and carthage in README.md

gregpardo:
Giving this little one a shot too.

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,0 +1,1 @@
+github "Alamofire/Alamofire" ~> 2.0.2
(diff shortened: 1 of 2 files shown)


orta:
:+1:

aschuch:
Shouldn't the Cartfile specify `~> 2.0` instead of `~> 2.0.2`? Not sure about the exact implications of this, but pinning to a major version seems to be the better choice here. Opinions?

gregpardo:
Yes I mentioned that [here](https://github.com/Moya/Moya/issues/246#issuecomment-146257866). I can make another pull request if needed unless someone wants to push a quick fix.

aschuch:
Go for it :)

PR #249, opened 2015-10-08
mapImage uses implicitly unwrapped optional.

ashfurrow:
A small change, not breaking. Adds compatibility with RxCocoa's `rx_image` property on `UIImageView`.


orta:
Cool, this makes sense. :+1:

PR #250, opened 2015-10-08
Fix for pull request #248. Use only major version in Cartfile

gregpardo:


esttorhe:
This looks OK to me; not too «savy» on `Carthage`; but not providing the last part of the `semver` will still work and will pick up minor releases?

aschuch:
In my experience with Carthage, the last version (`2.0.2`) is picked up.

esttorhe:
Thanks @aschuch.

I think it won't hurt to merge this change.

Thanks

PR #255, opened 2015-10-19
Update the Demo project for better Carthage compatibility

bojan:
The previous Demo project was built using an older version of CocoaPods that caused incompatibilities when using a Carthage build of Moya.

What happened was that Alamofire in the Demo project was built with a `DYLIB_COMPATIBILITY_VERSION` set by CocoaPods (3), instead of the original value of Alamofire (1).
Later when Moya was built and included into a Carthage project that also used Alamofire, the linker would complain that Moya was trying to load an Alamofire version that is not compatible.

This fix was suggested by this specific response in the CocoaPods repository https://github.com/CocoaPods/CocoaPods/issues/4308#issuecomment-145479575.


ashfurrow:
Fantastic, thanks a lot @bojan!

PR #261, opened 2015-10-22
Fixes issues with Carthage

bojan:
This is a rebuilt Demo and Pods project.

Somehow, the previous changes to the Pods project got reverted.
Also, there were some files missing in the RxSwift pod sources, which made the project uncompilable.

The Gemfile has also been updated since it was locked to a CocoaPods beta.


orta:
:+1: This looks good

PR #263, opened 2015-10-26
Update to latest ReactiveCocoa alpha (Xcode 7.1/Swift 2.1 only)

mhuusko5:
As the title says.

One note:
The podspec entry will have to remain as `"ReactiveCocoa", "= 4.0.0-alpha-3"` vs `"ReactiveCocoa", "~> 4.0.0-alpha-3"` until/unless https://github.com/CocoaPods/Specs/pull/13588 is merged. Not a big deal in this case.


petester42:
Could you separate the update to Xcode 7.1/Swift 2.1 from the OS X support?

mhuusko5:
Sorted.

CI is failing but for reasons unknown to me. Something to do with your guys' wacky multi-podpsec setup. ;)

aschuch:
I am getting 
```
A shell task failed with exit code 66:
xcodebuild: error: Scheme Moya is not currently configured for the build action.
```

when running `carthage update --no-use-binaries --platform iOS`

Moya.podspec
```diff
@@ -31,7 +31,7 @@ Pod::Spec.new do |s|
   s.subspec "ReactiveCocoa" do |ss|
     ss.source_files = "Moya/ReactiveCocoa/*.swift"
     ss.dependency "Moya/ReactiveCore"
-    ss.dependency "ReactiveCocoa", "~> 4.0-alpha.1"
+    ss.dependency "ReactiveCocoa", "= 4.0.0-alpha-3"
```
ashfurrow: I don't think the pod file needs to specify a specific version – using the `~> 4.0-alpha.1` syntax will let people pick whichever alpha they want. I think this specifier should be added to the Demo project's Podfile, though.

```rb
pod 'ReactiveCocoa', '~> 4.0-alpha.1'
```

mhuusko5: Unfortunately theres significant breaking changes in alpha 2 and 3 (which this pull request has changes/fixes for.. you just have to scroll a lot).
So, the point of this pull request is to support the latest alpha, with the expectation that this would be pushed as a 4.1.0 release.

ashfurrow: Ah, I see. Well, I'd still like to avoid binding to a specific version. Can we still use `"~> 4.0.0-alpha.3"` here, to accommodate for a 4th alpha?

Then again, if the 4th alpha itself has breaking changes... It might make sense to keep this as you have it. Suggestions, @orta?

orta: I don't think `~> 4.0.0-alpha.3` will do what you want, I think it will just equals. Yeah, probably have to do the exact version, given that they've broke API compat each time, seems logical to presume that in the future.

ashfurrow: Cool.

ashfurrow:
Cool, so we just need CI + merge from master + changelog update.

Moya.podspec
mhuusko5: Two things..
1) as you say, its possible/likely 4th alpha will have breaking changes
2) I tested `~> 4.0.0-alpha-3` and it resolves to `4.0.4-alpha-1`.. I guess CP prioritises the version number and ignore the alpha bit. as mentioned above, have opened https://github.com/CocoaPods/Specs/pull/13588 to change `4.0.0-alpha-1` to the contents of `4.0.4-alpha-1`, and kill `4.0.(1/2/3/4)-alpha-1`

ashfurrow:
Looks like ReactiveCocoa was updated in https://github.com/Moya/Moya/commit/535d7a1b03673f9097f89686b2845cc7685cbba8#diff-51d740005a289b6c12f9ba0fe81dc8db – does that sound right?

mhuusko5:
Yeah I guess @JRHeaton has made these changes since, and theyre on master. So I guess we can close this?

ashfurrow:
Cool, thanks for checking!

PR #264, opened 2015-10-26
updates reactive docs to include example

petester42:
Due to the talk in #177 I've added documentation on how to create the reactive providers. Should wrap up the reactive documentation.


ashfurrow:
:clap:

PR #267, opened 2015-10-26
Better compatibility with Carthage; changes for RAC v4 alpha 2+

JRHeaton:
Hi there. I really like Moya, yet it's currently not set up ideally for people wanting to install via Carthage, or even manually integrate with target dependencies. I've updated the project in a few ways to make things better for those people, and not affect the situation with CocoaPods. My motivation for doing so is that it's currently impossible to install [Mantle](https://github.com/Mantle/Mantle) and the latest ReactiveCocoa at the same time via CocoaPods. I'm attempting to build a framework that depends on Mantle, RAC, and [Reactive]Moya.

### Carthage Changes
- An Xcode project has been made with 3 shared iOS framework targets: `Moya`, `ReactiveMoya`, and `RxMoya`. 
- Each of the 3 targets has been set up to link against built `RAC`/`RxSwift`/`Alamofire` products in `Carthage/Build/iOS`, which will happen magically when installing Moya to another project, and can also be done locally for development using `carthage update --use-submodules --platform iOS` to install and build them.
- A proper `Cartfile` for installing deps either locally for development or through another project's `Cartfile`. This now includes all of the dependencies that are needed to build each of the 3 Moya framework targets.
- Now that a Carthage install of Moya doesn't need `Pods.xcodeproj` to work anymore (the symlink `_Carthage.xcodeproj` has been removed), there is no need for the pods-installed RAC to build to build Moya. This removes that redundancy and _saves time_.

### CocoaPods changes
For the most part, none. The `Demo` project remains in-tact and the same, still hosting the unit tests for this library and being set up ahead of time with `pod install` for CI and such.
- Updated RAC dependency to the latest version released.
- It made sense to move the shared source code of Moya to `Source/` instead of `Moya/`.

### What does this mean for maintenance?
- Source file references need kept up-to-date in the Xcode project with the framework targets.
- Align the dependency versions of `RAC`/`RxSwift`/`Alamofire` in the `.podspec` the `Cartfile`.

### In simple terms
- CocoaPods users of Moya will see no changes, other than a newer RAC version. 
- Carthage users of Moya now have the lovely advantage of Carthage itself resolving the deps of Moya.
- It's not much extra work for authors :)

### Tests
Unit tests are currently still in the Demo project as a separate target as before.

Also, I'm unsure about it for this project, but `Cartfile.resolved` may be unnecessary to track.


ashfurrow:
Hey there! Gotcha – a few things. You'll need to pull in changes from master (tests there [are currently :green_apple:](https://travis-ci.org/Moya/Moya)). If Carthage no longer relies on the Demo Xcode project at all, then we can stop checking in `Demo/Pods` altogether, and also stop duplicating changes in our pod specs for RxMoya and ReactiveMoya. 

At first glance, this change looks good to me :+1: We've been keen to avoid splitting in two, because we had previously tried splitting up the Moya _repo_, which didn't work. Keeping it all here would be fine imo. Let's see if anyone has any comments.

JRHeaton:
Cool, I've made the changes you mentioned. What purpose do/did the separate Rx/RAC Moya podspecs have? Also, what would be the correct company identifier to use for these targets' bundle identifiers?

ashfurrow:
Oh, hmm. Current ones are `com.ashfurrow.Demo` – not really appropriate anymore. Are the important in Carthage? If we just need one, `com.example` is fine.

The podspecs provide Carthage-compatible build targets in the Xcode project, but also complicate unit tests. I don't believe their needed after this PR – getting rid of them would be :100:

JRHeaton:
Heh, you commented as I updated that post. The identifier will be set as the bundle identifier of the built product. I'll go ahead and remove the podspecs from this branch then :)

petester42:
I like framework targets but I have some concerns with this move. I find that it might be hard to keep everything up to date with what is proposed here and might result in duplicate work. 

What happens when someone wants to contribute to the project? They change a file and then to make sure everything works they have to do `carthage bootstrap` to get those dependencies, make sure everything builds fine without error in the framework project, the run `pod install`, then run the tests in the demo project.

To me seems like a lot more work to make sure things are working fine. I think we should pick either having a project with framework targets or the current setup as the source of truth for the library. This would be only the place people would need go to contribute to the library (build framework, run tests, etc.). 

What does everyone think?

JRHeaton:
Well, there _really_ isn't much more work to do. Making sure the targets reference the right files is a matter of seconds of work, and only on the occasion that you actually add or remove a file. How often is that, and is the extra 15 seconds to open the project and drag in the file truly "a lot of work?" Definitely not any "duplicate" work that I can think of. The podspec in most cases should automatically work with file changes if the patterns still match said files. 

Also, think of the case where the collaborator is choosing to do his local development through the targets project, rather than the Demo project w/ a `pod install`. They are already adding/removing the files as they go, there is no additional work to do.

As for testing, I've offered to update the `.travis.yml` to make sure the targets all build as well, which will catch any slips on PRs if that were to ever happen. There really shouldn't be any reason that the library would fail to build in one configuration and not in the other, unless there is (as there has been) issues with any of the podspecs. It's still the same core source being compiled and linked against the same dependencies, it's just a matter of which package manager is grabbing them for us. 

I'm trying really hard to convince you, because I honestly believe this is a good change (and lately, it's been _necessary_ for my cause, since GitHub libs + CocoaPods don't play nicely, and the current setup was yielding issues with Carthage installs). I encourage you to actually pull down this branch if you get some time and check out the changes. I think you will see that it's not a great chore to ensure things are working fine.

aschuch:
I've just tried, works pretty well. Good work, @JRHeaton!

ashfurrow:
Yea I like this. Maybe could use some documentation? Time for a Contributing.md file?

aschuch:
Also I think that adding a CI step to verify carthage is a good idea. Does anyone know if its possible to run 2 CI steps, one for unit tests one for CP/Carthage install testing?

petester42:
I think a contributing.md is a good idea.

ashfurrow:
It’s possible to override Travis’ script command and do both. 

-- 
Ash Furrow
Sent with Airmail

On October 27, 2015 at 12:39:36 PM, Alexander Schuch (notifications@github.com) wrote:

Also I think that adding a CI step to verify carthage is a good idea. Does anyone know if its possible to run 2 CI steps, one for unit tests one for CP/Carthage install testing?

JRHeaton:
So as it turns out, the two extra podspecs may be required after all. With Rx and RAC variants only available now as subspecs of Moya, there isn't a way to install Moya/RxMoya/ReactiveMoya as separate modules into the Demo app. Trying to install both subspecs will build both RxMoya and ReactiveMoya sources under the 'Moya' pods project target, which will have conflicts (for example, the public `Image` typealias that exists in both source bases).  So is the best solution just to keep them?

ashfurrow:
This is why they were originally added – Carthage does not have any concept of a subspec, and the maintainers refuse to accept one to encourage smaller libraries. To get rid of them, we would need multiple Xcode projects for Carthage? If we need to keep them, what's the concrete difference with and without this PR to someone using Carthage?

This is, like, my least favourite part of software development. I'm sure others feel the same. Thanks for the hard work everyone – don't forget that this is volunteering, not your job. Don't burn out! If you need to step back or take a break, do it.

JRHeaton:
Heh, except I'm trying to use Moya at my job ;)

I think you misunderstood my last comment. This isn't an issue with Carthage, it's an issue with the Demo project and its test target needing to test all 3 modules, and _without_ separate podspecs, there's no way to install 2 different subspecs as separate targets. For example:
```ruby
pod 'Moya', :path => "../"
pod 'Moya/ReactiveCocoa', :path => "../"
pod 'Moya/RxSwift', :path => "../"
```

Now the sources of all three of those are trying to be built into one framework target by pods, which doesn't work.

There are a lot of refinements for Carthage on this PR, the most important one is that now dependencies of Moya are resolved _through Carthage_, and it isn't relying on an underlying pod install to be set up properly. Before, even with the dependencies listed in the Cartfile on master, that means those projects are having to be built twice. Once in the pods project, once in Carthage. This not only takes time, but lately it's been a 100% dead stop for my team trying to use Moya since the RAC podspec was broken for a while, and now Mantle's is as well, hence our usage of Carthage. I guess you could say this PR enables one to unhook from needing any work done from pods when you're using Carthage.

Regarding CI, I have been trying to think of a reasonable way to test Carthage stuff on Travis. Since iOS framework targets require code signing though, and some of our dependencies, like Alamofire, leave the default identity value as "iPhone Developer", Carthage skips building their targets, which is obviously an issue, as _our_ targets depend on them. Some of them do properly build for the "iphonesimulator" sdk, and I'm not sure how they set up their projects to do that, but I will investigate it.

ashfurrow:
Reading the rest of your comment, though the following with CocoaPods:

```rb
pod 'Moya', :path => "../"
pod 'Moya/ReactiveCocoa', :path => "../"
pod 'Moya/RxSwift', :path => "../"
```

Works fine.

ashfurrow:
Not clear on the need to use Carthage, but that's fine. I realize Mantle complicates things. 

Yeah, we know that the tests are using different frameworks. In an ideal, CocoaPods world, we would build _one_ Moya framework with all the subspecs. But we need to accommodate Carthage, so we instead have these extra podspecs for them. It leads to duplicated code in the tests because there are several different `MoyaTarget` protocols depending on which framework you've imported. 

This conversation is getting pretty lengthy – I'm not even really sure what we're trying to accomplish anymore :confounded: Would hopping on a Google Hangout help?

JRHeaton:
> In an ideal, CocoaPods world, we would build one Moya framework with all the subspecs.

What i'm saying is that this is not possible currently, because code from RxMoya conflicts with code from ReactiveMoya. 
For a concrete example, the definition of `Image` in both of these files:
- https://github.com/Moya/Moya/blob/master/Moya/RxSwift/Observable%2BMoya.swift#L5
- https://github.com/Moya/Moya/blob/master/Moya/ReactiveCocoa/SignalProducer%2BMoya.swift#L5

Basically, compiling all of the Moya source code in one build target will not work without some source modifications, so I put the separate podspecs back. That work, including reducing down unnecessary tests, can be done on a separate PR.

The goal is still the same for this PR: restructure things a bit so that the project "natively" supports Pods _and_ Carthage.  That part is done. Can we come up with a concrete list of things left to do that are required to consider this PR ready? I'm happy to do the work, I just need to know what's actually required. Otherwise, if you feel there's more to discuss, and perhaps too much for here, I'm not opposed to a call.

ashfurrow:
OK, cool. I think this PR is ready for final review.

One side-note:

> compiling all of the Moya source code in one build target will not work without some source modifications

This `typealias` is currently the only duplicated symbol – it should be moved into the ReactiveCore group and defined _once_. Other than that, you could use all subspecs in the same CocoaPods project.

ashfurrow:
Looked through the [raw diff](https://patch-diff.githubusercontent.com/raw/Moya/Moya/pull/267.diff) – all looks :+1:

ashfurrow:
OK, so we need contributing docs, deduplicated `Image` typeliases, and see if we need the extra podspes.

- https://github.com/Moya/Moya/issues/270
- https://github.com/Moya/Moya/issues/269
- https://github.com/Moya/Moya/issues/271

Anything I'm missing @JRHeaton ?

JRHeaton:
Looks correct to me :+1:. Thanks for going ahead and making issues for those.

PR #277, opened 2015-10-31
Allow specialization of request parameters at request time

JRHeaton:
Hi there. I'm proposing this small change to Moya after about a week of real world usage with it. In short, this PR lets one apply additional request parameters at time of calling `request()`, allowing more dynamic requests to be made without necessarily needing new (and special) targets. They are merged into the result of the endpoint closure's parameters.

### Example of why this is nice
So I will provide a current situation of mine as an example. I am working with an API that has a lot of optional parameters that apply to (mostly) any request. Rather than having target enum cases that all have an extra _n_ number of optional associated values that can be provided via `MoyaTarget.parameters`, it would be great to be able to provide additional params as the caller creating/invoking the request instead.

### API Changes
- A new, "designated" `MoyaProvider.request()` function that takes an optional parameter dictionary.
- A "convenience" `MoyaProvider.request()` function with the pre-existing signature, that calls the designated function with `nil` parameters.

Result: No breaking changes.

### Current alternative
With how the target and endpoint closure are the only things able to set parameters, I can only get this behavior by having a special target target enum case, like so:
```swift
public enum SweetAPI: MoyaTarget {
     ...
    indirect case Parameterized(target: SweetAPI, parameters: [String:AnyObject])

    ...
var parameters: [String:AnyObject]? {
    switch self {
    case let .Parameterized(target, params):
        return target.parameters.flatMap { original in
            original.mergeEntriesFromDictionary(params) // pseudo-code
        }
    default: return nil
    }
}
```
But this is less than ideal as it requires this sort of compositional target case to be present in the target type you're using to be able to.

### DIFF
--- a/Source/Moya.swift
+++ b/Source/Moya.swift
@@ -116,16 +116,20 @@ public class MoyaProvider<Target: MoyaTarget> {
     public func endpoint(token: Target) -> Endpoint<Target> {
         return endpointClosure(token)
     }
-
+    
     /// Designated request-making method. Returns a Cancellable token to cancel the request later.
-    public func request(target: Target, completion: Moya.Completion) -> Cancellable {
-        let endpoint = self.endpoint(target)
+    public func request(target: Target, parameters: [String:AnyObject]?, completion: Moya.Completion) -> Cancellable {
+        var endpoint = self.endpoint(target)
         let stubBehavior = self.stubClosure(target)
         var cancellableToken = CancellableWrapper()
-
+        
+        if case .Some(let parameters) = parameters {
+            endpoint = endpoint.endpointByAddingParameters(parameters)
+        }
+        
         let performNetworking = { (request: NSURLRequest) in
             if cancellableToken.isCancelled { return }
-
+            
             switch stubBehavior {
             case .Never:
                 cancellableToken.innerCancellable = self.sendRequest(target, request: request, completion: completion)
@@ -133,9 +137,9 @@ public class MoyaProvider<Target: MoyaTarget> {
                 cancellableToken.innerCancellable = self.stubRequest(target, request: request, completion: completion, endpoint: endpoint, stubBehavior: stubBehavior)
             }
         }
-
+        
         requestClosure(endpoint, performNetworking)
-
+        
         return cancellableToken
     }
 }
--- a/Source/Moya.swift
+++ b/Source/Moya.swift
@@ -142,6 +142,11 @@ public class MoyaProvider<Target: MoyaTarget> {
         
         return cancellableToken
     }
+
+    /// Convenience request-making method that takes no extra endpoint parameters
+    public func request(target: Target, completion: Moya.Completion) -> Cancellable {
+        return request(target, parameters: nil, completion: completion)
+    }
 }
 
 /// Mark: Defaults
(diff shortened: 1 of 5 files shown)


Source/Moya.swift
```diff
@@ -133,11 +137,16 @@ public class MoyaProvider<Target: MoyaTarget> {
...
-
+        
         requestClosure(endpoint, performNetworking)
-
+        
         return cancellableToken
     }
+
+    /// Convenience request-making method that takes no extra endpoint parameters
+    public func request(target: Target, completion: Moya.Completion) -> Cancellable {
```
ashfurrow: This could be rolled into the existing `request` function with default parameter value.

ashfurrow: But Is suppose that would be a breaking change.

ashfurrow:
Hmm. I understand the motivation for the change, but it does make me very uneasy. 

My goal when originally making Moya was to abstract away _everything_ I could into the `enum`. It's better, so my thinking went, that you must define everything in one spot _once_. This gives developers a quick-glance at what information each target needs to the your API. 

We [used to even have something like this](https://github.com/Moya/Moya/issues/74), but moved `parameters` into the `MoyaTarget` protocol to be more declarative. Really, what would be ideal is `enum` cases with default associated values (something I've heard the Swift team at Apple is working on).

Your approach of `flatMap` and `indirect` cases is really clever! Do you think adding documentation to demonstrate how to use this approach would be a good compromise?

Source/Moya.swift
JRHeaton: True. I just hadn't done that initially to prevent breaking existing code that uses Moya, whose parameter-less calls to `request()` will need updated to explicitly be `request(parameters: nil)` without this bridging function. If you're saying the breaking change is still okay with you, I will change it though.

ashfurrow: Yeah, breaking changes are fine with me as long as our next version is a major release. As long as we avoid braking semantic versioning, I'm ok with bumping our versions however makes sense for us.

JRHeaton:
Indeed, that is the unique beauty of Moya, I agree. It's just really hard to nail 100% of endpoint parameter configuration cases through associated values at the moment. And even with associated value defaults in the future, imagine an API where there are 10 optional parameters for 2/3 of endpoints. If that's 75 of 100, in total there are now 750 associated values you're switching on and checking against nil to decide whether to include them in the parameters calculated property in the enum. This is a lot more code and deters the ability to take a "quick glance."

Indirect target cases like I showed above _can_ work for transforming any property of a `MoyaTarget`, and eliminates the repetitive associated value switching in this scenario. The caller is still providing these parameters now, but the downside is that this functionality is restricted to `MoyaTarget` types that implement such indirect transformation cases (in other words, it's implementation opt-in per-API), which to me feels like a downside. You can't extend enums to add cases so you couldn't add this to an existing target type yourself, either. At best, you could create an intermediate target type, parameterized by the real target type, which applies transformations when it returns its `MoyaTarget` values. (ex: `TransformationTarget<GitHubAPI>.Parameterized(.Zen, myParams)`, but then you need a different provider to send transformed targets. I'd like to hear your thoughts on this if you're willing.

While I didn't see harm necessarily in providing the option for the caller to tweak target parameters _directly_, I do see how it conflicts with Moya's original principals, so I will go ahead and back down and close this. Thank you reviewing the code though and explaining the reasoning, and better potential alternatives. 😇

ashfurrow:
No problem! The issue yore trying to solve is rally interesting. It never occurred to me I think because I haven't worked with APIs that have many optional parameters. 

What do you think about opening an issue to discuss solutions?

JRHeaton:
I think that's a great idea. I'm gonna spend some time this afternoon experimenting with my ideas; I will create the issue later tonight after I've tried out some things and educated myself a bit more. Thanks again!

PR #279, opened 2015-10-31
Removes superfluous observable in RxSwift extension

### LINKED ISSUE #276: Redundant/superfluous SignalProducer in request()?
Howdy. Please forgive me if I am wrong here. I've been working with the Swift version (v4) of ReactiveCocoa for a week or so.

I was reading this code: https://github.com/Moya/Moya/blob/3cd402cfddf2b361b6251de6d24214758ae54511/Source/ReactiveCocoa/Moya+ReactiveCocoa.swift#L20-L47

and it looks a lot like there's a redundancy. The outer producer, when started, starts the inner signal, and seemingly just pipes the events from inner -> outer, and links the inner disposable to dispose when the outer does. The outer producer seems superfluous.

As an experiment, I reduced it down to this:
```swift
public func request(token: Target) -> SignalProducer<MoyaResponse, NSError> {
    
    // Creates a producer that starts a request each time it's started.
    return SignalProducer { [weak self] observer, requestDisposable in
        
        let cancellableToken = self?.request(token) { data, statusCode, response, error in
            if let error = error {
                observer.sendFailed(error as NSError)
            } else {
                if let data = data {
                    observer.sendNext(MoyaResponse(statusCode: statusCode!, data: data, response: response))
                }
                observer.sendCompleted()
            }
        }
        
        requestDisposable.addDisposable {
            // Cancel the request
            cancellableToken?.cancel()
        }
    }
}
```

And it still works as-expected, and unit tests pass. So am I missing a specific reason or behavior of producers that requires it to be the first way in some specific scenario? If there is, please explain if you're willing, I'd love to understand and improve my RAC-fu 👊🤓. Otherwise, I will PR this.

ashfurrow:
Removes some redundant code similar to #278.

Also, I added the changelog to the Xcode project so it's easier to find, and did a `pod install`.

@petester42 would you mind reviewing this one and merging when you're comfortable with it?

Fixes #276.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Next
 
 - Removed extraneous `SignalProducer` from ReactiveCocoa extension – @JRHeaton
+- Removed extraneous `deferred()` from RxSwift extension
 
 # 4.1.0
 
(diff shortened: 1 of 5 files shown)


petester42:
I made some formatting changes so that the two type of providers more closely match. If you are ok with this change it should be ready to merge.

ashfurrow:
Cool! I like the matching, but the empty return type is probably unnecessary. We don't specify `-> ()` when creating the obervable, either. Wanna take those out? Otherwise looks :100:

petester42:
yep will do. Im not a fan of those either.

ashfurrow:
Cool, great work!

PR #280, opened 2015-11-01
Don't code sign iOS framework targets built for simulator

JRHeaton:
This is required to build framework targets on a CI server with no code signing identities. I've opened a similar request [on Alamofire](https://github.com/Alamofire/Alamofire/pull/903) as well. RxSwift and Result are already set up like this, and ReactiveCocoa has the change on master, though not yet in a v4 release. Once all of the projects have releases with this change, we could add a step for travis to make sure the framework targets build properly.


ashfurrow:
Seems reasonable, thanks!

PR #286, opened 2015-11-08
Fix snippet sample code.

tomburns:
I'm working on wrapping my brain around Moya tonight, and noticed that some of the sample code is a little out of date syntax-wise. 

Here's one quick update so that this snippet works with the latest release; I'll try and round up any others I find.


petester42:
Cool. Thanks!

Issue #288, opened 2015-11-08
MoyaError (formerly: ReactiveCocoa Error Type)


tomburns:
I'm just getting started with Moya (and ReactiveCocoa, for that matter), but when using `request()`, `mapJSON()` etc to create a `SignalProducer` I find that I'm having to do a lot of mapping from the many underlying `NSError`s to a typed error enum in my code.

Is this something Moya could take care of by having an intermediate `ErrorType`, or is it best handled in my code? Personally I'd much rather see `request()` return something like `SignalProducer<MoyaResponse,MoyaError>` than `SignalProducer<MoyaResponse,NSError>`, but it's entirely possible I'm just holding it wrong :smile: 

Let me know either way; this seems like it'd make a decent first PR for me to tackle if you like the idea.

tomburns:
I think this would cause the RAC4 support to diverge a bit from the legacy `RACSignal` interface, but I'm not sure that's a bad thing either.

ashfurrow:
Cool, that approach sounds reasonable. @justinmakaila @petester42 do either of you have opinions about it? Whatever is more idiomatic for RAC 4. 

To be honest, I'd be OK with deprecating [`request()`](https://github.com/Moya/Moya/blob/136e3260e03a2d2e01ea6d9604e354ff3f9ed55d/Source/ReactiveCocoa/Moya%2BReactiveCocoa.swift#L39-L41) and shifting the responsibility to call `toRACSignal()` to the user. Thoughts?

petester42:
Im not sure how `RACSignal` would work with an error other than `NSError`. I have never tried it. But other than that everything looks good.

tomburns:
I believe anything that conforms to `ErrorType` can be bridged to an `NSError`, but honestly I don't know enough about the older `RACSignal` ReactiveCocoa interface to guarantee that'd work in this specific case.

ashfurrow:
I think they'll just get an error like they do now, it's just it'll be an `NSError` cast from this new type instead of an `NSError` we construct presently. I like that change, but we'll need to bump Moya up a major version to adhere to [semantic versioning](http://semver.org) :checkered_flag:

justinmakaila:
I think a custom error type is the way to go, because it gets rid of some guessing on the end-user side. I'm all for it

tomburns:
#289 is nearly merged, so I'm moving conversation back here to continue discussing which associated value's we'd like to see on the newly minted `MoyaError` type.

Here's what we have currently:

```
[...]

Issue #259, opened 2015-10-22
CI is Failing Intermittently


ashfurrow:
Not sure what caused it. Looks like a problem with Xcode (go figure).

```
The following build commands failed:
	CompileSwiftSources normal x86_64 com.apple.xcode.tools.swift.compiler
(1 failure)
2015-10-22 00:38:04.847 xcodebuild[769:3065] [MT] IDETestOperationsObserverDebug: (357FE833-30CF-44F9-9DD3-7F464EB8F250) Beginning test session with Xcode 7A220 on target <DVTiPhoneSimulator: 0x7f99c2cdecf0> {
		SimDevice: SimDevice : iPhone 6 (CEBBD79A-B6C1-4A02-BD33-56BBCBDED03E) : state={ Shutdown } deviceType={ SimDeviceType : com.apple.CoreSimulator.SimDeviceType.iPhone-6 } runtime={ SimRuntime : 9.0 (13A340) - com.apple.CoreSimulator.SimRuntime.iOS-9-0 }
} (9.0 (13A340))
2015-10-22 00:38:33.992 xcodebuild[769:4353] Connection peer refused channel request for "dtxproxy:XCTestManager_IDEInterface:XCTestManager_DaemonConnectionInterface"; channel canceled <DTXChannel: 0x7f99c33d1c20>
Testing failed:
	Command /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/swiftc failed with exit code 1
** TEST FAILED **
```

vinnybad:
For what it's worth, I'm seeing the same thing happen on our jenkins instance.

ashfurrow:
Misery loves company :)

ashfurrow:
Hmm, we're actually not seeing this problem anymore. Going to close the issue.

ashfurrow:
Looks like I spoke too soon.

vinnybad:
I had this issue when multiple builds were executing at the same time. Decreasing the number of executors to 1 resolved the issue. Not an ideal solution but this bought us some time to look for other solutions.

SRandazzo:
Little bit of snooping around led me to this issue on Travis-CI:
https://github.com/travis-ci/travis-ci/issues/4725

This appears to be a known bug. There is a potential workaround where you can a `sleep n` to the end of our travis.yml

Thoughts on doing the above hack?

ashfurrow:
Can't hurt!

[...]

Issue #198, opened 2015-08-23
Support ObjectMapper


lkuczborski:
It would be great if Moya supported [ObjectMapper](https://github.com/Hearst-DD/ObjectMapper) usage. It already can be accomplished with pure [Alamofire](https://github.com/Alamofire/Alamofire) with the help of [AlamofireObjectMapper](https://github.com/tristanhimmelman/AlamofireObjectMapper).

colinta:
Why just ObjectMapper?  Shouldn't the user decide what library to use?  This is a "middleware" decision, so I think that a separate `MoyaObjectMapper` library would be appropriate - and it could live in the **github.com/Moya** organization!  But there are plenty of other JSON <-> Object libraries:

- Argo
- ObjectMapper
- SwiftyJSON

The trouble I've run into is that it's impossible to have compile-time checking on these, because the endpoints map to *many* different return types.  We've solved this by having `Service` classes that define per-endpoint closures.  Tedious, but at the application-code level we know we'll get back the objects we asked for.

- Api.recentPosts -> [Post] (`PostService.recentPosts() { (posts, error) in }`)
- Api.post(id: 12) -> Post (`PostService.post(id: 12) { (post, error) in }`)
- Api.currentUser -> User? (`UserService.currentUser() { (user, error) in }`)

How are you using ObjectMapper + Moya right now?  Does it lend itself to becoming a generic library?

orta:
I'm not sure this is Moya's responsibility, object mapping in Swift is still painful and there's no obvious winner to back so we'd end up supporting a lot of different cases. I think @colinta's idea makes sense though, a separate pod that could act as a mapper.

FWIW though we didn't find it difficult at all to build our [own mapper](https://github.com/artsy/eidolon/blob/master/Kiosk/App/Networking/RAC%2BJSONAble.swift) that worked with [function chaining](https://github.com/artsy/eidolon/blob/master/Kiosk/App/AppViewController.swift#L76).

ashfurrow:
I don't think this is Moya's job, either, but Orta's right – we may able to be clever about completion blocks that would allow users to easily do this, even without the chaining. It could be a neat demo, case study, or example in the docs. I'll give some more though to what it might look like.

justinmakaila:
[...]

PR #289, opened 2015-11-08
Create and use MoyaError type for SignalProducers.

tomburns:
I took a blind stab at the change I was talking about in #288, mostly as an exercise for myself. I think this preserves the data presented previously in the NSErrors' userInfo, while providing a more swift-friendly interface for retrieving it.

As I said in the issue, I have no idea what I'm doing, but I'd love to hear your thoughts!


tomburns:
Hmm, tests are passing for me locally after that last commit. Let me know if there's something I can do to make Travis happy :)

ashfurrow:
Travis is failing for Moya intermittently (see: https://github.com/Moya/Moya/issues/259 ). I've restarted the build to see what happens, and I'll take a look at the code now.

Demo/DemoTests/SignalProducer+MoyaSpec.swift
```diff
@@ -183,7 +183,7 @@ class SignalProducerMoyaSpec: QuickSpec {
                 }
                 
                 expect(receivedError).toNot(beNil())
-                expect(receivedError?.domain).to(equal("\(NSCocoaErrorDomain)"))
+//                expect(receivedError?.domain).to(equal("\(NSCocoaErrorDomain)"))
```
tomburns: Forgot to comment earlier: I wasn't clear on a good replacement for this line

ashfurrow:
Yup, green now :smile: :green_apple: 

This looks great, fantastic work :100:

The `MoyaErrorCode` enum exists, but I feel like (given the direction Swift 2 has taken with error-handling) it may be time to retire it altogether and use just your new `MoyaError` enum. This would only change using Moya with ReactiveCocoa (maybe we could make corresponding changes for RxSwift). Any opinions on that? /cc @Moya/contributors 

There're a few tests I'd want to make sure we cover in `SignalProducer+MoyaSpec.swift` but we can sort those out later.

Updating the changeling is a good idea in the mean time – right at the top under "Next". Make sure to mark them with **bold breaking changes** and give yourself credit. Thanks again for the awesome pull request – we'll discuss it for a bit and get your changes into a release hopefully this week. Cheers! :coffee:

Demo/DemoTests/SignalProducer+MoyaSpec.swift
ashfurrow: Hmm, good question. I've not compared enum cases in Nimble matchers before. I looks like [a question others have had](https://github.com/Quick/Nimble/issues/212). If you want, take a stab at one of the solutions @jeffh suggests there – I can lend a hand to get these working, too.

petester42:
Im all for moving to a typed error instead of NSError. We could possibly use the typed error everywhere instead of just the reactive extensions. It would make things more consistent. The same could be done for `MoyaResponse` as I discussed [here](https://github.com/Moya/Moya/issues/198#issuecomment-154258823)

tomburns:
Pushed a simple comparison function for the error type; I wasn't comfortable shoving that type comparison into `Equatable` without further discussion. Also updated the changelog.

tomburns:
While we're looking at this change, it might be worth examining which associated values we'd like to have for each error. As an example: I've already added the status code as an `Int`, but it might be nice to have the expected range which resulted in the error as well? e.g. `.StatusCode(404,200...299,response)`

Source/ReactiveCore/MoyaError.swift
```diff
@@ -9,3 +9,59 @@ public enum MoyaErrorCode: Int {
...
+
+public enum MoyaError: ErrorType {
+    case ImageMapping(MoyaResponse)
+    case JSONMapping(MoyaResponse)
+    case StringMapping(MoyaResponse)
+    case StatusCode(Int,MoyaResponse)
+    case Data(MoyaResponse)
+    case Underlying(ErrorType)
+    
+    public func sameErrorType(otherError: MoyaError) -> Bool {
```
ashfurrow: Where there were a better way to structure this function, but I don't think there is. Is it useful to developers using Moya, or just us when testing? Either way, I think it might look cleaner inside an extension.

ashfurrow:
Agreed it's worth being thoughtful about their associated values. I'm not sure how I feel about the range specification. I _think_ I like it – anyone else have feels?

Source/ReactiveCore/MoyaError.swift
petester42: Making the error conform to Equatable would be a better solution in my opinion.

tomburns: This isn't equality per se though; to my mind that would involve the associated values matching as well.

Moving it to an extension makes sense; I'm trying to come up with non-testing use cases but nothing so far. Maybe we move it into the test target for now and see if we miss it?

tomburns:
To keep the associated values clean (especially given their lack of names) we could also consider specialized struts, like so:
```
public enum MoyaError: ErrorType {
    case StatusCode(StatusCodeErrorMetadata)
    [...]
}

struct StatusCodeErrorMetadata {
    let encounteredStatusCode: Int
    let expectedStatusRange: Range<Int>
    let response: MoyaResponse
}
```

That struct's name is a bit of a mouthful, but I like that it keeps us down to one associated value and the convention is fairly straightforward.

petester42:
You mentioned that you want to keep associated values clean "especially given their lack of names". You can give names to associated values inside an enum in swift if that's what you wanted to do. So like `case CoolCase(value: Sting)`

justinmakaila:
I think associated values are the way to go. They represent a domain, and each case can encapsulate any values associated with it.



—Justin Makaila

tomburns:
Ah, glad to learn that names work! No idea why I was so sure they didn't. 

+1 for associated values, then :joy:

Source/ReactiveCore/MoyaError.swift
ashfurrow: Agreed, this function works great. Moving it to the test target sounds great too, we can always move it back later.

tomburns:
I broke it again! :tada: 

Tests are passing locally for me; not sure what's up: 

```
tburns@Monarch ~/code/Moya [error-type *]
± % rake test                                                                                                                              !10072
set -o pipefail && xcodebuild -workspace 'Demo.xcworkspace' -scheme 'Demo' -configuration 'Debug' -sdk iphonesimulator -destination platform='iOS Simulator',OS='9.0',name='iPhone 6' build test | xcpretty -c --test
2015-11-08 23:02:13.836 xcodebuild[73178:2107923] [MT] IDETestOperationsObserverDebug: (FC2F7820-8995-42A0-8796-B9ADFD7D4C44) Beginning test session FC2F7820-8995-42A0-8796-B9ADFD7D4C44 with Xcode 7B91b on target <DVTiPhoneSimulator: 0x7f9b2d6ef3f0> {
		SimDevice: SimDevice : iPhone 6 (C5B3EBC8-B35E-44DE-8D89-72DCB622959D) : state={ Shutdown } deviceType={ SimDeviceType : com.apple.CoreSimulator.SimDeviceType.iPhone-6 } runtime={ SimRuntime : 9.0 (13A344) - com.apple.CoreSimulator.SimRuntime.iOS-9-0 }
} (9.0 (13A344))
..................................................................

	 Executed 66 tests, with 0 failures (0 unexpected) in 5.932 (5.959) seconds
```

petester42:
You didn't add the `MoyaError+MoyaSpec.swift` to the DemoTests target inside the demo project. Should work after you do that.

JRHeaton:
Very cool, I think this is a great way to handle errors 👍. 

A couple of thoughts:
1. With Swift's sweet pattern matching, I think you may able to trim some redundant code off of the `sameErrorType` function by switching on a tuple of the two values. Example:
```swift
switch (self, otherError) {
case (.ImageMapping, .ImageMapping), (.JSONMapping, .JSONMapping) /*, etc... */:
    return true
default: 
    return false
}
```
2. Since Alamofire fails with `NSError`, I think it may make sense to type the associated value on `Underlying` to be `NSError` rather than `ErrorType`. This prevents confusion that there could be any other type given and thus need to do runtime identification of the type that we know at compile time. This would require re-typing `EndpointSampleResponse.NetworkError`'s associated error type as well, so that `Moya.Completion` could have it's error type be `NSError`.

In short, the Alamofire/"underlying" layer only fails with `NSError`, and since none of these status code filtration or data mapping functions exist in the non-reactive Moya core, it is the only error type to be dealt with, so we can benefit from some type specificity here.

tomburns:
@JRHeaton The only nice thing about the current shape of that comparison function is that it's exhaustive, and will generate warnings if we add to the enum. Your version is certainly the more handsome of the two; what does the group think?

As an aside, the latest CI failure appears to be Travis flapping again.

petester42:
I like pattern matching more but that's just me. Also the switch can be exhaustive without default if that is what you are worried about. You just need to add all the combinations as cases. I don't think having a default case that returns false would be bad. Usually we would want to check that 2 errors are the same so if there is a new error and that function has not been updated then it will return false so tests would fail etc.

gregpardo:
I am also on board with moving away from NSError pattern. Can we just put the NSError into the metadata in case it's needed as well StatusCodeErrorMetadata @tomburns suggested?

JRHeaton:
Good point @tomburns, it's a bit of concision versus compile-time safety. I wish there was a prettier way to get the latter, but oh well.

Any thoughts on my first point above? Alamofire only uses `NSError`, so the `MoyaError` case that covers the domain of Alamofire errors should provide an associated `NSError`, not `ErrorType`.

ashfurrow:
I personally like ErrorType – it's convertible to NSError.

JRHeaton:
Just seems silly to declare things as ErrorType when we _know_ it's literally only _ever_ an NSError. Am I not understanding a benefit of doing it the other way? I'm not so much complaining as I am truly not understanding, I guess. This is even the motivation behind parametric error types for ReactiveCocoa signals – to not leave a signal that can throw deterministic error types to the open-ended `ErrorType` that could be anything, but to give a specific `ErrorType`-conforming type so that there is no nondeterminism about types at compile time. It _is_ convertible to NSError, but it also ALREADY is an NSError, hence my not understanding declaring it differently. And this only applies of course to the `Underlying` error case.

Think about the call point of the Alamofire request. The completion handler provided takes an NSError, which Moya gets back, and then (only ever) feeds as `Underlying`, which takes a more generic type than we know.

colinta:
Is there anything preventing Alamofire from converting to using `ErrorType` over `NSError`?  If they do that, Moya would have to push out a new version.  Not *at all* a bad thing, and this is the kind of version bump that makes semantic versioning so relevant, but that *might* be an argument in favor of the more generic `ErrorType`.

A way of justifying it philosophically might be: if we don't need / aren't consuming the features of `NSError`, then using the generic type indicates that we *aren't promising* what type of error is coming down the wire.  If this error is coming from a dependency, then this lack of promise is honest - we know we'll get an error, but we can't promise what it will be.

I'm just catching up on this discussion, but these are my initial thoughts.  Gonna re-read now.

ashfurrow:
The ability to write exhaustive catch blocks is actually a really good point. 

Let's try to remember that we're all after the same thing (writing great code) even if we disagree on how to get there. If it would help, I can schedule a Google Hangout to discuss some of the pros and cons.

gregpardo:
Should an issue be opened on Alamofire? I do believe it would be an API breaking change but I think it's probably worth pursuing. The docs on swift errors suggest using the ErrorType pattern is the way to go.

ashfurrow:
We should look into the history of their error handling first. The changelog might reveal why they have it this way.

gregpardo:
Coincidentally, their 'benefits to upgrading' to 3.0 is almost all about errors.

I think the part that may be of interest is the custom wrapping of any NSError to a type of our choosing. 

https://github.com/Alamofire/Alamofire/blob/0cb96d59ad47be1b6f5018188fcfd1d6b0ecbb0c/Documentation/Alamofire%203.0%20Migration%20Guide.md#benefits-of-upgrading

ashfurrow:
Cool, not sure where we stand on this one – anything stopping it from being merged? 

In the mean time, I'm going to merge https://github.com/Moya/Moya/pull/291 because it has no breaking changes, and it'd be cool to include its changes in a point release before this one (so people can have just those changes if they want).

tomburns:
I think this can be merged as a first step, unless we want to go all-in on the new error type all at once. If there are any pending changes I've missed, let me know and i'll try to address them.

tomburns:
Scanning the thread, the only things that I see as unresolved are what associated values we want to make available on each error, and naming them inside the tuples. Do we want to take care of that before merging, or just before releasing this into the wild?

ashfurrow:
Cooooool yeah sounds great! There've been changes on master since this PR was submitted, @tomburns could I ask you to merge from master?

tomburns:
Merge to master: :white_check_mark: 

If someone could give that a quick spot check I think we're good to go.

petester42:
After those small changes I think it should be good. It might cause you to have to update other parts of the codebase like the tests and the RxSwift extension though. If you need help with that, I'm here to help. :+1:

tomburns:
Thanks for the feedback! Good catches all around; I agree with all the proposed changes and I'll try to get this updated today.

tomburns:
I believe this is ready for further review; as it stands I was able to make the suggested changes without touching the Rx extensions or tests. I'm a little unhappy with that revised attempt at passing along the status code error (see my comment on the latest commit above) , so if anyone has any tips I'm all ears.

ashfurrow:
Cool, seems good to me. If someone else could give this a final one-over and merge, that'd be :100: Good work everyone :balloon:

petester42:
Looks good. There might be a little clean up needed to maybe remove `MoyaErrorDomain` and `MoyaErrorCode` if possible. If no one has other objections I will merge this.

PR #291, opened 2015-11-10
Refactor Reactive Extensions

petester42:
There was a lot of repetitive code in the reactive extensions for the mapping to different types. Decided to move those to an extension on `MoyaResponse` to be reusable. I also noticed that the tests for RACSignal were not in the test target so I fixed those up as well.

Changlog:

- Fix RACSignal tests not being run
- Update RACSignal tests to make them compile
- Create mapping helpers for MoyaResponse as an extension
- Change reactive extensions to use MoyaResponse mapping helpers
- Remove unneeded force unwrap in reactive extension's `request` method


ashfurrow:
Fantastic work, as usual! We could add **Breaking Change** to the changelog entry just to be super-clear. 

I've learned never to merge things before coffee; will do a final review later (unless someone beats me to it!).

Source/ReactiveCore/MoyaResponse.swift
```diff
@@ -20,6 +20,50 @@ public class MoyaResponse: NSObject, CustomDebugStringConvertible {
...
+    }
+    
+    /// Maps data received from the signal into a String.
+    func mapString() throws -> String {
+        guard let string = NSString(data: self.data, encoding: NSUTF8StringEncoding) else {
+            throw NSError(domain: MoyaErrorDomain, code: MoyaErrorCode.StringMapping.rawValue, userInfo: ["data": self])
+        }
+        return string as String
+    }
+}
```
orta: :+1: good stuff

petester42:
The only case I can think of this being a breaking change would be if the developer subclassed the provider and created a new request function that returned a subclass of `MoyaRespnse` instead. I believe that this would not work anymore but it would have before. I could try it out to see if that is really the case. Aside from that there are no API changes just additions.

ashfurrow:
Ah, clever!

PR #296, opened 2015-11-12
Deterministic stubbing 🎉

swizzlr:
Todo:
- [x] Update changelog

Allow injection of a date scheduler which will, when stubbing, be used instead
of the default behavior for executing stub closures.

This commit required reworking the internals of MoyaProvider, mostly factoring
out some common code to be shared by the subclass into a separate few methods
that are final and internal, and documented.


swizzlr:
To do: add tests

Source/Moya.swift
```diff
@@ -223,26 +241,17 @@ private extension MoyaProvider {
...
+    internal final func notifyPluginsOfImpendingStub(request: NSURLRequest, target: Target) {
+        let request = manager.request(request)
+        plugins.forEach { $0.willSendRequest(request, provider: self, target: target) }
     }
 }
 
-/// Private token that can be used to cancel requests
-private struct CancellableToken: Cancellable , CustomDebugStringConvertible{
+/// Internal token that can be used to cancel requests
+internal struct CancellableToken: Cancellable , CustomDebugStringConvertible{
```
ashfurrow: :+1:

Source/ReactiveCocoa/Moya+ReactiveCocoa.swift
```diff
@@ -4,12 +4,23 @@ import Alamofire
...
+            self.stubScheduler = stubScheduler
+            super.init(endpointClosure: endpointClosure, requestClosure: requestClosure, stubClosure: stubClosure, manager: manager, plugins: plugins)
+    }
+
     override public init(endpointClosure: EndpointClosure = MoyaProvider.DefaultEndpointMapping,
         requestClosure: RequestClosure = MoyaProvider.DefaultRequestMapping,
         stubClosure: StubClosure = MoyaProvider.NeverStub,
         manager: Manager = Alamofire.Manager.sharedInstance,
         plugins: [Plugin<Target>] = []) {
+            self.stubScheduler = nil
```
ashfurrow: Let's use one initializer with `nil` as a default value for the parameter?

ashfurrow:
Cool, looks great so far :cake

Source/ReactiveCocoa/Moya+ReactiveCocoa.swift
swizzlr: Can do but you'll lose API compatibility, that cool?

ashfurrow: yeah, this will be in our next major release anyway

Moya.xcodeproj/project.pbxproj
```diff
@@ -1830,6 +1831,7 @@
 				554B101E1BE7F11700F3D266 /* Release */,
 			);
 			defaultConfigurationIsVisible = 0;
+			defaultConfigurationName = Release;
```
swizzlr: Bed to kill al this

swizzlr:
SOMEBODY MERGE IT

ashfurrow:
I think you meant to say "this is ready for final review" 😉

I'm on my phone right now – anyone is free to merge this if it looks 👍 to them ☕️

JRHeaton:
Nice work. I have some confusion regarding the behavior of `createStubFunction`'s `canceled` parameter. I'm unfamiliar exactly with how `inout` parameters work in this type of context. When the closure returned from `createStubFunction` checks this bool, is it still checking the original variable in the caller's scope by reference? Or is it a copied value at this point, thus negating the intended `Cancellable` behavior?

Either way, it may be worthwhile to write a test that verifies this behavior pre-merge if we don't have one already. If we do, carry on 😇

swizzlr:
You know, I have no idea. I need to experiment and find out. I'll make a playground when I get to work (~1hr)

swizzlr:
I've ducked the issue by making the stub function take a cancellable token, and refactoring the token to be a reference type. While I was at it I gave it a canceled property and ensured writes to it were locked, let's see what this does to coverage.

swizzlr:
@JRHeaton care to take a look?

JRHeaton:
Clever! I'd say that seems like a better implementation. The only odd thing to me here is having a non-optional action closure on the token, but the initializer takes an optional one, defaulting to an empty `{ }` closure if initialized with `nil`. Perhaps the token could have an optional action closure (since it itself now handles a cancellation state boolean), and it could be invoked with Swift's optional function call syntax: `cancelAction?()` Feelings on this? It's a small thing, but would allow us to remove the icky force unwrap in `init()` :)

swizzlr:
Can do! Could also do an if let – it's really just because the nil-coalescing operator gave up the ghost. I'm trying to be conservative with API changes, but I keep forgetting this is an internal class.

ashfurrow:
Yeah, I like even the idea of a non-optional closure with a default init value of an empty closure.

swizzlr:
If we're getting into the nitty gritty I'm going to assume the non-optional version would take slightly more overhead in terms of space; or I could just be justifying the change I made and pushed before I saw your comment :)

ashfurrow:
I mean, I _prefer_ things to be non-optional whenever possible just to limit the surface area of the API. It's why the `Cancellable` tokens didn't keep track of their cancelled state until now :stuck_out_tongue: If you don't feel strongly against the idea, I don't feel _that_ strongly about it. 

Though I would like to keep things consistent within Moya and adhere to Apple's general practice of using `() -> Void` instead of `() -> ()` if that's okay.

swizzlr:
@ashfurrow sorted. Cancellable tokens should keep track of canceled state because otherwise they're really just a glorified completion handler.

ashfurrow:
> they're really just a glorified completion

Now you're getting it :wink:

swizzlr:
Don't furrow me bro

Issue #293, opened 2015-11-12
Change completion block type.


petester42:
Since we are going to merge a breaking change with #289 I'd like to propose that we change the completion block type.

From
`public typealias Completion = (data: NSData?, statusCode: Int?, response: NSURLResponse?, error: ErrorType?) -> ()`

To
`public typealias Completion = (response: MoyaResponse?, error: ErrorType?) -> ()`

This would allow people that do not use the reactive extensions to be able to use the helpful mapping functions and simplify the integrating with various json parsing libraries.

What do people think?

ashfurrow:
Great idea! Thoughts @Moya/contributors ?

gregpardo:
On board with this. This is much simpler. I was kind of wondering why it was broken up into 3 values related to response.

SRandazzo:
Like this a lot :+1:

justinmakaila:
Into it

esttorhe:
:100:

tomburns:
This is a great idea :+1:

jarrodrobins:
![giphy](https://cloud.githubusercontent.com/assets/1313663/11134697/e28bfeb6-89f2-11e5-8471-d781b4936e96.gif)

Sounds good to me. :+1:

swizzlr:
I think this should definitely be included in version 5.

petester42:
I've been playing with this idea a little bit and would like to suggest something a little different. How about the following:

`public typealias Completion = (result: () throws -> MoyaResponse) -> Void`

The advantage is that there is a clear distinction between the success and error paths. Which would look something like this:

```swift
GitHubProvider.request(.UserRepositories(username)) { result in
    do {
        let response = try result()
        print(response)
    } catch {
        print(error)
    }
}
```

ashfurrow:
Hmm, super-interesting! I like it. Anyone else have feelpinions?

swizzlr:
Well, I'd rather have a result type, because otherwise everyone's just going to be writing `let e = error as! MoyaError`, but what do I know checked exception bleh etc

swizzlr:
I have a massive PR incoming btw that addresses this issue plus a bunch of other stuff; this approach is better than our strange disjoint union of optional types.

JRHeaton:
> Well, I'd rather have a result type, because otherwise everyone's just going to be writing `let e = error as! MoyaError`

[...]

PR #297, opened 2015-11-12
Error Clean Up

petester42:
Before starting with #293 I just wanted to make sure that everything that came with #289 was finalized. There were a couple NSErrors that still existed so I cleaned it up to remove all usages of NSError and replace them with MoyaError.

Changes:
- Forward underlying errors when cast to MoyaError does not work
- Fix MoyaResponse extensions that were still throwing an NSError instead of MoyaError
- Change MoyaError to conform to `Equatable` in tests for cleaner syntax
- Update reactive extension tests to use MoyaError instead of NSError
- Update ReactiveCocoaMoyaProviderSpec tests to use the `SignalProducer` variant when possible.


Source/ReactiveCore/MoyaError.swift
```diff
@@ -10,11 +20,23 @@ public enum MoyaErrorCode: Int {
...
-    case ImageMapping(MoyaResponse)
-    case JSONMapping(MoyaResponse)
-    case StringMapping(MoyaResponse)
-    case StatusCode(MoyaResponse)
-    case Data(MoyaResponse)
-    case Underlying(ErrorType)
+internal extension MoyaError {
+    
+    // Used to convert MoyaError to NSError for RACSignal
+    var nsError: NSError {
```
ashfurrow: I think the Cocoa convention would make this `NSError` – I know it looks weird, but I think it makes sense to adhere.

ashfurrow:
Looks great! Thanks!

Source/ReactiveCore/MoyaError.swift
petester42: Would love to but I can't.
screen shot 2015-11-12 at 3 04 24 pm

ashfurrow: lololol

Well then.

swizzlr:
I've opened a PR against this branch to try to resolve some of the code coverage issues: https://github.com/petester42/Moya/pull/1

petester42:
Should be good to merge now. If anyone wants to look it over.

JRHeaton:
Just a thought, but does it make sense to implement `Equatable` in this way? The Swift docs declare this protocol as being for "value equality", which I feel this is not, as this expression could be true: `.Underlying(x) == .Underlying(y)` (where `x != y`). Seems to be typical convention to implement equality checking for associated value enum cases to compare the associated values, provided they're equatable (example: https://github.com/ReactiveCocoa/ReactiveCocoa/blob/fd64bf6ea7a83dec13fe42244db470fd9641a9a1/ReactiveCocoa/Swift/Event.swift#L94-L111), which I think `MoyaResponse` could conform to to make this work for this error type. 

As for checking the _case_ equality, that's still very much useful and necessary, in this case for the tests, but it felt more appropriate as a function on the enum IMO. May be confusing for users of Moya seeking true value equality checking on this type, which may have been the reason for @tomburns' [initial reservation about implementing it this way](https://github.com/Moya/Moya/pull/289#issuecomment-154882315).

petester42:
Equality is just what we define to be equal. In this case since we are just using it for tests its simpler to just check the error type since in the tests that use it the response is going to be the same since it is hard coded. If we were going to make it public then ya I agree. @tomburns wanted it to be exhaustive and comparing the responses too without a default case is not fun.

JRHeaton:
Sure, just trying to stay aligned with observed language conventions 😊
From Swift `Equatable` docs:
> Instances of conforming types can be compared for value equality using operators == and !=.

I think it's more responsible of us to not pollute the type with a non-conventional `Equatable` implementation just to be able to use `==` in a test rather than a (more appropriately named) function on the type. We don't need value equality for this type at this point, but we may in the future, or users of Moya may want to extend to implement it themselves, and would be stuck as the type already conforms with behavior they don't expect.

swizzlr:
`Merging #297 into master will increase coverage by +9.07% as of 20e82cf`

😱

swizzlr:
Do however agree with -1 on default cases in switches, and not comparing underlying values, but while it's internal to the framework it matters not to consumers of the API.

JRHeaton:
Well I didn't say anything needed changed with the switch statement or how it was implemented, just that this (again, perfectly valid) _case_ comparison should change in no way _other than_ taking it away from `Equatable` where it doesn't fit.

And it _is_ public, unless you mean unreleased. It publicly conforms to `Equatable`, using up that conformity in a non-standard/sane way, which was my point. Again finding myself at a loss of understanding, so I'll ask more explicitly: why is it thought that two "same-operation" errors from two entirely different responses are "equal?" Or two completely different `NSError` objects in `.Underlying(err)`? I remain with my opinion on this, because IMO `.Underlying(errorA) == .Underlying(errorB)` evaluating true in any scenario other than where `errorA == errorB` seems lacking in sanity to me. As someone (also a user of Moya) who would potentially want equality checking with the behavior I describe, and would be forced to implement it OUTSIDE of `Equatable`.

petester42:
Good points all round. I didn't think it would have caused such a big issue since that file is not even in Moya. Yes it is public but it is only part of the test target. I'll see what I can do about it later today.

petester42:
Since the MoyaError Equatable method was only used by tests to compare the types of errors, I changed it to be a matcher for Nimble. Makes more sense since it's only for tests. 

@JRHeaton want to review this and then merge?

ashfurrow:
> changed it to be a matcher

Great idea! :100:

swizzlr:
Looks damn good to me! We just need to resolve the merge conflicts.

JRHeaton:
Looks good once conflicts are fixed 👍

petester42:
That should do it :smile:

PR #299, opened 2015-11-13
Fix bundle version for Carthage etc

swizzlr:
Guess no one has yet tried to submit ReactiveMoya with Carthage yet.

This commit removes the MOYA_VERSION env var as it was out of date and not used everywhere. I don't think it's needed.

### DIFF
--- a/Moya.xcodeproj/project.pbxproj
+++ b/Moya.xcodeproj/project.pbxproj
@@ -1419,7 +1419,6 @@
 				INFOPLIST_FILE = "Supporting Files/Info.plist";
 				INSTALL_PATH = "$(LOCAL_LIBRARY_DIR)/Frameworks";
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks @loader_path/Frameworks";
-				MOYA_VERSION = 4.0.3;
 				PRODUCT_BUNDLE_IDENTIFIER = org.cocoapods.RxMoya;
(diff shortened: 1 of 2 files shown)


ashfurrow:
Can you merge in from master and push?

petester42:
I don't think this will solve the problem. Doesn't the framework bundle version have to match the app version now?

esttorhe:
If i'm not mistaken that was an error from  that should be fixed by now.

petester42:
It wasn't fixed last week when I uploaded a beta. Maybe with the update to iTunes connect that just came out it is fixed,

swizzlr:
I can personally confirm that the matching version thing was a bug and was def fixed for me. There's no good reason to declare the framework version, so I think 1 is a safe bet.

The issue this resolves was that some targets had no MOYA_VERSION defined, thus causing a nil short bundle ID, and thus failing ITC checks.

petester42:
Ah ok. Thanks for the fix. That was probably my fault. Forgot to add MOYA_VERSION to the other targets when I created them.

swizzlr:
No worries sport; was there a reason why you wanted MOYA_VERSION? I can put it back in if that makes the library easier to use for more people.

petester42:
I have no need for it. I didn't add it. I just updated the project to have more targets. Just didn't add it to the newer ones I guess.

swizzlr:
Well, looks like we're good to go with this?

PR #301, opened 2015-11-15
Deprecate RAC2 support

### LINKED ISSUE #298: Deprecate RAC2 support?
Not insignificant time is being spent maintaining the RACSignal extensions, as well as the non-RAC3+ provider. We should consider officially marking these as deprecated for some time, before eventually removing them. 

Not sure if anyone else is using them or if its just me, and I won't be for much longer. Open for discussion!

petester42:
As the title states, ReactiveCocoa 2 support has been deprecated.

Closes #298


Source/ReactiveCore/MoyaError.swift
```diff
@@ -9,9 +9,10 @@ public enum MoyaError: ErrorType {
     case Underlying(ErrorType)
 }
 
-// If RACSignal support gets removed we can also remove everything under here
+@available(*, deprecated, message="This will be removed when ReactiveCocoa 4 becomes final. Please visit https://github.com/Moya/Moya/issues/298 for more information.")
 public let MoyaErrorDomain = "Moya"
 
+@available(*, deprecated, message="This will be removed when ReactiveCocoa 4 becomes final. Please visit https://github.com/Moya/Moya/issues/298 for more information.")
 public enum MoyaErrorCode: Int {
```
swizzlr: I should think that, in fact, we can merely rename these types, remove the domain and take away the underlying `Int` – these types are still useful and ought to remain.

petester42: If the Int is taken away then it is just what MoyaError is. I don't really see a use for both MoyaErrorCode and MoyaError when we remove RAC 2 support. The only use of MoyaErrorCode is to generate an NSError for RAC 2.

orta:
Probably worth mentioning in the README too

orta:
Form maybe clarifying that its 3 and above, rather than saying what isn't supported

Source/ReactiveCore/MoyaError.swift
ashfurrow: Something like https://github.com/Moya/Moya/blob/master/Source/ReactiveCore/MoyaError.swift ?

ashfurrow:
Agreed with Orta – some clarification in the readme would be good, but that's not a blocker. I'll open an issue.

PR #302, opened 2015-11-16
The Great Renaming (and retyping)

swizzlr:
(Net reduction in LoC!)

I got carried away closing #293.

Significant changes:
- Completion handler now returns `Response?, ErrorType?`
- `Plugin`s are now a protocol
- `ReactiveCore` is gone and now part of Moya in general
- `MoyaTarget` is now `TargetType`
- `MoyaResponse` is now `Response`
- `MoyaProvider` is now `NetworkResourceProvider` (discussion welcome on improving this)
- `sampleResponseClosure` parameter is now an escaping autoclosure, which makes the API a lot cleaner to use.
- Test classes are now `final`, as all classes should be.
- `Response` is now `final` (see above bullet), and will be changed to a `struct` once we drop RACSignal support.
- `Endpoint` is now a struct, and the `SampleResponseClosure` typealias has been removed for simplicity's sake
- `EndpointSampleResponse` now takes a non-optional ErrorType, or a `Response` object as opposed to a code and data.


swizzlr:
I believe these changes to be good because they make the library a bit more "Swift". Modules provide a namespace, so we don't need to prefix stuff, nor do we need to create our own namespace.

Plugins sorely needed to be a protocol, but were quite objc. I sometimes feel Moya is really wanting to have that magic dynamism that Swift rightly disallows. Turns out that if we get rid of the provider param, we can make plugins protocols. I can't see a good use for it either. Could be replaced with an `AnyObject` param.

We should either make the completion in-type a result, or a throwing func, just as suggested by @petester42, right now this is just weird.

ashfurrow:
Okie doke, wow! You're right this is a _huge_ PR :sweat_smile: That's ok, thanks for the work! I'd like to discuss a ew top-level issues before we discuss nitty-gritty details. 

I'm good with all the renames except `MoyaProvider` – I know it's not necessarily "Swift", but I really like having the library name somewhere in there. Maybe it's just me ¯\\_(ツ)_/¯

The changes to `EndpointSampleResponse` look good, too. The `final` test classes, I've never seen that but fi that's what's suppose to be, that's fine. 

I think adding @petester42's suggestions about throwing closures becomes necessary with these new changes. We can wait until after this is merged before doing it though – already a big enough PR :)

This'll need a merge from master and all of these points need to be described in the changelog, but yeah I agree with the general thrust. Thanks for the work! Smaller PRs next time! :cake:

colinta:
I would add that I have always appreciated the names `MoyaTarget` and `MoyaResponse`. Reminds where those objects originated. 

In my apps I have often defined Target and Response types - those names seem too generic to "claim" by a library.

ashfurrow:
Yeah good point, even if the compiler lookup can succeed most of the time using type inference, my brain's type inferrer is less powerful.

swizzlr:
You ever change one thing, and then change something else, and then before you know it you're crying on the bathroom floor and the compiler is screaming in the kitchen?

swizzlr:
I agree with the points about target and provider. Let me know if you think I should do it for response too. Generally speaking, Swift is moving in a direction where namespaces are a thing (cf Alamofire.Request etc) but I can totally see how this could be irritating and make the APIs less usable.

swizzlr:
I'm going to factor this into individual PRs, so we do this right!

ashfurrow:
Sounds good! Looking forward to the PRs!

PR #304, opened 2015-11-16
Remove Moya namespace class in favour of the module namespace.

swizzlr:

PR #305, opened 2015-11-16
Rename MoyaResponse -> Response

swizzlr:
Depends on #304


ashfurrow:
Cool, looks good. This depends on #304? Want to update the PR description?

Also, a changelog entry would be :cake:

swizzlr:
Am I correct in reading this as "that would be cake"

ashfurrow:
That would be a correct literal interpretation, yes.

colinta:
I thought we were going to keep the `Moya-` prefix on these types, what's the advantage of removing it? Advantages to keeping it:

Unambiguous origination. A `MoyaResponse` came from Moya. Where does `Response` come from? I wish Alamofire used a prefix on these generic names, too. 

No possibility of namespace collisions. `import Moya` will bring `Response` in as a top-level type, no need to prefix `Moya.` - so what does it mean when you create your own Response type in a file that imports Moya?

I think, and this is just a "dissenting opinion" that this change is an unnecessary breaking change.

ashfurrow:
@colinta those are some good points. It seems like Swift libraries are going one of two ways: either they're sticking with common Objective-C-style conventions like prefixes or they're going 100% prefix-less with no prefixes. Alamofire has gone the latter route. 

I agree with the downsides. In the case where there are two `Response` types, you can delineate between the two by using the `Moya.` prefix (we do this when using Alamofire [here](https://github.com/Moya/Moya/blob/e0b2b2a9108d3b9df8e6e671a91e49422a0ef20c/Source/Moya.swift#L14) for example).

Seems we're at a crossroads. I honestly don't have a strong opinion either way, but I recognize that this is a big shift and it deserves thoughtful consideration. Trying not to spam @Moya/contributors but I think this deservers further discussion.

colinta:
It does seem that way, and it just occurred to me that this change doesn't *remove* the `Moya.` prefix, it just makes it optional. With that in mind, I'm less opposed to it.

gregpardo:
https://github.com/raywenderlich/swift-style-guide

According to this we should just use namespaces. I'm okay with that. If there is a collision you just have a little more to type.

JRHeaton:
I don't see any real downsides to prefix-less type names in frameworks. As has been established, with references to the full type name, including the module name prefix, ensures there's always a way to disambiguate references to same-name types that exist in separate modules.

In the case where a user of Moya (or any other Swift module) with a "colliding" type name doesn't want to keep prefixing with the module name, they can always use a typealias.

ashfurrow:
Cool, can someone hit the "merge" button?

colinta:
Wheeeeeee!

colinta:
Do we "Delete branch" afterwards?

ashfurrow:
Sure!

steam:
@colinta loves to merge!

swizzlr:
THE CI WAS FAILING

swizzlr:
NOOOOOOO WE

(:sly smile:)

PR #307, opened 2015-11-16
Plugins are now a protocol.

swizzlr:
Dependes on #305


ashfurrow:
@swizzlr is it possible to turn off the code coverage check for the patch? It's consistently failing, and I don't think it's super-necessary.

ashfurrow:
Also for some reason it's still trying to merge in commits that were already merged. Can you merge in from master when you get a chance?

swizzlr:
Code coverage check is a nice way to remind you of technical debt, IMO, and I've got a fix coming up now.

swizzlr:
See it prodded me to add test coverage to some stuff and I did and look we have test coverage hooray

ashfurrow:
Yeah totally, I'm just worried someone sending their first PR will be intimidated by a failing CI :(

Source/Moya.swift
```diff
@@ -5,30 +5,10 @@ import Alamofire
...
-            return .OPTIONS
-        case PATCH:
-            return .PATCH
-        case TRACE:
-            return .TRACE
-        case .CONNECT:
-            return .CONNECT
-        }
+public enum Method: String {
+    case GET = "GET", POST = "POST", PUT = "PUT", DELETE = "DELETE", OPTIONS = "OPTIONS", HEAD = "HEAD", PATCH = "PATCH", TRACE = "TRACE", CONNECT = "CONNECT"
```
ashfurrow: Swift 2 will infer the raw values for enums with underlying String types. So we can use this:

```swift
case GET, POST, PUT, DELETE, OPTIONS, HEAD, PATCH, TRACE, CONNECT
```

Source/Moya.swift
```diff
@@ -5,30 +5,10 @@ import Alamofire
...
+    public var method: Alamofire.Method {
+        return Alamofire.Method(rawValue: rawValue)!
```
ashfurrow: :grimacing: 

The crash operator, eh? Well, I'd definitely want some kind of comment explaining that. Maybe we could coalesce to something instead of force-unwrapping? `GET` maybe?

Source/Moya.swift
swizzlr: RLY

swizzlr: OMG

ashfurrow: ya rly

Source/Moya.swift
swizzlr: There's nothing meaningful to coalesce to – it's a programmer error for these methods to not to map to Alamofire methods, IMO. I'd rather it crash than start sending `GET`s when you're expecting `PUT`.

ashfurrow: I just don't trust Alamofire not to change something :stuck_out_tongue: Maybe coalesce to `GET`? Dunno. Whatever.

ashfurrow:
@swizzlr Anything I can do to help on this?

Source/Moya.swift
aaroncrespo: Also when you dry up those Strings you can do `Alamofire.Method(rawValue: String(self)) ` followed by the open question of coalesce or  Force. But the case is safer...

swizzlr:
Sorry, week got mad. To do is to address your crashing concerns and string inference, add some code coverage to alamofire method conversion and the method logger plugin. After that i think my codecov duty will be done.

ashfurrow:
Cool. One thing, @swizzlr: there is absolutely no apologize necessary – we're grateful for your contributions and you should not feel like you've let anyone down.

Source/Moya.swift
petester42: I am trying finish up this pull request but I really don't like the use of rawValue here. Explicitly mapping the methods seems better and offer compile time checks that all cases are handled. Also no `!`

ashfurrow: Could we make our enum use an underlying type from another enum? Like AF's?

petester42: is there any plans to use this mapping function? Because I just noticed it is never used. We could just remove it :stuck_out_tongue:

ashfurrow: Oh hey, you're right!

petester42:
I think this should be good to go now.

ashfurrow:
Looks 👌 to me! Would someone open an issue to update our docs around plugins?

PR #310, opened 2015-11-20
Change completion closure to return a result type

petester42:
This addresses issue #293. It changes the `Completion` closure to return a `Return` type instead of multiple optional parameters.

There is an issue though. Since we are using `Moya.Something` in places now some frameworks don't currently build. Since the module name for `ReactiveMoya` does not match `Moya` those frameworks don't build. This is even an issue with the current version but I don't see how to fix it.

Anyone have an idea on how to fix this. I tried setting the module name to Moya for all the frameworks but then you get a warning:
```
warning: PRODUCT_MODULE_NAME may not be overridden for framework target 'ReactiveMoya iOS'
```


ashfurrow:
This looks :100: great work!

For the module name problem, since that exists currently, we should probably open a separate issue about it? 

Thanks again, @petester42. We'll need to update our docs, but we can open an issue for that after we merge this. Can someone else from @Moya/contributors take a look and merge if they're happy?

gregpardo:
Quite impressive! I looked it over and it looks like this will be a great direction going forward. I'm going to go ahead and merge it so I can get some participation here.

PR #315, opened 2015-11-23
Fix Reactive Namespace

### LINKED ISSUE #313: Build failures in RAC/Rx framework targets due to prefix-less type names
Hi guys. Today I pulled master and tried building all targets on all platforms in the project, to make sure all was gooby. I noticed that ReactiveMoya and RxMoya targets have incurred some errors as a result of the prefix-less type name changes. 

Errors like this are happening on uses of types like `Moya.Completion` because `Moya` is only actually the module name in the "base" Moya, and the base source is shared. The reason this can't work is because Swift doesn't currently provide a way to reference things by the "current module." A lot of the decls in the source don't actually require the prefix where we use it, but some do, due to ambiguity, etc.

From what I can see, here are a few ways to fix this:
- Convert the Rx/ReactiveCocoa additions be extension frameworks that themselves link against Moya
- Revert type names back to having the Moya prefix
- Change the reactive framework targets' module names to `Moya` (though this would be kinda weird, linking to `RxMoya.framework` but using `import Moya`)

I had previous said I saw no drawbacks to moving to this naming convention, so I apologize 🙁 but I'd like to hear thoughts on what the best option for resolution is.

petester42:
This aims to fix #313 I tried various things but this solution seems to be the one that affects everything the least. What I did is that for the reactive schemes I added a prebuild action that changes all references in the source files from `Moya.` to `ReactiveMoya.` or `RxSwift` and also added a post build action that changes it back to `Moya.`

It's sort of hacky but it works properly. I have a lot of thoughts on how we are supporting carthage currently which seems like a bigger headache than it should be or worth. I'll open an issue about changes we could make to make everyone's life easier. 

Here are some of the things I tired for those interested:
- Changing the module name to `Moya` which works but since the output name has to be `Moya.framework` there will only be 1 framework in the carthage folder since the other ones get overwritten
-  Changing the module name to `Moya` and renaming the framework after it has been built in a post build action. This also works in Xcode but carthage references the framework name when moving it to the carthage folder so it fails with `file not found` since the name it wants is `Moya.framework`
- Rename instances of `Moya.` in the source files for `ReactiveMoya` and `RxMoya` on the fly before compiling and rename them back after compiling


swizzlr:
The inherent issue is that Carthage just builds targets in an xcodeproj; based on the issue, I'd recommend option 1, which was to make extension frameworks that sugar the respective libraries for Moya. I'm pretty sure that's what Apple would recommend in this situation if we were back in submodules. Until such time as we have weak linking in Swift, I think that would be the best option.

petester42:
The problem with option 1 is that cocoapods support would be broken. If we want to keep subspecs the way they are then we can't change the source files to using `import Moya` for reactive extensions. If we use option one then I believe that we also will need 3 distinct pods rather than 1 with subspecs.

ashfurrow:
@petester42 I quite like your approach here. Seems pretty reasonable, given the constraints we're in. 

> supporting carthage currently which seems like a bigger headache than it should be or worth.

You took the words right out of my mouth...

PR #316, opened 2015-11-24
Update documentations

wangshengjia:
This PR update some documentations which is not matching the API correctly.


ashfurrow:
Looks great, thanks!

PR #324, opened 2015-12-01
Add tests for OS X and tvOS

### LINKED ISSUE #265: Platform Tests on CI
We currently only test iOS on continuous integration, but ideally we should be able to test against all the platforms we support.

petester42:
This should close #265 with the omission of watchOS since it can't have test targets as of now.

Changes:
- Renames DemoTests folder to Tests
- Add test targets for OS X and tvOS
- Update Podfile to have the new targets
- Update Rakefile to run all the new test targets
- Add `ImageJPEGRepresentation` method for tests since there is no `UIImageJPEGRepresentation` equivalent for OS X


petester42:
URGH! Don't know why this intermittently fails seems like #259 is still an issue. Whenever it fails it happens to be the `uses a custom Alamofire.Manager for session challenges` test that fails.

ashfurrow:
When I've seen problems like this, it's typically "the nth test always fails", irrespective of which test that is.

petester42:
You think its a quick thing or a travis thing?

ashfurrow:
Likely a combination – I've seen it on other projects where Quick/Nimble only fail to work on slow hardware like Travis, and only sometimes. 

Want to restart that build to get the green check in our commit history?

petester42:
I had done it already. cross your finger lol

petester42:
nope. :cry:

ashfurrow:
Looks like the error we're seeing here is different from #265. Could this be a legitimate test failure? It's only happening on the pr tests – @petester42 when you have a chance, could you merge in changes from master and see if the test continues to pass locally? That way we can verify if it's a true failure or a Travis problem.

ashfurrow:
Nope, just intermittent ¯\\_(ツ)_/¯

I'm merging!

Issue #326, opened 2015-12-01
Is that possible to intercept a request before sending it using plugin?


mrahmiao:
I conformed PluginType to intercept the request, the `NetworkLoggerPlugin` do logged the changes I've made. However, I can see nothing changed in outgoing requests through Charles app.

Here is my code:

```swift
// Plugin
struct InterceptPlugin: PluginType {
  func willSendRequest(request: RequestType, target: TargetType) {
    (request.request as? NSMutableURLRequest)?.allHTTPHeaderFields = ["Hello": "World"]
    (request.request as? NSMutableURLRequest)?.HTTPBody = "RxMoya".dataUsingEncoding(NSUTF8StringEncoding)
  }
  
  func didReceiveResponse(result: Result<Response, Error>, target: TargetType) {
    
  }
}

// Provider

lazy var provider: RxMoyaProvider<HTTPBin> = {
    return RxMoyaProvider<HTTPBin>(plugins: [InterceptPlugin(), NetworkLoggerPlugin()])
}()

```

wangshengjia:
You sure about ```(request.request as? NSMutableURLRequest)?``` will not give a ```nil``` cause ```NSURLRequest``` not same as ```NSMutableURLRequest``` ?

mrahmiao:
I've printed the request and the mutable request do exist. 

```swift
print(request.request)           // Optional(<NSMutableURLRequest: 0x7fa0fbf12e30> { URL: http://httpbin.org/post })
print(request.request != nil)    // true
```

This is what the logger logged:

```
["Moya_Logger: [02/12/2015 05:53:18] Request: <NSMutableURLRequest: 0x7fd5134ae1c0> { URL: http://httpbin.org/post }", "Moya_Logger: [02/12/2015 05:53:19] Request Headers: [\"Content-Type\": \"application/x-www-form-urlencoded; charset=utf-8\", \"Hello\": \"World\"]", "Moya_Logger: [02/12/2015 05:53:19] HTTP Request Method: POST"]
```

wangshengjia:
Hmm.. you are right.

Apparently, it's because ```willSendRequest``` get called after ```request.resume()```... (It should not !)

So the workaround for now is using the approach mentioned by @ashfurrow in issue #325, it will work.

mrahmiao:
Yes, the workaround in issue #325  can be used to solve my interception problem.

[...]

PR #328, opened 2015-12-02
Fix a bug that unable to intercept requests using plugins

mrahmiao:
See issue #326

### DIFF
--- a/Source/Moya.swift
+++ b/Source/Moya.swift
@@ -175,13 +175,15 @@ public extension MoyaProvider {
 internal extension MoyaProvider {
     
     func sendRequest(target: Target, request: NSURLRequest, completion: Moya.Completion) -> CancellableToken {
-        let request = manager.request(request)
         let plugins = self.plugins
-        
+
         // Give plugins the chance to alter the outgoing request
         plugins.forEach { $0.willSendRequest(request, target: target) }
-        
-        // Perform the actual request
+
+        // Perform actual request
+        let request = manager.request(request)
+
+        // Add completion block to obtain response
         let alamoRequest = request.response { (_, response: NSHTTPURLResponse?, data: NSData?, error: NSError?) -> () in
             let result = convertResponseToResult(response, data: data, error: error)
             // Inform all plugins about the response
@@ -217,8 +219,8 @@ internal extension MoyaProvider {
     
     /// Notify all plugins that a stub is about to be performed. You must call this if overriding `stubRequest`.
     internal final func notifyPluginsOfImpendingStub(request: NSURLRequest, target: Target) {
-        let request = manager.request(request)
         plugins.forEach { $0.willSendRequest(request, target: target) }
+        _ = manager.request(request)
     }
 }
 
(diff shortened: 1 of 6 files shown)


wangshengjia:
Btw, I suppose if you can create the PR based on another branch instead of ```master```, that would be great. :tada:

mrahmiao:
Close this PR,  create a new branch in my local repo and then create a PR?

mrahmiao:
It seems that I forget to make the conformance to RequestType of NSURLRequest public.

mrahmiao:
@wangshengjia I've used my local repo in a new project and everything works fine.

wangshengjia:
Ah... my bad.. I didn't see it. It should compile without any error then.

On the other hand.. it seems to that you can not just fake methods without implementing the details. It turns out this solution will likely break the implementation of ```willSendRequest``` in ```CredentialsPlugin```

Why not just set ```startRequestsImmediately``` to false by default and call ```resume()``` manually.

ashfurrow:
Really cool PR, thanks! I don't mind it coming from master at all, personally. Could you add an entry to the changelog when you get a chance?

Thanks again! 🎉

Source/Moya.swift
```diff
@@ -285,3 +287,18 @@ private struct CancellableWrapper: Cancellable {
...
 /// Make the Alamofire Request type conform to our type, to prevent leaking Alamofire to plugins.
 extension Request: RequestType { }
+
+/// Make `NSURLRequest` conform to `RequestType` so that it can be sent with `PluginType` method `willSendRequest(_:target:)`.
+extension NSURLRequest: RequestType {
+    public var request: NSURLRequest? {
+      return self
+    }
+
+    public func authenticate(usingCredential credential: NSURLCredential) -> Self {
```
ashfurrow: These two methods make me feel a bit weird, since they don't do anything but should. I realize it's a bit of a chicken-and-the-egg problem :hatching_chick: Is this what @wangshengjia was talking about with calling `resume` manually?

wangshengjia: This approach is a little bit weird and will likely break the ```CredentialsPlugin```.

Personally I suppose we can just set ```startRequestsImmediately``` to ```false``` and add ```alamoRequest.resume()``` right behind where we created ```alamoRequest```

ashfurrow: Makes sense!

mrahmiao: Setting `startRequestsImmediately` to `false` will not work since the request will be wrapped in `NSURLSessionDataTask` after `manager.request(request)` is being called. At the time any changes to the original request would not affect the data task.

mrahmiao:
@wangshengjia Credential is the missing part~

Source/Moya.swift
wangshengjia: Nice catch ...

petester42: What about trying to change the task's `currentRequest`. Would that work?

wangshengjia: Probably non ...

wangshengjia:
@mrahmiao Great, thx

mrahmiao:
@ashfurrow Alter the changelog and make a PR?

ashfurrow:
You can update the changelog, commit, and push to this branch and github will add it to this pr ⭐️

mrahmiao:
It seems that separating two kinds of plugins (ordinary plugins and plugins used to change requests) is a much more appropriate way to solve this bug since almost no related point is exposed by Alamofire.

PR #329, opened 2015-12-02
Updated to ReactiveCocoa v4.0-alpha.4

mpurland:


ashfurrow:
Fab, looks great! Want to add an entry to the changelog?

mpurland:
Sure, will do after I look into this build issue.

mpurland:
@ashfurrow Can you trunk push this updated podspec for the latest ReactiveCocoa alpha? It's needed in order to set the version in the Podfile for the Demo project in Moya. 

See https://github.com/mpurland/Specs/blob/rac-update/Specs/ReactiveCocoa/4.0.4-alpha-4/ReactiveCocoa.podspec.json

mpurland:
For reference: https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2590

ashfurrow:
Sure, can do. Where did you get the original to modify? We're using https://github.com/mhuusko5/Podspecs as a source for Ruby-based pod specs. /cc @mhuusko5

Also, I'm afraid to say that the ReactiveCocoa developers are unlikely to do anything but close your issue. They've declined repeated offers from myself and others to help maintain the podspec in their repo, and choose to be poor sports instead.

mpurland:
I used the podspec for 4.0.3-alpha-3 as a base since that's the last one you added. 4.0.4-alpha-1 is a much older version actually...

mpurland:
Is it because they're pushing for carthage instead?

ashfurrow:
Makes sense – the Ruby ones are a bit easier to reason about. 

I'm not sure why they insist on declining, but it's been that way for as long as I can remember – far before Carthage.

mpurland:
Not sure what the next step here is. The carthage build works, but the Demo project will not build because it depends on CocoaPods and the Moya podspec which can't reference a branch or commit.

ashfurrow:
The next step is we need to push a podspec to trunk. Could you update the podspec on https://github.com/mhuusko5/Podspecs and send him a PR? Then we can push to trunk, then we can merge this :tada:

mpurland:
Done https://github.com/mhuusko5/Podspecs/pull/1

mhuusko5:
Pushed to trunk

mpurland:
Re-pushed commit, now let's see if travis builds it...

mpurland:
@ashfurrow This should be good now?

Changelog.md
```diff
@@ -2,6 +2,7 @@
 
 # 5.0.0
 
+- Updated to latest ReactiveCocoa v4.0-alpha.4
```
ashfurrow: Hey! Sorry for the confusion, but this should go in the "Next" section, since 5.0.0 is already released. Does that make sense?

mpurland: Sure

ashfurrow:
Looks good to me, if someone else wants to review + merge after the changelog is updated, that'd be :100: 

Thanks again @mpurland – great work!

mpurland:
Ready to :boat:

ashfurrow:
Thanks!

ashfurrow:
I've added you as a contributor to Moya, so you've got full push access :tada: You can read more about that here: https://github.com/Moya/contributors

I've also released 5.1.0 of Moya so you can use your new commits from Trunk! Thanks again!

mpurland:
@ashfurrow Thanks. Glad to help out now and in the future.

mpurland:
@ashfurrow Any reason why 4.0.2 is the Latest release and every release since is just a pre-release?

Also, any thoughts on submitting .framework.zip to releases for carthage compatibility so that Moya doesn't need to build (there are a lot of dependencies)?

ashfurrow:
@mpurland of ReactiveCocoa? Not sure. 

I have no feelings towards Carthage support. If that's a thing we can do (easily) to help them, sure sounds good :+1: My only requirement is that it gets automated: https://github.com/Moya/Moya/blob/e069bc22e28b2d69cfcfbbae3cb15304d112fbd6/Rakefile#L27-L57

petester42:
the problem that I've encountered with the prebuilt binaries from Carthage is that when a new version of swift/Xcode comes out it has to be recompiled right now. I saw myself using the "-no-use-binaries" option a lot to circumvent that since even though the source didnt have errors it just doesn't work and has to be rebuilt. 

Right now I would advise against distributing prebuilt binaries until that problem with swift is solved.

mpurland:
I've run into that issue as well. I hope it will be addressed soon.

mpurland:
@ashfurrow The above issue about tags and releases was encountered in #332.

It looks like the Rakefile will need to be adjusted to publish releases vs. just tags if you would still only want to cut releases that way.

ashfurrow:
Oh I see, so specifying GitHub instead of Git in Carthage means very different things, one uses GitHub-specific thingies and the other just uses tags? I did not know that – seems silly. Anyway to create a "release" from the command line?

mpurland:
There is a dependency for rakefiles that I've seen referenced that can be used.

mpurland:
I think this is what I saw earlier: https://github.com/mpalmer/github-release

PR #340, opened 2015-12-10
[Rakefile] Use most recent sdk version

petester42:
I found that `rake test` was failing on my machine since I didn’t have the iOS 9.1 simulator installed after updated to Xcode 7.2. Setting the sdk without a version will use the most recent one which should be fine.


ashfurrow:
Cooooool

Issue #338, opened 2015-12-08
Moya's Result type shadows Result's Result type


mhuusko5:
Now that Moya has a Result type, it isn't possible to use Result's Result type in a file that imports Moya
Same thing as happened when Alamofire added a Result type

Might it be better to just add Result as a dependency? It's quite a small/stable one, and unlike usual dependencies, it's *less* troublesome to the end user..

ashfurrow:
Doesn't type inference tell Swift which to use? It's also possible to use Moya.Result. I'm not opposed to adding a dependency if we need it.

petester42:
There is a technical reason we can't. Since alamofire uses its own result we can't use the standard result library because we need to namespace it. Sadly you can't do `Result.Result` as a type for the completion block. Compiler no like. 

Is there a thing like `hiding` in Haskell for swift that I don't know of?

mhuusko5:
Unfortunately as @petester42 mentions, Result.Result doesn't work because of an ongoing Swift quirk where you cant explicitly access a type that has the same name as its module.

So, Moya.Result can be access inferred or explicitly, but if Moya is imported in a file, inferred use of Result is ambiguous/won't compile, but Result.Result can't be accessed either.

And there's no `hiding` or such, as far as I know :(

It's a weird position to be in, because this could be fixed in the language tomorrow, or.. not tomorrow. I personally think it's worth integrating the defacto Result.Result, but I haven't been bitten by adding a dep to a large OSS library before.

petester42:
As for your original issue. You can use Result's Result type just prefix the Moya one like so `Moya.Result` where the two clash.

mhuusko5:
Moya.Result makes it so *those* uses are unambiguous. It doesn't make it so I can use Result.Result. Those instances are still ambiguous (because can't prefix).

ashfurrow:
Sounds like the easiest solution is just to add a dependency to Result. I'm totally OK with that – anyone have time to send a PR?

mhuusko5:
Little busy atm, but should be able to some time this week. Perhaps that will also give some time for others (@Moya/contributors), if anyone has strong opinions about additional dependencies, to chime in.

ashfurrow:
Fab idea. This would also be a great "first time" contribution, too!

petester42:
You might need to do some hacks to make it work. I tried added this dependency when we moved to result but the compiler was not happen. Maybe moving the completion typealias to another file would work. Any file with `import Alamofire` will cause issues.

[...]

PR #344, opened 2015-12-14
Move to antitypical Result type

petester42:
As discussed in #338, this move to using the standard result type instead of creating our own. Due to Alamofire having its own result type I had to move some things around to remove the name clashing. I had to move things to other files since the antitypical result type cannot be prefixed since the name of the framework is `Result` and using `Result.Result` gives a compiler error.

This should work with cocoapods just fine but there seems to be issues with carthage. ReactiveCocoa versions give errors for tvOS and watchOS targets since the minimum deployment target is now 2.1 and 9.1 respectively instead of 2.0 and 9.0. The same thing happens with the RxSwift version for tvOS. Maybe there was a change in Xcode 7.2 that causes this problem or these frameworks don't have the proper setting for the minimum deployment target anymore. This needs investigation.


ashfurrow:
Cool, thanks for looking into this. We could certainly just bump up out minimum deployment versions. Anyone else free to help on this?

Austinate:
I can try to bring proper solution

petester42:
The update to ReactiveCocoa 4 RC1 fixes the reactive cocoa issues. The RxSwift issue still happens but I'm pretty sure its on their side now.

mpurland:
Did bumping to RC1 fix the issue?

petester42:
@mpurland yes it did. thanks. The same issue is happening with RxSwift but I sent them a fix.

petester42:
This should be good to go with the update to rxswift :tada:

ashfurrow:
Looks good to me!

Issue #350, opened 2015-12-17
Stubbed responses still call the network.


crenwick:
When I stub a provider, like
```swift
let SomeProvider = MoyaProvider<SomeAPI>(stubClosure: MoyaProvider.ImmediatelyStub)
```
the code in my app is stubbed fine, however it still calls the API. 

Is there an easy and clear way to not have stubbed responses make any network calls? 

These are typically backends that I'm building simultaneously with the app, so these stubbed calls are now creating unwanted data on my endpoint or creating strange glitches for endpoints that are still under development.

Apologies if this has already been addressed -- I just couldn't find a clear answer in the issues or docs.

wangshengjia:
Before all, there is nothing to apology, and thank you for addressing this issue :100: 

The problem is probably from [here](https://github.com/Moya/Moya/blob/master/Source/Moya.swift#L220), `manager.request(urlRequest)` will call `resume()` immediately.

In my opinion, this is a bug and should be fixed. I'll look forward and propose a PR to fix this.

swizzlr:
That's unusual! And also bad! That shouldn't happen! A PR to fix this would be most welcomed. Hopefully there's a way to retrieve the AF request object without side-effecting.

ashfurrow:
I've opened an issue on Alamofire to get this sorted out. We'll see how it goes.

The problem appears to be how we structure the authentication plugins. Maybe we need to rethink the plugin architecture. Actually calling the network while stubbing seems like a serious issue and we should prioritize.

ashfurrow:
@cnoon has provided valuable feedback: https://github.com/Alamofire/Alamofire/issues/986 I believe we should follow their advice and use a custom manager instead of the shared one (while still allowing devs to pass in their own manager if they want). 

Thoughts? Feels?

wangshengjia:
Hmm, solution is surprisingly simple😂 ... but it do solve our issue.  It's OK for me. I'll update PR #354 if there is no objection.

ashfurrow:
Sounds good, thanks! In the AF issue, they had some guidelines about how to create a manager with good defaults. 

We also need to update our documentation to make it clear that we don't use the shared manager anymore. Since this is a breaking change, the next Moya version will be 6.0.0.

wangshengjia:
Yeah, in short terms, I think it would be suffice to modify ade1589 with default configuration:
```swift
    public final class func DefaultAlamofireManager() -> Manager {
[...]

PR #354, opened 2015-12-22
Fix an issue that stubbed responses still call the network.

wangshengjia:
This PR suppose to fix issue #350 by calling `request.resume()` manually.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,7 @@
 # Next
 
+- **Breaking Change** pass a built-in `DefaultAlamofireManager` as parameter's default value instead of passing the singleton `Alamofire.Manager.sharedinstance` when initialize a `provider`
+
 # 5.3.0
 
 - Updates to RXSwift 2.0.0
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,6 +4,7 @@
 
 - Updates to RXSwift 2.0.0
 - Moves to use Antitypical/Result
+- Fixes issue that stubbed responses still call the network.
 
 # 5.2.1
      
--- a/Source/Moya.swift
+++ b/Source/Moya.swift
@@ -119,7 +119,7 @@ public class MoyaProvider<Target: TargetType> {
 
 public extension MoyaProvider {
     
-    // These functions are default mappings to endpoings and requests.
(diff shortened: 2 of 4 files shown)


Source/Moya.swift
```diff
@@ -150,6 +150,12 @@ public extension MoyaProvider {
     public final class func DefaultRequestMapping(endpoint: Endpoint<Target>, closure: NSURLRequest -> Void) {
         return closure(endpoint.urlRequest)
     }
+
+    public final class func DefaultAlamofireManager() -> Manager {
+        let manager = Alamofire.Manager.sharedInstance
+        manager.startRequestsImmediately = false
```
ashfurrow: Hmm, I'm not in love with the idea of changing Alamofire's singleton's state, but I'm not sure I see another option.

wangshengjia: Yeah.. I see what you mean. Me too actually, I'm not sure about another option. 

Either we find another way to build an `Alamofire.Request` ourself which is obviously not a good idea, or we can just pass a `NSURLRequest` instance into plugin's delegate method instead of `Alamofire.Request`. In latter case, we gain a safer plugin mechanism (as I mentioned [here](https://github.com/Moya/Moya/issues/326#issuecomment-162242355)), but we will lose a little flexibility, it'll break `CredentialsPlugin`.

 ¯\\\_(ツ)\_/¯ , still seeking ...

wangshengjia:
As discussed in #350 , I've updated PR with some more commits. Would you take a look?

Changelog.md
```diff
@@ -1,9 +1,12 @@
 # Next
 
+- **Breaking Change** pass a built-in `DefaultAlamofireManager` as parameter's default value instead of passing the singleton `Alamofire.Manager.sharedinstance` when initialize a `provider`
+
 # 5.3.0
 
 - Updates to RXSwift 2.0.0
 - Moves to use Antitypical/Result
+- Fixes issue that stubbed responses still call the network.
```
ashfurrow: This would go up in the Next section, eh?

wangshengjia: Ah.. yes, my bad. I forgot to move(or remove) this line 😅

Issue #358, opened 2015-12-27
Optional parameters


Skornos:
I have a couple of endpoints with optional parameters specified like this

```
public enum MyAPI {
    ...
    case Content(Int, Int?, String?)
    ...
}
```

The problem is that `public var parameters` expects a type `[String: AnyObject]?` so if I pass optional value I have to unwrap it with `!` operator which causes Exception if the parameter is `nil`.

I was thinking about filtering all the params from the dictionary where the value is `nil` or is there any other way? Like for example specifying `parameters` as `[String: AnyObject?]?`

Thanks.

ashfurrow:
Hmm, good question! I think the filtering idea you have could work, but it's not ideal. Like, Moya should handle this somehow. Let me think about it, maybe someone else has a better idea. In the meantime, filtering is your best bet.

Skornos:
Thanks for such quick answer.

I have found another workaround. Assigning `nil` to initialized dictionary is equal to removing the key from the dictionary so instead of naming all the params in the initializer you create new empty dictionary of type `[String: AnyObject]?` and than you assign each parameter like `parameters["foo"] = nil`.

ashfurrow:
Ah, that makes sense! We should add that as an example to our documentation. Would you like to send a pull request? :wink:

Skornos:
I will have time for PR hopefully anytime soon :+1:

ashfurrow:
Cool, thanks! Take your time, no rush! :christmas_tree:

mpurland:
With a separated out TargetType for each endpoint I take this approach:

```swift
var parameters: [String: AnyObject]? {
    let optionalParameters: [String: AnyObject?] = ["value": nonOptionalValue, "optionalValue": optionalValue]
    return optionalParameters.mapMaybe { $0 }
}
```

where mapMaybe is to take a [String: AnyObject?] to [String: AnyObject]

As from Swiftz:
```swift
func mapMaybe<Value2>(f : Value -> Value2?) -> [Key : Value2]
```

jasl:
I've an idea, but may break design goal of Moya: Providing a non-generic `MoyaProvider`

## `enum` isn't a good way for complex endpoints

For example, there is an endpoint

endpoint: api/v1/topics

param | type | optional | sample | desc
--------- | ----- | ----------- | ----------| ------
offset | Integer | Yes | 20 | 
limit | Integer | Yes | 20 | 
[...]

PR #376, opened 2016-01-20
Proving a refactor way

jasl:
@ashfurrow here's the codes to answer #358  but it's including some other changes. that's change nothing interfaces for end-users but `stubClosure`.

PS: codes may rough, and I'm just test the Demo and pass all cases

This PR including:
- [Break] Pass `StubBehavior` value to provider directly, no need by `stubClosure`
- [Break] `Endpoint` now non-generic
- [Break] Introducing backends for provider. Now provider itself become lightweight, backend do the really job.
- [Break] add toNSMutableRequest() method to Endpoint, deprecate urlRequest property
- [Break] use `ConvertToEndpoint` function to build `Endpoint`, remove `Provider#endpoint`
- Add a non-generic version provider
- `Endpoint` has `parameterEncoding` property but can not be setting except in endpointClosure, let's made it into TargetType, and set a default `.URL` as Moya current does

### DIFF
--- a/Source/Endpoint.swift
+++ b/Source/Endpoint.swift
@@ -12,7 +12,7 @@ public enum EndpointSampleResponse {
 
 
 /// Class for reifying a target of the Target enum unto a concrete Endpoint.
-public class Endpoint<Target> {
+public class Endpoint {
     public typealias SampleResponseClosure = () -> EndpointSampleResponse
 
     public let URL: String
@@ -39,7 +39,7 @@ public class Endpoint<Target> {
     }
 
     /// Convenience method for creating a new Endpoint with the same properties as the receiver, but with added parameters.
-    public func endpointByAddingParameters(parameters: [String: AnyObject]) -> Endpoint<Target> {
+    public func endpointByAddingParameters(parameters: [String: AnyObject]) -> Endpoint {
         var newParameters = self.parameters ?? [String: AnyObject]()
         for (key, value) in parameters {
             newParameters[key] = value
@@ -49,7 +49,7 @@ public class Endpoint<Target> {
     }
 
     /// Convenience method for creating a new Endpoint with the same properties as the receiver, but with added HTTP header fields.
-    public func endpointByAddingHTTPHeaderFields(httpHeaderFields: [String: String]) -> Endpoint<Target> {
+    public func endpointByAddingHTTPHeaderFields(httpHeaderFields: [String: String]) -> Endpoint {
         var newHTTPHeaderFields = self.httpHeaderFields ?? [String: String]()
         for (key, value) in httpHeaderFields {
             newHTTPHeaderFields[key] = value
@@ -59,7 +59,7 @@ public class Endpoint<Target> {
     }
     
     /// Convenience method for creating a new Endpoint with the same properties as the receiver, but with another parameter encoding.
-    public func endpointByAddingParameterEncoding(newParameterEncoding: Moya.ParameterEncoding) -> Endpoint<Target> {
+    public func endpointByAddingParameterEncoding(newParameterEncoding: Moya.ParameterEncoding) -> Endpoint {
         
         return Endpoint(URL: URL, sampleResponseClosure: sampleResponseClosure, method: method, parameters: parameters, parameterEncoding: newParameterEncoding, httpHeaderFields: httpHeaderFields)
     }
@@ -77,7 +77,7 @@ extension Endpoint {
 }
 
 /// Required for making Endpoint conform to Equatable.
-public func ==<T>(lhs: Endpoint<T>, rhs: Endpoint<T>) -> Bool {
+public func ==(lhs: Endpoint, rhs: Endpoint) -> Bool {
     return lhs.urlRequest.isEqual(rhs.urlRequest)
 }
 
--- a/Source/Endpoint.swift
+++ b/Source/Endpoint.swift
@@ -57,17 +57,27 @@ public class Endpoint {
 
         return Endpoint(URL: URL, sampleResponseClosure: sampleResponseClosure, method: method, parameters: parameters, parameterEncoding: parameterEncoding, httpHeaderFields: newHTTPHeaderFields)
     }
-    
+
     /// Convenience method for creating a new Endpoint with the same properties as the receiver, but with another parameter encoding.
     public func endpointByAddingParameterEncoding(newParameterEncoding: Moya.ParameterEncoding) -> Endpoint {
-        
+
         return Endpoint(URL: URL, sampleResponseClosure: sampleResponseClosure, method: method, parameters: parameters, parameterEncoding: newParameterEncoding, httpHeaderFields: httpHeaderFields)
     }
 }
 
 /// Extension for converting an Endpoint into an NSURLRequest.
 extension Endpoint {
+    // TODO: Compatibility, this should better be a method, not a property
     public var urlRequest: NSURLRequest {
+        let request: NSMutableURLRequest = NSMutableURLRequest(URL: NSURL(string: URL)!)
+        request.HTTPMethod = method.rawValue
+        request.allHTTPHeaderFields = httpHeaderFields
+
+        return parameterEncoding.toAlamofire.encode(request, parameters: parameters).0
+    }
+
+    // let's return a mutable request, that could be modify before it's be performed
+    func toNSMutableURLRequest() -> NSMutableURLRequest {
         let request: NSMutableURLRequest = NSMutableURLRequest(URL: NSURL(string: URL)!)
         request.HTTPMethod = method.rawValue
         request.allHTTPHeaderFields = httpHeaderFields
@@ -84,6 +94,6 @@ public func ==(lhs: Endpoint, rhs: Endpoint) -> Bool {
 /// Required for using Endpoint as a key type in a Dictionary.
 extension Endpoint: Equatable, Hashable {
     public var hashValue: Int {
-        return urlRequest.hash
+        return self.toNSMutableURLRequest().hash
     }
 }
--- a/Source/Moya.swift
+++ b/Source/Moya.swift
@@ -55,10 +55,10 @@ public class MoyaProvider<Target: TargetType> {
     public let plugins: [PluginType]
     
     /// Initializes a provider.
-    public init(endpointClosure: EndpointClosure = MoyaProvider.DefaultEndpointMapping,
-        requestClosure: RequestClosure = MoyaProvider.DefaultRequestMapping,
-        stubClosure: StubClosure = MoyaProvider.NeverStub,
-        manager: Manager = MoyaProvider<Target>.DefaultAlamofireManager(),
+    public init(endpointClosure: EndpointClosure = DefaultEndpointMapping,
+        requestClosure: RequestClosure = DefaultRequestMapping,
+        stubClosure: StubClosure = NeverStub,
+        manager: Manager = DefaultAlamofireManager(),
         plugins: [PluginType] = []) {
             
             self.endpointClosure = endpointClosure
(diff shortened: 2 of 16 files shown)


jasl:
This PR mainly extract duties from `Provider`, it's has too many jobs before: build `Endpoint`s, do requests, decide how to mock requests, that's made things complex, like mock requests, it's no need in production.

Now, Provider only handle build `Endpoint` and as a interface to end-user to requesting targets, let backend do the really job, it's a protocol `MoyaProviderBackendType` so we can replace it easily.
There an additional advantage, we separate the really requests and mock requests to specified backends, and in test env, use mock backend, and use normal backend in production. and Moya now could decoupling with Alamofire.

Then, the provider become lightweight, so maintain a non-generic version become possible.

jasl:
Endpoint no need to be generic, it's a intermediate product from `Target` to `Request`, when building `Endpoint` by `Target`, all job relates Target should be done in `endpointClosure`, like add authenticate header as your mentioned, so in `requestClosure` no need to know the `Endpoint` where's from, so Endpoint no need to generic.

colinta:
While I really like a lot of what I see here, I see that not only does this refactor the Provider code (and adding the Backend looks like a good idea), it also reverts the StubClosure back to the StubBehavior.

In short, I think there are too many breaking changes to have this considered as just a refactor - merging this would require a huge changelog entry.

I would rather see this broken into 2 or 3 PRs: One to just introduce the `MoyaBackend`, another to discuss reverting `StubClosure` to `StubBehavior`, and a third to move `Moya.swift` code into `Provider.swift`.  Does any one else feel like this PR is just *too huge* to get merged in?

jasl:
@colinta 
I'm a new guy to learning Swift and iOS developing, my project using Moya and I feel a little inconvenient, have described at #358 

please don't merge it I'm no confident in my Swift skill, and it's not polish yet.

This PR is mainly to prove, I can made a non-generic provider and keeping interfaces compatible, and is it keeping Moya's philosophy? does the new provider design more reasonable?

ashfurrow:
Aye, @colinta I encouraged @jasl in https://github.com/Moya/Moya/issues/358 to open a PR to discuss some of the changes. I'll have time hopefully to look it over tomorrow!

colinta:
Yup, to be clear, I really *like* the parts related to making the provider non-generic.  My worry is that the other changes introduced here will make it hard to keep the discussion focused on just that one thing.  Thanks for being open, though, and I look forward to seeing where this goes!

jasl:
@colinta No matter, after discussion I can cherry-pick the good part

jasl:
just adding two changes:
- `Endpoint` is a struct instead of class
- `TargetType` has `toEndpoint()` to build `Endpoint`, `Provider#endpoint` is no used has removed

jasl:
And, I have a question: Why modify `Endpoint` returns a new `Endpoint` not itself? seems no worry about side effect here.

jasl:
@ashfurrow 
Here is the last wave I can imagine at this time:
- `Endpoint` has `parameterEncoding` property but can not be setting except in `endpointClosure`, let's made it into `TargetType`, and set default is `.URL` as Moya currently does
- [Break] add `toNSMutableRequest method()` to `Endpoint`, deprecate `urlRequest` property

Here I have another question:
[How does it work to make Endpoint to Request asynchonous?](https://github.com/Moya/Moya/commit/847bf33f76e9743dc2e04df629bbe9803de92c24#diff-ab556bf607d84c3809645f443db9b8ceR151)

jasl:
actually backend still needs to know plugins, target, endpoint, but it's should only know `Request`, I'm looking for solution

jasl:
@colinta 
could you tell me when we need to customize `stubClosure`?
I'm trying to keeping `stubClosure` as current Moya does, but I found it seems impossible.
If you can told me usage of `stubClousure` I may find another way to keep that

ashfurrow:
@jasl Yeah, sometimes we want to stub immediately and sometimes we want a delay. Always in unit tests, of course.

ashfurrow:
Could be a use-case for an `enum`:

```swift
struct StubBehavior {
    case .NoStubbing
    case .ImmediateStubbing
    case .CustomStubbing(closure: T -> Int)
}
```

Or something like that. I don't feel like the overhead of the current usage is bad at all, not sure how much extra utility we'd get changing the stub parameters (again lol) given everyone would have to update their code.

Source/Moya.swift
```diff
@@ -21,7 +21,33 @@ public protocol TargetType {
...
     var sampleData: NSData { get }
+
+    func toEndpoint() -> Endpoint
+}
+
+public extension TargetType {
+    func toEndpoint() -> Endpoint {
+        let url = self.baseURL.URLByAppendingPathComponent(self.path).absoluteString
+        return Endpoint(URL: url, sampleResponseClosure: {.NetworkResponse(200, self.sampleData)}, method: self.method, parameters: self.parameters, parameterEncoding: parameterEncoding)
+    }
```
ashfurrow: But in a protocol extension, if we only have a reference to the type as `TargetType`, won't this _always_ execute this instead of the function specified in the actual type?

Source/Endpoint.swift
```diff
@@ -74,16 +75,25 @@ extension Endpoint {
 
         return parameterEncoding.toAlamofire.encode(request, parameters: parameters).0
     }
+
+    // let's return a mutable request, that could be modify before it's be performed
+    func toNSMutableURLRequest() -> NSMutableURLRequest {
```
ashfurrow: I think we should probably just get Alamofire to create the `NSURLRequest`, then call `mutableCopy` on it. Does that make sense?

jasl: the Moya current flow is `Endpoint` -> `NSURLRequest` (by `RequestClosure` definition) then `NSURLRequest` -> `AlamoRequest` (by [sendRequest](https://github.com/Moya/Moya/blob/master/Source/Moya.swift#L166))

`toNSMutableURLRequest` is used in `RequestClosure` [for example](https://github.com/Moya/Moya/blob/master/docs/Endpoints.md#request-mapping) 

I just keeping the behavior, and I think that's fine, I've read Alamofire's Request, most except authenticate are no use on preparation (authenticate can be done by plugin), and that's help with decoupling with Alamofire (although it's no necessary)

Source/Moya.swift
jasl: Yes, that's my mistake, maybe we should create a function `convertToEndpoint` instead.
anyway, just don't let `Provider` do this.

Source/Endpoint.swift
ashfurrow: Makes sense, thanks for the clarification.

Source/Moya.swift
```diff
@@ -21,7 +21,33 @@ public protocol TargetType {
...
+}
+
+public extension TargetType {
+    var parameterEncoding: ParameterEncoding { return .URL }
+    var sampleData: NSData { return "".dataUsingEncoding(NSUTF8StringEncoding)! }
```
ashfurrow: Same problem as above here, though I really don't like the idea of allowing developers to bypass this anyway. Moya has a strong opinion on unit testing: it's a good idea and everyone should do it. If someone wants to add something like this to their `TargetType` type, that's fine, but they should feel bad about it :stuck_out_tongue:

ashfurrow:
So yeah, lots of interesting ideas in here, some I like and some I don't. I'm not sure I'm the best arbiter of what should/shouldn't be considered though – Moya is too close to me and I don't think I can be impartial.

@Moya/contributors If you have time, taking a look through this PR and grabbing ideas to consider would be :100: Thanks again @jasl – I think from this PR we can break out into smaller issues to consider the pros and cons of each idea individually. Does that make sense?

jasl:
@ashfurrow 
I've consider this, This behavior can be keep, and moving to `StubBehavior`, even we can enhence it like `mockEndpoint(Endpoint, response)` I think that's more flexible on testing, better than 'sampleDate` (I don't have critical reason, but testing code in releasing codebase seem weird to me)

ashfurrow:
Actually, I take back my comment about always stubbing in unit tests :speak_no_evil: [Eidolon](https://github.com/artsy/eidolon) uses stubbed responses for an offline mode available to OSS developers who don't have access to our API, as well as Artsy developers when they're offline. It's often useful for development, though I agree in _production_ it does seem weird.

jasl:
@ashfurrow 
`StubBackend` could do stubbing responses, just using it in *testing* and *development*, and I've an idea could move all codes for stubbing, but that's another discussion.

At last, thank you for reviewing my huge PR, and thank for your blog and other open source works!

jasl:
Just reverting all about `TargetType` default values, and update PR description.

Source/Moya.swift
jasl: has revert these default value, but keeping `parameterEncoding`, I've not updated test cases yet

colinta:
Taking another stab at analyzing this, but again it's tricky because there are *many* ideas all getting thrown into one PR.  I would still love to see this broken out into 2 (or more) PRs.  So far, I've identified these changes:

1. Refactor Endpoint so that it doesn't neet `<Target>`.
   - This was the original goal, wasn't it?  And it looks like this would be a straightforward change.
2. Adding MoyaBackend - this looks like a nice way to separate the "real vs stubbed" response behavior.
   - Adds `Source/Backend.swift`, and refactors `Source/Moya.swift`
   - btw, looks like `MoyaProviderStubBackend` doesn't use or need the `manager` instance
3. Revert `StubClosure` to `StubBehavior`.  The most important reason, I think, for the `StubClosure` is because the `stubbedData` needs to be called lazily.  If this can be accomplished with `StubBehavior`, I think I'd be fine with it.

colinta:
###### Some more questions

- What does `MoyaCommonProvider` do? I don't see this class used anywhere, but it is defined in `Provider.swift`.
- `DefaultCommonEndpointMapping` and `DefaultEndpointMapping` do the same thing, can this be simplified to just having `DefaultEndpointMapping`?
- `toNSMutableURLRequest` is only used to calculate a hash value (```public var hashValue: Int { return self.toNSMutableURLRequest().hash```), why not use `self.urlRequest` there?
- I see that `CancellableToken` has been made public, but I think it can remain private if you change the return type to `Cancellable` in `MoyaProviderBackend`

###### Most importantly

I labeled this `enhancement`, but then that begged the question: *what* is being enhanced here?  I see a lot of changes (too many in my opinion, to be very honest), but I can't point at any one thing that is greatly *improved* by these changes. @jasl what is the most important feature that is being added?  @ashfurrow what do you see as the best ideas here?

jasl:
First, I apologize for my terrible English, I'll try my best to explain my motivation and my idea.

## Motivation

I'm looking for a solution to build a Restful endpoints abstraction for [my learning project](https://github.com/jasl/RubyChinaAPP) which is a forum's client, by read many articles and libraries I found Moya seems fit my needs.

When I progressing my plan, I meet some difficulties:
- There are 38 endpoints need to declare
- Some endpoints are complex that may has optional value, mutually exclusive field, required fields
- too many endpoints in ONE enum made it to long, difficult to organize

(PS: My company's app has about 100+ endpoints, and more complex than mine)

So, I tried to using `struct` to [organize endpoints](https://github.com/jasl/RubyChinaAPP/blob/master/RubyChinaApp/Libraries/Endpoints/Topics.swift), It looks better, I use nested struct to simulate namespace (that mapping Restful resources), and resources' path can [reusable](https://github.com/jasl/RubyChinaAPP/blob/master/RubyChinaApp/Libraries/Endpoints/Topics.swift#L12).

But, there is an another problem, Moya's provider is generic, and it's designed for endpoints which declared in a enum, like [Eidolon](https://github.com/artsy/eidolon/blob/master/Kiosk/App/Networking/ArtsyAPI.swift). If I still using `struct`, the code should like:
```swift
let provider = MoyaProvider<TheSite.Topics.Listing>()
provider.request(TheSite.Topics.Listing) { res in
  // callback
}
```
That's weird, can't be reused, and difficult to building a higher abstraction for provider to encapsulating OAuth2 signing and some configuration for Request [[sample]](https://github.com/jasl/RubyChinaAPP/blob/master/RubyChinaApp/Libraries/Provider.swift)

So the only problem is Provider's generic contraint, and it's possible to remove without side effect, so I copy Moya 5.3 to my project, and [just remove the generic contraint](https://github.com/jasl/RubyChinaAPP/blob/master/NetworkAbstraction/NetworkAbstraction/APIProvider.swift#L52), all problem solved perfectly!

Then, I wrote my thinking in #358, and @ashfurrow replied me.

## The goal

The goal is simple, but complex to reach, that **making a non-generic Provider possible**.

## How to

By theory, most of Provider's codes should be shared, but I tried many ways, failed.

Finally, I found a way, to short:

- Clearify data flow that how Provider sending a request
- Made Provider light-weight
- Introducing Backend to handle request

## About StubClousure

That's my mistake, because I promise keeping current APIs, In new Provider design [I need to decide what backend should be use](https://github.com/Moya/Moya/pull/376/files#diff-b96c1bf37472f8e8d65f16aaf540ed10R35), and `stubClousure` can't be evaluate here.

Actually, `stubClousure` should be setting to `StubBackend` directly like `StubBackend(stubClousure: {})`, that's make sense.

Anyway, It's should break current Provider's initializer.

## Answer questions

### What does MoyaCommonProvider do? I don't see this class used anywhere, but it is defined in Provider.swift.

`MoyaCommonProvider` is a non-generic Provider, you can use it when you using struct or class to organizing your endpoints.

### DefaultCommonEndpointMapping and DefaultEndpointMapping do the same thing, can this be simplified to just having DefaultEndpointMapping?

DefaultCommonEndpointMapping and DefaultEndpointMapping is do the same thing, but DefaultEndpointMapping can't be used for non-generic Provider, and some people want [do something works which needs to know TargetType's type](https://github.com/Moya/Moya/blob/master/docs/Endpoints.md#from-target-to-endpoint-)

But, I think `DefaultEnpointMapping` is no need, you can just override `toEndpoint` (It's declared in `TargetType`) to do the same job.

### I see that CancellableToken has been made public, but I think it can remain private if you change the return type to Cancellable in MoyaProviderBackend

I don't have idea about this, I think made its public may help for create new Backend in project?

jasl:
ping @colinta

Source/Moya.swift
jasl: @ashfurrow I've do some research in Playground, `toEndpoint()` is safe.

```swift
protocol FooType {
    func foo() // that's important
}

extension FooType {
    func foo() {
        print("foo type")
    }
}

class FooClass: FooType {
    func foo() {
        print("foo class")
    }
}

struct FooStruct: FooType {
    func foo() {
        print("foo struct")
    }
}

func bar(foo: FooType) {
    let fooType = foo as FooType
    fooType.foo()
}

let foo = FooClass()
foo.foo() //=> foo class

(foo as FooType).foo() //=> foo class

bar(foo) //=> foo class

let fooS = FooStruct()
fooS.foo() //=> foo struct

(fooS as FooType).foo() //=> foo struct
```

jasl: It seems by overriding can do the same job in `endpointClosure`

ashfurrow:
Hi again @jasl, we've created a first-party approach in #427 that I think might work. Does that satisfy your needs?

jasl:
It seems still made `MoyaProvider` can't be reused, it specify **the** struct type

Actually we can do the same thing before that PR

```swift
let provider = MoyaProvider<StructAPI>()
provider.request(StructAPI()) { res in
  // callback
}
```
now

```swift
let provider = MoyaProvider<StructTarget<StructAPI>>()
provider.request(StructTarget(StructAPI())) { res in
  // callbace
}
```

That's not simplify our codes, but make the things unclear(wrap `StructAPI` into `StructTarget`)

ashfurrow:
Ah, I see. So you're looking to use _any_ struct that conforms to the `TargetType`, not just a specific one. That makes sense. I've opened https://github.com/Moya/Moya/pull/430 to try and fix that. Would love your feedback :cake: 

We need to balance the needs of different users, and while I understand wrapping the struct in an enum like this may not be the best approach from everyone's perspective, we've had to make compromises. I've been working with other maintainers for weeks on this issue and I feel like this is the best approach. If you like, I'd be happy to have a Skype call and chat about it, and hear more about your ideas to make Moya even better. Thanks again for the PR :bow:

jasl:
@ashfurrow 

My English speaking quite badly :scream:  My Skype is my email `jasl9187 at hotmail.com`, I'll try to explain my thoughts as I can.

Actually, after this PR, I forked Moya and build my own version for research [MoyaX] (I'm sorry to rename the project, I'm not want to divide, just because that's can made me away from compatibility concerns)(https://github.com/jasl/moyax).

jasl:
I polished this PR in my fork, and I fork eidolon with MoyaX works quite well, [codes here](https://github.com/jasl/eidolon/tree/MoyaX).
PS: It's may not pass compiling because I force push MoyaX master, and not change test cases, they're too big.

ashfurrow:
Don't worry about renaming at all, it's open source because we want people to experiment however they like! I'll add you on Skype and see when you're free.

ashfurrow:
After chatting with @jasl today, I've moved their feedback into #434 to discuss point-by-point. We'll borrow the ideas we can and bring them into Moya one-at-a-time, so I'm going to close this PR. However, that doesn't mean we should stop discussing the ideas. There are some really neat ones here, and I've asked in #434 that we list specific ideas to bring over, so please continue the conversation there :bow:

PR #396, opened 2016-02-01
Updated Cartfile and Moya.podspec to use RAC4 release 

Austinate:
Only problematic part i see now — only 4.0.0 RC2 is available at Pods Spec repo


Austinate:
Also probably we should bump Moya version since RC4 update but i'm not sure yet, so decided to delay that change

ashfurrow:
Cool. This is only failing because the RAC podspec wasn't pushed until like an hour ago. I'm going to re-run CI.

Austinate:
:+1:

ashfurrow:
Cool, this has been released as 6.1.3 on CocoaPods and as a GitHub release.

PR #401, opened 2016-02-05
Adds response computed variable

ashfurrow:
This is useful when you just need the (possible) network response but don't care about the specific Error (see [this example](https://github.com/artsy/eidolon/pull/584/files#diff-ce82ff0f55f023dbcf5e8ab92d76c934R90) of usage).

It feels silly to have to write it, but it makes sense. May as well only write the silly code once :smile:

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,7 @@
 # Next
 
+- Adds `response` computed property to `Error` type, which yields a Response object if available.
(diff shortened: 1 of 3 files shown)


sunshinejr:
:+1:

ashfurrow:
Thanks!

PR #405, opened 2016-02-11
Updates example with fix to String extension by changing to Optional

Legoless:
The `stringByAddingPercentEncodingWithAllowedCharacters` method of `String` returns an optional, thus the compilation fails using the provided example. Changed the `URLEscapedString` property to also return an optional.

### DIFF
--- a/docs/Examples.md
+++ b/docs/Examples.md
@@ -18,8 +18,8 @@ Let's take a look at what that might look like.
 
 ```swift
 private extension String {
-    var URLEscapedString: String? {
-        return self.stringByAddingPercentEncodingWithAllowedCharacters(NSCharacterSet.URLHostAllowedCharacterSet())
+    var URLEscapedString: String {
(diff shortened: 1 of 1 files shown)


sunshinejr:
Hey! Thanks for the PR! 🎉

You are right, it returns `Optional` and we should change it, since it is compile-time error. Although in the Demo project [we actually use it a little bit differently](https://github.com/Moya/Moya/blob/master/Demo/Demo/GitHubAPI.swift#L21-L25). I think we should stick to one version and depending on which one we choose, make changes to Docs & Demo respectively.

ashfurrow:
Agreed, making it the same as the Demo project would probably be best I think (that's how I use it in production ¯\\_(ツ)_/¯). Thanks @Legoless!

Legoless:
Updated to match the Demo.

sunshinejr:
Perfect! thanks again @Legoless  🎉

Issue #403, opened 2016-02-10
Add http body


ivanruizscm:
I have an endPoint:

```swift
var endpointClosure = { (target: MyCustomApi) -> Endpoint<MyCustomApi> in
    let endpoint: Endpoint<MyCustomApi> = Endpoint<MyCustomApi>(
        URL: url(target),
        sampleResponseClosure: {.NetworkResponse(200, target.sampleData)},
        method: target.method,
        parameters: target.parameters
    )
    switch target {
    case .Person(let json):
        //How to add HTTPBody here?  endpoint.urlRequest is not mutable, so i can't change the request.
    default:
        return endpoint
    }
}

enum NextVisitApi {
    case Person(someJSON: JSON)
}
```

ashfurrow:
Typically HTTP bodies come from parameters, looks like you have one that's specified in `JSON`. So to set the HTTP body to the json, you just need to tell Moya which parameter encoding to use. 

```swift
var endpointClosure = { (target: MyCustomApi) -> Endpoint<MyCustomApi> in
    let endpoint: Endpoint<MyCustomApi> = Endpoint<MyCustomApi>(
        URL: url(target),
        sampleResponseClosure: {.NetworkResponse(200, target.sampleData)},
        method: target.method,
        parameters: target.parameters
    )
    switch target {
    case .Person(let json):
        return endpoint.endpointByAddingParameterEncoding(.JSON)
    default:
        return endpoint
    }
}
```

And that should work. You can check out the different options for parameter encoding [here](https://github.com/Moya/Moya/blob/aece4ebe3ad3cab946d8df93de4ee72c7024c8e7/Source/Moya%2BAlamofire.swift#L9).

Let me know if that helps :+1:

Austinate:
Also i'd like to add [this link](https://github.com/Moya/Moya/blob/23c4d506c5be9b94d9ecf16cb1f70893ad0b4f41/docs/Endpoints.md#request-mapping)
Hope it will be helpful

ashfurrow:
Yeah, good point! The request-mapping is later-stage stuff in the [pipeline](https://github.com/Moya/Moya/tree/master/docs#documentation), but it's there that I'll often play around with stuff I can't get working at higher levels.

ivanruizscm:
This is correct?

var endpointClosure = { (target: NextVisitApi) -> Endpoint<NextVisitApi> in
    let endpoint: Endpoint<NextVisitApi> = Endpoint<NextVisitApi>(
[...]

PR #406, opened 2016-02-12
endpointByAdding as a convenience method for long method chaining.

sunshinejr:
So after seeing issue #403 and  [this](https://github.com/Moya/Moya/issues/403#issuecomment-182784469) piece of code, and especially:

```swift
return endpoint.endpointByAddingParameterEncoding(.URL).endpointByAddingHTTPHeaderFields(["Content-Type":"application/x-www-form-urlencoded; charset=UTF-8"])
```

I decided to make a convenience method that we can call without chaining it. 
Please let me know if there is something I can improve. Thanks! 🎉

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,7 @@
 # Next
 
 - Adds `response` computed property to `Error` type, which yields a Response object if available.
+- Adds convenience `endpointByAdding` method.
 
 # 6.1.3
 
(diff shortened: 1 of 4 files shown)


ashfurrow:
Cool! I love this idea, and the implementation is :100: Good use of `nil` to represent a "missing value" :+1:

PR #404, opened 2016-02-10
Added URLEncodedInURL to ParameterEncoding

pedrofjfmartins:
POST requests were forced to insert parameters in the body and did not have the option to encode them in the query string.

PR #409, opened 2016-02-15
Parameter encoding

sunshinejr:
> So this is only a proposition since I don't know if there is much against using `ParameterEncoding` as a `typealias` compared to our own enum.

In this PR I tried to remove explicitly typed enum for `ParameterEncoding` that only is a bridge between us and `Alamofire`. I think that `public typealias` sounds fine (at least for me :bomb:) and I've made few changes. 

From positives, we can forget about PRs like #404 and upgrading our code everytime `Alamofire` adds something. What was left, however, was the equality operator in our unit testing. BUT! Thanks to [this stackoverflow answer](http://stackoverflow.com/a/34769898/2340462) we can forget (at least for now) about it. If there are some negative aspects, please let me now.

I'm really eager to hear from your guys! Thanks 🎉


ashfurrow:
Interesting! I had not thought of just using a `typealias`! My only concern would be making sure that Moya users don't have to be aware of Alamofire, but the unit tests don't `import Alamofire`, so it _should_ be fine. Cool!

Anyone else see any issues? Otherwise, I'd just want to make sure we document the change in the changelog.

sunshinejr:
@ashfurrow Exactly my thoughts, was worried that user would have to import Alamofire, but it seems fine. This was just a suggestion, because I thought that there was a specific reason we didn't try typealias yet. So if we end up doing this, I will definitely write up the changelog + in my opinion we could remove the AlamofireMoyaMappingSpec test since this should be now done by the Alamofire itself.

sunshinejr:
Okay I've updated the PR to prepare for merge 😇

Changelog.md
```diff
@@ -3,6 +3,7 @@
 - Adds `response` computed property to `Error` type, which yields a Response object if available.
 - Added URLEncodedInURL to ParameterEncoding.
 - Adds convenience `endpointByAdding` method.
+- **Breaking Change** Remove our own implementation of `ParameterEncoding` and make it a `public typealias` of `Alamofire.ParameterEncoding`.
```
ashfurrow: Is it a breaking change? No one could have been relying on the underlying value of our enum because it didn't _have_ one, and we've kept the name the same, so existing extensions anyone had would be fine. Thoughts?

ashfurrow: I suppose _if_ someone was overriding the `Endpoint` class or otherwise relying on access to the underlying Alamofire functionality... I just don't want to release a major version unnecessarily.

sunshinejr: I was suggested by the removed `toAlamofire()` function which our `ParameterEncoding` had. 🤔But it was internal so I'm fine with removing the breaking change alias ✌️

ashfurrow: Yeah, internal means no one could have been using it. I vote to remove the "breaking change" warning.

sunshinejr: Agreed ✌️Removed the warning.

ashfurrow:
Awesome, thanks so much!!

PR #420, opened 2016-03-04
Fix endpoint setup when adding a header or parameters without one of them

### LINKED ISSUE #419: Updating to 6.2 makes the request duration really longer and (sometimes) never call the completion
After updating to Moya 6.2, I'm getting some really weird issues:
- Some requests never come back (Charles, a HTTP tracker, get a response 200), but Moya never call the completion. Apparently, that happens only with the GET method, the POST request is responding well
- The request duration has increased a lot: 58ms (6.1.3) -> 10s (6.2). These numbers are from Charles

I confirm that these both bugs has been creating in the 6.2. Rolling back to v6.1.3 solve both of them.
I also look at the CHANGELOG, but didn't see any breaking changes that could create these weird behaviours. 

Any ideas?

tbaranes:
That PR fixed #419. Apparently, set an empty array as an endpoint parameters creates all this weird bugs. I have strictly no idea why, and if one you has an answer, I'd love to hear it! :monkey_face: 

The small changes I did to fix my both issues (related to #419) is just to let the current parameters / headers if we aren't adding any values to them.

I make some tests, didn't find anything that could break. What do  you think?

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,7 +1,5 @@
 # Next
 
-# 6.1.1
-
 - Fixed endpoint setup when adding `parameters` or `headers` when `parameters` or `headers` or nil
 
 # 6.2.0
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,9 @@
 # Next
 
+# 6.1.1
+
+- Fixed endpoint setup when adding `parameters` or `headers` when `parameters` or `headers` or nil
+
 # 6.2.0
 
 - Adds `response` computed property to `Error` type, which yields a Response object if available.
--- a/Source/Endpoint.swift
+++ b/Source/Endpoint.swift
@@ -62,23 +62,25 @@ public class Endpoint<Target> {
     }
     
     private func addParameters(parameters: [String: AnyObject]?) -> [String: AnyObject]? {
-        var newParameters = self.parameters
-        if let unwrappedParameters = parameters where unwrappedParameters.count > 0 {
-            newParameters = self.parameters ?? [String: AnyObject]()
-            unwrappedParameters.forEach { (key, value) in
-                newParameters?[key] = value
-            }
+        guard let unwrappedParameters = parameters where unwrappedParameters.count > 0 else {
+            return self.parameters
+        }
+        
+        var newParameters = self.parameters ?? [String: AnyObject]()
+        unwrappedParameters.forEach { (key, value) in
+            newParameters[key] = value
         }
         return newParameters
     }
     
     private func addHTTPHeaderFields(headers: [String: String]?) -> [String: String]? {
-        var newHTTPHeaderFields = self.httpHeaderFields
-        if let unwrappedHeaders = headers where unwrappedHeaders.count > 0 {
-            newHTTPHeaderFields = self.httpHeaderFields ?? [String: String]()
-            headers?.forEach { (key, value) in
-                newHTTPHeaderFields?[key] = value
-            }
+        guard let unwrappedHeaders = headers where unwrappedHeaders.count > 0 else {
+            return self.httpHeaderFields
+        }
+        
+        var newHTTPHeaderFields = self.httpHeaderFields ?? [String: String]()
+        headers?.forEach { (key, value) in
+            newHTTPHeaderFields[key] = value
         }
         return newHTTPHeaderFields
     }
--- a/Source/Endpoint.swift
+++ b/Source/Endpoint.swift
@@ -79,7 +79,7 @@ public class Endpoint<Target> {
         }
         
         var newHTTPHeaderFields = self.httpHeaderFields ?? [String: String]()
-        headers?.forEach { (key, value) in
+        unwrappedHeaders.forEach { (key, value) in
             newHTTPHeaderFields[key] = value
         }
         return newHTTPHeaderFields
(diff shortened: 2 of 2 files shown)


Source/Endpoint.swift
```diff
@@ -55,20 +55,33 @@ public class Endpoint<Target> {
...
+        let newHTTPHeaderFields = addHTTPHeaderFields(httpHeaderFields)
         let newParameterEncoding = parameterEncoding ?? self.parameterEncoding
-        
         return Endpoint(URL: URL, sampleResponseClosure: sampleResponseClosure, method: method, parameters: newParameters, parameterEncoding: newParameterEncoding, httpHeaderFields: newHTTPHeaderFields)
     }
+    
+    private func addParameters(parameters: [String: AnyObject]?) -> [String: AnyObject]? {
+        var newParameters = self.parameters
+        if let unwrappedParameters = parameters where unwrappedParameters.count > 0 {
+            newParameters = self.parameters ?? [String: AnyObject]()
```
sunshinejr: Seems like we are assigning 2x `self.parameters` to the `newParameters` variable in one case. I would just assign new array to it if its nil, otherwise do nothing :panda_face:

Source/Endpoint.swift
```diff
@@ -55,20 +55,33 @@ public class Endpoint<Target> {
...
+            unwrappedParameters.forEach { (key, value) in
+                newParameters?[key] = value
```
sunshinejr: In that case we are _sure_ we have the `newParameters` array (via check few lines above), so I would use `!` (but it's me) :smiling_imp:

Source/Endpoint.swift
```diff
@@ -55,20 +55,33 @@ public class Endpoint<Target> {
...
+            }
+        }
+        return newParameters
+    }
+    
+    private func addHTTPHeaderFields(headers: [String: String]?) -> [String: String]? {
+        var newHTTPHeaderFields = self.httpHeaderFields
+        if let unwrappedHeaders = headers where unwrappedHeaders.count > 0 {
+            newHTTPHeaderFields = self.httpHeaderFields ?? [String: String]()
```
sunshinejr: Same as above with 2x assigning.

Source/Endpoint.swift
```diff
@@ -55,20 +55,33 @@ public class Endpoint<Target> {
...
+            headers?.forEach { (key, value) in
```
sunshinejr: We can use `unwrappedHeaders` here :)

Source/Endpoint.swift
tbaranes: True, but I prefer not using the `!` if that's not mandatory :innocent:

Source/Endpoint.swift
tbaranes: If we use `unwrappedHeaders`, we have to cast the value in `String`. That's why I make the choice of not using it, but I can change it if you prefer

sunshinejr:
Hey! 

First of all, thanks for the time spent on resolving the issue! Helps a lot 🙌 I've made some small comments, but other than that we are good to go!

Source/Endpoint.swift
tbaranes: Right, I'm gonna improve that point

Source/Endpoint.swift
sunshinejr: Sure, no problem with that!

Source/Endpoint.swift
tbaranes: Nice catch, gonna change this!

tbaranes:
Just edit following your comments. Let me know if you need anything else!

Source/Endpoint.swift
```diff
@@ -55,19 +55,34 @@ public class Endpoint<Target> {
...
+        let newHTTPHeaderFields = addHTTPHeaderFields(httpHeaderFields)
+        let newParameterEncoding = parameterEncoding ?? self.parameterEncoding
+        return Endpoint(URL: URL, sampleResponseClosure: sampleResponseClosure, method: method, parameters: newParameters, parameterEncoding: newParameterEncoding, httpHeaderFields: newHTTPHeaderFields)
+    }
+    
+    private func addParameters(parameters: [String: AnyObject]?) -> [String: AnyObject]? {
+        guard let unwrappedParameters = parameters where unwrappedParameters.count > 0 else {
+            return self.parameters
+        }
+
```
sunshinejr: Perfect 💯

Source/Endpoint.swift
```diff
@@ -55,19 +55,34 @@ public class Endpoint<Target> {
...
+    }
+    
+    private func addHTTPHeaderFields(headers: [String: String]?) -> [String: String]? {
+        guard let unwrappedHeaders = headers where unwrappedHeaders.count > 0 else {
+            return self.httpHeaderFields
+        }
         
         var newHTTPHeaderFields = self.httpHeaderFields ?? [String: String]()
-        httpHeaderFields?.forEach { (key, value) in
+        headers?.forEach { (key, value) in
```
sunshinejr: Are you sure that if we use `unwrappedHeaders` instead of `headers?` we will have to cast? Tried that one in Xcode and seems like it's working just fine, but it might be just me ¯\_(ツ)_/¯

tbaranes: You are right, just retry, it's working. I probably dreamt this warnings :expressionless:

Changelog.md
```diff
@@ -1,5 +1,9 @@
 # Next
 
+# 6.1.1
```
sunshinejr: One last comment, this header has to go 😥

tbaranes: Wuuups, on it.

sunshinejr:
@tbaranes Thanks again for the changes! You did great job in finding that one and fixing it. 💯 When the tests pass I will merge it. Cheers! 🎉

sunshinejr:
@ashfurrow We will need the invite for @tbaranes :beers:

ashfurrow:
Done 🎉 

@tbaranes no pressure to accept the invitation – if you'd like more information on what that means, check out our [contributor guidelines](https://github.com/Moya/contributors) and feel free to reach out with any questions.

tbaranes:
Thanks! Love to join your great project :tada:

Issue #408, opened 2016-02-12
Using Structs (working fork)


Matthijn:
First of, `Moya` looks great. At least most of it does. But the one single `enum` bugs me a bit. You have to define a single `API endpoint` in multiple parts of the same file. When the `API` grows, the `enum` gets really big since it is hard to divide it into multiple files. 

I've made some changes in my [fork of Moya](https://github.com/Matthijn/Moya), and perhaps this idea can be discussed here and maybe (when the idea is liked) it can be integrated in `Moya` itself. But it would break existing code using `enums` in the way I've implemented it currently.

The benefits of using `structs` are:

- You can easily divide your `API` definition accros multiple files
- All information regarding one single `API Endpoint` is contained within one place

Also I've made some extra changes which makes using the structs even nicer. 

- Automatic generating of the `parameters` in the request based on the structure of the `struct` using `reflection` (optional, you can still define your own parameters if you choose so)
- Easy support for urls with formatting like: /user/{id}

I've got a full [README.md](https://github.com/Matthijn/Moya) on the Github page with some example code. I'd love to hear all of your opinions on it. 

The biggest change in regard to the original version of `Moya` is the addition of the [Smoya.swift](https://github.com/Matthijn/Moya/blob/master/Source/Smoya.swift) file. Which is short for `StructuredMoya`. In there some `protocols` and `extensions` have been defined.

Next to that most of `Moya` remains the same. But there are some changes regarding `generics` to hook up the right `enum`. This has been removed and it accepts any `TargetType` now.

Also the `baseURL` has been removed from the `TargetType` and you must add that in your own `endpoint`. (The default `endpoint` has been removed). Since defining the `baseURL` in every `struct` seemed odd to me.

Again, I'd love to get some feedback. And am curious if people like this approach.

ashfurrow:
Hey, this is pretty cool! Thanks for the kind words, too. 

I'll have more time this weekend to look closer at it. I agree there are limitations in centralizing things, especially as APIs grow. I've broken things into multiple enums in a project, and [we've been talking about structs](https://github.com/Moya/Moya/pull/376) for a little bit. The use of the reflection API is really clever.

I'm not sure if we can "turn the ship" at this point (the structure of the enums is... pretty engrained into Moya), but I'd love to borrow some of the ideas from this. We should then consider linking to it from our docs, if you're interested in maintaining it as a separate project.

Matthijn:
[...]

PR #424, opened 2016-03-10
Structs in Moya.

sunshinejr:
As per issue #408, I'm sharing with you the solution I've come up with to the topic, which is structs in Moya. The general idea was a perfect start by @Matthijn followed up by a really good direction by @colinta 

## Motivation
We would like to keep existing enum setup and additionally support structs based setup for bigger projects or for any other reason. Specifically, what should be possible with this PR is that we can still use old configuration of enums and it won't break, but additionally we can use Moya with structs configuration, for instance:

```swift
GitHubStructProvider.request(.UserProfile(UserProfileResource(name: name))) { result in
            
}
```

## Implementation
My idea was to create protocol that will be a default protocol for enum when using it with wrapped structs, I called it `ServiceType`. Then make `TargetType` conform to it protocol so we can reuse code for both implementations.

## Issues

There were few issues that occurred during the implementation. Changes I've made created some problems with generic's type inferring so I had to move Defaults (`DefaultEndpointMapping`, `DefaultRequestMapping`, `DefaultAlamofireManager` and stubbing defaults) to another non-generic class and I called it `MoyaDefaults`, which is a **breaking change** in case someone used them via `MoyaProvider.DefaultAlamofireManager` for instance. Couldn't find a way for not-creating another struct without many changes in code so if you have an idea I would love to hear it. 

## Demo
I've also added an example of usage to the current demo. As per our conversation I also believe we could move it to second demo. In short we can now define our struct like the one below:

```swift
public enum GitHubStruct: ServiceType {
    case Zen(ZenResource)
    case UserProfile(UserProfileResource)
}

public struct ZenResource: TargetType {
    public let path = "/zen"
    public let baseURL = NSURL(string: "https://api.github.com")!
    public let method: Moya.Method = .GET
    public let parameters: [String: AnyObject]? = nil
    public let sampleData = "Half measures are as bad as nothing at all.".dataUsingEncoding(NSUTF8StringEncoding)!
}
```

## Conclusion
This is really a first step and we could improve it I believe, so I'm curious to hear what you think about it.

Cheers 🎉

### DIFF
--- a/Demo/Demo/GitHubStructAPI.swift
+++ b/Demo/Demo/GitHubStructAPI.swift
@@ -1,0 +1,16 @@
+import Foundation
+import Moya
+
+let GitHubStructProvider = MoyaProvider<GitHubStruct>(plugins: [NetworkLoggerPlugin(verbose: true, responseDataFormatter: JSONResponseDataFormatter)])
+
+public enum GitHubStruct: ServiceType {
+    case Zen(ZenResource)
+}
+
+public struct ZenResource: TargetType {
+    public let path = "/zen"
+    public let baseURL = NSURL(string: "https://api.github.com")!
+    public let method: Moya.Method = .GET
+    public let parameters: [String: AnyObject]? = nil
+    public let sampleData = "Half measures are as bad as nothing at all.".dataUsingEncoding(NSUTF8StringEncoding)!
+}
--- a/Demo/Demo/GitHubStructAPI.swift
+++ b/Demo/Demo/GitHubStructAPI.swift
@@ -5,6 +5,7 @@ let GitHubStructProvider = MoyaProvider<GitHubStruct>(plugins: [NetworkLoggerPlu
 
 public enum GitHubStruct: ServiceType {
     case Zen(ZenResource)
+    case UserProfile(UserProfileResource)
 }
 
 public struct ZenResource: TargetType {
@@ -13,4 +14,24 @@ public struct ZenResource: TargetType {
     public let method: Moya.Method = .GET
     public let parameters: [String: AnyObject]? = nil
     public let sampleData = "Half measures are as bad as nothing at all.".dataUsingEncoding(NSUTF8StringEncoding)!
+}
+
+public struct UserProfileResource: TargetType {
+    public var path: String {
+        return "/users/\(name.URLEscapedString)"
+    }
+    public var baseURL: NSURL {
+        return NSURL(string: "https://api.github.com")!
+    }
+    public var method: Moya.Method {
+        return .GET
+    }
+    public var parameters: [String: AnyObject]? {
+        return nil
+    }
+    public var sampleData: NSData {
+        return "Half measures are as bad as nothing at all.".dataUsingEncoding(NSUTF8StringEncoding)!
+    }
+    
+    public let name: String
 }
--- a/Demo/Tests/MoyaProviderSpec.swift
+++ b/Demo/Tests/MoyaProviderSpec.swift
@@ -7,7 +7,7 @@ class MoyaProviderSpec: QuickSpec {
     override func spec() {
         var provider: MoyaProvider<GitHub>!
         beforeEach {
-            provider = MoyaProvider<GitHub>(stubClosure: MoyaProvider.ImmediatelyStub)
+            provider = MoyaProvider<GitHub>(stubClosure: MoyaDefaults.ImmediatelyStub)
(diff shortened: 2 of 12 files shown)


Matthijn:
I must say I like this solution. Haven't checked into this a while. But this looks promising.

Demo/Tests/MoyaProviderSpec.swift
```diff
@@ -7,7 +7,7 @@ class MoyaProviderSpec: QuickSpec {
     override func spec() {
         var provider: MoyaProvider<GitHub>!
         beforeEach {
-            provider = MoyaProvider<GitHub>(stubClosure: MoyaProvider.ImmediatelyStub)
+            provider = MoyaProvider<GitHub>(stubClosure: MoyaDefaults.ImmediatelyStub)
```
colinta: I'm really liking this PR!  However, I'm gonna ask for a few changes to go into a separate PR, like the change from `MoyaProvider` to `MoyaDefaults`.  This is a breaking change, and I'm hoping we can support structs without any breaking changes.

sunshinejr: Thanks 🎉 Sure, will do this today. ✌️

Source/Moya.swift
```diff
@@ -65,12 +87,16 @@ public class MoyaProvider<Target: TargetType> {
...
     
     /// Returns an Endpoint based on the token, method, and parameters by invoking the endpointsClosure.
-    public func endpoint(token: Target) -> Endpoint<Target> {
+    public func endpoint(token: TargetType) -> Endpoint<TargetType> {
         return endpointClosure(token)
     }
     
     /// Designated request-making method. Returns a Cancellable token to cancel the request later.
     public func request(target: Target, completion: Moya.Completion) -> Cancellable {
+        guard let target = target.resource else {
```
ashfurrow: Could you expand on the use of `.resource` here, I'm not sure I understand it or `ServiceType` but I'd like to know more!

Demo/Demo/GitHubStructAPI.swift
```diff
@@ -0,0 +1,37 @@
...
+    }
+    public var parameters: [String: AnyObject]? {
+        return nil
+    }
+    public var sampleData: NSData {
+        return "Half measures are as bad as nothing at all.".dataUsingEncoding(NSUTF8StringEncoding)!
+    }
+    
+    public let name: String
+}
```
ashfurrow: Are these two different to demonstrate flexibility? I like flexibility, and this looks cool, but I think using the syntax from `ZenResource` for all of these properties (except `path`) would help encourage users to follow that pattern where they can. I like opinionated code :wink:

Source/Moya.swift
sunshinejr: Of course! So the general idea of `ServiceType` is that this is the core of **StructsAPI**. Every enum that wants to use Moya with structs would have to conform to `ServiceType`. We then believe that when you conform to that protocol, you know what are you doing (I mean that is the risk I was talking about on hangout). So as a follow-up, I continue to believe the user and I try to fetch the associated object as it is a Resource (TargetType). But! I've made `TargetType` to conform to `ServiceType` as a default. So if you are using **EnumAPI**, it basically means that we don't need to look for associated object, we are the associated object, we are a TargetType (Resource). 

I don't know if that seems good to you, but let me know, I will try harder express my mind 🤕

Demo/Demo/GitHubStructAPI.swift
sunshinejr: Yeah, I've added this as a second option with using properties of a struct as a parameters. Sure, will do!

Source/Moya.swift
ashfurrow: OK, that makes sense, thanks. I guess I was thinking of something like this:

```swift
enum StructTargetType: TargetType {
  public var path: String {
    switch self {
      case .Struct(let t): return t.path
    }
  }
  // and so on
}
```

So that the struct-based approach totally piggy-backed on the existing enum structure, instead of having the existing enum structure conform to something new. 

That was totally just _my_ interpretation, it may not be possible. You've played with this more than I have :smile:

colinta: What Ash describes here is also what I was imagining - start with what we have, instead of adding anything new, just update the examples to show that *struct based endpoints is already possible*.

ashfurrow:
Going to close this in lieu of #427.

Issue #189, opened 2015-08-19
Add examples to documentation


ashfurrow:
Our documentation is thorough, but not instructive. We need to do a better job of providing examples and real-world scenarios. 

(Continued from #188.)

> In general I would shoot for more complete examples. Newcomers as I am one, lack guidance. For example it wasn't quite clear at first for me if I have to implement my own Provider. Turns out it was sufficient to use the default one with a closure as an initialization parameter. —@jk

ashfurrow:
Might make sense for a more in-depth demo in a separate repo on this organization.

bontoJR:
This is definitely something I would like to see. :)

At the moment I am trying to figure out how to have a parameter always passed to every request, in this case is an `access_token`. It took me some time to understand how to use this library and I had to check the whole source code to master it, I know they are not so much lines, but still... also the RxSwift's version requires some extra documentation b/c it looks like is outdated, I think. :+1:

scottrhoyt:
We have some real-ish examples in the documentation now. Do you think that is enough to close this issue?

PR #426, opened 2016-03-14
Move Defaults and Stubbing from MoyaProvider to MoyaDefaults.

sunshinejr:
While preparing for Structs API that potentially could be coming to Moya, we decided that the PR #424 was too large and, what is more important, it had breaking change that is extracted to this PR. 

In this PR we extracted from `MoyaProvider`:
- Defaults: `DefaultAlamofireManager`, `DefaultRequestMapping`, `DefaultEndpointMapping`
- Stubbing: `NeverStub`, `ImmediatelyStub`, `DelayedStub`

and moved them to the new class `MoyaDefaults`.

For more information you can take a look at description of #424. Thanks! 🎉

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,7 @@
 # Next
 
-- Fixed endpoint setup when adding `parameters` or `headers` when `parameters` or `headers` or nil
+- **Breaking Change** move Defaults (`DefaultAlamofireManager`, `DefaultRequestMapping`, `DefaultEndpointMapping`) and Stubbing (`NeverStub`, `ImmediatelyStub`, `DelayedStub`) from `MoyaProvider` to the new class `MoyaDefaults`.
+- Fixed endpoint setup when adding `parameters` or `headers` when `parameters` or `headers` or nil.
 
 # 6.2.0
 
--- a/Demo/Podfile.lock
+++ b/Demo/Podfile.lock
@@ -1,5 +1,5 @@
 PODS:
-  - Alamofire (3.2.0)
+  - Alamofire (3.2.1)
   - Moya (6.2.0):
     - Moya/Core (= 6.2.0)
   - Moya/Core (6.2.0):
@@ -53,7 +53,7 @@ EXTERNAL SOURCES:
     :path: ../
 
 SPEC CHECKSUMS:
-  Alamofire: 4da478599fccddff361205c8929333d7fe18dceb
+  Alamofire: f11d8624a05f5d39e0c99309b3e600a3ba64298a
   Moya: a5e57a405584177027daab240f4d47baa0722562
   Nimble: 79d40f4d69d47314229bbabacaa02b8838c779b9
   OHHTTPStubs: d2bf6a0f407620d67e75a1d3f12d86a2b1dd15c0
--- a/Demo/Tests/MoyaProviderSpec.swift
+++ b/Demo/Tests/MoyaProviderSpec.swift
@@ -7,7 +7,7 @@ class MoyaProviderSpec: QuickSpec {
     override func spec() {
         var provider: MoyaProvider<GitHub>!
         beforeEach {
-            provider = MoyaProvider<GitHub>(stubClosure: MoyaProvider.ImmediatelyStub)
+            provider = MoyaProvider<GitHub>(stubClosure: MoyaDefaults.ImmediatelyStub)
         }
         
         it("returns stubbed data for zen request") {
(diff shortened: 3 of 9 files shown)


ashfurrow:
Makes sense! We should double-check if documentation needs updating. 

So if I understand correctly, we'll hold off on merging this and take these changes _out_ of #424? Then:

1. Merge #424.
1. Release a patch for structs in Moya.
1. Merge this.
1. (Possibly wait for other breaking changes.)
1. Release a major version for this.

Does that sound right?

sunshinejr:
So we basically _can't_ (at least not with the changes I proposed) introduce #424 without introducing `MoyaDefaults`. Normally it would go the way you suggested, but this time we kinda have to do it the other way around, which is:

1. Merge this.
2. Make sure everything is running correctly.
3. Merge #424 which is the StructsAPI on its own (without `MoyaDefaults` introduction).
4. (Possibly wait for other breaking changes.)
5. Release a major version for this.

If something is still unclear I'm happy to answer! 😉

colinta:
Well no, we could remove the changes that reference `MoyaDefaults` from #424.  That's what I had intended by my comment.

You could either rebase #424, and remove the `MoyaDefaults` code, or create a new commit, and this PR *should* override that change.

sunshinejr:
Yeah I could rebase the #424 and delete the last commit which is `MoyaDefaults`. What I meant was that merging #424 before this one will also include the changes from here.

ashfurrow:
Right, I _think_ what @colinta means is that we can rebase #424 to remove any reference to `MoyaDefaults` so that we have two, totally independent pull requests. Then we can have a release between #424 and this one, so that people can adopt structs without adopting this (possibly breaking) change. Does that make sense?

sunshinejr:
Yeah, makes sense! But, I don't know if it possible. #424 relies on `MoyaDefaults`, because there was problems with type inferring 😔

ashfurrow:
Those do happen >.< Could I ask you to take a look and see what problems they are? Maybe we can find a less-than-ideal, non-breaking change and move to this (which is a good idea) as soon as possible. Does that sound ok?

sunshinejr:
Sure, will take a look 👀

sunshinejr:
So, the problem with generics is that we have 2 options:
- we have TargetType as a MoyaProvider's Target,
- we have ServiceType which has 0..infinity TargetTypes

What leaves us with the case that our base protocol is ServiceType. Swift's generics seems to have a problem when an actual TargetType is passed:



And we can "repair" it by specifying the TargetType explicitly. From:
```swift
MoyaProvider<GitHub>(endpointClosure: MoyaProvider.DefaultEndpointMapping)
```

To:
```swift
MoyaProvider<GitHub>(endpointClosure: MoyaProvider<GitHub>.DefaultEndpointMapping)
```

But it is still a breaking change. 😔

ashfurrow:
Interesting! I guess I was thinking of https://github.com/Moya/Moya/pull/424 as having only _one_ struct-based enum type that Moya gives for developers to put their structs in (noted in [this](https://github.com/Moya/Moya/pull/424#discussion_r56069092) comment, sorry if I was unclear). Would that solution fix this problem, since we wouldn't need a ServiceType and could continue with the existing implementation? 

Thanks again for all your work on this, it's really great and we all really appreciate it!

sunshinejr:
That would be most likely possible, but I don't know if it suits really #424 tho? Maybe we could leave #424 with the current implementation and create new one, which will only add new Demo project or new class that show struct setup. Also, we agreed on a new Demo project, right? DemoStructs seems all right? Do we need some extra changes in old Demo? You got the documentation part, right?

And thank you, I really appreciate that! ☺️

ashfurrow:
Yup, I've got documentation covered but we need to finalize the implementation first :smile: I don't know if a new demo project is necessary (beyond #189) if we can just add a side-by-side demo of using structs in the existing project, your call.

If we _can_ use the single-case-enum I mentioned in #424, maybe it's worth trying out a PR with just that idea? I can whip something quick together this afternoon, I'd like to hear your thoughts on it. Let me know :cake:

sunshinejr:
Yeah, go for it 🎉 Adding it to the existing Demo is perfectly fine with me 👌

sunshinejr:
I'm closing this one. This doesn't need to be introduced since we went with another option of structs API.

ashfurrow:
Makes sense :+1:

PR #427, opened 2016-03-15
Adds sample implementation for struct-based enum

ashfurrow:
This is an alternative implementation of #408 (in lieu of #424).

Basically, Moya provides an `enum` type called `StructTarget` that has a single case, `.Struct` with an associated, generic value. That value is the "real" `TargetType`, the `struct` that our users would use. So instead of

```swift
MoyaProvider<MyStructAPI>
```

You'd use

```swift
MoyaProvider<StructTarget<MyStructAPI>>
```

Pros:

- Simple implementation
- No breaking changes

Cons:

- More verbose for developers using structs

Curious to know what you think!

### DIFF
--- a/Source/Moya.swift
+++ b/Source/Moya.swift
@@ -32,35 +32,30 @@ public enum StructTarget<T: TargetType>: TargetType {
     }
 
     public var path: String {
-        switch self {
-        case .Struct(let t): return t.path
-        }
+        return target.path
     }
 
     public var baseURL: NSURL {
-        switch self {
-        case .Struct(let t): return t.baseURL
-        }
+        return target.baseURL
     }
 
     public var method: Moya.Method {
-        switch self {
-        case .Struct(let t): return t.method
-        }
+        return target.method
     }
 
     public var parameters: [String: AnyObject]? {
-        switch self {
-        case .Struct(let t): return t.parameters
-        }
+        return target.parameters
     }
 
     public var sampleData: NSData {
-        switch self {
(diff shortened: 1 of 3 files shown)


colinta:
This is more in line with what *I* had pictured - simple, more verbose, but doesn't introduce or rely on any new features.

I must give props to @sunshinejr for the cool use of reflection, but I would actually steer clear of that in a library like Moya.  Keep it simple, keep it explicit, let individual developers choose to implement something more interesting & complex.

But that's just my opinion! :smiley:

Source/Moya.swift
```diff
@@ -24,6 +24,45 @@ public protocol TargetType {
     var sampleData: NSData { get }
 }
 
+public enum StructTarget<T: TargetType>: TargetType {
```
sunshinejr: Nice one! I really like that 🙌 One thing I can add to this is that we can create a stored property for example that will give us the target, e.g.:
```swift
private var target: T {
    switch self {
        case .Struct(let t): return t
    }
}
```

and then we can skip the switch stuff in other properties with just e.g. `return target.path`.

ashfurrow: That's a great idea!

ashfurrow:
d8c017582e3

sunshinejr:
Nice! What would be the really simple example of usage tho? I wanna make sure I'm on the same page :smile:

ashfurrow:
Cool! I'll create an issue to update docs and demo app.

PR #430, opened 2016-03-16
Removes generic type requirement on StructTarget

ashfurrow:
Based off of work done in #427, refined from [this comment](https://github.com/Moya/Moya/pull/376#issuecomment-197149515). 

This removes the generic parameter for `StructTarget` and instead lets users create one with _any_ struct that conforms to the `TargetType`, so one provider can be used with many different struct targets.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,7 @@
 # Next
 
-- Fixed endpoint setup when adding `parameters` or `headers` when `parameters` or `headers` or nil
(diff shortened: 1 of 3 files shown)


sunshinejr:
Seems fine to me, @ashfurrow! 🎉 I don't know, but maybe we could add the info about structs to Changelog with this PR?

ashfurrow:
Good thinking :+1:

Issue #432, opened 2016-03-16
Update docs with more detailed examples.


sunshinejr:
We need more examples in our docs page. I'm thinking about making a Table of contents in our [Examples.md](https://github.com/Moya/Moya/blob/master/docs/Examples.md), where we could make few examples. One would be the easiest, the most basic setup with `.GET` request, then we could make another one with POST, another one with custom and specific examples of endpoints, stubs, and we could even add the setup with structs (vide #428). Also the **Examples** document feels like too much text at once, we could make something like description of _how_ you can setup Moya and _why_, and then the examples step by step. 

This is just an idea but we can brainstorm together 🤔

What are your thoughts? 🎉

ashfurrow:
I like this idea, but I don't want the docs to get overwhelming. Maybe we should expand from a single `Examples.md` and do something more like "TargetExamples.md" etc? That way each document stays bite-sized.

Seems like a bit of a duplicate from https://github.com/Moya/Moya/issues/189 , what do you think? Might be worth having a milestone around documentation to gather all the ideas in one place. 

Probably related to #428 and #369 as well.

sunshinejr:
First, milestone is a good choice, it would be much easier to read (also yeah, kinda a duplicate, sorry about that). About examples I agree that multiple files are a way to go, also we could have one file that will guide new user through examples, like in this file you can learn basic and the simplest usage, then here you have this and that because reasons. Or maybe do it in the main docs? Hmm 🤔

ashfurrow:
I think this might be a good structure:

```
docs/
  Targets.md
  ... etc ...
  Examples/
    Readme.md
    TargetExamples.md
    ... etc ...
```

Then the readme in the new directory can be a guide to the other documents, like [what we have now](https://github.com/Moya/Moya/tree/master/docs).

sunshinejr:
I _really_ like that one! 🙌

Dschee:
I would appreciate more detailed examples a lot.

I'm currently trying to implement an API the Moya way but I'm struggling with all the options and plugins and how to actually integrate them. Specifically it's not clear to me from reading the current docs which parts are required and what the default behavior of Moya is if only the required things are implemented. Also I can't figure out how to add a logging plugin right now – seems like I have to dig much deeper into Moya than I wanted to. I think things would be much easier if multiple examples were provided.

sunshinejr:
Thank you for your input, @Dschee ! 🎉 

[...]

PR #438, opened 2016-03-21
Improve Examples and Provider documentation

Dschee:
This **improves the documentation** of the Examples.rd and Provider.rd files by adding **more details**, structuring by params (**clarity**) and being more explicit about Moyas **default behavior** and **minimum usage requirements**.

I did these changes after trying to use Moya for a project and got stuck with it after reading the documentation alone. Once I looked into the code all my questions got answered quickly – but not everybody will do that. So I in this PR I try to be more explicit about the parts that confused me and try to provide **more detailed** and **practical examples**.

See also #432.

### DIFF
--- a/docs/Examples.md
+++ b/docs/Examples.md
@@ -32,15 +32,15 @@ extension MyService: TargetType {
     }
     var method: Moya.Method {
         switch self {
-        case .Zen, .ShowUser(_):
+        case .Zen, .ShowUser:
             return .GET
-        case .UpdateUser(_, _, _):
+        case .UpdateUser:
             return .PATCH
         }
     }
     var parameters: [String: AnyObject]? {
         switch self {
-        case .Zen, .ShowUser(_):
+        case .Zen, .ShowUser:
             return nil
         case .UpdateUser(_, let firstName, let lastName):
             return ["first_name": firstName, "last_name": lastName]
--- a/docs/Examples.md
+++ b/docs/Examples.md
@@ -3,72 +3,100 @@ Examples
 
 So how do you use this library? Well, it's pretty easy. Just follow this
 template. First, set up an `enum` with all of your API targets. Note that you
-can include information as part of your enum. Let's look at a simple example.
+can include information as part of your enum. Let's look at a common example. First we create a new file named `MyService.swift`:
 
 ```swift
-enum GitHub {
+enum MyService {
     case Zen
-    case UserProfile(String)
+    case ShowUser(id: Int)
+    case UpdateUser(id: Int, firstName: String, lastName: String)
 }
 ```
 
 This enum is used to make sure that you provide implementation details for each
-target (at compile time). The enum *must* conform to the `TargetType` protocol.
-Let's take a look at what that might look like.
+target (at compile time). You can see that parameters needed for requests can be defined as per the enum cases parameters. The enum *must* additionally conform to the `TargetType` protocol. Let's get this done via an extension in the same file:
 
 ```swift
-private extension String {
-    var URLEscapedString: String {
-        return self.stringByAddingPercentEncodingWithAllowedCharacters(NSCharacterSet.URLHostAllowedCharacterSet())!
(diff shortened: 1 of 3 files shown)


docs/Examples.md
```diff
@@ -3,72 +3,100 @@ Examples
...
         }
     }
     var method: Moya.Method {
-        // all requests in this example will use GET.  Usually you would switch
-        // on the enum, like we did in `var path: String`
-        return .GET
+        switch self {
+        case .Zen, .ShowUser(_):
+            return .GET
+        case .UpdateUser(_, _, _):
```
ashfurrow: If you don't need an enum's associated values, you can also just use `case .UpdateUser:` and it should work :+1:

docs/Examples.md
```diff
@@ -3,72 +3,100 @@ Examples
...
+            return .PATCH
```
ashfurrow: Patch eh? Nice, one of my favourite HTTP obscurities. For the sake of keeping the code familiar, do you think we should use a more typical put instead?

docs/Providers.md
```diff
@@ -123,17 +133,21 @@ let manager = Manager(
...
 ```
 
+### plugins:
+
 Finally, you may also provide an array of `plugins` to the provider. These receive callbacks
 before a request is sent and after a response is received. There are a few plugins
 included already: one for network activity (`NetworkActivityPlugin`), one for logging
 all network activity (`NetworkLoggerPlugin`), and another for [HTTP Authentication](Authentication.md).
 
+For example you can enable the logger plugin by simply passing `[NetworkLoggerPlugin()]` alongside the `plugins` parameter of your `Endpoint`. Note that a plugin can also be configurable, for example the already included `NetworkActivityPlugin` requires a `networkActivityClosure` parameter. The configurable plugin implementation looks like this:
```
ashfurrow: Could be worth linking to the [auth docs](https://github.com/Moya/Moya/blob/master/docs/Authentication.md#basic-http-auth) for further reading on the CredentialsPlugin.

ashfurrow:
This is fantastic, @Dschee, thank you for taking the time to improve our documentation! I've made a few inline suggestions above – I'm happy to merge as-is and submit a PR with my feedback, none of these are blockers.

docs/Examples.md
Dschee: Cool, you're right, I was using it like that already but thought that might be a Swift side effect and actually undocumented. But now you say it, I can clean it up. Thanks!

docs/Examples.md
Dschee: Sure, but let's not use PUT either then as it's not very RESTful (which I think is kind of the standard right now) so what about changing the `.UpdateUser` case up to a `.CreateUser` and make it a POST? That should be familiar enough. I've changed the code to do this.

docs/Providers.md
Dschee: Isn't the `Authentication.md` linked just at the end of the section above? Shouldn't this be sufficient? You can of course add an additional link but I'm not sure what you are thinking of exactly.

Dschee:
I've answered all your inline comments and tried to update the code accordingly. I hope all issues are fixed now. Thanks for considering this PR, @ashfurrow! :tada:

docs/Examples.md
```diff
@@ -3,72 +3,100 @@ Examples
 
 So how do you use this library? Well, it's pretty easy. Just follow this
 template. First, set up an `enum` with all of your API targets. Note that you
-can include information as part of your enum. Let's look at a simple example.
+can include information as part of your enum. Let's look at a common example. First we create a new file named `MyService.swift`:
 
 ```swift
-enum GitHub {
+enum MyService {
```
orta: :+1:

docs/Providers.md
ashfurrow: Ah, yes! Must've missed it.

ashfurrow:

PR #439, opened 2016-03-23
Feature/multipart upload

justinmakaila:
Added the ability for multipart uploads. It's not *really* sick because of the ugly button on the toolbar, but it works. Let's get this baby reviewed, refined, and into rotation ;)

### DIFF
--- a/Cartfile.resolved
+++ b/Cartfile.resolved
@@ -1,4 +1,4 @@
 github "Alamofire/Alamofire" "3.2.1"
 github "antitypical/Result" "1.0.2"
-github "ReactiveX/RxSwift" "2.2.0"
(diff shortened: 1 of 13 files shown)


justinmakaila:
This doesn't build on Carthage for the same reason master doesn't (Something about "Result" being built with a different version of the compiler). Looking into it now.

ashfurrow:
Thanks Justin, let us know when this is ready for a final look over :+1:

justinmakaila:
Uhh, strange issue... The project doesn't build if `Demo/` is present, probably due to the shared schemes in the `xcworkspace`. @ashfurrow, @orta, Do they have to be shared for Cocoapods?

Issue #114, opened 2015-04-24
Implement uploading of multpart/form-data


ptiz:

ashfurrow:
Great idea! I think Alamofire supports this already, so there is an opportunity to leverage their code while being clever about how we expose the functionality.

ptiz:
I am afraid it is not, out of the box. But it would be great if there was something like Alamofire.upload at least.

ashfurrow:
OK, cool. We can start there then :)

aschuch:
There is [Alamofire support](https://github.com/Alamofire/Alamofire#uploading) for uploading and downloading files, so this would be straight forward I guess.

I anyone already working on this? Might need it in the following weeks so I might as well hack on it if you want.

ashfurrow:
Help on this would be very much appreciated. I'm catching up on some work for the remainder of the week, and won't have time to look at it for a while.

zdavison:
Also looking for this fairly soon. @aschuch do you have any progress on this? If not I can take a look.

aschuch:
@zdavison I did not start on it yet. I probably don't get time to do it anytime soon though... :disappointed:

zdavison:
@aschuch that's ok, i'll take a look here, we need it in the next week or so.

ashfurrow:
This is a fundamentally new behaviour for Moya. I'm trying to figure out the best way to add it to our existing API without overly complicating things. 

If we want to use the existing `request()` function, then we would need to have a way to distinguish targets between upload and regular HTTP requests. We _could_ add `.Upload` to the `Moya.Method` enum, which would be easy, but would break parity with the HTTP verbs and the `Alamofire.Method` enum. 

On the other hand, if we add a new optional property to `MoyaTarget` for uploading (say, we call it `uploadData`), we could ignore the method of the target and assume it's uploading based on the presence of a non-`nil` `uploadData`. 

In either case, we would still need to figure out the progress callback. Maybe we could incorporate it into the cancellable token somehow maybe? 

Neither option really sits with me. We could, instead, have a new `upload` method on the `MoyaProvider` class. We would still need the optional `uploadData` property, the API would become _slightly_ more complicated, and it would be some duplicated code in `MoyaProvider` and its subclasses. 

Anyone have any feels about this?

alcarvalho:
If only Swift had support for default parameters to Enums, we could have `.POST` and others with a optional parameter with a default to `nil`. 

[...]

PR #441, opened 2016-03-24
Feature/multipart upload multipartuploadable

justinmakaila:
This builds off of `feature/multipart-upload` a little bit, but makes it a bit more flexible. I think this one is better suited for production.

It better introduces the concept of an upload, rather than just the ability to perform multipart uploads.

This was heavily influenced by the chatter on #114, @guidohendriks and @startupthekid

### DIFF
--- a/Cartfile.resolved
+++ b/Cartfile.resolved
@@ -1,4 +1,4 @@
 github "Alamofire/Alamofire" "3.2.1"
 github "antitypical/Result" "1.0.2"
-github "ReactiveX/RxSwift" "2.2.0"
+github "ReactiveX/RxSwift" "2.3.1"
 github "ReactiveCocoa/ReactiveCocoa" "v4.0.1"
--- a/Demo/Demo.xcodeproj/project.pbxproj
+++ b/Demo/Demo.xcodeproj/project.pbxproj
@@ -55,6 +55,8 @@
 		5EC197EF1A1BB16D00F4DFD4 /* Images.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 5EC197EE1A1BB16D00F4DFD4 /* Images.xcassets */; };
 		5EC197F21A1BB16D00F4DFD4 /* LaunchScreen.xib in Resources */ = {isa = PBXBuildFile; fileRef = 5EC197F01A1BB16D00F4DFD4 /* LaunchScreen.xib */; };
 		5EC198081A1BB24600F4DFD4 /* GitHubAPI.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5EC198071A1BB24600F4DFD4 /* GitHubAPI.swift */; };
+		7CC30DBE1CA30E10002C8282 /* HTTPBinAPI.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7CC30DBD1CA30E10002C8282 /* HTTPBinAPI.swift */; };
+		7CC30DC01CA3101D002C8282 /* Providers.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7CC30DBF1CA3101D002C8282 /* Providers.swift */; };
 		EEFDA00B2FC44BE035958929 /* Pods_MoyaTests_tvOS.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = DCD314EAFE25D3A0CE7997D8 /* Pods_MoyaTests_tvOS.framework */; };
 /* End PBXBuildFile section */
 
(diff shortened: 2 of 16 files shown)


justinmakaila:
This still requires some tests for the other upload types, but multipart is working :+1:

GuidoHendriks:
Looks really good, nicely done! 😎 Will have a good look at this soon, if only I had time I would have given it a shot myself.

justinmakaila:
@GuidoHendriks There's still time, I'm sure there are bugs I'm missing cause I've been looking at it a bit too long

justinmakaila:
Turns out that this builds on Carthage, unlike the last one. You have to run `carthage update --no-use-binaries`, because most of the distributed binaries were built with the old compiler /facepalm

petester42:
Ya. I find that it is always better to set that on

aschuch:
This looks really good to me. I think we should also consider adding a way to get progress information for uploads. Do you think an optional closure on the `request` method would be ok?

justinmakaila:
@aschuch I'm down with that. I think it'd be coolest if we could get it like this:
```swift
let target = API.Target.SomeUploadTarget
APIProvider.request(target)
  .progress { bytesSent, totalBytes in
    let percentage = bytesSent / totalBytes
    self.progressView.value = percentage
  }
```
But that would require some contextual knowledge of the target (mostly that it's an upload).

startupthekid:
I might be off base here but what about for multipart uploads, we create a `Response` subclass that has a progress method? That way all the normal methods on `Response` would work and for multipart, you'd get the progress you need.

Something like:

```swift
func request(token: Target) -> SignalProducer<UploadResponse, Error> {}
```

justinmakaila:
@startupthekid I think you're on to something. Do @Moya/contributors see any issues with potentially introducing a new `Response` type?

colinta:
I also like this implementation!  Feels very "Moya-esque" :smile: 

The "Response-subclass w/ `progress` method is a good idea to me, too.

Thanks for tackling this one!

startupthekid:
That does mean though that `Response` can't be a `final` class. That should be ok though, makes Moya more extensible, especially if someone wants to provide a custom response class for debugging purposes or what have you.

Another interesting direction could be making a `ResponseType` protocol that `Response` provides the implementation for and then an `UploadResponseType` that conforms to `ResponseType` and adds a few handy progress methods. 

```swift
public protocol ResponseType: CustomDebugStringConvertible, Equatable {
    var statusCode: Int { get }
    var data: NSData { get }
    var response: NSURLResponse?

   ....
}

public protocol UploadResponseType: ResponseType {
    func progress(...)
}
```

justinmakaila:
+1 to the `ResponseType` protocol.

Not sure how to integrate the progress though, because the response type wouldn't be available until after the request completes.

startupthekid:
Instead of on `Response` it could be on `Request` and then `request` could send an intermediary request object that informs progress. Not sure how to go about implementing that though.

justinmakaila:
We may have to introduce an `UploadRequest` or something along those lines... I'm imagining we'd just  add a `.progress { ... }` closure to the `Alamofire` call site which invokes the `progress` provided by `Moya`.

jasl:
Or we may add `progress` [here](https://github.com/Moya/Moya/blob/master/Source/Plugin.swift#L19) and add a closure named `willPerformRequest(request: RequestType)` and call it before the request is sending out (for upload, we call it in `encodingCompletion`

jasl:
But I think a better solution needs a breaking change, the `ParameterEncoding` is conflict with multipart-form-data, actually multipart-form-data is a kind of parameter's encoding, but Moya currently using Alamofire's `ParameterEncoding`

startupthekid:
@jasi that's a good point. It may be better to use our own encoding that mirrors Alamofire.

@justinmakaila something else I'm concerned about is supporting parameters in an upload request.
In the scenario where someone uploads a file and needs to attach that reference to another model, that can't currently be done in Alamofire. 

For example I'm using feathersjs on my backend to supports before and after hooks:

```javascript
blobService.before({
  all: sharedAuthHooks,
  // Verify that the postID parameter exists and that it refers to a real post.
  create(hook) {
    const { params } = hook;
    const { postID } = params;
    if (typeof postID === 'undefined') {
      return Promise.reject(new errors.BadRequest('Invalid postID parameter'));
    }
    // Find a post with that id, error will be thrown if none exists
    return postService.get(postID).then(() => hook);
  }
});


blobService.after({
  // After a blob is created, attach the id to the post.
  create(hook) {
    const { result, data } = hook;
    const { id } = result;
    const { postID } = data;
    // Update the associated post with the blob id
    return postService.patch(postID, { blob: id }, {}).then(() => hook);
  }
});
```

Before a blob (a file object) is created, it verifies that a postID parameter is sent and that that post exists. After the creation occurs, the blob id is then attached to the post. But with Alamofire, you can't attach parameters to a file upload.

The reason I'm worried about that is because in applications that will introduce multiple points of failure in similar situations. Instead of relying on a single source of truth on the backend where parameters are verified and the appropriate models are updated, each client has to ensure that post-creation, the correct updating takes place. The more we can do to mitigate that risk that better but I'm not sure how with Alamofire currently.

justinmakaila:
@startupthekid I think it should be possible for us to utilize the `encodingCompletion` invocation to add `parameters` from the target. All I'm doing is taking the encoded request and immediately kicking it off.

jasl:
If we refactor the `ParameterEncoding`, keeping the same behaviour both with normal request and multipart upload is possible and simple
I've done this for my Moya fork, [details here](https://github.com/jasl/MoyaX/blob/master/Sources/Backends/AlamofireBackend.swift#L70-L134)

justinmakaila:
The only thing I don't like about mirroring the `Alamofire.ParameterEncoding` to a custom `Moya.ParameterEncoding` (rather than the typealias) is that we tie ourselves to `Alamofire` updates. As their `ParameterEncoding` changes, we have to update ours. I'd rather us be able to abstract it in `Moya` land, transforming it into `Alamofire` land.

jasl:
That's may not be an issue, for Cocoapods or Charthage, we can lock the Alamofire's version, and for 3.x, Alamofire would unlikely change `ParameterEncoding` (and `ParameterEncoding` should be a stable component according to its commit logs)

jasl:
That's should not be an issue, for Cocoapods or Charthage, we can lock the Alamofire's version, and for 3.x, Alamofire would unlikely change `ParameterEncoding` (and `ParameterEncoding` should be a stable component according to its commit logs)

justinmakaila:
Closing in favor of #483 and f3f7570dcf875d2c95cae41397176c4577cfc555

PR #402, opened 2016-02-07
[WIP] Moves releases & testing to Fastlane

ashfurrow:
Fixes #378.

Currently having problems running tests for iOS... getting an error when I try to import Quick:

```
❌  /Users/ash/bin/Moya/Demo/Tests/RACSignal+MoyaSpec.swift:1:8: module file's minimum deployment target is ios9.0 v9.0: /Users/ash/Library/Developer/Xcode/DerivedData/Demo-grgybcslxdyyuxecwdhpshvabelw/Build/Intermediates/CodeCoverage/Demo/Products/Debug-iphonesimulator/Pods-MoyaTests-iOS/Quick.framework/Modules/Quick.swiftmodule/x86_64.swiftmodule

import Quick
```

I'm thinking it might have to do with having iOS 9.1 for iOS/watchOS tests and iOS 9.0 for tvOS tests. I'm downloading the iOS 9.0 simulator now to see. If that's the case, we can just run all the tests on 9.0.

Also, now that releasing versions of Moya will be easier, I'd like to add others to Trunk as pod owners so they can push, too; I don't want to become a bottleneck on the project.

PR #452, opened 2016-03-30
Fixes CI

ashfurrow:
Fixes #451.

PR #453, opened 2016-03-30
New GH release strategy

### LINKED ISSUE #378: Move releases to Fastlane
We have a Rakefile now, which is good, but Fastlane is better. Specifically because it has a [GitHub Release Action](https://github.com/fastlane/fastlane/blob/master/docs/Actions.md#github-releases) that would make my life easier :sweat_smile:

ashfurrow:
The gem we were using doesn't work, and I don't have time to finish #402. This is a good-enough solution. Next step after this is to add more owners to the pod so I'm not a bottleneck in cutting releases.

Closes #378.


ashfurrow:
(Expecting CI to fail on this, until #452 is done.)

ashfurrow:
Closing in favour of #456

PR #456, opened 2016-03-30
[Rakefile] Automate GH releases

### LINKED ISSUE #455: Automate GitHub releases
CocoaPods do this on their OS X app: https://github.com/CocoaPods/CocoaPods-app/blob/1eb14dce9c7a4d0d42a8dd534e25f5451942fc78/Rakefile#L1049-L1060 We should be able to use the same approach (store a GitHub access token somewhere and use it to compose an HTTP response). Then again, that means either sharing a token or requiring all deployers to make their own, which might end up being more effort than #453.

segiddins:
Closes #455 (presuming you didn't also want any sort of artifact uploads)


ashfurrow:
Cool! And we just throw in our .netrc file creds? Super neat!

segiddins:
Yeah, you just need to have GH creds in `.netrc`

PR #462, opened 2016-04-03
Open up option to use other queue instead of main queue

kokhong20:
## Open up option to use other queue instead of main queue

Moya default callback is on main thread and it is not possible to assign a specific queue for the response callback.

This pull request added an option to assign a specific queue.

This is useful when we want to perform other tasks in the background thread like saving these data to database.


ashfurrow:
Looks fine to me! Thanks for the contribution. 

I'm concerned that, as-is, this is a breaking change (if someone was subclassing the provider and overriding this method, the changing method signature would break thei override). Do you think, for now, we could introduce a second request function, instead of using the default value for a parameter?

A changelog entry and test would be awesome, but testing this might be tricky. We can merge without one and add it later.

kokhong20:
I think using the same function with default parameter is better. If introduce another function like 
`requestWithQueue` is quite weird and might have duplicate code too.

I have added the test case but only able to test whether callback is at main thread as `dispatch_get_current_queue` is deprecated. should be good enough.

ashfurrow:
I think using the same function is better, too, and I want to do that when we move to our next major release, but I don't want to break our [semantic versioning](http://semver.org) by breaking anyone's code.

kokhong20:
Ok. I have make another function.

Source/Moya.swift
```diff
@@ -103,6 +103,28 @@ public class MoyaProvider<Target: TargetType> {
...
+            }
+        }
+        
+        requestClosure(endpoint, performNetworking)
+        
+        return cancellableToken
+    }
+    
     /// Designated request-making method. Returns a Cancellable token to cancel the request later.
     public func request(target: Target, completion: Moya.Completion) -> Cancellable {
```
ashfurrow: The new function looks great! What do you think about replacing the body of this one, the fold function, with:

```swift
return request(target, queue: nil, completion: completion)
```

And just piggy-back on the new function?

Source/Moya.swift
```diff
@@ -103,6 +103,28 @@ public class MoyaProvider<Target: TargetType> {
         return endpointClosure(token)
     }
     
+    /// Designated request-making method with queue option. Returns a Cancellable token to cancel the request later.
+    public func requestWithQueue(target: Target, queue:dispatch_queue_t? = nil, completion: Moya.Completion) -> Cancellable {
```
sunshinejr: I think there is no need to use name `requestWithQueue`, we can just call it  `request`, but with one parameter more.

ashfurrow: Sounds right.

Source/Moya.swift
sunshinejr: Here we can just call the new method with `queue = nil` so we avoid code duplication.

nuudles:
What do you think about also performing the `stubRequest` in the specified queue? Getting it off of the main thread would be helpful for me for some of my requests, which include dynamically generating thumbnails of images.

Source/Moya.swift
kokhong20: Ya, that is a better approach! 👍

sunshinejr:
Good point @nuudles! 👏  What do you think about it @kokhong20? Is it okay with you? Also it would be probably the last point before we can merge this one I think.

sybarite:
@ashfurrow It would be great to have this feature in a non breaking way in the current release @kokhong20 has worked an good enough backwards compatible solution.

@kokhong20 Can you resolve merge issues to get this merged in the current release? @ashfurrow If he does not respond should I open up another PR with a backwards compatible implementation and can we start the process or merging it to the main repo?

ashfurrow:
Looks cool, I'm fine with releasing this as a non-breaking change. Complicating things slightly is the fact that `master` currently _has_ non-breaking changes already. I'll take care of releasing this this weekend. What I'll probably do is go back to the most recently released commit, merge this in, make a release, then merge _that_ commit into master (closing this PR). That way everybody wins :tada:

ashfurrow:
This has been released as Moya 6.5.0 and should be accessible from CocoaPods and Carthage. I've also had to manually merge into master, a few mew conflicts but nothing too serious. The merge commit is  https://github.com/Moya/Moya/commit/350bff8edbdb35007bdb8fd909aceef7d9d83d6b

@kokhong20 thank you for your contribution! I've invited you to join the Moya GitHub organization – no pressure to accept! If you'd like more information on what that means, check out our [contributor guidelines](https://github.com/Moya/contributors) and feel free to reach out with any questions.

PR #481, opened 2016-05-11
Multipart file upload

leonereveel:
First of all, thanks @ashfurrow for this amazing library!

I need to support file upload using Moya and after look at #114 discussion, I decided to implement my own solution.

I just added a *MultipartTargetType* protocol and made some extensions to *MoyaProvider* where the Target matches the new type.

Here is an example:

```Swift
enum MyAwesomeAPI {
  case EndpointWithImageUpload(NSURL)
}
// Extend from MultipartTargetType
extension MyAwesomeAPI: MultipartTargetType {
   var path: ...
   var parameters: ...
   var method: ...
   var sampleData: ...
   var multipartBody:[MultipartFormData] {
        switch self {
        case .EndpointWithImageUpload(let imageUrl):
                return [MultipartFormData(provider: .File(contentFile), name: "files", mimeType:"image/jpeg", fileName: "photo.jpg")]
        default:
            return []
        }
    }
}
// Call the api
let (progress, response) = RxMoyaProvider<MyAwesomeAPI>()
    .request(MyAwesomeAPI.CallWithImageUpload(fileUrl))

// Do whatever you want with 'progress' and 'response' observables
```
I had some trouble to compile and test the project. I can create some tests and add to this PR if some one help me with that.

Thank you and enjoy!

PR #483, opened 2016-05-12
Multipart upload

leonereveel:
This the same implementation as #481 but is based on master branch.

I need to support file upload using Moya and after look at #114 discussion, I decided to implement my own solution.

I just added a *MultipartTargetType* protocol and made some extensions to *MoyaProvider* where the Target matches the new type.

Here is an example:

```Swift
enum MyAwesomeAPI {
  case EndpointWithImageUpload(NSURL)
}
// Extend from MultipartTargetType
extension MyAwesomeAPI: MultipartTargetType {
   var path: ...
   var parameters: ...
   var method: ...
   var sampleData: ...
   var multipartBody:[MultipartFormData] {
        switch self {
        case .EndpointWithImageUpload(let imageUrl):
                return [MultipartFormData(provider: .File(contentFile), name: "files", mimeType:"image/jpeg", fileName: "photo.jpg")]
        default:
            return []
        }
    }
}
// Call the api
RxMoyaProvider<MyAwesomeAPI>()
    .requestWithProgress(MyAwesomeAPI.CallWithImageUpload(fileUrl))
    .filterCompleted()
    .mapToJSON().....
```

`requestWithProgress` method returns a `ProgressResponse` that contains the progress and the response result when the http call is completed. Use the method `filterCompleted()` to get the actual `Response` and `filterProgress()` to get the progress of the Upload.

I had some trouble to compile and test the project. I can create some tests and add to this PR if some one help me with that.

Thank you and enjoy!


leonereveel:
Hey @ashfurrow! I improved the API a little bit. I created a `ProgressResponse` and added some `filter` methods to get the response and the result. I created a new method called `requestWithProgress` to not collide with actual `request` method. After all this work I thought: What do not make some more changes and add progress to all HTTP calls? What do you think?

ashfurrow:
Cool! Want to email me at ash@ashfurrow.com to schedule a time that works?

ashfurrow:
Cool, this looks awesome. I think we have the following TODOs to get this wrapped up:

- [ ] Add ReactiveCocoa extension for requesting.
- [ ] Add documentation for multipart requests.
- [ ] Add tests for new functionality.

I'll try to take a look at this this weekend.

hirad:
@leonereveel: I just needed multipart upload and started using your fork. 

Once I changed my `TargetType` to conform to `MultipartTargetType`, I had to change all my existing references to `request` to include `progress: nil` parameter because of the ambiguity.

I tried [experimenting](https://github.com/leonereveel/Moya/pull/1) with a slightly different API that I think is simpler. What do you guys think?

Source/Moya.swift
```diff
@@ -4,6 +4,28 @@ import Result
 /// Closure to be executed when a request has completed.
 public typealias Completion = (result: Result<Moya.Response, Moya.Error>) -> ()
 
+public typealias ProgressBlock = (progress:ProgressResponse) -> Void
```
colinta: For consistency with the rest of the code, I think we should add a space after `:` (`progress: ProgressResponse`).  Doesn't have to be part of this PR, we could clean up afterwards. 😃

Source/Moya.swift
```diff
@@ -4,6 +4,28 @@ import Result
...
+
+public struct ProgressResponse {
+    public let totalBytes:Int64
```
colinta: (same thing - on all these types)

Source/Moya.swift
hirad: Agreed!

colinta:
Definitely good stuff in this PR - and I like @hirad's changes in his experimental branch, FWIW.

leonereveel:
Cool guys! Let me know if you need any contribution. @hirad Maybe adding another method only to upload can solve this problem. I did this way because I'm using RxSwift all over my app and these changes fit perfectly. But we can find a better solution.

Source/Moya.swift
```diff
@@ -240,6 +315,59 @@ public extension MoyaProvider {
 
 internal extension MoyaProvider {
     
+    private func sendUpload(target: Target, request:NSURLRequest, multipartBody:[MultipartFormData], progress: Moya.ProgressBlock? = nil, completion: Moya.Completion) -> CancellableWrapper {
+        var cancellable = CancellableWrapper()
+        let plugins = self.plugins
+        
+        let multipartFormData = { (form:RequestMultipartFormData) -> Void in
+            let params = decodeHTTPBody(request.HTTPBody)
```
hirad: This part is a little iffy to me. What happens if the parameter encoding wasn't `.URL`? Then the value in `params` here will be malformed. I'm not sure what'd be the best way to handle encoding of additional parameters in a multipart upload.

One option might be wrapping Alamofire's `ParameterEncoding` and providing our own encoding option for multipart? 

In any case, I think decoding the already-encoded value is problematic. We have access to all the raw values anyway (through `target.parameters` or better yet `self.endpoint(target).parameters`). I think we should use that directly.

leonereveel: Agreed!

leonereveel: @hirad I merged your pull request and added the parameter encoding using Alamofire's URL encoding. I don't think we need to support other encoding types on multipart form.

AndrewSB:
@leonereveel this looks great! What's left before we can use it?

leonereveel:
@AndrewSB I need to do some tests, which I will have time next week. Community review and merge.

colinta:
I'm def happy to review this, I'm excited to (finally!) get uploads into Moya.  Please `@colinta` when it's ready (and `@Moya/contributors`).

leonereveel:
Hey! I did merged master branch and did some improvements. The App I'm developing is almost in production and everything is working so far. The only possible problem I may face is this one:

```swift
manager.upload(request, multipartFormData: multipartFormData) {(result:MultipartFormDataEncodingResult) in
    switch result {
    case .Success(let alamoRequest, _, _):
        // Give plugins the chance to alter the outgoing request
        plugins.forEach { $0.willSendRequest(alamoRequest, target: target) }
        
        // Perform the actual request
        alamoRequest
            .progress { (bytesWritten, totalBytesWritten, totalBytesExpected) in
                progress?(progress:ProgressResponse(totalBytes: totalBytesWritten, bytesExpected: totalBytesExpected))
            }
            .response(queue: queue) { (_, response: NSHTTPURLResponse?, data: NSData?, error: NSError?) -> () in
                let result = convertResponseToResult(response, data: data, error: error)
                // Inform all plugins about the response
                plugins.forEach { $0.didReceiveResponse(result, target: target) }
                completion(result: result)
        }
        
        alamoRequest.resume()
        
        cancellable.innerCancellable = CancellableToken(request: alamoRequest)
    case .Failure(let error):
        completion(result: .Failure(Moya.Error.Underlying(error as NSError)))
    }
}
```
This code executes when Alamofire is making the multipart-form and the cancellable token is set only after the multipart is made. I'm returning a token that is not set yet and if I try to cancel it may be null. How can I return a cancellable token in this case? Anybody knows how I can solve this? Maybe a semaphore can do the job, but lock the main thread is not a good option.

orta:
^ Looks like Swift-lint is running once per target, might want to have an option there

leonereveel:
I merged from master and improved the code a little bit. Can anyone help me to test it? @colinta?

colinta:
Cool!  I will not be able to today, but definitely will do tomorrow.

ashfurrow:
@orta weird, we only run `bundle exec danger

colinta:
Very cool, and I found it very easy to use.  I've got some refactors for you, I'll make a PR into this branch.  I've added some code to the Demo project that uploads a file to Giphy using their public API.

Some comments:

1. Spacing in your code is a little different from the rest of Moya: `func(arg:Type)` instead of `func(arg: Type)`.  I've changed all the instances of this that I found; this follows Swift code conventions, and is internally consistent.
2. I did find an interesting issue: if you don't specify `queue: dispatch_queue_t?`, the response is not called on the main thread - that part is OK, I suppose.  If you *do* specify the queue (e.g. `queue: dispatch_get_main_queue()`), the *progress block* is *not called on this queue*, which surprised me.  I'll include a fix for this in my PR.
3. The design is great; really easy to create a multipart endpoint, easy to add progress.

👍

colinta:
For anyone interested, my recommended changes (and additions to the Demo app) are here: <https://github.com/leonereveel/Moya/pull/3>

leonereveel:
Thank you for your contribution @colinta. I have only two questions for you. I'm a little concerned about this code here:

```swift
internal struct CancellableWrapper: Cancellable {
    internal var innerCancellable: CancellableToken? = nil

    private var isCancelled = false

    internal func cancel() {
        innerCancellable?.cancel()
    }
}
```

I check if this cancellable was canceled before starts uploading **(line 421)**. 
1) As far as I know, all struct values are passed by value (https://developer.apple.com/swift/blog/?id=10). Event if I get this value and changed it, my closure won't "see" the change.
2) I didn't see in any place that `cancel()` method will change the variable `isCancelled` to `true`.

colinta:
Hmm, you've got a good point there! @ashfurrow did you read that ☝️?

I see that the specs test the cancel-ability of this code, though... I'll look at this for a bit more.  Definitely fishy.

colinta:
@ashfurrow Should `isCancelled` map to `innerCancellable?.canceled ?? false`?

colinta:
@leonereveel There is a small difference in your code:

```
// Moya existing code
216:  cancellableToken.innerCancellable = self.sendRequest
...
// your code
271:  cancellableToken = self.sendUpload
```

leonereveel:
I need to do that because the request is not returned immediately. Alamofire builds the multi-part on background and returns the request, where I set the `innerCancellable`, that why it's different.

colinta:
Right - actually I think that *is why* you need to assign the result of `sendUpload` (the "innerCancellable") to the original `cancellableToken`.  If you assign it again (as you are here), it will never be returned to the user - they will be calling `cancel()` on an object whose `innerCancellable` is `nil`.

In `sendUpload` you return another `CancellableWrapper`, and *I think* this is correct - the wrapper in `request()` will call `cancel()` on that wrapper, which will then call `cancel()` on the Alamofire `Cancellable`.

colinta:
I'll try it in the Demo app - calling `cancel()` immediately should make its way into Alamofire, right?

leonereveel:
If define `CancellableWrapper` as `class`, this shouldn't be a problem. The object will be returned as reference and the `innerCancellable` property will be updated after Alamofire builds the multipart. The `cancel` method on that class should set `isCancelled` to true event if `innnerCancellable` was not set yet.

colinta:
No it's still not as simple as that, but I'm learning more about what's going on.

So because `performNetworking` is asynchronous (or it *can* be - in the case of using OAuth, for example), assigning to `cancellableToken` inside of `performNetworking` will have no effect on the code that calls `request`.

```swift
func request() -> String {
    var cancellable = "foo"
    inOneSecond {
        cancellable = "bar"
    }
    return cancellable
}

request() // always "foo"
```

*BUT*, I see that `multipartFormData` is not called synchronously, and so I think the issue here is that `innerCancellable` doesn't have a default value - something that can assign and store the `isCancelled` value.

I propose this: for one, I agree that it *seems like* `CancellableWrapper` should be a class, not a struct, because we're treating it as a reference. Next, I think that `innerCancellable` should be a `Cancellable`, not a `CancellableToken`.  Then, we can assign a default value to `innerCancellable` that can be "canceled" while the request is being constructed.  Because until `innerCancellable` is assigned, the request isn't cancellable, which seems like a bug.

Something like this:

```swift
// change to a class, for reference semantics
internal class CancellableWrapper: Cancellable {
    // calling cancel() on this initial instance will just update the canceled value
    internal var innerCancellable: Cancellable = SimpleCancellable()

    // renamed 'isCancelled' for consistency
    var canceled: Bool { return innerCancellable.canceled ?? false }

    internal func cancel() {
        innerCancellable.cancel()
    }
}

internal class SimpleCancellable: Cancellable {
    var canceled = false
    func cancel() {
        canceled = true
    }
}
```

colinta:
I'm creating #530 to address this issue - and I've got a branch that combines #530 and #483 here: <https://github.com/colinta/Moya/tree/multipart-after-refactor>

leonereveel:
Good solution @colinta!  👍 

I'm looking forward to see everything merge into master branch (upload and this refactoring). I'm glad to see that this discussion solved another possible bug.

leonereveel:
@ashfurrow @colinta  If you need any help to document this, let me know. I can do that this weekend.

ashfurrow:
This is fantastic work and a great example of open source collaboration. Thanks everyone!

Wondering about the new property in the TargetType protocol: `var multipartBody: [MultipartFormData]? { get }`. Could we have a protocol extension that has a default implantation that returns nil? Or would that override the user's implementation?

leonereveel:
I tried that @ashfurrow and it override user's implementation. One of my commits was only to remove the extension.

ashfurrow:
💯

leonereveel:
Fixed multipart-parameters encoding. I didn't know that this must be encoded differently.

ashfurrow:
Tests pass, not sure why Danger failed but I'm gonna merge 💯

ashfurrow:

colinta:
Sweet! I agree about the collaboration. Great feature, and tons of people working on it - awesome!

PR #479, opened 2016-05-04
Change .Underlying to NSError, add error handling docs

Paldom:
According to #478

PR #492, opened 2016-05-23
Remove redundant casting to NSError from tests.

sunshinejr:
So because #479 was merged, we don't really need casting with `as NSError` anymore, because it always succeeds. 🐼


ashfurrow:
Failures unrelated to this PR, merging 👍

PR #493, opened 2016-05-24
[Readme] Update usage section.

sunshinejr:
I've made some changes to Readme due to the feedback. Basically some people didn't really get why their code didn't work with just `MoyaProvider`, while using Reactive Extensions. Because of that, I've added example of provider initializing before the request. The next step I took was to add explicit info about different setup needed for `RxSwift` and `ReactiveCocoa`, instead of linking to them before their sections. Other than that I've made some fixes to URLs that we didn't change after the examples upgrade.

Also, I've changed `GitHubProvider` name in `RxSwift` and `ReactiveCocoa` docs to just `provider` like it is [here](https://github.com/Moya/Moya/blob/master/docs/Providers.md), to be more consistent. 

As always, let me know what do you think! :tada:

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -77,9 +77,10 @@ github "Moya/Moya"
 Usage
 ---
 
-After [some setup](docs/Examples), using Moya is really simple. You can access an API like this:
+After [some setup](docs/Examples/Basic.md), using Moya is really simple. You can access an API like this:
 
 ```swift
+provider = MoyaProvider<GitHub>()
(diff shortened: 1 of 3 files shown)


sunshinejr:
Hmm, also what is going on with the tests? Anybody from @Moya/contributors knows? 🤔

ealeksandrov:
Good cleanup!
About red CI - let's create separate issue for it? I am seeing the same failure for most test runs recently.

PR #496, opened 2016-05-25
Changed MoyaResponse in RxSwift docs to Moya.Response

Legoless:
At first I got confused when reading the documentation, because `MoyaResponse` does not exist (`RxMoyaProvider` does on the other hand). Changed the docs to match the class location which is `Moya.Response`.


ealeksandrov:
Good catch! Can you also make the same update to `ReactiveCocoa.md`?

Legoless:
Done. ;)

ealeksandrov:
Awesome, thanks!

PR #499, opened 2016-05-27
[Changelog] Update breaking change alias.

sunshinejr:
In the **Next** section the **breaking change** wasn't bold and was kinda misleading me to think we are not having any breaking changes right now. Also we were doing the alias in the beginning of the change lately, so I've moved it as well, to make it more coherent.


ashfurrow:
👍

PR #507, opened 2016-06-14
added explicit url creation to endpoint closure doc

NickAger:
Replaced:
  `url(target)`
with:
  `let url =
target.baseURL.URLByAppendingPathComponent(target.path).absoluteString`

as the current distribution of Moya does not include `func url`.

Alternatively `func url` could perhaps be added to Moya as a helper
method, perhaps as an extension method e.g.:

```swift
/// Endpoint closure `url` helper
extension Endpoint {
    public class func url(route: TargetType) -> String {
        return
route.baseURL.URLByAppendingPathComponent(route.path).absoluteString
    }
}
```

...which would remove some code duplication in the examples.

### DIFF
--- a/docs/Endpoints.md
+++ b/docs/Endpoints.md
@@ -61,7 +61,8 @@ analytics.
 
 ```swift
 let endpointClosure = { (target: MyTarget) -> Endpoint<MyTarget> in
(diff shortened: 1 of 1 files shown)


sunshinejr:
Neat, thanks! 🎉 Build failure is not connected with the PR so I'm gonna merge it. 🐼

ashfurrow:
Hey there! Thanks a lot for contributing to Moya! I've invited you to join the Moya GitHub organization – no pressure to accept! If you'd like more information on what that means, check out our [contributor guidelines](https://github.com/Moya/contributors) and feel free to reach out with any questions.

Issue #489, opened 2016-05-19
Decision about deprecated RAC methods.


sunshinejr:
This topic was discussed in #298, but it was long time ago and was mostly for deprecation. And because we have many breaking changes and it is time for breaking, I think it is the best time to decide what to do with `RACSignal` extensions. We have many warnings in Moya about deprecation of methods that we should take care of:

> This will be removed when ReactiveCocoa 4 becomes final. Please visit https://github.com/Moya/Moya/issues/298 for more information.

Personally I think that the extension for `Moya.Error` with `.nsError` property could stay, because we have questions about the `.nsError` property once a few weeks. Other than that I want to know what other @Moya/contributors think! 🐼

aschuch:
I completely agree. We should keep `.nsError` and get rid of RAC 2.0 with the next major release.

petester42:
I don't think we should keep the nsError property. It really doesn't add anything. You can get all the same information from `Error`. The advantage I see NSError is that can be printed nicely to the console but that can easily be added to `Error`

ashfurrow:
I'm 👍 for getting rid of RAC 2 support, and I'm keen to remove the `nsError` property, since it just serves to erase the `MoyaError` type, and was only necessary for Objective-C interop.

ealeksandrov:
+100 for removing deprecated RAC support.
I'm also for removing `nsError`. We have good docs now and it is really duplicate of existing functionality.

sunshinejr:
I'm okay with removing the `nsError` property as long as we document the error section, because this is really a common question here.

ealeksandrov:
We already have updated docs: [ErrorTypes.md](https://github.com/Moya/Moya/blob/master/docs/Examples/ErrorTypes.md)
But I agree maybe it worth short mention or link in Readme.

justinmakaila:
I think we should remove the `nsError` property. After all, we're really coercing our already good errors into NSError for legacy support and dated patterns. I think keeping it just adds to the confusion and is what is leading people to ask about it.

sunshinejr:
Okay, I think we came to the agreement that it should be thrown out from Moya. Is anyone from @Moya/contributors willing to PR these changes? It would include erasing deprecated methods only, if we decide to expand documentation on the `Error` matter, I think it should go into another PR anyways.

sunshinejr:
Closing in favor of #508.

PR #508, opened 2016-06-14
Drop RACSignal support.

sunshinejr:
So after the discussion in #489, I took a stab at removing the support for old version of ReactiveCocoa. As always I'm open for any discussion, so if you think something could be improved, is redundant or is lacking, please let me know! 🐼


ashfurrow:
Awesome! CI failure is unrelated (really need to fix that...) gonna merge!

PR #530, opened 2016-07-08
Refactor Cancellables

colinta:
Unless I'm reading this code incorrectly, there are times when calling `cancel()` on the `cancellableToken` will have no effect.

1. Assigning `innerCancellable` to a property of a struct should have no effect once `cancellableToken` is returned to the caller.  The reason it "works" so far, is because `performNetworking` is *usually* called synchronously.
2. The `isCancelled` property on `CancellableWrapper` always returns `false`, so the request would always be sent, *even if* we fixed (1)

To address these issues:

1. Make `CancellableWrapper` a class, so that assigning `innerCancellable` asynchronously in `performNetworking` will have the desired effect.
2. Rename `isCancelled` to `canceled` for consistency w/ `Cancellable / CancellableToken`
3. Assign a default value to `innerCancellable` that keeps track of `canceled`, so that requests can be aborted before they are sent.

### DIFF
--- a/Demo/Tests/MoyaProviderSpec.swift
+++ b/Demo/Tests/MoyaProviderSpec.swift
@@ -165,6 +165,22 @@ class MoyaProviderSpec: QuickSpec {
                 
                 expect(receivedError).toNot( beNil() )
             }
+
+            it("returns success when request is not cancelled") {
+                var receivedError: ErrorType?
+
+                waitUntil { done in
+                    let target: GitHub = .UserProfile("ashfurrow")
+                    let token = provider.request(target) { result in
+                        if case let .Failure(error) = result {
+                            receivedError = error
+                        }
+                        done()
+                    }
+                }
+
+                expect(receivedError).to( beNil() )
+            }
         }
 
         describe("a provider with a custom endpoint resolver") {
--- a/Demo/Tests/MoyaProviderSpec.swift
+++ b/Demo/Tests/MoyaProviderSpec.swift
@@ -183,6 +183,113 @@ class MoyaProviderSpec: QuickSpec {
             }
         }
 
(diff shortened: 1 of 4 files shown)


colinta:
These are internal changes, no changelog entry necessary, I think.

If the changes are good, I would like to add specs around cancellables.  Async specs - UG.

ashfurrow:
Looks 💯 Sorry for radio silence on the other PR, on vacation this week. This looks good to merge when you're ready, I'll review the comments on the other tomorrow.

colinta:
Some final notes as I let specs run on CI:

- Renamed `canceled` to `cancelled`, because it's spelled with two LLs throughout the rest of the code.
- Prior to this PR, the only place `canceled` was "exposed" was in `CancellableToken` as an internal property, so no breaking changes.  In `CancellableWrapper` it was `isCancelled`, and this was a private var.
- Updated specs accordingly
- Added specs to test cancel behavior, and here's the (strange? unintended?) logic that is tested:
  - if the request is cancelled immediately, the completion handler is not called
  - if the request is cancelled *before the request is sent*, the completion handler is not called
  - if the request is cancelled *after the request is sent*, the completion handler is called with a `.Failure`

In most cases, the completion handler *will be called*, because most people use the default `requestClosure`, which starts the request before `request` returns the `Cancellable`

It's strange, to me, that the completion handler works this way.  Anyone else? CC @Moya/contributors

colinta:
```Executed 102 tests, with 0 failures (0 unexpected) in 3.764 (3.935) seconds```

It's `danger` that is causing circleci to fail 😢 

@ashfurrow GTG?

ashfurrow:
Agreed, that's surprising behaviour. I'll open a separate issue to fix that, this looks great to merge.

Issue #534, opened 2016-07-12
Without the "Carthage/Checkouts" directory, the `.gitmodules` file causes issues when installing with Carthage


justinmakaila:
While trying to install the latest on master, I kept getting an error saying the following:
```
...
*** Checking out Alamofire at "3.4.1"
...
*** Checking out Result at "2.1.2"
*** Checking out ReactiveCocoa at "4bd8d7839d70e41ecb5a58e99bf3a06917b4c7d8"
...
*** Checking out RxSwift at "2.6.0"
*** Checking out Moya at "6c0880a4dee1f4014c8f7ed221de06ebafe5a566"
...
Parse error: expected submodule commit SHA in output of task (ls-tree -z 6c0880a4dee1f4014c8f7ed221de06ebafe5a566 Carthage/Checkouts/Alamofire) but encountered: 
```
This error only occurred if I tried to update `Moya` (I had it pointing to my `multipart-upload` branch). I noticed that the change coming from my branch to master was the removal of the `Carthage/Checkouts` file in `47a226adb44a0983976b26892435159490147bdb`.

My fix was to run `carthage update`, cancel the command once the check out finished, followed by:
```fish
$ cd Carthage/Checkouts/Moya
$ rm -rf .gitmodules
$ cd ../../../; and carthage build Moya
```

The `Carthage` source checks a dependencies `.gitmodules` file for the dependencies, and then attempts to copy them into the project. Since our `.gitmodules` file local directories within the `Moya` directory, it's causing the `git ls-tree` command to return an empty string, killing the `carthage` process.

@Moya/contributors is there any reason we have those `.gitmodules` hanging around anymore? They appear to be left over from someone running `carthage update --use-submodules`

justinmakaila:
Removed `.gitmodules` file on my fork and currently testing. Will file PR and leave merging to you guys

ashfurrow:
This is good to close by #536?

PR #536, opened 2016-07-12
Removed .gitmodules

justinmakaila:
Related to #534. Tested in local repo and got through `carthage update` no problem.


ashfurrow:
Looks like it was checked in here, in October 2015: https://github.com/Moya/Moya/commit/9ee73026ba8e2bf4394de7d795ba42b778b3fea7#diff-8903239df476d7401cf9e76af0252622 so the problem this PR solves has been around since about [version 4.1.0](https://github.com/Moya/Moya/blob/master/Changelog.md#410) of Moya. @justinmakaila Do you think this is something worth including in the changelog? Like, would it be useful for Carthage users to have a note in there about it?

justinmakaila:
@ashfurrow I don't think so, everything looks fine for downloading the build with carthage with and without submodules. I think it made it in as a mistake and became a problem somewhere between then and now as `Carthage` progressed.

ashfurrow:
Okay cool. The only CI failure is a missing changelog, so I'm going to merge (and then submit a PR changing that "no changelog" error to a warning...)

orta:
> Okay cool. The only CI failure is a missing changelog, so I'm going to merge (and then submit a PR changing that "no changelog" error to a warning...)

Might want to check for App-level changes before failing instead

ashfurrow:
Hmm, good idea. I added the error originally to help us be more diligent about our changelog entries. I still worry that if we keep the error, we might intimidate new contributors with a failed CI build instead of a warning. And longtime contributors are [even getting called out](https://github.com/Moya/Moya/pull/530#issuecomment-231434087). Going to move discussion over to https://github.com/Moya/Moya/pull/537

PR #548, opened 2016-07-14
Changelog update & Danger config update

### LINKED ISSUE #546: Nothing about multipart file uploads in 7.0.0 release notes?
Didn't see anything about #483 in 7.0.0's [notes](https://github.com/Moya/Moya/releases/tag/7.0.0), was that on purpose?

ashfurrow:
We forgot to add a changelog entry for #483, this change now fails a build if library source code has changed without a changelog entry. This means that refactors that don't change user functionality will need to be included, too. I'm okay with that, refactors might introduce bugs and changelog entries could really help someone. 

@colinta Any thoughts on this?

Fixes #546.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -7,6 +7,7 @@
 - **Breaking Change** Implements inflights tracking by adding `trackInflights = true` to your provider.
 - **Breaking Change** Changes `MoyaProvider.RequestClosure` to have `Result<NSURLRequest, Moya.Error> -> Void` instead of `NSURLRequest -> Void` as a `done` closure parameter.
 - **Breaking Change** New community guidelines.
+- New multipart file upload.
 - New cURL-based logging plugin.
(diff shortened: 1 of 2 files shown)


colinta:
Good point - we'll probably need to be diligent about telling ourselves and others about that (e.g. if it's a refactor we should point out "include an entry that summarizes the changes, even if they aren't breaking).

ashfurrow:
An expanded message is a good idea 👍

ashfurrow:
Empty Moya message is a known issue: https://github.com/ashfurrow/danger-swiftlint/issues/10

@colinta does that message look good?

colinta:
👍 I haven't gotten around to looking at the documentation, to see where a description of uploading can go.  That's enough of a mention that people can see that it's supported.

orta:
Yep

PR #553, opened 2016-07-17
Adds Download and Upload Task type support to Moya

ceyhuno:
- Adds Download.Request Download type
- Adds Upload.File Upload type
- Puts Multipart Upload type in UploadType enum
- Removes "**multipartBody**" from TargetType protocol and adds a "**task**" instead
- Adds new provider and and a button to Demo app to demonstrate download
- An _Alamofire_ response is now considered _Successful_ even if doesn't have any data with it and we return a _Moya_ response from it using an empty _NSData_ when converting

### DIFF
--- a/Demo/Demo/GitHubAPI.swift
+++ b/Demo/Demo/GitHubAPI.swift
@@ -52,8 +52,8 @@ extension GitHub: TargetType {
             return nil
         }
     }
-    public var multipartBody: [MultipartFormData]? {
-        return nil
+    public var task: Task {
+        return .Request
     }
     public var sampleData: NSData {
         switch self {
--- a/Source/Moya.swift
+++ b/Source/Moya.swift
@@ -38,6 +38,20 @@ public enum StubBehavior {
     case Delayed(seconds: NSTimeInterval)
 }
 
+public enum UploadType {
+    case File(NSURL)
+    case Multipart([MultipartFormData])
+}
+
+public enum DownloadType {
+    case Request(DownloadDestination)
+}
+
+public enum Task {
+    case Request
+    case Upload(UploadType)
+    case Download(DownloadType)
+}
 
 public struct MultipartFormData {
     public enum FormDataProvider {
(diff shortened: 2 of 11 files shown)


ashfurrow:
Awesome thanks! Ignore the SwiftLint errors, we're having problems with it on CI. I can review the code shortly 👍

Demo/Demo/GitHubAPI.swift
```diff
@@ -52,8 +52,8 @@ extension GitHub: TargetType {
             return nil
         }
     }
-    public var multipartBody: [MultipartFormData]? {
-        return nil
+    public var task: Task {
```
ashfurrow: This is smart, good thinking.

Source/Moya.swift
```diff
@@ -449,8 +469,8 @@ internal extension MoyaProvider {
 public func convertResponseToResult(response: NSHTTPURLResponse?, data: NSData?, error: NSError?) ->
     Result<Moya.Response, Moya.Error> {
     switch (response, data, error) {
-    case let (.Some(response), .Some(data), .None):
-        let response = Moya.Response(statusCode: response.statusCode, data: data, response: response)
+    case let (.Some(response), data, .None):
+        let response = Moya.Response(statusCode: response.statusCode, data: data ?? NSData(), response: response)
```
ashfurrow: This makes sense to me, since "failing on empty data" was something Moya was doing differently from Alamofire, so this brings us more in line with their conventions.

ashfurrow:
Cool, I feel like this is good to merge. @leonereveel @colinta you two wrote the original multipart upload code, how does this look to you? Shame we'll have two major versions in quick succession, but I think it's worth it.

colinta:
I didn't *want* to like this, because I assumed it was bike shedding, but as soon as I started digging in, I realized that it was a really good way to design the "how am I sending / receiving this request" aspect. So yeah, I'm on board! 👍

ashfurrow:
Yeah, same here 😄 Okay, gonna wait for others to have input, otherwise I think we're good to merge!

leoneparise:
Good design! Now we have support for download and upload with progress. 👍  Great work!

ashfurrow:
Fab, gonna merge!

Issue #559, opened 2016-07-23
Stickers


ashfurrow:
Moya's gotten pretty big, I reckon it's time we get some stickers 🎉 There are a few aspects to this issue:

- [x] Pick a sticker medium (probably through [stickermule](https://www.stickermule.com)). Thinking a single-colour [transfer sticker](https://www.stickermule.com/products/transfer-stickers) would be the coolest
- [x] Make a sticker design based on the chosen medium ☝️.
- [x] Figure out how to collect addresses for contributors to mail them stickers
- [x] Order the stickers.
- [x] Mail them out.

This is a long-term ticket, I'm thinking we'll try to get it wrapped up sometime before the end of 2016.

esttorhe:
I totally understand going with single-colour but oh boy; i'm going to miss the colors 😉

ashfurrow:
Yeah, I'm leaning more and more towards the colours too 😄

orta:
IMO, think the colours too - can work with https://www.stickermule.com/products/clear-stickers 
Moya logo has too thin lines to pull off the ones CP/Artsy use

ashfurrow:
Agreed, I think either the clear ones or maybe the [square ones](https://www.stickermule.com/products/square-stickers) could work. I think the square ones would go well if we used the GitHub avatar rather than the entire logo:

![](https://raw.githubusercontent.com/Moya/Moya/master/web/logo_square.png)

I'm not sure which one, either that or the actual logo would be more recognizable:

![](https://raw.githubusercontent.com/Moya/Moya/master/web/logo_github.png)

orta:
Agree re:square

Re: full logo might want type e.g.

![screen shot 2016-07-23 at 11 48 54 pm](https://cloud.githubusercontent.com/assets/49038/17081683/0ab56b34-5130-11e6-8215-b5aefe19bd75.png)

specifically done bad on purpose to not think of design

ashfurrow:
It's like "lorem ipsum" but for visual design 😆 

Cool. Okay, maybe a straw poll? Vote by reacting to this comment on if you think a square logo or a full logo would be more recognizable in sticker format.

| Square Logo | Full Logo |
|--------------|----------|
| 🎉                  | ❤️           |

ashfurrow:
Cool, calling it: square logo. Which means the design is finished :tada:

I'll take care of ordering them, the tricky bit is figuring out how to collect addresses. @orta you have experience with this. Google Forms?

[...]

PR #560, opened 2016-07-23
Adds readme for web directory

ashfurrow:
Loosely related to #559.

WIP because GitHub is choosing not to render the `svg` tag.


ashfurrow:
@orta suggestions for the best way to display these colours? This is what I'm going for:

screen shot 2016-07-23 at 1 56 36 pm

orta:
My only known hack here doesn't work http://www.tizag.com/htmlT/bgcolor.php - https://gist.github.com/orta/62febeb573df8e52965a5af7ddf5ba17

orta:
Images are the best way to go

orta:
though the purple is massive, and the rest are the right sized, would recommend throwng in image optim too

ashfurrow:
Okay, i think this is ready for review 👍

orta:
Better fix all them swiftlintsssssssssssssssssssss

ashfurrow:
Yup, working on it now. It's also running more than once.

orta:
but nah, everything looks good from me.

One option for improvement though. You could do all the images at 2x size, then set the width/height to be half that. Then you have retina images. This is how I do it in most of my web projects.


screen shot 2016-07-23 at 9 29 03 pm

with 

screen shot 2016-07-23 at 9 28 23 pm

instead of this at full size:

screen shot 2016-07-23 at 9 29 28 pm

ashfurrow:
okaaaaay

orta:
It's a free TIL ;)

orta:
Yep - that should do it

ashfurrow:
I've opened https://github.com/Moya/Moya/pull/561 to address the SwiftLint problems.

PR #561, opened 2016-07-24
SwiftLint fixes

### LINKED ISSUE #270: Contributing docs
Adding/removing files specifically is fairly non-straightforward. /cc @JRHeaton

ashfurrow:
Goodness, lots in here. SwiftLint us run with Danger on ever pull request we get. This is good, but there are a few problems:

- SwiftLint yielded lots of problems, which are now fixed.
- SwiftLint is run multiple times (see [this issue](https://github.com/ashfurrow/danger-swiftlint/issues/16)). So I updated the plugin to just run once, but on all files. This is fine as long as we don't merge warnings.
- SwiftLint was complaining about `Moya.swift` being over 400 lines of code long. Yikes. I split it out into multiple files.
- Updated Carthage's `Moya.xcodeproj` to reflect new files.
- Added a Danger failure for added/deleted library files without a change to the Carthage Xcode project. Fixes #270.
- Changelog for all this, and fixed formatting for a previous entry.


Dangerfile
```diff
@@ -18,5 +18,12 @@ if has_app_changes && no_changelog_entry && not_declared_trivial
   fail("Any changes to library code need a summary in the Changelog.")
 end
 
+added_swift_library_files = git.modified_files.grep(/Source.*\.swift/).empty?
+deleted_swift_library_files = git.deleted_files.grep(/Source.*\.swift/).empty?
+modified_carthage_xcode_project = !(git.deleted_files.grep(/Moya\.xcodeproj/).empty?)
+if (added_swift_library_files || deleted_swift_library_files) && modified_carthage_xcode_project
+  fail("Added or removed library files require the Carthage Xcode project to be updated.")
+end
```
orta: smart move, please add some comments above ( remember that this is shown on Danger.Systems - so assume no understanding of the problem domain )

Source/Target.swift
```diff
@@ -0,0 +1,94 @@
...
+        self.name = name
+        self.fileName = fileName
+        self.mimeType = mimeType
+    }
+
+    public let provider: FormDataProvider
+    public let name: String
+    public let fileName: String
+    public let mimeType: String
+}
```
orta: debatable, but it'd be nice to have inline docs for a bunch of these classes - jazzy'll pick them up

ashfurrow:
Going to rebuild so Danger runs.

Dangerfile
ashfurrow: Ah, good point. I added a note to the readme but didn't think of danger.systems.

Source/Target.swift
ashfurrow: Yeah, this PR really focuses on splitting existing code out. I think if we merge as-is, then we should open an issue to thoroughly document stuff for Jazzy.

orta: chill to me

orta: just naggin'

ashfurrow:
Cool, pushed the update to the Dangerfile, and no warnings! So that's nice.

ashfurrow:
🍏

orta:
🚢

Issue #551, opened 2016-07-15
RxMoya.TargetType vs Moya.TargetType


AndrewSB:
I'm refactoring my networking code, and I've found a compiler error that I had previously (badly) worked around.

I have an `OnlineProvider` similar to the one in Artsy's iOS app. It's class definition looks like: 
`class OnlineProvider<Target where Target: RxMoya.TargetType>: RxMoyaProvider<Target>`

The part I don't like here, is the `Target where Target: RxMoya.TargetType` because it makes me define my `TargetType` enum as `enum MyAPI: RxMoya.TargetType`. I'd much rather have `Target where Target: Moya.TargetType` and `enum MyAPI: Moya.TargetType`

So in attempt to change that, I've been able to refactor `MyAPI` to conform to `Moya.TargetType`, but I'm now having an issue with my `OnlineProvider`:

<img width="975" alt="screen shot 2016-07-15 at 11 24 19 am" src="https://cloud.githubusercontent.com/assets/3814772/16884514/b69e0810-4a7e-11e6-8819-73343257f43f.png">
I'm not sure, but I *think* that's because `RxMoyaProvider` wants to be specialized with a `RxMoya.TargetType`, not a `Moya.TargetType` and because `RxMoya.TargetType` doesn't inherit from `Moya.TargetType`, the compiler treats them as totally separate protocols (with the exact same definition, albeit)?

ashfurrow:
Yeah, seems likely. We've run into problems in the past on our Demo app (`import`ing more than one version of Moya). You're using Carthage, eh?

AndrewSB:
unfortunately

assuming that if you're using cocoapods, there is no `RxMoya.TargetType` since cocoapods doesn't actually build modules?

ashfurrow:
Well, CocoaPods will include all the specified subspecs in one module, named `Moya`. So if you have `pod 'Moya/RxSwift`, that will pull all the RxSwift Moya files _and_ the regular Moya files into one module. Carthage doesn't have an equivalent of subspecs (they're stance is that you should use multiple frameworks instead – which we considered but decided against).

_Maybe_ you could do something like

```swift
typealias TargetType = RxMoya.TargetType
```

And then you wouldn't see `RxMoya.`?

AndrewSB:
@ashfurrow playing with this right now, will hopefully update soon with what I ended up doing

AndrewSB:
[...]

PR #563, opened 2016-07-25
Make RxMoya & ReactiveMoya dependent on Moya

AndrewSB:
related to #551


ashfurrow:
Cool! This is for Carthage support? Could you summarize the changes+motivation behind them in the PR? (For those of us not in the Carthage know-how.)

AndrewSB:
Will do @ashfurrow! I still need to finish this up - another 2 or 3 commits before this works - will get a short write up here when I get around to those commits. Hopefully tomorrow morning

I'll also need to verify that this doesn't break the existing Cocoapods install - I can try a `pod install`, but it would be awesome if a contributor who's more familiar with Cocoapods (maybe you 😛) also vets this PR when it's done

petester42:
I'll review it when you're done. 😉

AndrewSB:
@petester42 instead of using a build script, it may be more idiomatic to use a compiler directive - similar to how RxCocoa uses `RX_NO_MODULE` https://github.com/ReactiveX/RxSwift/blob/master/RxCocoa/Common/DelegateProxyType.swift#L10

What do you think? I haven't used a compiler directive before, but if it seems like a good idea I'll give implementing it a shot

petester42:
Interesting idea. Sounds good to me.

justinmakaila:
omg you're my hero. this has been on my docket for too long. Excited to check this out when you're done @AndrewSB!

ashfurrow:
Awesome! Sounds exciting, and like it's an involved process. Definitely worth a blog post, I'm sure we're not the only library to have this problem.

AndrewSB:
Just got around to finish implementing this, what I've done is importing the `Moya` framework into each of the Reactive & Rx source files, and remove all Moya source files from the Reactive and Rx targets.

Currently I'm importing Moya into the Reactive & Rx files like so:
```
#if !COCOAPOD
import Moya
#endif
```
where I'm expecting Cocoapods to define a debug flag named `COCOAPOD` kind of like how it's explained in [this SO post](http://stackoverflow.com/questions/24111854/in-absence-of-preprocessor-macros-is-there-a-way-to-define-practical-scheme-spe/24112024#24112024).

I'm relatively sure Cocoapods doesn't define that flag, so some other workaround must be found. I've just created an issue on https://github.com/CocoaPods/CocoaPods asking if there's anything I can do.

Also currently testing this to make sure it works through a carthage install

AndrewSB:
Ok, I think this is now at a point where it's usable in Carthage, testing currently.

AndrewSB:
This works for me through Carthage!

Heres an excerpt from the diff upgrading to this branch in my app 🎉
```diff
diff --git a/Proxy/Internal/Network/Networking.swift b/Proxy/Internal/Network/Networking.swift
index d9bcd86..bafb067 100644
--- a/Proxy/Internal/Network/Networking.swift
+++ b/Proxy/Internal/Network/Networking.swift
@@ -9,7 +9,7 @@ import RxMoya
 
 private let reachabilityManager = ReachabilityManager()
 
-class OnlineProvider<Target>: RxMoyaProvider<Target> where Target: RxMoya.TargetType {
+class OnlineProvider<Target>: RxMoyaProvider<Target> where Target: TargetType {
 
     fileprivate let online: Observable<Bool>
 
@@ -17,7 +17,7 @@ class OnlineProvider<Target>: RxMoyaProvider<Target> where Target: RxMoya.Target
          requestClosure: @escaping RxMoyaProvider<Target>.RequestClosure = RxMoyaProvider.DefaultRequestMapping,
          stubClosure: @escaping RxMoyaProvider<Target>.StubClosure = RxMoyaProvider.NeverStub,
          manager: Manager = RxMoyaProvider<Target>.DefaultAlamofireManager(),
-         plugins: [RxMoya.PluginType] = [NetworkLoggerPlugin(verbose: true)],
+         plugins: [PluginType] = [NetworkLoggerPlugin(verbose: true)],
          online: Observable<Bool> = reachabilityManager.reach) {
 
         self.online = online
@@ -28,7 +28,7 @@ class OnlineProvider<Target>: RxMoyaProvider<Target> where Target: RxMoya.Target
                    plugins: plugins)
     }
 
-    override func request(_ token: Target) -> Observable<RxMoya.Response> {
+    override func request(_ token: Target) -> Observable<Response> {
         let actualRequest = super.request(token)
 
         return online
@@ -43,7 +43,7 @@ class OnlineProvider<Target>: RxMoyaProvider<Target> where Target: RxMoya.Target
 }
 
 protocol NetworkingType {
-    associatedtype T: RxMoya.TargetType, ProxyAPIType
+    associatedtype T: TargetType, ProxyAPIType
     var provider: OnlineProvider<T> { get }
 }
 
@@ -55,7 +55,7 @@ struct Networking: NetworkingType {
 extension Networking {
 
     // public interface
-    func request(_ token: ProxyAPI) -> Observable<RxMoya.Response> {
+    func request(_ token: ProxyAPI) -> Observable<Response> {
         let actualRequest = self.provider
             .request(token)
             .filterSuccessfulStatusCodes()
@@ -99,15 +99,15 @@ extension Networking {
 
 private extension Networking {
 
-    static func endpointsClosure<T>(_ target: T) -> RxMoya.Endpoint<ProxyAPI> where T: RxMoya.TargetType, T: ProxyAPIType {
+    static func endpointsClosure<T>(_ target: T) -> Endpoint<ProxyAPI> where T: TargetType, T: ProxyAPIType {
         let url = "\(target.baseURL.absoluteString)\(target.path)"
-        let sampleResponse: RxMoya.Endpoint.SampleResponseClosure = { .networkResponse(200, target.sampleData) }
+        let sampleResponse: Endpoint.SampleResponseClosure = { .networkResponse(200, target.sampleData) }
 
-        let endpoint = RxMoya.Endpoint<ProxyAPI>(URL: url,
-                                                 sampleResponseClosure: sampleResponse,
-                                                 method: target.method,
-                                                 parameters: target.parameters,
-                                                 parameterEncoding: target.parameterEncoding)
+        let endpoint = Endpoint<ProxyAPI>(URL: url,
+                                          sampleResponseClosure: sampleResponse,
+                                          method: target.method,
+                                          parameters: target.parameters,
+                                          parameterEncoding: target.parameterEncoding)
 
         let baseHeaderFields = [
             "Accept-Version": "2016-07-05",
@@ -120,7 +120,7 @@ private extension Networking {
         // networking is the epitome of side effects + shared global state 😰
         let authenticationHeaderFields: [String: String]? = target.requiresAuth ? ["Authorization": "Bearer \(AuthToken.local()!.accessToken)"] : nil // swiftlint:disable:this line_length
 
-        return endpoint.endpointByAddingHTTPHeaderFields(baseHeaderFields + authenticationHeaderFields)
+        return endpoint.adding(newParameters: baseHeaderFields + authenticationHeaderFields)
     }
 
 }
```

AndrewSB:
@petester42 @ashfurrow @justinmakaila I'm sorry this took so long, but I think it's finally done!

Things left to do:
- [ ] maybe add docs?
- [ ] check cocoapods install
- [ ]  code review from you guys 😄 

Really excited to have this working now!

petester42:
This looks good but I'll review it more throughly tomorrow night. I have a question though. Why are we using the `COCOAPODS` flag in the source files? 

I believe a better solution would be to provide our own flag to do the same thing so that was we are not dependent on cocoapods having this flag. We can simply add our own flag to `Other Swift Flags` by adding `-DNEEDS_MOYA` or something to our targets. That way we check for `#if NEEDS_MOYA` in the code instead. It's basically the same but I like that better.

Moya.xcodeproj/project.pbxproj
```diff
@@ -304,18 +188,18 @@
 			isa = PBXFrameworksBuildPhase;
 			buildActionMask = 2147483647;
 			files = (
-				55B5D5C21C1F6F5500A11B82 /* Result.framework in Frameworks */,
-				0630A35F1BE1C96B00AD4789 /* Alamofire.framework in Frameworks */,
+				465D2C5C1D457C460059D1B6 /* Moya.framework in Frameworks */,
+				0630A3611BE1C96B00AD4789 /* (null) in Frameworks */,
```
petester42: This empty framework can be removed from the targets that have it.

Source/RxSwift/Observable+Moya.swift
```diff
@@ -1,5 +1,8 @@
 import Foundation
 import RxSwift
+#if !COCOAPODS
```
petester42: I think we should use our own swift flag instead of piggybacking on the cocoapods one. Something like `IMPORTS_MOYA` or something.

AndrewSB:
I talked to orta about it https://github.com/CocoaPods/CocoaPods/issues/5889#issuecomment-247737624, he said that Cocoapods defines a env variable, that I think (but am not sure) can be used in code with a `#if` check.

We could also define our own flag, but rn that just seems like extra work to maintain for no added cost. If theres another reason to define the flag, then we should jump on it 😄

Moya.xcodeproj/project.pbxproj
AndrewSB: removed all of them 👍 
thanks for catching that!

Source/RxSwift/Observable+Moya.swift
AndrewSB: I talked to orta about it https://github.com/CocoaPods/CocoaPods/issues/5889#issuecomment-247737624, he said that Cocoapods defines a env variable, that I think (but am not sure) can be used in code with a `#if` check.

We could also define our own flag, but rn that just seems like extra work to maintain for no added cost. If theres another reason to define the flag, then we should jump on it 😄

petester42: Fair enough. It's really just preference so it's all good!

petester42:
As far as your other tasks I think everything has already been addressed. You can merge this whenever you feel comfortable with it.

AndrewSB:
Thanks for the review @petester42 👍 I'm using this branch in prod with one of my apps, so I feel pretty comfortable with it too.

I do want to make sure that this doesn't break the cocoapods install though. @ashfurrow @justinmakaila do either of you use Moya through Cocoapods?

justinmakaila:
@AndrewSB I use Carthage as my dependency manager, and I believe @orta and @ashfurrow are the resident CocoaPods experts

AndrewSB:
@justinmakaila alright, I'll wait for someone else to try this then, I don't want to merge this into master and have it break for everyone depending on this library through Cocoapods

Source/RxSwift/Observable+Moya.swift
ashfurrow: I think `#if !COCOAPODS` is fine, personally.

ashfurrow:
I don't see any CocoaPods breaking changes, I think we're good to merge.

AndrewSB:
awesome, I'll merge 😄

AndrewSB:
oh, @ashfurrow: you mentioned that it would be cool to write a blog post about this change. If you think it still merits one, I'd love to write/help write it. You can email me (asbreckenridge@me.com) to talk about it 👍

BasThomas:
Circle is complaining, is this a problem? @AndrewSB

ashfurrow:
Nah, CI fails like that intermittently. Can you trigger a rebuild on master?

BasThomas:
Rebuilding from their web interface immediately fails the build again, it seems. 😕

PR #565, opened 2016-07-26
Fixes Error 65 intermittent failures on CI

ashfurrow:


ashfurrow:
I bet it's failing because we have different simulators to boot. We'll have to add it as part of the Rake task I guess.

ashfurrow:
Nah, we're currently only using one iOS simulator (tvOS still not supported) and OS X. I changed the UUID to point to the simulator we're actually using. We'll see.

PR #568, opened 2016-07-29
Modifies mapJSON to take optional failsOnEmptyData parameter

### LINKED ISSUE #494: Moya says that response is nil after coming back to the app from lock screen
Hi,

Here is the case:
- return to the app from the lock screen;
- nearly immediately (1-2 seconds) send the HTTP request via moya;

Actual Result:
- moya returns error `Moya.Error.Underlying` and says that the response (NSHTTPURLResponse) is `nil`. NSError contains - `error NSError?    domain: "NSURLErrorDomain" - code: 18446744073709550611 0x000000012c5d9700`

Note:
The code where this debug info caught:
`Moya.swift`
Func: `internal func convertResponseToResult(response: NSHTTPURLResponse?, data: NSData?, error: NSError?) -> Result<Moya.Response, Moya.Error>`

It could be steadily reproduced **only within 1-2 seconds after returning to the application from the lock screen**.

Please help.

spxrogers:
[Enhancement (bug? fix)] modified function `Response.mapJSON` to take an optional paramenter "failsOnEmptyData:" to give users control if they want to get an empty response or throw an error if the response data is empty.

fixes #494

### DIFF
--- a/Demo/Tests/ErrorTests.swift
+++ b/Demo/Tests/ErrorTests.swift
@@ -52,6 +52,32 @@ class ErrorTests: QuickSpec {
             }
         }
 
+        describe("mapping a result with empty data") {
+            let response = Response(statusCode: 200, data: NSData())
+
+            it("fails on mapJSON with default parameter") {
+                var succeeded = false
+                do {
+                    let _ = try response.mapJSON()
+                } catch {
+                    succeeded = true
+                }
+
+                expect(succeeded).to(beTruthy())
+            }
+
+            it("returns default non-nil value on mapJSON with overridden parameter") {
+                var succeeded = true
+                do {
+                    let _ = try response.mapJSON(failsOnEmptyData: false)
+                } catch {
+                    succeeded = false
+                }
+
+                expect(succeeded).to(beTruthy())
+            }
+        }
+
         describe("Alamofire responses should return the errors where appropriate") {
             it("should return the underlying error in spite of having a response and data") {
                 let underlyingError = NSError(domain: "", code: 0, userInfo: nil)
--- a/Demo/Tests/ErrorTests.swift
+++ b/Demo/Tests/ErrorTests.swift
@@ -56,14 +56,14 @@ class ErrorTests: QuickSpec {
             let response = Response(statusCode: 200, data: NSData())
 
             it("fails on mapJSON with default parameter") {
-                var succeeded = false
+                var mapJSONFailed = false
                 do {
                     let _ = try response.mapJSON()
                 } catch {
-                    succeeded = true
+                    mapJSONFailed = true
                 }
 
-                expect(succeeded).to(beTruthy())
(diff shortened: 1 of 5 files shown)


spxrogers:
two things worth noting:
1. `failesOnEmptyData:` is a named parameter to promote overriding this behavior explicit as possible when calling.
2. I chose to return an `NSNull` item instead of an `NSArray` or `NSDictionary` because i guess it felt more representative of an _empty_ result. 

open for discussion/input from other @Moya/contributors if they have differing opinions 😸

Demo/Tests/ErrorTests.swift
```diff
@@ -52,6 +52,32 @@ class ErrorTests: QuickSpec {
...
 
+        describe("mapping a result with empty data") {
+            let response = Response(statusCode: 200, data: NSData())
+
+            it("fails on mapJSON with default parameter") {
+                var succeeded = false
+                do {
+                    let _ = try response.mapJSON()
+                } catch {
+                    succeeded = true
```
ashfurrow: The wording here confused me initially (using `succeeded` to indicate that an error was successfully caught). What do you think about changing this variable to be something like `mapJSONFailed` or similar?

Source/RxSwift/Observable+Moya.swift
```diff
@@ -37,9 +37,9 @@ extension ObservableType where E == Response {
     }
 
     /// Maps data received from the signal into a JSON object. If the conversion fails, the signal errors.
-    public func mapJSON() -> Observable<AnyObject> {
+    public func mapJSON(failsOnEmptyData failsOnEmptyData: Bool = true) -> Observable<AnyObject> {
```
ashfurrow: Well done with including the call-site parameter name, I would've forgotten that.

ashfurrow:
@spxrogers This is great, thanks! Can you add a note to the changelog about the new parameter?

ashfurrow:
@orta there was a Danger failure [on the build](https://circleci.com/gh/Moya/Moya/358) that looked like:

```
$ bundle exec danger
sh: -c: line 0: unexpected EOF while looking for matching `"'
sh: -c: line 1: syntax error: unexpected end of file

Danger has failed this build. 
Found 1 error.

bundle exec danger returned exit code 1
```

I've not seen that before, anything I can do to troubleshoot/reproduce?

Demo/Tests/ErrorTests.swift
spxrogers: completely agree with you on that. 👍

spxrogers:
ah! yes. i'll commit the new changelog item along with the test case variable update later this evening.

ashfurrow:
Thanks a lot! No pressure to rush!

ashfurrow:
Cool, looks good to merge! @spxrogers I think you'll need to merge from master first, feel free to merge after that 👍

PR #579, opened 2016-08-10
Documentation: obsolete requestClosure syntax in Endpoints.md

vzsg:
Hi.

The aforementioned part of the documentation was using two different, but nevertheless obsolete versions of the RequestClosure. To help others avoid hitting the same wall as I did this morning, here's a small patch that brings them up to date.

I also added a force type cast as otherwise it wouldn't compile with Swift 2.2.

### DIFF
--- a/docs/Endpoints.md
+++ b/docs/Endpoints.md
@@ -132,6 +132,8 @@ Instead of modifying the request, you could simply log it, instead.
 
 ```swift
 let requestClosure = { (endpoint: Endpoint<GitHub>, done: MoyaProvider.RequestResultClosure) in
+    // Using the `as!` forced type cast operator is safe here,
+    // as `mutableCopy()` will always return the correct type.
     let request = endpoint.urlRequest.mutableCopy() as! NSMutableURLRequest
 
     // Modify the request however you like.
--- a/docs/Endpoints.md
+++ b/docs/Endpoints.md
@@ -125,18 +125,18 @@ that hits the network. It has a default value of `MoyaProvider.DefaultRequestMap
 which simply uses the `urlRequest` property of the `Endpoint` instance. 
 
 This closure receives an `Endpoint` instance and is responsible for invoking a
-its argument of `NSURLRequest -> Void` with a request that represents the Endpoint.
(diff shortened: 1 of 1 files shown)


ashfurrow:
Thanks for the PR! I like it, not too sure about _recommending_ using a force cast, even though I just looked and it's exactly what my code does too 🙈 What do you think about a 

```swift
guard let request = endpoint.urlRequest.mutableCopy() as? NSMutableURLRequest else { return }
```

Or something similar? I think if we keep the force cast, we should add a short comment explaining it will always succeed. Thoughts?

BasThomas:
I prefer a `guard`. It wouldn't be optional if it could never fail.

vzsg:
I have mixed feelings about using a `guard` here.

Sure, under any other circumstance, forcing is a bad smell; but here it's a direct consequence of how `NSMutableCopying` works. As such, it cannot actually fail.

BasThomas:
But a `guard` does not do any harm in that case, either.

colinta:
I have the same feeling as @vzsg on this: using `guard` implies that we don't know what's going on, and so we're protecting ourselves from something; i.e. `request.mutableCopy()` *might* return *another type* of object.

But that's not how that method works – we all know that it is a reasonable guarantee that we will get back an `NSMutableURLRequest`.

OTOH if this code was written with `guard` I would probably say nothing; it doesn't *hurt* to have a guard there.  But there *is* benefit to showing intention with force-unwrapping.  It says "we know what we're going to get, we trust the system".  I'm OK with that. </twocents>

colinta:
Anecdotal support of this argument: Alamofire uses a force-unwrap to convert `NSURLRequest` to `NSMutableURLRequest` https://github.com/Alamofire/Alamofire/blob/master/Source/Alamofire.swift#L72

vzsg:
That's a nice catch. (It would be pretty nice if this was a public extension, or Moya called the closure with a mutable URLRequest in the first place.)
Also the NSMutableURLRequest documentation states that NSURLSession and co. all use mutableCopy internally. I'd trust the system.

Is that okay if I go with the "add a comment" route?

ashfurrow:
Adding a comment sounds like the best approach, thanks for all the discussion everyone!

ashfurrow:

Issue #580, opened 2016-08-11
Text and Image added to "files" in multipart form data


mattfxyz:
Using Charles, I noticed anything in the MultipartFormData array is added to the files section of the request. When testing the same request Postman, text is added to "fields" and the image is added to "files". 

Is this Moya's intended behavior? It seems like only files should go to files

mattfxyz:
I did some digging, it looks like by setting the `mimeType` to "", Moya tells Alamofire that it's a file type, adding it to the files section of the multipart/form-data request. 

I'm happy to provide a fix.

spxrogers:
a PR would be awesome @mattfxyz :)

leoneparise:
@mattfxyz Is it fixed?

mattfxyz:
This change never made it into Swift 2.3 or 2.2. Are there plans to release an update or can I merge into the release branch?

mattfxyz:
I have a fix in my fork. Can we create a new release branch?

BasThomas:
Should be possible, can you create a PR?

mattfxyz:
What branch? Should I just make a new 2.3 one?

BasThomas:
I suppose so, and then release a `7.something` release for CP / Carthage. @ashfurrow, is this possible?

ashfurrow:
Makes sense to me. You should be able to branch off from the tags of our releases.

mattfxyz:
I created a release-7.0.4 branch and opened a PR to migrate the changes into it.

PR #581, opened 2016-08-12
Call appropriate Alamoefire RequestMultipartFormData append method

mattfxyz:
MultipartFormData objects in Moya will call an AlamoFire method `appendData` to build the form for the request. The `appendBodyPart` method in AlamoFire is not overloaded with nil defaults for mimeTypes or filenames, rather it calls a different `appendBodyPart` method that never assigns them to the form. However, when adding a mimeType (even if it's an empty String) AlamoFire places the content in the "files" section of the request. JSON and String data belong in the "fields" section. This means anything added in a MultipartFormData are added to "files". 

### Fix 

- `mimeType` and `fileName` were made into optionals and the default values were set to `nil`, rather than `""`. The loop that iterates over the array of MultipartFormData will call the appropriate AlamoFire function based on the mimetype and filename. This will case a **break** when a consumer is reading the mimeType property. 
- A helper append function was crated for each data type (data, file, stream) that checks the values of the mimeType and fileName before calling the AlamoFire method. 

I brought this up in Issue #580

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,7 @@
 # Next
 
+- **Breaking Change** `fileName` and `mimeType` are now optional properties on a MultipartFormData object.
+- Correct Alamofire `appendBodyPart` method id called in MultipartFormData.
 - **Breaking Change** Removes `multipartBody` from TargetType protocol and adds a `task` instead.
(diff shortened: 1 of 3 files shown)


ashfurrow:
Fantastic, looks great to me!

ashfurrow:

mattfxyz:
Closes #828

PR #583, opened 2016-08-16
Header documentation enhancements

BasThomas:
- Fixed some typos
- Fixed a duplicated header doc
- Added `backticks` to more clearly show types


BasThomas:
I guess that's not actually needed? Or should I add a `improved header documentation` in the Changelog?

ashfurrow:
This is great, thanks for the PR! Yeah, a changelog entry like the one you suggested would've been great, don't worry about it.

PR #592, opened 2016-08-26
Add pod install hint for Demo app

### LINKED ISSUE #590: Demo requires manual pod installation
Is this intended? I think it would be better to include the pods directly, so users trying the demo can skip this step.

BasThomas:
Fixes #590


colinta:
👍

PR #593, opened 2016-08-29
add suport to swift 2.3

gustavogervasio:
add support to swift 2.3

### DIFF
--- a/Source/Moya+Defaults.swift
+++ b/Source/Moya+Defaults.swift
@@ -3,8 +3,15 @@ import Alamofire
 /// These functions are default mappings to `MoyaProvider`'s properties: endpoints, requests, manager, etc.
 public extension MoyaProvider {
     public final class func DefaultEndpointMapping(target: Target) -> Endpoint<Target> {
-        let url = target.baseURL.URLByAppendingPathComponent(target.path).absoluteString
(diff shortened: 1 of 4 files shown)


ashfurrow:
Makes sense to me! Good to merge when 🍏

Source/Moya+Defaults.swift
```diff
@@ -3,8 +3,8 @@ import Alamofire
 /// These functions are default mappings to `MoyaProvider`'s properties: endpoints, requests, manager, etc.
 public extension MoyaProvider {
     public final class func DefaultEndpointMapping(target: Target) -> Endpoint<Target> {
-        let url = target.baseURL.URLByAppendingPathComponent(target.path).absoluteString
-        return Endpoint(URL: url, sampleResponseClosure: {.NetworkResponse(200, target.sampleData)}, method: target.method, parameters: target.parameters)
+        let url = target.baseURL.URLByAppendingPathComponent(target.path)!.absoluteString
```
ashfurrow: Hmm, what if we replaced this with a check for a `nil` URL? Something like a `guard` that `else`es to a `fatalError`, then we can avoid the force unwraps?

gustavogervasio: I liked the idea.

PR #594, opened 2016-08-29
swift 2.3

gustavogervasio:
add support to swift 2.3

### DIFF
--- a/Source/Moya+Defaults.swift
+++ b/Source/Moya+Defaults.swift
@@ -3,8 +3,17 @@ import Alamofire
 /// These functions are default mappings to `MoyaProvider`'s properties: endpoints, requests, manager, etc.
 public extension MoyaProvider {
     public final class func DefaultEndpointMapping(target: Target) -> Endpoint<Target> {
-        let url = target.baseURL.URLByAppendingPathComponent(target.path).absoluteString
-        return Endpoint(URL: url, sampleResponseClosure: {.NetworkResponse(200, target.sampleData)}, method: target.method, parameters: target.parameters)
+        guard let stringURL = target.baseURL.URLByAppendingPathComponent(target.path)?.absoluteString  else {
+               fatalError("\(target) url is not valid")
+        }
+        return Endpoint(URL: stringURL, sampleResponseClosure: {.NetworkResponse(200, target.sampleData)}, method: target.method, parameters: target.parameters)
+        /*
+        if let url = target.baseURL.URLByAppendingPathComponent(target.path) {
+            if let stringURL = url.absoluteString {
+                return Endpoint(URL: url!, sampleResponseClosure: {.NetworkResponse(200, target.sampleData)}, method: target.method, parameters: target.parameters)
+            }
+        }
+        fatalError("\(target) url is not valid")*/
     }
 
     public final class func DefaultRequestMapping(endpoint: Endpoint<Target>, closure: RequestResultClosure) {
--- a/Changelog.md
+++ b/Changelog.md
@@ -11,6 +11,10 @@
 - Added `supportsMultipart` to the `Method` type, which helps determine whether to use `multipart/form-data` encoding.
 - Added `PATCH` and `CONNECT` to the `Method` cases which support multipart encoding.
 
+# 7.0.1
+
+- Add support to swift 2.3
+
 # 7.0.0
 
 - **Breaking Change** Drops support for `RACSignal`.
--- a/Demo/Demo.xcodeproj/project.pbxproj
+++ b/Demo/Demo.xcodeproj/project.pbxproj
@@ -369,9 +369,11 @@
 					};
 					5EC197DF1A1BB16D00F4DFD4 = {
 						CreatedOnToolsVersion = 6.1.1;
+						LastSwiftMigration = 0800;
 					};
 					5EC197F61A1BB16D00F4DFD4 = {
 						CreatedOnToolsVersion = 6.1.1;
+						LastSwiftMigration = 0800;
 					};
 				};
 			};
@@ -873,6 +875,7 @@
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks";
 				PRODUCT_BUNDLE_IDENTIFIER = "com.ashfurrow.$(PRODUCT_NAME:rfc1034identifier)";
 				PRODUCT_NAME = "$(TARGET_NAME)";
+				SWIFT_VERSION = 2.3;
 			};
 			name = Debug;
 		};
@@ -886,6 +889,7 @@
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks";
 				PRODUCT_BUNDLE_IDENTIFIER = "com.ashfurrow.$(PRODUCT_NAME:rfc1034identifier)";
 				PRODUCT_NAME = "$(TARGET_NAME)";
+				SWIFT_VERSION = 2.3;
 			};
 			name = Release;
 		};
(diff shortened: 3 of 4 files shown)


Source/Moya+Defaults.swift
```diff
@@ -3,8 +3,12 @@ import Alamofire
 /// These functions are default mappings to `MoyaProvider`'s properties: endpoints, requests, manager, etc.
 public extension MoyaProvider {
     public final class func DefaultEndpointMapping(target: Target) -> Endpoint<Target> {
-        let url = target.baseURL.URLByAppendingPathComponent(target.path).absoluteString
-        return Endpoint(URL: url, sampleResponseClosure: {.NetworkResponse(200, target.sampleData)}, method: target.method, parameters: target.parameters)
+        if let url = target.baseURL.URLByAppendingPathComponent(target.path) {
```
ashfurrow: Cool, much better! What do you think about this:

```swift
guard let url = target.baseURL.URLByAppendingPathComponent(target.path),
      let stringURL = url.absoluteString else {
    fatalError(...)
}

return Endpoint(URL: stringURL, sampleResponseClosure: {.NetworkResponse(200, target.sampleData)}, method: target.method, parameters: target.parameters)
```

This keeps the "happy path" unindented, what do you think?

gustavogervasio: It is better 👍

gustavogervasio: @ashfurrow the test didn't pass
❌  /Users/distiller/Moya/Source/Moya+Defaults.swift:6:15: initializer for conditional binding must have Optional type, not 'NSURL'

        guard let url = target.baseURL.URLByAppendingPathComponent(target.path),
              ^

Do you known what?

ashfurrow: Huh! It must not be an optional. Try this?

```swift
guard let stringURL = target.baseURL.URLByAppendingPathComponent(target.path).absoluteString  else {
    fatalError(...)
}
```

`URLByAppendingPathComponent` should return a non-optional URL. Try it out and see if you can get it to work, let me know if I can help :+1:

gustavogervasio: It is not work, because URLByAppendingPathComponent return a NSURL?

BasThomas: In that case, adding a `?` should just work.

```swift
guard let stringURL = target.baseURL.URLByAppendingPathComponent(target.path)?.absoluteString  else {
    fatalError(...)
}
```
Right?

gustavogervasio: Yes. But do you like to use this?
I don't like use "?".
In this case i prefer:
```
        if let url = target.baseURL.URLByAppendingPathComponent(target.path) {
            if let stringURL = url.absoluteString {
                return Endpoint(URL: url!, sampleResponseClosure: {.NetworkResponse(200, target.sampleData)}, method: target.method, parameters: target.parameters)
            }
        }
        fatalError("\(target) url is not valid")

```

What do you think?

BasThomas: In that case, I'd do the following:

```swift
if let
    url = target.baseURL.URLByAppendingPathcomponent(target.path),
    stringURL = url.absolutestring {
        return Endpoint(URL: url, sampleResponseClosure: { .NetworkResponse(200, target.sampleData) }, method: target.method, parameters: target.parameters)
}
fatalError("\(target) url is not valid")
```

Notes:
- `url` shouldn't need force unwrapping, as it has been unwrapped previously, right?
- we might want to go with an inverted code path here, using `guard` to fail early.

gustavogervasio: It is true.

charlag:
Excuse me, I may be wrong, but why Xcode version in CircleCI is set to "7.3.1"? Tests fail because of that I suppose. Looks like they support Xcode 8 already https://circleci.com/blog/xcode-8-0-beta-support-for-ios-builds/

BasThomas:
@charlag because this is a PR to add support for Swift 2.3. We are currently using Swift 2.2, which also means using Xcode 7.3.1. We may need to change the CircleCI file though? (cc @ashfurrow @gustavogervasio)

ashfurrow:
Xcode 8 supports both Swift 2.3 and Swift 3.0, right? May we do need Xcode 8 then?

BasThomas:
Sorry if I wasn't clear - that's what I was trying to say. 

Swift 2.2 == Xcode 7
Swift 2.3 == Xcode 8
Swift 3.0 == Xcode 8

marcelofabri:
Is there any blockers on this besides updating `circle.yml`?

charlag:
Is there a plan on updating to 2.3/3.0? Xcode is GM now. I want to help with migration but don't know what is the right way to do it.

sunshinejr:
@charlag I've opened PR with help wanted on the Swift 3.0 migration for Xcode GM Seed. You can take a look #608, we need some help!

delta98:
any update on this being released?

ashfurrow:
I can merge+release this for Swift 2.3 before releasing a separate version for 3.0, _or_ we can just do it for 3.0. What does everyone think?

marcelofabri:
I think we should have a Swift 2.3 release 💯

charlag:
2.3 should be awesome. I don't think there will be much support for 3.0 until the middle of October and Moya is one if my 'blockers' (I still love it!).

ashfurrow:
Okay, will reply from @gustavogervasio's branch and push a tag. Wish me luck!

ashfurrow:
Rather after #621, I think master is updated for 2.3.

ashfurrow:
I've cherry-picked the commits from #621 into the 7.0.1 tag and everything looks good. This means we can release _just_ Swift 2.3 support and not also the breaking changes currently on master. I'm downloading the Xcode 8 GM to test locally before releasing. Thanks everyone!

marcelofabri:
@ashfurrow You probably want to bump Alamofire version to 3.5.0 in the podspec as well 💪 
(that PR only bumps in Cartfile)

ashfurrow:
Cool, will do :+1: What's the trick to get Xcode 8 to use Swift 2.3?

marcelofabri:
It should prompt you to migrate if `SWIFT_VERSION` is not set on the project (which #621 set).

ashfurrow:
Wow, that was a lot of work. I've got 7.0.1 released on Carthage and CocoaPods, would someone be able to verify it's working for them as expected?

I'm going to close this PR.

marcelofabri:
It seems it's working with CocoaPods, but apparently (at least one) breaking change got in: https://github.com/Moya/Moya/blob/7.0.1/Source/Target.swift#L10

ashfurrow:
😢 @marcelofabri can you check for other breaking changes? If that's the only one, we can replace the `7.0.0` tag and hopefully no one will be the wiser.

ashfurrow:
I've pushed the release up to a `release-7.0.1` branch.

marcelofabri:
@ashfurrow I think all of them got in, since there are no commits to master after the tag 😬

ashfurrow:
Okay, well let's stop the bleeding. Hold on.

ashfurrow:
I've pushed a 7.0.1 tag with an identical commit to 7.0.0 so no one's projects will break. I'll take another look tomorrow, thanks for pointing out the issue!

ashfurrow:
Cherry-picking commits turned out to be too complicated, so instead I branched off of the 7.0.0 tag and redid the changes for Swift 2.3: https://github.com/Moya/Moya/compare/7.0.1...release-7.0.1-take-two . I'm going to lint the podspec locally before releasing, I'll let you know.

ashfurrow:
7.0.2 has been released, could someone take a look and verify it's all good?

marcelofabri:
All good from what I've seen 👍 🎉

ashfurrow:
Great, thanks!

PR #599, opened 2016-09-04
Added a snippet of code to better understand how to use a plugin

filograno:
I have added a snippet of code to clarify the use of a Plugin and done some textual modification to better understand what a plugin is.


BasThomas:
Woo, nice!

PR #612, opened 2016-09-09
fix for swift3 variadic params

AndrewSB:
Two downsides to the approach I went with:

1. Its a breaking change, but that's understandable, given this is a source change
2. In the print output, the items may be printed as an array, instead of a variadic list. Not sure how to get around that, I looked at function currying and partially applying the print function, but it didn't lead anywhere


sunshinejr:
Wow, fast! Thanks! Looks fine to me! About that problem with printing array, would it work that instead of using array, make the items parameter the last one in tuple, so it can accept variadic parameter?

AndrewSB:
Oh that might! Let me try it

AndrewSB:
Yeah, that worked @sunshinejr. Awesome insight!
If this looks good to you, then I think it's ready to merge 😄

sunshinejr:
💯

PR #608, opened 2016-09-08
[Swift 3.0] Update for Xcode GM Seed.

sunshinejr:
Hey guys! 🎉  I tried to update Moya to the latest Swift 3.0 shipped with Xcode GM Seed. As there is work that still needs to be done, I would use some help with:

- [x] `sendAlamofireRequest` method in [Moya+Internal](https://github.com/Moya/Moya/blob/swift-3.0/Source/Moya%2BInternal.swift#L187-L221). (@sunshinejr, PR: #613)
- [x] [NetworkLoggerPlugin](https://github.com/Moya/Moya/blob/swift-3.0/Source/Plugins/NetworkLoggerPlugin.swift). (@AndrewSB, PR: https://github.com/Moya/Moya/pull/612)
- [x] [ReactiveCocoa extensions with Tests](https://github.com/Moya/Moya/tree/swift-3.0/Source/ReactiveCocoa). (@ashfurrow, PR: #647)
- [x] [Tests](https://github.com/Moya/Moya/tree/swift-3.0/Demo/Tests). (@sunshinejr, PR: #625)
- [x] Continuous Integration.
- [x] Any other things you may find.

I think together we can make it better 👊Cheers!

PR #613, opened 2016-09-09
[Swift 3.0] Update to new Alamofire Requests & Progress API.

sunshinejr:
So due to [Alamofire#1455](https://github.com/Alamofire/Alamofire/pull/1455) change, we have to update our progress blocks as well. For now I've just updated the new Alamofire's Progress API to our existing Progress API. If we should update to two blocks instead of one (`downloadProgress`, `uploadProgress`) is eligible for discussion. 

More info #608.

Cheers! 🐼

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,5 @@ Carthage/Build
 Carthage/Checkouts
 Cartfile.resolved
 Demo/Pods
+.ruby-version
+.ruby-gemset
--- a/Demo/Demo.xcodeproj/project.pbxproj
+++ b/Demo/Demo.xcodeproj/project.pbxproj
@@ -508,7 +508,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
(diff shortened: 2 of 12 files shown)


edwardvalentini:
this looks great!

edwardvalentini:
Oh no - alamofire deprecated the Int64 variants for the progress handlers

orta:
sounds like this is definitely a major version bump then 👍

edwardvalentini:
I'm working on a PR for this right now.

edwardvalentini:
even though 7.0.0 was released recently - does it make sense to move to 8.0.0 as @orta suggests?

BasThomas:
I too think that makes the most sense, yeah.

edwardvalentini:
I'm testing #618 #613 #608 on my own fork and everything seems to be working smoothly 👍

sunshinejr:
Okay, I think that it can be merged to the main Swift-3.0 repo, and make sure that Moya builds, and if there are any issues about this PR we can make a new one with fixes.

PR #614, opened 2016-09-09
Podspec update alamofire

edwardvalentini:
Alamofire updated their podspec to 4.0.0 on master which causes errors if we still reference 4.0.0-beta.2


sunshinejr:
Tried it out and it seems fine! Thanks!

PR #618, opened 2016-09-11
Fix for Alamofire deprecation of (Int64,Int64,Int64) style progress blocks and Alamofire 4.0.0 readiness

edwardvalentini:
This PR fixes the progress handlers to use Progress instead of the [now deprecated by Alamofire]-style (Int64,Int64,Int64) variant.


sunshinejr:
Thanks!

PR #630, opened 2016-09-14
Test our Carthage support in CI

orta:
We should get the CI to keep track of our Carthage support, I debated putting this in danger but it takes a very long time to run, and when I ran it locally Xcode crashed :-/

screen shot 2016-09-14 at 6 11 16 pm

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -104,3 +104,17 @@ task :release, :version do |task, args|
                    name: version,
                    body: changelog.split(/^# /)[2].strip)
 end
+
+desc 'Run a local copy of Carthage on this current directory.'
+task :carthage_test do
+  # make a folder, put a cartfile in and make it a consumer
+  # of the root dir
+
+  Dir.mkdir("carthage_test")
+  File.write(File.join("carthage_test", "Cartfile"), "git \"file://#{Dir.pwd}\"")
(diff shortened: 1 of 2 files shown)


orta:
Maybe it'll be better on CI

ashfurrow:
Great idea!

orta:
Dang, an extra 35m to run on each CI run. Wonder if it's worth adding Jenkins _just_ for doing this instead of putting it on the real CI.

Anyway, we _know_ master is broken, this just validates it. Happy to either rebase with master isn't broken.

ashfurrow:
I'm not surprised master isn't working, the 7.0.x stuff branched off from the 7.0.0 tag, and all the Swift 3 stuff is in its own branch, so master has been languishing for a bit. 

We could consider only testing one platform with Carthage, instead of iOS, macOS, watchOS, and tvOS. It might be worth the time tradeoff, not sure. What do you think?

orta:
I'd be OK with just doing iOS, if we can - once some of the other PRs are in, I'll look at the docs to see if I can do that

ashfurrow:
Great work! I think keeping Carthage support working is an acceptable tradeoff for a slower CI, thanks Orta.

Issue #518, opened 2016-06-29
Bool value in post and put parameter is always 0


lucasp90:
Hi,

I'm trying to send a request with a boolean value, so I configured my Target like this:

``` swift
public enum MyAPI {
   ...
  case Foo(NSNumber, String, String, String, Bool)
  ...
}
```

My parameters:

``` swift
  public var parameters: [String: AnyObject]? {
    switch self {
   ...
    case .Foo(_, let data1, let data2, let data3, let boolData):
      return [
        "data1": data1,
        "data2": data2,
        "data3": data3,
        "somethingBool": Int(boolData)
      ]
    default:
      return nil
    }
  }
```

Problem:
My request is always sending false as the value.

jasl:
What about

```swift
  public var parameters: [String: AnyObject]? {
    switch self {
   ...
    case .Foo(_, let data1, let data2, let data3, bool data4):
      return [
        "data1": data1,
        "data2": data2,
        "data3": data3,
        "somethingBool": data4 ? 1 : 0
      ]
    default:
      return nil
    }
  }
```

ashfurrow:
You should be able to just have `"somethingBool": data4` as an element in the returned dictionary, right?

lucasp90:
@ashfurrow: Sorry, I forgot to mention it. That was my first attempt, and it also sends false there.
I've tried to make the request via postman and it succeeded with the proper value. Maybe the specification of the request helps to detect if it's a server error or Moya's configuration one.

ashfurrow:
Thanks for the follow-up. How are you detecting the value, on the server side or client side? Could you try using the [network activity plugin](https://github.com/Moya/Moya/blob/master/docs/Plugins.md#network-activity-indicator) to log the request before it's sent?

lucasp90:
@ashfurrow I'm detecting it on the client side. The parameter is being sent as 0 or 1 but I guess I'm missing to track the __real__ request body being sent to the server. I'll try to use that plugin and see what happens

lucasp90:
@jasl I've tried to do what you've said and got the same result.

I've finally managed to save the value by sending the value as a string: `"true"` or `"false"`, but I'm not really sure what's causing the boolean to fail.

```swift
  public var parameters: [String: AnyObject]? {
    switch self {
   ...
    case .Foo(_, let data1, let data2, let data3, bool data4):
      return [
        "data1": data1,
        "data2": data2,
        "data3": data3,
        "somethingBool": "\(data4)"
[...]

PR #637, opened 2016-09-15
Added parameter example using Target's associated values

lucasp90:
Hi!

This adds a simple example of using Target's associated values as parameters for our requests. I've  chosen an example with a `Bool` value according to what you've pointed out at issue #518.

### DIFF
--- a/docs/Targets.md
+++ b/docs/Targets.md
@@ -13,6 +13,7 @@ public enum GitHub {
     case Zen
     case UserProfile(String)
     case UserRepositories(String)
+    case Branches(String, Bool)
 }
 ```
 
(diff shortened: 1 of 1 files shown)


docs/Targets.md
```diff
@@ -90,6 +96,8 @@ public var sampleData: NSData {
         return "{\"login\": \"\(name)\", \"id\": 100}".dataUsingEncoding(NSUTF8StringEncoding)!
     case .UserRepositories(let name):
         return "[{\"name\": \"Repo Name\"}]".dataUsingEncoding(NSUTF8StringEncoding)!
+    case .Branches(_, _):
```
ashfurrow: In cases like this, where there's nothing in the parens, you can use `case .Branches` without the parens at all.

lucasp90: Great! Didn't know about it. Fixed it 👍

ashfurrow:

PR #643, opened 2016-09-16
Swift package manager support

AndrewSB:
Things we're still waiting for:

- [x] Ability to explicitly import Foundation (**pushing this off** for a future PR/release)
- [x] Ability to import libraries through dynamic/static libraries (currently spm doesn't build dependencies, it just embeds the source files in the Moya xcodeproj (**pushing this off** for a future spm release, spm doesn't support this as of now)
- [x] Merging https://github.com/Moya/Moya/pull/698, so Moya has a swift package manager compatible layout
- [x] Alamofire fixing their swift package manager installation (https://github.com/Alamofire/Alamofire/issues/1815)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,6 @@ Carthage
 Demo/Pods
 .ruby-version
 .ruby-gemset
+# Swift Package Manager
+.build
+Packages
--- a/Changelog.md
+++ b/Changelog.md
@@ -8,6 +8,7 @@
 - Re-add `MultiTarget` to project.
 - Adopted an SPM-compatible project structure.
 - Moved tests to Moya.xcodeproj.
+- Supported the swift package manager
 
 # 8.0.0-beta.6
 
--- a/Package.swift
+++ b/Package.swift
@@ -7,6 +7,12 @@ let package = Package(
             name: "Moya"
         ),
         Target(
+            name: "ReactiveMoya",
+            dependencies: [
+                .Target(name: "Moya")
+            ]
+        ),
+        Target(
             name: "RxMoya",
             dependencies: [
                 .Target(name: "Moya")
@@ -14,10 +20,13 @@ let package = Package(
         )
     ],
     dependencies: [
-        .Package(url: "https://github.com/AndrewSB/RxSwift", versions: Version(3, 0, 2)...Version(4, 0, 0)),
-        .Package(url: "https://github.com/Alamofire/Alamofire", versions: Version(4, 0, 0)...Version(5, 0, 0))
+        .Package(url: "https://github.com/Alamofire/Alamofire", majorVersion: 4),
+        .Package(url: "https://github.com/ReactiveCocoa/ReactiveSwift", "1.0.0-alpha.3"),
+        .Package(url: "https://github.com/ReactiveX/RxSwift", majorVersion: 3)
     ],
     exclude: [
         ".build",
+        "Demo",
+        "Tests"
     ]
 )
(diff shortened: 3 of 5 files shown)


AndrewSB:
Both ReactiveSwift & RxSwift vend through spm now, trying to see if I can get Alamofire to work, running into some issues related to https://github.com/Alamofire/Alamofire/issues/1544

After that we should be ready to go 🎉

VivienGiraud:
Any update ?

AndrewSB:
@VivienGiraud still waiting on the folks at Alamofire. 

I opened https://github.com/Alamofire/Alamofire/issues/1815, since they previously closed https://github.com/Alamofire/Alamofire/issues/1544

scottrhoyt:
I'd love to get this working as well. However, short of some fragile hackery, I don't see it as a possibility while SPM doesn't support setting the deployment target and [doesn't support non Mac/Linux platforms or depending on Apple Frameworks](https://github.com/apple/swift-package-manager/blob/master/Documentation/Reference.md#depending-on-apple-modules).

The best long term solution would be hassle the SPM folks to add this in time for the next Swift version.

If SPM support is high priority, the other option is to depend instead on an Alamofire fork with the offending functionality removed (if possible) until the underlying issue in SPM is resolved. This could even be the beginning of replacing Alamofire with a Moya-specific `URLSession` wrapper. That would obviously come with it's own list of pros and cons.

AndrewSB:
@scottrhoyt: I was able to use spm to deploy an iOS app (I only had one private dependency), but it was very similar to using a submodule. `swift package update` pulled down the files and built a binary (that I ignored), and then I dragged the source files into my iOS xcodeproj

not ideal, but it worked

scottrhoyt:
Thanks @AndrewSB. That is good to know. Like you said, you were using it very similarly to a submodule. I suspect you could get Moya to build just fine using a similar method. Even though `swift build` might fail, you can drag the resolved dependencies' source files into an Xcode project and build from there. To get `swift build` to work is another matter.

The core of the issue holding up Alamofire seems to be a dependency on a newer version of `Foundation` than the version SPM defaults to. The solution proposed by the Alamofire maintainers is to basically use Xcode configurations to override this behavior. A pure SPM solution would involve being able to depend upon specific Apple frameworks within `Package.swift`--this is what currently doesn't exist.

If the goal of getting SPM working for Moya is to add Linux support (as opposed to another macOS-based dependency manager), then the `Foundation` dependencies within Alamofire will likely only cause even more headaches. This is because of the how `Foundation` is deployed on Linux (bundled with Swift and not available via SPM) and how it is out of sync with `Foundation` on Apple platforms in terms of both API naming conventions and missing functionality. This makes smaller projects a hassle to port to Linux and maintain that support. I can imagine Alamofire would be a much larger undertaking.

scottrhoyt:
Hey @AndrewSB, I took a look at this and realized it just needed one small change to build correctly via SPM. Might be a good idea to add an SPM build job to CI to catch these kind of things in the future (could be a separate PR), but this is looking good! 👍

AndrewSB:
I'm going to retarget this PR to come off #885.
Hopefully we'll have Alamofire cooperating soon so we can merge this in 👌

scottrhoyt:
Try a `swift build` with that commit that I pushed. I think it builds fine now.

AndrewSB:
I'll check it out

AndrewSB:
Woah, that was awesome. Cool, I'll cherry pick two commits onto Moya's `master` and we should be ready to go here

PR #671, opened 2016-09-22
Refactor code, remove warnings

sendyhalim:
Refactor code to remove xcode warnings

### DIFF
--- a/Source/Moya+Internal.swift
+++ b/Source/Moya+Internal.swift
@@ -203,7 +203,7 @@ private extension MoyaProvider {
         }
 
         // Perform the actual request
-        if progressCompletion != nil {
+        if let _ = progressCompletion {
             switch progressAlamoRequest {
             case let downloadRequest as DownloadRequest:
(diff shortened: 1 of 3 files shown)


Source/Moya+Internal.swift
```diff
@@ -203,7 +203,7 @@ private extension MoyaProvider {
         }
 
         // Perform the actual request
-        if let progress = progressCompletion {
+        if progressCompletion != nil {
```
BasThomas: What about `if let _ = progressCompletion` Instead of the `nil` check?

ashfurrow: I personally prefer `if let _ = ` as well, wasn't sure if it was just me 😅

BasThomas: 🙌

sendyhalim: Sure, that's better! 😄

BasThomas:
I suppose that changelog entry is not needed here. 😉

PR #674, opened 2016-09-24
Update ReactiveSwift to 1.0.0-alpha.2

wongzigii:


wongzigii:
@ashfurrow Would you like to rebuild it manually?

ashfurrow:
Hmm, getting a legitimate failure this time, but I don't understand. Your PR is exactly what I would've done, not sure why it's failing. Keep it open and I'll investigate, thanks again!

wongzigii:
Seems like Circle is complaining about RxSwift. I saw `error: assigning non-escaping parameter 'selector' to an @escaping closure` in this build.

wongzigii:
I've notice Moya is ignoring `Carthage.resolve` and `Carthage.parivate`. Should we move them back?

ashfurrow:
That's a good question! @justinmakaila can you chime in on this?

petester42:
For a library I don't see the resolved file being necessary since we don't really need to care which exact version we are using. Not having the resolved file means that if we specify ~> 2.2 then when someone builds Moya they will always get the most recent 2.2.x version without us having to run carthage update to update the resolved file.

At least, that is what I understood from reading the description of the resolved file.

wongzigii:
> Not having the resolved file means that if we specify ~> 2.2 then when someone builds Moya they will always get the most recent 2.2.x version without us having to run carthage update to update the resolved file.

Partly, that is what `~>` use for version requirement compatibility in Cartfile, since Carthage's compatibility is determined according to Semantic Versioning.

IMO, `Carthage.resolve` file is supplementary which describes the versions of Moya's *dependencies*.

Imagine this scenario: You saw Moya released a new version with patch, and you run `Carthage update` to bump your project to a new stage. However, some dependency of Moya have new commits. (And this is why we should avoid using `github "SOME_DEP" "master/branch"` or `github "SOME_DEP" "https://github.com/XXX.git"` unless you indeed need it.)  You saw errors and ticketed an issue. Unfortunately, we don't exactly know why you are facing those errors, for the reason that our dependencies are totally different.

FYI: [cartfileresolved](https://github.com/Carthage/Carthage/blob/master/Documentation/Artifacts.md#cartfileresolved)

Correct me if I am wrong. :)

AndrewSB:
For what it's worth, I think @wongzigii has the right idea. We should probably be using the semantic version of a library instead of (basically) a commit hash (in this case `master`).

@wongzigii if you update the tag ReactiveSwift's release to `alpha 2`; confirm that it doesn't break anything in `ReactiveMoya`'s target, and no one from @Moya/contributors has an issue with it, I think it would make sense to merge this in.

wongzigii:
Circle CI timeout: pod install took more than 10 minutes 😢

ashfurrow:
@wongzigii That timeout happens sometimes, I've restarted the build 👍

ashfurrow:
Just realized we forgot to merge this, I'll deal with conflicts and merge today, thanks again 👍

ashfurrow:

PR #685, opened 2016-09-26
A better typed urlRequest property of an Endpoint?

KelvinJin:
Currently the `urlRequest` [property](https://github.com/Moya/Moya/blob/master/Source/Endpoint.swift#L92) of a `Endpoint` is a IUO (Implicitly Unwrapped Optional). But this doesn't seems to be correct. Let's say if we think the `urlRequest` definitely exists (not true) then we should delete the `excalmatory mark` to make it non-optional at all. On the other hand, if we think the `urlRequest` might not exist (sometimes when the url is invalid or the encoding fails), then we should explicitly mark it as optional.

Although we could assume the user should be responsible for the risks, I like the latter assumption better. This PR made the `urlRequest` optional. And because the `hashValue` relies on the property, I have to change that as well. ATM, I'm using the URL as a backup if the request doesn't exist. (Might need to consider the parameters/parameterEncoding as well. method/headers won't effect.)

We can discuss the problem here or in a separate issue.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,7 @@
 
 - **Breaking Change** Throw dedicated `Error.jsonMapping` when `mapJSON` fails to parse JSON
 - Removed the unused `StreamRequest` typealias that was causing watchOS failures.
+- `urlRequest` property of `Endpoint` is now truly optional. The request will fail if the `urlRequest` turns out to be nil and a `requestMapping` error will be returned together with the problematic url.
 
 # 8.0.0-beta.1
 
--- a/Demo/Podfile.lock
+++ b/Demo/Podfile.lock
@@ -1,16 +1,16 @@
 PODS:
   - Alamofire (4.0.0)
-  - Moya (8.0.0-beta.1):
-    - Moya/Core (= 8.0.0-beta.1)
-  - Moya/Core (8.0.0-beta.1):
+  - Moya (8.0.0-beta.2):
+    - Moya/Core (= 8.0.0-beta.2)
+  - Moya/Core (8.0.0-beta.2):
     - Alamofire (~> 4.0.0)
     - Result
-  - Moya/ReactiveCocoa (8.0.0-beta.1):
+  - Moya/ReactiveCocoa (8.0.0-beta.2):
     - Moya/Core
     - ReactiveSwift
-  - Moya/RxSwift (8.0.0-beta.1):
+  - Moya/RxSwift (8.0.0-beta.2):
     - Moya/Core
-    - RxSwift (= 3.0.0-beta.1)
+    - RxSwift
   - Nimble (5.0.0-alpha.30p1)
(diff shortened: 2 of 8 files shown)


ashfurrow:
This is a great idea and a swell PR. We (I?) added the IUO because it was easy, but this is definitely the better approach. Thanks!

I've restarted CI, it failed due to a version mismatch in the Podfile.lock. Do the tests pass locally?

ashfurrow:
CI is running on https://circleci.com/gh/Moya/Moya/608 , I just had to rerun without the cache.

nanoxd:
We should add tests to be sure there aren't any regressions as well as document the change and a migration plan for existing users (if any).

ashfurrow:
Agreed, a changelog entry should be sufficient though – this isn't any more intrusive than the other breaking changes we've got on master.

nanoxd:
I should read more closely at filenames 😅.

KelvinJin:
Updated the error type to include the corresponding url(path) and updated the change log as well. I would want to test the error case but have no idea where to start. Ideally I should create a new `TargetType` with some faulty url or by sending some faulty parameters. (by using some wired characters?) Right now we only have `Github` available.

ashfurrow:
This is great, I'll take care of the test and merge :+1:

ashfurrow:
In this case, I did the following to resolve the conflicts:

```sh
git remote add KelvinJin https://github.com/KelvinJin/Moya.git
git fetch KelvinJin
git checkout KelvinJin/master
git checkout -b merge-685
git pull origin/master
```

git resolved the changelog conflicts, I just had to resolve a conflict with a unit test from #688. I did need to run `bundle exec pod update ReactiveSwift`, since CocoaPods was giving an error. Then I `git add . ; git commit` and used the default merge message. 

Let me know if anyone has any questions about this workflow, it might be worth writing up in a blog post sometime.

KelvinJin:
@ashfurrow 
I just realise there's a struct called `ObjectIdentifier` which can help us get the `hashValue` for `AnyObject`. And this brings up another question, since the `hashValue` would be the same even when the `URLRequest` has different `httpMethod` as shown below.

```Swift
let path = "www.google.com"
let url = URL(string: path)!
var request1 = URLRequest(url: url)
request1.httpMethod = "GET"
let request2 = URLRequest(url: url)
request1.httpMethod = "POST"
request1.hashValue // -506241750264670446
request2.hashValue // -506241750264670446
```
 And the current implementation in the `requestNormal` function seems to treat the two requests the same as long as the have the same `hashValue` (as the key of the dictionary). Would this be a problem? 

I only had some experience developing web apps and I remembered you can use the same url endpoint with different methods to do different things such as `GET` & `DELETE`.... I could be wrong.

ashfurrow:
No problem, two different objects can have the same hash value without any major consequences (as long as two equal objects don't have different hash values, it should be fine).

Issue #684, opened 2016-09-26
Swift 3: lowercase HTTP verb enums?


BasThomas:
Should we lowercase these, too?

https://github.com/Moya/Moya/blob/366a3142825957169a2807269382bb868d54f00c/Source/Target.swift#L52

spxrogers:
AF did 🙃 
https://github.com/Alamofire/Alamofire/blob/e8bca7e40a861327366c5ccba42a688d17ec482a/Source/ParameterEncoding.swift#L30-L40

sunshinejr:
👍

ashfurrow:
Makes sense, does anyone have time to submit a PR?

nanoxd:
✋ I can do it if no one else has time today

Issue #556, opened 2016-07-19
Any possibility for network engine abstraction


jcampbell05:
In our app we use Firebase, we would love to abstract this to the point where we can use alamofire or firebase without the caller knowing anything.

As it is we will have to either have an abstraction of an abstraction or have the code know it's calling Moya or non-moya code.

ashfurrow:
Hmm, this is an interesting question. There's no reason we can't/shouldn't do this, it's just never been a priority before, so it's definitely worth considering. I'm thinking we abstract the functions we call from Alamofire into protocols, and extend Alamofire types to conform to those protocols. Then we just need a way to specify a different set of types that also conform.

There's a bit of hassle where we're currently using `typealias` to create Moya-specific types from Alamofire ones (for things like request methods, GET, POST, etc) that would need a little bit more code, but nothing too dramatic. 

I've got a flight tomorrow, might take a stab at this on the plane. In the meantime, you could explore subclassing Moya's `Provider` class to give you similar behaviour. Please let us know if you have ideas for implementation, too.

jasl:
[My fork](https://github.com/jasl/MoyaX) has done this.

ashfurrow:
I forgot about that, thanks!

ashfurrow:
I took an initial stab at this, it's quite an overwhelming change. We actually lean on Alamofire a lot, include use of `typealias` to create Moya types that are really just Alamofire types. If we do this, I would want someone to be able to theoretically use Moya without having any reference to Alamofire code at all.

@jasl how up-to-date is your fork? Moya has added multipart form uploading and a few other large structural changes recently, would it work with those changes?

jasl:
@ashfurrow I sync all commits to <https://github.com/jasl/Moya> and it's a fork of Moya.

I'm afraid my fork has large different with current Moya, because I've done myself version multipart-upload, new Swift 3 naming convention, refactor example, non-generic provider etc  before

ashfurrow:
Okay, thanks for the quick response!

I tried to tackle this from top-down, like replacing our references to `Alamofire.Manager` with a new protocol defined in Moya, but _everything_ broke immediately, and I was no where near finished an hour later. I think a better approach would be to remove the `typealias`s in `Moya+Alamofire.swift` first, then replace use of `Manager` with a new protocol. I'll see what I can do later.

jasl:
[...]

PR #689, opened 2016-09-26
Changes HTTP verbs enum to lowercase

pedrovereza:
As discussed in #684

```swift
public enum Method: String {

    case options = "OPTIONS"
    case get     = "GET"
    case head    = "HEAD"
    case post    = "POST"
    case put     = "PUT"
    case patch   = "PATCH"
    case delete  = "DELETE"
    case trace   = "TRACE"
    case connect = "CONNECT"
```

### DIFF
--- a/Source/Target.swift
+++ b/Source/Target.swift
@@ -49,14 +49,23 @@ public enum StructTarget: TargetType {
 
 /// Represents an HTTP method.
 public enum Method: String {
-    case GET, POST, PUT, DELETE, OPTIONS, HEAD, PATCH, TRACE, CONNECT
+
+    case options = "OPTIONS"
+    case get     = "GET"
+    case head    = "HEAD"
+    case post    = "POST"
+    case put     = "PUT"
+    case patch   = "PATCH"
+    case delete  = "DELETE"
+    case trace   = "TRACE"
+    case connect = "CONNECT"
 
     public var supportsMultipart: Bool {
         switch self {
-        case .POST,
-             .PUT,
-             .PATCH,
-             .CONNECT:
+        case .post,
+             .put,
+             .patch,
+             .connect:
             return true
         default:
             return false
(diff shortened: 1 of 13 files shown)


BasThomas:
Was just busy doing this, thank god I had a GitHub window open 🙈

pedrovereza:
@BasThomas Oh, my bad! Totally forgot to mention I was working on it 😅

Source/Target.swift
```diff
@@ -49,14 +49,23 @@ public enum StructTarget: TargetType {
 
 /// Represents an HTTP method.
 public enum Method: String {
-    case GET, POST, PUT, DELETE, OPTIONS, HEAD, PATCH, TRACE, CONNECT
+
+    case options = "OPTIONS"
```
BasThomas: Any reason for swapping this order here? Not that it matters that much, just curious

BasThomas:
No worries!

Source/Target.swift
pedrovereza: No reason, was using [Alamofire as a reference](https://github.com/Alamofire/Alamofire/blob/e8bca7e40a861327366c5ccba42a688d17ec482a/Source/ParameterEncoding.swift#L30-L40)

Should I reorder it back to how it was?

sunshinejr: Hmm, didn't really check the code so forgive me, but is there a reason we cannot use typealias for existing enum from Alamofire?

pedrovereza: Tried this and seems to work fine.

```swift
public typealias Method = Alamofire.HTTPMethod

extension Method {
    public var supportsMultipart: Bool {
        switch self {
        case .post,
             .put,
             .patch,
             .connect:
            return true
        default:
            return false
        }
    }
}
``` 

Only downside I can see here is that it's yet another dependency on Alamofire. But given the discussion in #556, not sure if that's an actual issue.

BasThomas: Yeah, we might want to just use a typealias here.

pedrovereza: Done 😉

PR #694, opened 2016-09-29
Added Important Note

fear731:
When I initially tried to initialize all parameters within one line, I found out that the parameters won't be removed from the request if their value is ```nil```, nor unwrapped if they contain a value.
Added a hint to the example to avoid that wrong implementation for others.

### DIFF
--- a/docs/Examples/OptionalParameters.md
+++ b/docs/Examples/OptionalParameters.md
@@ -53,7 +53,7 @@ You **have to** add optional parameters like shown above, one per line. Optional
 	public var parameters: [String: Any]? {
 	    switch self {
 	    case .users(let limit):
-	        let params: [String: Any] = ["limit" = limit]
+	        let params: [String: Any] = ["limit": limit]
 	        return params
         default:
             return nil
--- a/docs/Examples/OptionalParameters.md
+++ b/docs/Examples/OptionalParameters.md
@@ -50,6 +50,7 @@ You **have to** add optional parameters like shown above, one per line. Optional
 
 ```swift
 //...
+	// This won't work!
 	public var parameters: [String: Any]? {
 	    switch self {
 	    case .users(let limit):
--- a/docs/Examples/OptionalParameters.md
+++ b/docs/Examples/OptionalParameters.md
@@ -13,13 +13,13 @@ extension MyService: TargetType {
 	public var parameters: [String: Any]? {
 	    switch self {
 	    case .users(let limit):
-	        var params: [String: Any] = [:]
-	        params["limit"] = limit
-	        return params
-        default:
-            return nil
-        }
-    }
+		var params: [String: Any] = [:]
+		params["limit"] = limit
+		return params
+	    default:
+	        return nil
+	}
+}
 //...
 }
 ```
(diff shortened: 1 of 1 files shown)


docs/Examples/OptionalParameters.md
```diff
@@ -42,3 +42,24 @@ extension MyService: TargetType {
...
+Important Note
+--------------
+You **have to** add optional parameters like shown above, one per line. Optional parameters won't be removed in case of ```nil``` if you try to initialize them within one line, e.g.:
+
+```swift
+//...
+	public var parameters: [String: Any]? {
+	    switch self {
+	    case .users(let limit):
+	        let params: [String: Any] = ["limit" = limit]
```
sunshinejr: Hmm, does this line compile for you?

sunshinejr: Hmm, does this line compile for you?

sunshinejr:
Hey! Thanks for the PR! Although at the first glance I don't get the error here, so I've made a comment about the line that is confusing me the most.

docs/Examples/OptionalParameters.md
fear731: Of course not, sorry. Fixed the typo.

sunshinejr: Oh, right, no probs, confused me a bit huh

docs/Examples/OptionalParameters.md
```diff
@@ -42,3 +42,24 @@ extension MyService: TargetType {
...
+You **have to** add optional parameters like shown above, one per line. Optional parameters won't be removed in case of ```nil``` if you try to initialize them within one line, e.g.:
+
+```swift
+//...
+	public var parameters: [String: Any]? {
+	    switch self {
+	    case .users(let limit):
+	        let params: [String: Any] = ["limit": limit]
+	        return params
+        default:
```
sunshinejr: Could you align vertically with `case .users`?

sunshinejr:
Well, I didn't know about this caveat, not sure if this is intended behaviour of Swift, but glad you discovered it and shared with us! Only one more comment from me and we are ready to merge. Thanks again!

docs/Examples/OptionalParameters.md
fear731: There you go :)

docs/Examples/OptionalParameters.md
```diff
@@ -42,3 +42,24 @@ extension MyService: TargetType {
...
+Important Note
+--------------
+You **have to** add optional parameters like shown above, one per line. Optional parameters won't be removed in case of ```nil``` if you try to initialize them within one line, e.g.:
+
+```swift
+//...
+	public var parameters: [String: Any]? {
+	    switch self {
+	    case .users(let limit):
+	        let params: [String: Any] = ["limit": limit]
```
ashfurrow: Since this code is a demonstration of how _not_ to do things, adding a comment like `// This won't work!` above this line might be a good idea. Thoughts?

fear731: Done 👍

ashfurrow:

PR #704, opened 2016-10-05
Issue template

BasThomas:
I made a first draft for an issue template. What do you think? Let me know!

I have kept it as easy and bare as possible, as to not 'scare away' any people opening an issue.


orta:
Yep - great work @BasThomas - this is a good start, and if people want to build off it, send more PRs 👍

BasThomas:
Sounds good!

PR #705, opened 2016-10-05
[Alamofire] Use .default instead of inits in Parameter Encoding.

sunshinejr:
Hey guys! 🎉 So I've changed the way we use Alamofire's `ParameterEncoding` in Moya, because it seems like we should use the `.default` parameter of given type, rather than `init` (per docs [here](https://github.com/Alamofire/Alamofire/blob/master/README.md#parameter-encoding), also from implementation these are aliases, but who knows what brings the future!). I hope I didn't overlook anything and we are good to go! 😅


sunshinejr:
I believe CI fail is not connected with the changes so gonna merge it 👍

Issue #706, opened 2016-10-05
Moya version for Swift X.X


sunshinejr:
Hey guys! So I was wondering if we should make a note somewhere in the documentation / README, that specific version of Moya support X version of Swift. I wasn't sure if there is need for something like this so I am asking you: what do you think about it? 👊

BasThomas:
I think that's a good idea - I don't see it doing any harm, anyway 😃

sunshinejr:
Great! Closing this one in favor of #710 which should implement the idea.

PR #710, opened 2016-10-08
[README] Add Moya version vs Swift version table

sunshinejr:
Hey again! 🎉 This is a product of #706, at first I wanted to make another page for it, but it seems like a small table in Installation section should be sufficient. But if you have more ideas/concerns around it, please share with me! 👊


BasThomas:
Reworded and fixed some typos; LGTM otherwise! Thanks for adding this :)

Issue #668, opened 2016-09-22
Increase code coverage


ashfurrow:
Our code coverage is okay, at a solid 80%. We've held this for a while, but we could do a bit better. Our [current coverage](https://codecov.io/gh/Moya/Moya/tree/master/Source) has some places we could improve:

<img width="1154" alt="screen shot 2016-09-21 at 9 37 47 pm" src="https://cloud.githubusercontent.com/assets/498212/18734570/d21c2478-8043-11e6-8c54-86b7d0d3c683.png">

I'll take a look sometime, this is probably a longer-term task though.

Viniciuscarvalho:
@ashfurrow this task requires expanding the use of different classes? Or use more RxSwift and others?

ashfurrow:
Both actually, the graph above is just an outline, you can see individual files' coverage in the link.

Viniciuscarvalho:
@ashfurrow There are some files that are not being found for example Moya + Internal, is giving 404 error.

ashfurrow:
Hmm. Sounds like it could be a configuration issue, do you have experience with CodeCov? Could be worth a look.

ashfurrow:
Looks like CodeCov can't access file names with + in them, like https://codecov.io/gh/Moya/Moya/src/master/Source/Moya%2BInternal.swift . I've contacted their support team to clear it up.

ashfurrow:
We're in the high 80's, CodeCov seems to be working. Gonna close this, thanks everyone 👍

PR #716, opened 2016-10-14
Fix download progress bug

### LINKED ISSUE #715: Request from targets of task type download never call completion block
Requests of the task type download never call the completion block when executed. This behaviour can be seen in the demo project. Tapping the download button will show a progress bar but the bar will never disappear when the request finishes. The expected result would be that the progress bar turns green on success and disappears when the request has finished.

The reason why this happens is that the `DownloadRequest` type does not inherit from the `DataRequest` type. In the function `sendAlamofireRequest` in `Moya+Internal.swift`, [the current request is cast to `DataRequest` which then handles the callbacks](https://github.com/Moya/Moya/blob/master/Source/Moya%2BInternal.swift#L220-L231). Since `DownloadRequest` does not inherit from `DataRequest` the cast always fails for those the download type and the proper callback logic never gets called.

petester42:
Closes #715

As detailed in #715, the callbacks of the download requests were never being called because the `DownloadRequest` does not inherit from the `DataRequest`.

What was changed was that there is now an different cast for download requests. This is based on the [code that is found in the Alamofire example] (https://github.com/Alamofire/Alamofire/blob/master/Example/Source/DetailViewController.swift#L104-L112). The second casts fixes the completion block never getting called issue but the progress is still wrong.

I noticed when trying to write tests for this that the progress object will never have its completed property be true since it requires a response to be present. In all the code for progress tracking a response is never set so what I did is that on request completion update the progress object to have a response. Doing so has the desired effect of having the progress object returning completed.

This also contributes to increasing the code coverage from #668.

### DIFF
--- a/Source/Moya+Internal.swift
+++ b/Source/Moya+Internal.swift
@@ -216,32 +216,23 @@ private extension MoyaProvider {
             default: break
             }
         }
+        
+        let completionHandler: RequestableCompletion = { response, request, data, error in
+            let result = convertResponseToResult(response, request: request, data: data, error: error)
+            // Inform all plugins about the response
+            plugins.forEach { $0.didReceiveResponse(result, target: target) }
+            progressCompletion?(ProgressResponse(response: result.value))
+            completion(result)
+        }
+        
+        if var dataRequest = progressAlamoRequest as? Requestable {
+            dataRequest = dataRequest.response(queue: queue, completionHandler: completionHandler)
 
-        if var dataRequest = progressAlamoRequest as? DataRequest {
-            dataRequest = dataRequest.response(queue: queue, completionHandler: { handler in
-                let result = convertResponseToResult(handler.response, request: handler.request, data: handler.data, error: handler.error)
-                // Inform all plugins about the response
-                plugins.forEach { $0.didReceiveResponse(result, target: target) }
-                completion(result)
-            })
-
-            if let dataRequest = dataRequest as? T {
-                progressAlamoRequest = dataRequest
-            }
-        } else if var dataRequest = progressAlamoRequest as? DownloadRequest {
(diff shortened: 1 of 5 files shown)


Source/Moya+Internal.swift
```diff
@@ -228,6 +228,18 @@ private extension MoyaProvider {
...
+                let result = convertResponseToResult(handler.response, request: handler.request, data: nil, error: handler.error)
+                // Inform all plugins about the response
+                plugins.forEach { $0.didReceiveResponse(result, target: target) }
+                progressCompletion?(ProgressResponse(response: result.value))
+                completion(result)
+            })
+            
+            if let dataRequest = dataRequest as? T {
+                progressAlamoRequest = dataRequest
+            }
```
ashfurrow: Looks like this `else if` block looks very similar to the one above it, it's just the `progressCompletion` block that's called here and not above. Do you think there's a better abstraction we could use to avoid the repetition?

ashfurrow:
Great PR, thanks! The only feedback I have is in the comment above. If you like, we can address it now, or we can open an issue to address it later.

Source/Moya+Internal.swift
ashfurrow: For example, we could create a protocol with the required methods from both `DataRequest` and `Download Request`, and then extend those types to conform to that protocol. Then our conditional `if let` could cast to the protocol. What do you think?

petester42: I think that would be a good idea. I'll look into it.

Ernest-su:
still not call .completed in RxMoyaProvider.

BasThomas:
@Ernest-su, what do you mean? Is this a possible bug? If you create an issue for that, please do!

petester42:
That's possible. I don't think that scenario has been tested.

PR #717, opened 2016-10-15
Adding ability to fully mock an NSHTTPURLResponse

### LINKED ISSUE #701: Allow stubbing of NSURLResponse within sampleResponseClosure
We are using Moya to build a caching layer on top of a restful service.  The way we uniquely identify requests in our system is via their URL, which we store into our data store (using Realm).   Specifically, we are using the `response?.URL?.absoluteString` property of `Response`.  We would like to test this functionality, but the only way to stub a response seems to be with a `EndpointSampleResponse`, which always has a `response` of `nil`.  

Is there any way we can open the response property up for stubbing?  Something like
```swift
/// Used for stubbing responses.
public enum EndpointSampleResponse {

    /// Note the additional NSURLResponse field
    case NetworkResponse(Int, NSData, NSURLResponse?)

    /// The network failed to send the request, or failed to retrieve a response (eg a timeout).
    case NetworkError(NSError)
}

```

gregkerzhner:
Fixes #701.


ashfurrow:

PR #719, opened 2016-10-16
Adds tests for #717 and cleans up changelog

ashfurrow:
Related to #668.

PR #721, opened 2016-10-16
Lowercase class funcs

BasThomas:
As discussed [here](https://github.com/Moya/Moya/pull/719#discussion_r83555707), class funcs mimicking enum cases have now been lowercased.

Please check if I missed any?

I also cleaned up the changelog a bit.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,10 +1,8 @@
 # Next
 
-- **Breaking Change** Throw dedicated `Error.jsonMapping` when `mapJSON` fails to parse JSON.
 # 8.0.0-beta.3
 
-- **Breaking Change** Throw dedicated `Error.jsonMapping` when `mapJSON` fails to parse JSON
-- Removed the unused `StreamRequest` typealias that was causing watchOS failures.
+- **Breaking Change** Throw dedicated `Error.jsonMapping` when `mapJSON` fails to parse JSON.
 - **Breaking Change** Renamed `endpointByAddingHTTPHeaders` to `adding(newHttpHeaderFields:)`.
 - **Breaking Change** Renamed `endpointByAddingParameters` to `adding(newParameters:)`.
 - **Breaking Change** Renamed `endpointByAddingParameterEncoding` to `adding(newParameterEncoding:)`.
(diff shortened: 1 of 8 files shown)


BasThomas:
We could also consider changing `.neverStub` and friends to exclude the `Stub` part, making it `.never`.

ashfurrow:
Nah, I like `neverStub` since it's not a member of a `Stub` type, but rather a provider.

BasThomas:
That's what I was wondering. In context `.never` makes sense, but out of context it doesn't.

ashfurrow:
We could make a new `Stub` type within the provider, might be overkill. What do you think?

Demo/Tests/ReactiveCocoaMoyaProviderTests.swift
```diff
@@ -60,10 +60,10 @@ class ReactiveCocoaMoyaProviderSpec: QuickSpec {
...
 
             class TestProvider<Target: TargetType>: ReactiveCocoaMoyaProvider<Target> {
-                init(endpointClosure: @escaping EndpointClosure = MoyaProvider.DefaultEndpointMapping,
-                    requestClosure: @escaping RequestClosure = MoyaProvider.DefaultRequestMapping,
-                    stubClosure: @escaping StubClosure = MoyaProvider.NeverStub,
-                    manager: Manager = MoyaProvider<Target>.DefaultAlamofireManager(),
+                init(endpointClosure: @escaping EndpointClosure = MoyaProvider.defaultEndpointMapping,
+                    requestClosure: @escaping RequestClosure = MoyaProvider.defaultRequestMapping,
+                    stubClosure: @escaping StubClosure = MoyaProvider.neverStub,
+                    manager: Manager = MoyaProvider<Target>.defaultAlamofireManager(),
```
sunshinejr: Hmm, should we also consider making `defaultAlamofireManager` a variable? Looking by the init this is the only place we use brackets.

sunshinejr: On the other hand this is another breaking change...

ashfurrow: For context, this was originally written as a function, since it returns a new manager instance every time (rather than an identical closure). We can totally revisit that though.

sunshinejr: Oh, right! Just was thinking out loud. Let's just leave it as it is for now 👍

BasThomas:
Yeah, I think that's overkill. Let's just leave it like this for now.

Changelog.md
```diff
@@ -2,19 +2,20 @@
...
+- **Breaking Change** Throw dedicated `Error.jsonMapping` when `mapJSON` fails to parse JSON.
+- **Breaking Change** Renamed `endpointByAddingHTTPHeaders` to `adding(newHttpHeaderFields:)`.
+- **Breaking Change** Renamed `endpointByAddingParameters` to `adding(newParameters:)`.
+- **Breaking Change** Renamed `endpointByAddingParameterEncoding` to `adding(newParameterEncoding:)`.
+- **Breaking Change** Renamed `endpointByAdding(parameters:httpHeaderFields:parameterEncoding)` to `adding(parameters:httpHeaderFields:parameterEncoding)`.
+- **Breaking Change** Changed HTTP verbs enum to lowercase.
 - `urlRequest` property of `Endpoint` is now truly optional. The request will fail if the `urlRequest` turns out to be nil and a `requestMapping` error will be returned together with the problematic url.
 - **Breaking Change** Made RxMoya & ReactiveMoya frameworks dependant on Moya framework, making them slimmer and not re-including Moya source in the Reactive extensions. ([PR](https://github.com/Moya/Moya/pull/563))
+- **Breaking Change** Made some `class func`s [mimicking enum cases](https://github.com/Moya/Moya/blob/master/Source/Moya.swift#L117-L133) lowercased.
```
ashfurrow: This will need to go above the `8.0.0-beta.3` section.

Issue #720, opened 2016-10-16
Moya 8.0.0 final


marcelofabri:
I haven't found it anywhere and it seems that the Moya 8 milestone is empty: https://github.com/Moya/Moya/issues?q=is%3Aopen+is%3Aissue+milestone%3A%22Moya+8+%2F+Swift+3%22, so what is blocking Moya 8.0.0 to be released? 

My guess would be `ReactiveCocoa` and `RxSwift`. If so, couldn't we just pin their versions to their latest betas for now and once they're final we can make a new release. Thoughts?

Another (much more difficult) option would be making these subspecs projects themselves so that they can have their own release cycle. Of course that would break everyone's Podfiles 😅

ashfurrow:
Hmm, not sure where that milestone came from actually. I agree it's time to consider releasing Moya 8.

We're still ironing out regular bugs and other breaking changes for Swift 3, personally I'm fine with keeping it in beta for now. Especially since if we go final with beta versions of RxSwift/RAC, and then there's another RxSwift/RAC beta with breaking changes, then we need to have a new major version to avoid breaking semantic versioning.

sunshinejr:
I also think that we should stay in beta as long as our dependencies are. No need to rush and ReactiveSwift isn't even out of alpha. Also, I've removed the milestone to avoid confusion.

marcelofabri:
My main issue with staying in beta is that the main reason seems to the optional subspecs dependencies and that only Moya 8 is compatible with Swift 3 - which is out for a while. I don't like the idea of "being stuck" because of RxSwift/RAC as none of those are dependencies of my project.

I know it's unfair, but just tried to expose my concern as others might have the same feeling.

I opened this issue to check if they were any other blockers and how I could help 😊 

Also, since we're here, it seems that there're a lot of changes since beta 2. Could we have a new beta? 😅

ashfurrow:
It's a fair point, thanks for bringing it up. I think a new beta makes at least makes sense, I'll try to get it done today.

ashfurrow:
Cool, beta 3 is up and pushed for Carthage and CocoaPods 👍

ashfurrow:
RxSwift's updated to 3.0 final, so we're just waiting on Reactiveswift now. I'll open an issue asking for a timeline.

marcelofabri:
Thanks for the update, Ash! 💯

BasThomas:
Closing in favor of #758

PR #737, opened 2016-10-23
Dependency updates

ashfurrow:
Updates dependencies to their latest versions, including a final release of RxSwift. See #720 for more info.

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,3 +1,3 @@
 github "Alamofire/Alamofire" ~> 4.0.0
-github "ReactiveCocoa/ReactiveSwift" "1.0.0-alpha.2"
-github "ReactiveX/RxSwift" "3.0.0-rc.1"
+github "ReactiveCocoa/ReactiveSwift" "1.0.0-alpha.3"
+github "ReactiveX/RxSwift" "3.0.0"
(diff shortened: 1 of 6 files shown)


Cartfile
```diff
@@ -1,3 +1,3 @@
 github "Alamofire/Alamofire" ~> 4.0.0
-github "ReactiveCocoa/ReactiveSwift" "1.0.0-alpha.2"
-github "ReactiveX/RxSwift" "3.0.0-rc.1"
+github "ReactiveCocoa/ReactiveSwift" "1.0.0-alpha.3"
+github "ReactiveX/RxSwift" "3.0.0"
```
AndrewSB: I think this should either be `~> 3.0.0` or `== 3.0.0`
https://github.com/Carthage/Carthage/blob/master/Documentation/Artifacts.md#version-requirement

AndrewSB: It works as is now, but `""` tell carthage to use `git rev-parse`, which the dependency resolver often has issues with (à la https://github.com/Carthage/Carthage/issues/733)

AndrewSB:
CI errd out for some strange reason, attempting a rebuild

pedrovereza:
Attempting another rebuild now that #745 seems to have saved `cocoapods/repo/master`

pedrovereza:
And it got past `pod install`. I'm so happy :tada:

PR #745, opened 2016-10-27
[CI] Caches pods repo

pedrovereza:
We've been facing a lot of time-out issues when running `bundle exec pod install` in Circle CI.
This is an attempt to fix it (or at least make it happen less often).

Suggested in https://discuss.circleci.com/t/cocoapods-timing-out/7494


ashfurrow:
Thanks for taking care of this!

PR #751, opened 2016-11-02
Use String(data:encoding:) instead of NSString(data:encoding:)

devxoul:


devxoul:
@ashfurrow Sure! Just updated it :)

PR #758, opened 2016-11-07
Add required TargetType protocol method to Basic.md

austinlouden:
`public var task: Moya.Task { get }` is now required, this adds it to the setup documentation.

### DIFF
--- a/docs/Examples/Basic.md
+++ b/docs/Examples/Basic.md
@@ -66,7 +66,10 @@ extension MyService: TargetType {
         }
     }
     var task: Task {
-        return .request
+        switch self {
+            case .zen, .showUser, .createUser, .showAccounts:
+                return .request
+        }
     }
     var multipartBody: [MultipartFormData]? {
         // Optional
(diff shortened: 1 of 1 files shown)


austinlouden:
@AndrewSB Your example seems more clear to me, I can update. I think adding other Task types may be too complex for `Basic.md`, but I don't have a strong preference either way.

AndrewSB:
I'll approve after that update 😄 
I think you're right about adding other types of tasks making `Basic.md` too complicated

BasThomas:
Not sure if this needs a comment explaining what this is needed for?

AndrewSB:
@BasThomas is that because `Task` feels less cut and dry than something like `Method` or `Path`?

I feel as if a user can just deep link into the `Task` type and see the documentation for the type there - if you still think docs in the basic usage make sense, go ahead and add that comment 👌.

I can review and then merge it in after you've made that edit 🙃

BasThomas:
I think you're right - the documentation on the type should be enough.

ashfurrow:

PR #761, opened 2016-11-09
Updates readme image design

ashfurrow:
This is the readme image, the first point at which people interact with Moya, so it's gotta be awesome. This PR centres the text, aligns the coloured boxes, and gives some whitespace around the images.


pedrovereza:
Nice! 💅

Issue #766, opened 2016-11-13
Moya doesn't take into account the requestRetrier configuration of a session manager


jeryRazakarison:
In the case of the requestRetrier, that is because the requestRetrier delegate method is **only** called when there is an **Alamofire error** in the response. (you can see it here https://github.com/Alamofire/Alamofire/blob/master/Source/SessionDelegate.swift#L434)
When you directly use Alamofire, this error would appear if you call `validate()` on a DataRequest.

BUT when using Moya, you can't call `validate()` on a DataRequest (since it is hidden), and thus Alamofire does not consider there is an error, and the requestRetrier `should` process is not launched.

You can test that your requestRetrier would still work with Moya by calling validate() on the DataRequest here:
https://github.com/Moya/Moya/blob/master/Source/Moya%2BInternal.swift#L187
would be
`let alamoRequest = manager.request(request as URLRequestConvertible).validate()`
Using this **hot fix**, your requestAdapter and requestRetrier work as before (as when you were using Alamofire directly).

But the real improvement would be for Moya to allow us in the switch/case configuration (in the `TargetType` protocol) to pass true or false, if we want or not the Alamofire `validate()` appended.

BasThomas:
Did you mean to make a separate issue for this? (#700)

jeryRazakarison:
I thought it was a good idea, because the first author of issue #700 spoke about cache policy problems specifically.
The `requestRetrier` issue was introduced later by _DanteCore_, but his question and my answer (and PR) don't help the initial author.

BasThomas:
Cool, makes sense!

ashfurrow:
Fixed by #767.

Issue #700, opened 2016-09-30
Question: Moya not respecting SessionManager configuration for requests?


pschneider:
Hi all,

I'm currently trying to implement a cache solution with a custom URLCache (and modify some cache methods like in this example (http://stackoverflow.com/a/19847811), Moya/Alamofire and therefor I'm using a custom SessionManager with my provider which configures the URLCache to use.

Especially this part is interesting:

``` swift
// ...
URLCache.shared = myCustomURLCache
sessionConfig.urlCache = URLCache.shared
sessionConfig.requestCachePolicy = .returnCacheDataElseLoad
// ...
```

I pass the final manager to my `MoyaProvider`, but it seems, that the send request does not have the correct `RequestCachePolicy` set.

Now if I put some output in 
`func sendRequest(_ target: Target, request: URLRequest, queue: DispatchQueue?, progress: Moya.ProgressBlock?, completion: @escaping Moya.Completion) -> CancellableToken {`

I'm seeing this:

``` swift
print(manager.session.configuration.requestCachePolicy.rawValue) // correctly set
print(request.cachePolicy.rawValue) // wrong (always `useProtocol...`)
let alamoRequest = manager.request(request as URLRequestConvertible)
print(alamoRequest.request?.cachePolicy.rawValue) // wrong
```

I don't know if this is intended or maybe a bug in Alamofire as Moya is using the correctly configured `manager` to build the `alamoRequest`.

Is this behavior intended in any way?

DanteCore:
Also have this problem. I tried to set custom SessionManager with RequestRetrier that was introduced in Alamofire 4, but implementation of RequestRetrier protocol not invokes

jeryRazakarison:
In the case of the requestRetrier, that is because the requestRetrier delegate method is **only** called when there is an **Alamofire error** in the response. (you can see it here https://github.com/Alamofire/Alamofire/blob/master/Source/SessionDelegate.swift#L434)
When you directly use Alamofire, this error would appear if you call `validate()` on a DataRequest.

BUT when using Moya, you can't call `validate()` on a DataRequest (since it is hidden), and thus Alamofire do not consider there is an error, and the requestRetrier `should` process is not launched.

You can test that your requestRetrier would still work with Moya by calling validate() on the DataRequest here:
https://github.com/Moya/Moya/blob/master/Source/Moya%2BInternal.swift#L187
would be
`let alamoRequest = manager.request(request as URLRequestConvertible).validate()`
[...]

Issue #718, opened 2016-10-16
Which baseURL to use for a stub-only provider?


benpackard:
There are occasions where a base URL isn't available and the `sampleData` is relied on for all requests. For example:
- The app is in development and the API isn't available yet. 
- The app allows the user to switch to a 'demo mode' at runtime, where the provider relies only on `sampleData`.

What is the correct `NSURL` to provide as the `baseURL` in such instances? A simple `NSURL()` will result in fatal error `url is not valid` when the url is constructed by Moya:

``` swift
// Moya.swift
guard let stringURL = target.baseURL.URLByAppendingPathComponent(target.path)?.absoluteString else {
    fatalError("\(target) url is not valid")
}
```

The URL won't be accessed so should I try just sticking `google.com` in there? Even then, I still need to account for the fact that `NSURL(string: String)` returns an optional and `baseURL` does not, so if I either have to force unwrap, or fall back to `NSURL()` eventually:

``` swift
var baseURL: NSURL {
    return appDefaults.selectedShow.apiURL ?? NSURL(string: "http://google.com") ?? NSURL()
}
```

This all feels like I'm on the wrong track - is there something I'm missing? What is the best practice here?

I'm using Moya 7.0.2 on Swift 2.3. Thanks.

ashfurrow:
What a great question! Usually for example URLs that aren't meant to ship, I use example.com – this sounds like a good alternative?

benpackard:
Thanks Ash - still feels slightly fishy to have to provide a URL that should never be used, plus a fallback (`NSURL()`) that returns a known fatal error. What would the ramifications be of making `baseURL` optional, I wonder? Likely a can of worms, I guess?

ashfurrow:
It would be a can of worms, unfortunately. The trouble is that somewhere – either inside Moya or outside – someone needs to deal with the fact that a call to `NSURL(string: )` might return `nil`. We've decided it's better to give the user of the library the flexibility to fail as they like, instead of having Moya fail or work unexpectedly.

It's a balance, there are arguments for both approaches. Do you think adding documentation of this rationale might help explain it to other Moya users? Where should we put that documentation?

benpackard:
I think that's the right call. In my own example, if I don't have a URL string I'm debating whether to throw my own fatal error, let Moya do so, or output a warning and use `example.com` and it makes sense that I get to determine that.

[...]

PR #767, opened 2016-11-13
Alamofire automatic validation configuration

jeryRazakarison:
Sometimes, users will want to use alamofire's automatic validation on some requests. For exemple, when they want to take advantage of the new `RequestRetrier`.
This functionality will allow to optionally configure `Alamofire` validation on some requests (you can see an example in the documentation).
This also solves the `RequestRetrier` issue mentioned in #700 and opened as a separate issue here: #766.

The `Alamofire` validation configuration has been set to `false` by default so that `Moya` users don't have to change their current configuration.
And if `Alamofire` validation configuration is not set up, the behaviour remains the same.

### DIFF
--- a/Source/Moya+Internal.swift
+++ b/Source/Moya+Internal.swift
@@ -174,17 +174,20 @@ private extension MoyaProvider {
     }
 
     func sendUploadFile(_ target: Target, request: URLRequest, queue: DispatchQueue?, file: URL, progress: ProgressBlock? = nil, completion: @escaping Completion) -> CancellableToken {
-        let alamoRequest = target.validate ? manager.upload(file, with: request).validate() : manager.upload(file, with: request)
+        let uploadRequest = manager.upload(file, with: request)
+        let alamoRequest = target.validate ? uploadRequest.validate() : uploadRequest
         return self.sendAlamofireRequest(alamoRequest, target: target, queue: queue, progress: progress, completion: completion)
     }
 
     func sendDownloadRequest(_ target: Target, request: URLRequest, queue: DispatchQueue?, destination: @escaping DownloadDestination, progress: ProgressBlock? = nil, completion: @escaping Completion) -> CancellableToken {
-        let alamoRequest = target.validate ? manager.download(request, to: destination).validate() : manager.download(request, to: destination)
+        let downloadRequest = manager.download(request, to: destination)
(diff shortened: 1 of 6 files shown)


jeryRazakarison:
Yes, since this enhancement is only needed in specific cases, maybe it should not belong to the Basic Usage documentation.

Looking at the TargetType protocol, it will be pretty self-explanatory. What do you propose? Do we put a sample code in the issues instead?

BasThomas:
We could include an example in the example app maybe? Also, we've been considering an [`Advanced.md`](https://github.com/Moya/Moya/issues/718#issuecomment-254076697) file, maybe it would fit there?

jeryRazakarison:
Yeah, you're right, I'm adding in a few hours the new option in the demo app, and in a new file Advanced.md.
Should I also add the change in the Changolog.md?

BasThomas:
Yes, that'd be awesome!

Source/Moya+Internal.swift
```diff
@@ -174,17 +174,17 @@ private extension MoyaProvider {
     }
 
     func sendUploadFile(_ target: Target, request: URLRequest, queue: DispatchQueue?, file: URL, progress: ProgressBlock? = nil, completion: @escaping Completion) -> CancellableToken {
-        let alamoRequest = manager.upload(file, with: request)
+        let alamoRequest = target.validate ? manager.upload(file, with: request).validate() : manager.upload(file, with: request)
```
BasThomas: Although I am not sure about naming, you might want to do this step in two lines. Something like

```swift
let initialRequest = manager.upload(file, with: request)
let alamoRequest = target.validate ? initialRequest.validate() : initialRequest
```

So we can prevent code duplication and possible errors due to that in the future. What do you think?

BasThomas: This also applies to the `sendDownloadRequest` and `sendRequest` functions, of course.

jeryRazakarison: Yes, this is totally right. I have made the changes.

jeryRazakarison:
Hi, can you explain me what I can do to satisfy codecov/patch and codecov/project?

ashfurrow:
I'm not going to worry about our code coverage going down by 0.08%, this looks great, merging 👍

ashfurrow:

BasThomas:
Thanks @jeryRazakarison! 😀

PR #768, opened 2016-11-14
Update SubclassingProvider.md to Swift 3

BasThomas:
Can someone review this? I am not sure if the `UserInfo` and `Router` types are purely  made for the example here, or if these are in any way connected to `RxSwift`?


ashfurrow:
I think they're specific to this example, not related to RxSwift for sure. Thanks for updating this 👍

PR #773, opened 2016-11-17
Cleanup header documentation and error messages

BasThomas:


BasThomas:
Oh no, seems like all of these have been commited incrementally... ( #773, #774, #775, #776, #777 )
Any idea how I can fix this?

ashfurrow:
No worries, can we just merge them in that order?

pedrovereza:
I'll merge this since it's the first one (and the most simple), then we see if the others need to be fixed 😉

PR #778, opened 2016-11-17
Add mapString:withKeyPath methods

bchrobot:
Convenience method for pulling a single String value from a JSON response

### DIFF
--- a/Source/Response.swift
+++ b/Source/Response.swift
@@ -83,10 +83,10 @@ public extension Response {
         guard let keyPath = keyPath else { return try mapString() }
 
         guard let jsonDictionary = try mapJSON() as? NSDictionary,
-            let object = jsonDictionary.value(forKeyPath:keyPath) as? String else {
+            let string = jsonDictionary.value(forKeyPath:keyPath) as? String else {
                 throw Error.jsonMapping(self)
         }
 
-        return object
+        return string
     }
 }
--- a/Source/Response.swift
+++ b/Source/Response.swift
@@ -77,4 +77,16 @@ public extension Response {
         }
         return string
     }
+
+    /// Maps received data at key path into a String.
+    public func mapString(atKeyPath keyPath: String?) throws -> String {
+        guard let keyPath = keyPath else { return try mapString() }
+
+        guard let jsonDictionary = try mapJSON() as? NSDictionary,
+            let object = jsonDictionary.value(forKeyPath:keyPath) as? String else {
+                throw Error.jsonMapping(self)
+        }
+
+        return object
+    }
 }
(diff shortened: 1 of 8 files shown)


Source/Response.swift
```diff
@@ -77,4 +77,16 @@ public extension Response {
         }
         return string
     }
+
+    /// Maps received data at key path into a String.
+    public func mapString(atKeyPath keyPath: String?) throws -> String {
```
sunshinejr: Hmm, here we don't have default value for `keyPath` parameter, but for RxSwift & ReactiveSwift extensions we have `nil` as default. I would rather stick to one, but I'm not sure which one is better, what do you think?

sunshinejr:
Hey there @bchrobot! Thank you so much for the PR 🎉 I've only got one comment but other than that it looks 💯

Source/Response.swift
bchrobot: That is a good point. I am inclined to drop the default parameters. There is no benefit to allowing `atKeyPath` to be left off as there is already a mapString method that takes no parameters.

bchrobot: Using a default of nil would allow the mapString methods to be refactored as one method, but I don't think that adds much value.

BasThomas: Hmm, why not? I think using one function is more clear. This would also prevent possible code duplication.

Source/Response.swift
```diff
@@ -77,4 +77,16 @@ public extension Response {
...
+        guard let keyPath = keyPath else { return try mapString() }
+
+        guard let jsonDictionary = try mapJSON() as? NSDictionary,
+            let string = jsonDictionary.value(forKeyPath:keyPath) as? String else {
+                throw Error.jsonMapping(self)
```
sunshinejr: Also I'm not sure whether it should be `.jsonMapping` error or `.stringMapping` error. The function name still is `mapString` so I'm more sold on the latter.

Source/Response.swift
sunshinejr: Normally I would stand with @BasThomas, but in this case, where the implementation is different, I think we can use the solution proposed by @bchrobot.

bchrobot: I will take another look at combining them. At first pass I wasn't happy with the nested `guard` statements and didn't immediately see a cleaner way of handling it.

BasThomas:
Thanks @bchrobot! 🙌

pedrovereza:
@ashfurrow Where is @bchrobot invitation to join us?? 😄

ashfurrow:

ashfurrow:
Weird, the webhook delivery failed. Re-delivering the webhook payload seemed to work, thanks for the bump 👍

ashfurrow:
Looks like GitHub was having some trouble yesterday:

screen shot 2016-11-22 at 11 14 47 am

Looking in my inbox, I got a notification of the error on Heroku, so if it happens again I'll know 👍

PR #784, opened 2016-11-20
[Feature] MultiTarget.

sunshinejr:
Heya guys! 👋

So a few months back we introduced the `StructsAPI`(#427), but we didn't really document that nor made a demo. Maybe it was for the better? Because we can use structs really without introducing this API at all. What we couldn't do was to avoid creating multiple providers for multiple targets. 

So I thought about this and came up with an idea. I've renamed our `StructTarget` to `MultipleTarget`, because you can use this with any type you want, that conforms to `TargetType`. Additionally, I've made a new target in `Demo` project, called `DemoMultiTarget`, that shows how you can use any of the network API that we have (`Giphy`, `GitHub` etc.) with just one provider and the `MultipleTarget`. I've made a new target because it would be hard to see through everything that is now in the `Demo` target & VC. That way (with usage of new Shared directory), we can reuse most of the code and make two clear examples of different usages. 

The last thing I did was to make an example file & link it in the Readme.

I think this could help large projects in splitting logic (or at least I'm pretty sure I will use it in my bigger projects) & make it easier to maintain. 

I think that's it! Please comment anything you find strange/bad/superbad in the PR 😈

Cheers! 🎉


sunshinejr:
Thanks Ash! 🙇

BasThomas:
Any idea why Danger didn't run here? Only saw the linter issues here: https://github.com/Moya/Moya/pull/787#issuecomment-262306591. (Fixed those in 7d8489f, FWIW)

PR #787, opened 2016-11-22
Update deployment target to iOS 8.0, macOS 10.10

### LINKED ISSUE #782: Moya can support iOS8?
Alamofie 4.1.0 has Updated the iOS Example app deployment target to iOS 8.0; Why Moya does not support iOS8？

BasThomas:
Fixes #782


pedrovereza:
Oh, should we add a note in the changelog as well?

BasThomas:
Will do, good catch! Another question... it seems like we can drop the deployment target on macOS to 10.10 as well (now 10.11). Do we want to do that as well?

ashfurrow:
That would be lovely!

BasThomas:
Hmm, I think we need to update the Cartfile and Cartfile.resolved as well? This still points to Alamofire 4.0 / 4.0.1, which does not support iOS 8.

AndrewSB:
@BasThomas taking a look now

For testing in the future, we can create a (or many) test cases that run in parallel with our CI? Maybe using Travis just to check if carthage is still working?

For the test cases, they would just be single view applications that import Moya & Alamofire, each with a different deployment target (test one can be iOS 8, test two can be iOS 10), and the only the thing travis will test is to see if the targets compile.
We could also test cocoapods in the same fashion.

Not sure if this is going overboard though- let me know what you think?

pedrovereza:
> For the test cases, they would just be single view applications that import Moya & Alamofire, each with a different deployment target (test one can be iOS 8, test two can be iOS 10), and the only the thing travis will test is to see if the targets compile.
We could also test cocoapods in the same fashion.

We use cocoapods to install pods in the Demo app before executing tests. Is it fair to say this covers  cocoapods?

I'm not familiar with carthage, but #630 added a step in the build called `rake:test carthage`, which runs `carthage bootstrap --platform 'iOS'`. Doesn't it cover carthage support? 

*Apologies if I miss understood your suggestions* 😅

BasThomas:
Hmm, strange how CI didn't fail with Carthage then. Because carthage indeed seems te be tested during CI. 🤔

AndrewSB:
You didn't misunderstand them at all @pedrovereza

The issue is that `carthage bootstrap --platform 'iOS'` would succeed, since Carthage just pulls down the dependencies and builds them, it doesn't test to see if they'd all compile under the same swift or iOS version (like how cocoapods does in it's podspec).

To test `carthage` for iOS 8, we'd need to create an target that targets iOS 8, build dependencies with carthage, `import Moya` & `import Alamofire`, and then try to compile the project.

pedrovereza:
@AndrewSB Oh, now I get it, thanks! 👍

BasThomas:
What might be a simpler fix is to add a `Danger` rule to check if the `cartfile` and `cartfile.resolved` have been updated if the `podfile` or `podfile.lock` has been updated and vice versa?

pedrovereza:
@BasThomas Yea, sounds like a good first step.

AndrewSB:
@BasThomas that sounds like a good first step to me too 🙃

PR #797, opened 2016-11-27
Update Carthage to support iOS 8 and macOS 10.10

BasThomas:


AndrewSB:
@BasThomas this looks good to me (thank you for making it 🙃g, can you run a `carthage update --no-build` to update rxswift in the resolved file, and to make sure everything works?

ufosky:
I see Alamofire latest version is 4.2.0. Why this use 4.1.0.

BasThomas:
As we use `4.1.0` via CocoaPods as well. We might be able to up this to `4.2.0`, but let's bump it to `4.1.0` for now to keep them in sync.

BasThomas:
Hmm, do we want to update both CocoaPods and Carthage to `Alamofire 4.2.0`? Now I've locked Carthage to `4.1.0`, but not really for a reason. This might be confusing in the future?

BasThomas:
Closing in favor of #799

AndrewSB:
#799 looks good to me 😄

PR #804, opened 2016-11-29
Update documentation's Endpoint param URL to url.

BasThomas:
Thanks for catching this, @rlam3!


BasThomas:
How does this increase the coverage? 🤔

Issue #648, opened 2016-09-17
Rename ReactiveCocoa subspec


ashfurrow:
We don't depend on ReactiveCocoa anymore (in Swift 3), just its new cousin: [ReactiveSwift](https://github.com/ReactiveCocoa/ReactiveSwift). It's a really cool project they're undertaking, separating ReactiveCocoa into two libraries. 

Since we now have a different dependency, I think it's worth considering renaming the `Moya/ReactiveCocoa` subspec to `Moya/ReactiveSwift`. We'd also need to rename directories, Swift extension file names and extension names, and document it as well.

orta:
This is a major breaking change, will need docs in the changelog etc on how to migrate blah blah blah

BasThomas:
Do we want to take this on for `8.0.0`? This might be the best time to do so.

ashfurrow:
Makes sense to me. I was thinking we could include it as two subspecs, and deprecate the old one in favour of the new. To avoid repeating ourselves in the podspec, we can have the old deprecated subspec simply have a dependency on the new one. Does that make sense?

BasThomas:
That does make sense, but I am not sure how to do that myself. @Moya/contributors, anyone that wants to take a stab at this?

AndrewSB:
8's going to be a big one 😮

ashfurrow:
Yup. I think we should be about good to go. Anything else we need for 8.0?

BasThomas:
Wait for an official release for ReactiveSwift is about it, I think.

BasThomas:
#805 has been merged. 🤗

ashfurrow:
Cool cool. I don't see any more 1.0 issues on their repo, and they just had a new pre-release of 1.0 ship two days ago, so hopefully it won't be too long.

BasThomas:
Let's hope so. 😀

BasThomas:
Seems like there's now one issue still open for [1.0](https://github.com/ReactiveCocoa/ReactiveSwift/milestone/1) there, but I bet it won't take too long anymore.

PR #805, opened 2016-11-29
Rename ReactiveCocoa to ReactiveSwift

BasThomas:
See #648


BasThomas:
If anyone wants / has an idea on how to take [this](https://github.com/Moya/Moya/issues/648#issuecomment-263372164), please do!

> I was thinking we could include it as two subspecs, and deprecate the old one in favour of the new. To avoid repeating ourselves in the podspec, we can have the old deprecated subspec simply have a dependency on the new one. Does that make sense?

BasThomas:
Could we simply do this? Without needing the `ss.source_files`?

```
s.subspec "ReactiveCocoa" do |ss|
    ss.dependency "Moya/ReactiveSwift"
end
```

BasThomas:
The current Podfile seems to work. Did what [I thought](https://github.com/Moya/Moya/pull/805#issuecomment-263545349) would work. Can anyone confirm? `pod lib lint` runs fine (although it takes a really long time 😮)

ashfurrow:
Yup, it takes a long time because it needs to run every permutation of every subspec. This is great, the only change I would add is the modify the Podfile used in the Demo project to reference the new subspec. Great work!

BasThomas:
Done. 👍

PR #813, opened 2016-12-01
Allow plugins to modify requests and responses.

scottrhoyt:
## Purpose

While the endpoint and request closures are useful for modifying requests, they
provide an inconvenient solution for many use cases--notably authentication--for
a number of reasons:
  * Only one closure of each type can be provided, making it less ideal when
    multiple atomic modifications need to be made to a request.
  * Overriding these closures require duplicating the work that the default
    closures do and knowledge of more internal Moya API.
  * It is sometimes ambiguous as to which closure you should choose for your
    task.

## Solution

Plugins can be extended with ability to modify requests and results. This would
bring the concept more in line with the concept of web framework middleware and
provide a convenient interface for Moya users and 3rd party plugin creators.
Ultimately, the ability to interact with requests and responses will make
Plugins much more useful.

To do this, we can add two functions to `PluginType`:
```swift
func prepareRequest(_ request: URLRequest, target: TargetType) -> URLRequest
func processResponse(_ result: Result<Response, Error>, target: TargetType) -> Result<Response, Error>
```

*Note: function names were chosen to be consistent with existing interface but
there is an opportunity to better conform to the Swift API Guidelines*

These functions are called before `willSendRequest` and after `didReceiveResponse`
respectively. Like the other protocol functions, empty implementations are
provided by default.

## Example

The following is an example of supplying an access token via a plugin and only
authorizing targets that need it.

```swift
protocol AuthorizedTargetType: TargetType {
  var needsAuth: Bool { get }
}

struct AuthPlugin: PluginType {
  let token: String

  func prepareRequest(_ request: URLRequest, target: TargetType) -> URLRequest {
    guard let target = target as? AuthorizedTargetType, target.needsAuth else {
      return request
    }

    var request = request
    request.addValue("Bearer " + token, forHTTPHeaderField: "Authorization")
    return request
  }
}

let provider = MoyaProvider<Target>(
  plugins: [
    AuthPlugin(token: "eyeAm.AJsoN.weBTOKen"),
    NetworkLoggerPlugin()
  ]
)
```

### *Add another closure for simpler request modification*

While this would solve the problem of request modification not being
straightforward, it still wouldn't be clean to chain modifications together.
Also, there wouldn't be a simple interface to encourage 3rd party plugin creation.

### *Make the existing plugin functions capable of modifying request/results*

This would not be easily possible due to the way the `willSendRequest` abstracts
Alamofire's `Request`. In addition, adding the extra lifecycle calls makes it
easier to create consistent behavior (e.g. always log requests after they have
been modified by plugins).

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,7 @@
 
 - **Breaking Change** Renamed `ReactiveCocoaMoyaProvider` to `ReactiveSwiftMoyaProvider`.
 - Renamed the `ReactiveCocoa` subspec to `ReactiveSwift`.
+- `PluginType` can now modify requests and responses through `prepareRequest` and `processResponse`
 
 # 8.0.0-beta.5
 
--- a/Source/Moya+Internal.swift
+++ b/Source/Moya+Internal.swift
@@ -12,10 +12,18 @@ public extension MoyaProvider {
         let stubBehavior = self.stubClosure(target)
         let cancellableToken = CancellableWrapper()
 
+        // Allow plugins to modify response
+        let pluginsWithCompletion: Moya.Completion = {
+            result in
+            var result = result
+            self.plugins.forEach { result = $0.processResponse(result, target: target) }
+            completion(result)
+        }
+
         if trackInflights {
             objc_sync_enter(self)
             var inflightCompletionBlocks = self.inflightRequests[endpoint]
-            inflightCompletionBlocks?.append(completion)
+            inflightCompletionBlocks?.append(pluginsWithCompletion)
             self.inflightRequests[endpoint] = inflightCompletionBlocks
             objc_sync_exit(self)
 
@@ -23,14 +31,14 @@ public extension MoyaProvider {
                 return cancellableToken
             } else {
                 objc_sync_enter(self)
-                self.inflightRequests[endpoint] = [completion]
+                self.inflightRequests[endpoint] = [pluginsWithCompletion]
                 objc_sync_exit(self)
             }
         }
 
         let performNetworking = { (requestResult: Result<URLRequest, Moya.Error>) in
             if cancellableToken.isCancelled {
-                self.cancelCompletion(completion, target: target)
+                self.cancelCompletion(pluginsWithCompletion, target: target)
                 return
             }
 
@@ -40,9 +48,12 @@ public extension MoyaProvider {
             case .success(let urlRequest):
                 request = urlRequest
             case .failure(let error):
-                completion(.failure(error))
+                pluginsWithCompletion(.failure(error))
                 return
             }
+
+            // Allow plugins to modify request
+            self.plugins.forEach { request = $0.prepareRequest(request, target: target) }
 
             switch stubBehavior {
             case .never:
@@ -54,7 +65,7 @@ public extension MoyaProvider {
                         self.inflightRequests.removeValue(forKey: endpoint)
                         objc_sync_exit(self)
                     } else {
-                        completion(result)
+                        pluginsWithCompletion(result)
                     }
                 }
                 switch target.task {
@@ -79,7 +90,7 @@ public extension MoyaProvider {
                         self.inflightRequests.removeValue(forKey: endpoint)
                         objc_sync_exit(self)
                     } else {
-                        completion(result)
+                        pluginsWithCompletion(result)
                     }
                 }, endpoint: endpoint, stubBehavior: stubBehavior)
             }
--- a/Source/Moya+Internal.swift
+++ b/Source/Moya+Internal.swift
@@ -14,9 +14,8 @@ public extension MoyaProvider {
 
         // Allow plugins to modify response
         let pluginsWithCompletion: Moya.Completion = { result in
-            var result = result
-            self.plugins.forEach { result = $0.processResponse(result, target: target) }
-            completion(result)
+            let processedResult = self.plugins.reduce(result) { $1.processResponse($0, target: target) }
+            completion(processedResult)
         }
 
         if trackInflights {
@@ -52,7 +51,7 @@ public extension MoyaProvider {
             }
 
             // Allow plugins to modify request
-            self.plugins.forEach { request = $0.prepareRequest(request, target: target) }
+            let preparedRequest = self.plugins.reduce(request) { $1.prepareRequest($0, target: target) }
 
             switch stubBehavior {
             case .never:
@@ -69,19 +68,19 @@ public extension MoyaProvider {
                 }
                 switch target.task {
                 case .request:
-                    cancellableToken.innerCancellable = self.sendRequest(target, request: request, queue: queue, progress: progress, completion: networkCompletion)
+                    cancellableToken.innerCancellable = self.sendRequest(target, request: preparedRequest, queue: queue, progress: progress, completion: networkCompletion)
                 case .upload(.file(let file)):
-                    cancellableToken.innerCancellable = self.sendUploadFile(target, request: request, queue: queue, file: file, progress: progress, completion: networkCompletion)
+                    cancellableToken.innerCancellable = self.sendUploadFile(target, request: preparedRequest, queue: queue, file: file, progress: progress, completion: networkCompletion)
                 case .upload(.multipart(let multipartBody)):
                     guard !multipartBody.isEmpty && target.method.supportsMultipart else {
                         fatalError("\(target) is not a multipart upload target.")
                     }
-                    cancellableToken.innerCancellable = self.sendUploadMultipart(target, request: request, queue: queue, multipartBody: multipartBody, progress: progress, completion: networkCompletion)
+                    cancellableToken.innerCancellable = self.sendUploadMultipart(target, request: preparedRequest, queue: queue, multipartBody: multipartBody, progress: progress, completion: networkCompletion)
                 case .download(.request(let destination)):
-                    cancellableToken.innerCancellable = self.sendDownloadRequest(target, request: request, queue: queue, destination: destination, progress: progress, completion: networkCompletion)
+                    cancellableToken.innerCancellable = self.sendDownloadRequest(target, request: preparedRequest, queue: queue, destination: destination, progress: progress, completion: networkCompletion)
                 }
             default:
-                cancellableToken.innerCancellable = self.stubRequest(target, request: request, completion: { result in
+                cancellableToken.innerCancellable = self.stubRequest(target, request: preparedRequest, completion: { result in
                     if self.trackInflights {
                         self.inflightRequests[endpoint]?.forEach { $0(result) }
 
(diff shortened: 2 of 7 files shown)


BasThomas:
This is great! Some questions:

- I don't think this is a breaking change right now, right?
- On the following:
> *Note: function names were chosen to be consistent with existing interface but
there is an opportunity to better conform to the Swift API Guidelines*

Do you have an idea on where we can improve? As we haven't yet released `8.0.0` final yet, we'd love to iron out these Swift 3-issues and names.

- Danger will complain about this in a minute probably, but could you add these changes to the Changelog and maybe add your awesome example to the documentation?

Thanks again for taking this on!

scottrhoyt:
No problem!

> *  I don't think this is a breaking change right now, right?

Nope, should be good to go. The behavior of the existing Plugins haven't been modified and default implementations are provided so that Moya-consumer plugins should conform by default.

> * On the following:
> > Note: function names were chosen to be consistent with existing interface but
> > there is an opportunity to better conform to the Swift API Guidelines

Nothing big here. Mostly `didReceiveResponse` and `processResponse` seemed awkwardly named when the first parameter is of type `Result<Response, Error>` and not `Response`. You could consider `didReceiveResult` and `processResult`. Alternatively, a scheme like this could work:

```swift 
prepare(request: URLRequest, target: TargetType) -> URLRequest
willSend(request: URLRequest, target: TargetType)
didReceieve(result: Result<Response, Error>, target: TargetType)
process(result: Result<Response, Error>, target: TargetType) -> Result<Response, Error>
```

I went ahead and made an entry into the Changelog and added the example.

BasThomas:
Thanks for the changes! @Moya/contributors, do we want to take a look at the function signatures for `8.0.0`? See the above comment by @scottrhoyt.

pedrovereza:
:+1: on the changes and functions renaming.

@ashfurrow thoughts?

docs/Plugins.md
```diff
@@ -1,36 +1,39 @@
 Plugins
 =======
 
-Moya plugins are used to perform some side effect whenever a request is sent or received. Plugins are defined by the `PluginType` protocol, which can receive callbacks when:
-- a request is going to be sent;
+Moya plugins are used modify requests and responses or perform side-effects. A
+plugin is called when:
+- a request is being prepared.
```
sunshinejr: Hmm, dunno about this "when request is being prepared". Should the user be worried that Moya is still preparing the request? I'd go with something that describes it more clearly. Your docs for functions are good enough imo.

docs/Plugins.md
```diff
@@ -1,36 +1,39 @@
...
+- a request is about to be sent.
 - a response has been received.
+- a response is being processed.
```
sunshinejr: The same situation here (as for preparing).

sunshinejr:
The idea behind upgrading Plugins is great, and the implementation is 💯, good job @scottrhoyt! 

However I'm not sure about `prepareRequest`, for some reason it feels like duplicating functionality (`requestClosure`). We might get a lot more questions about what is better for that use case and so on. 

Also, about dropping the `request`, `response` part of the plugin function names - great idea, I'm up for that, but maybe let's do it in another PR to avoid chaos.

BasThomas:
Would definitely do the renaming in another PR, yeah.

Source/Plugin.swift
```diff
@@ -8,16 +8,24 @@ import Result
...
 
     /// Called after a response has been received, but before the MoyaProvider has invoked its completion handler.
     func didReceiveResponse(_ result: Result<Moya.Response, Moya.Error>, target: TargetType)
+
+    /// Called to modify a result before completion
+    func processResponse(_ result: Result<Moya.Response, Moya.Error>, target: TargetType) -> Result<Moya.Response, Moya.Error>
 }
 
 public extension PluginType {
+    func prepareRequest(_ request: URLRequest, target: TargetType) -> URLRequest { return request }
```
scottrhoyt: My hope was that the type information would be clear enough, but I can definitely see both sides. We can't use `RequestType` for `prepareRequest`, but it would also be possible to change `willSendRequest` to pass just the `URLRequest` of the `Alamofire.Request` to be more consistent. This would eliminate the need for `RequestType` but it would also mean that `CredentialsPlugin` would need to be modified. That might be a good thing now though so that there is consistency in where in the lifecycle requests are being modified.

docs/Plugins.md
scottrhoyt: I understand that confusion. Should this just be reverted or is there another change you have in mind?

sunshinejr: Maybe we can be more descriptive about events and in the parenthesis add the function name? So we can describe it freely and still have reference in the line. What do you think?

Also I've just noticed that you missed "to" in:
> plugins are used **to** modify requests

Source/Plugin.swift
sunshinejr: Hmm, would it possible to remove the `RequestType` with modification of `CredentialsPlugin`? Not sure because didn't really investigate it that much.

scottrhoyt:
@sunshinejr , the intention here is to replace the use of `requestClosure` for a large number of use cases. What I'm not a fan of about `requestClosure` is that if you make use of it, you are overriding default behavior and must offer a replacement for that behavior--even if it is trivial. It also isn't very composable with respect to chaining request modifications which discourages creating extensions for Moya to handle common use cases.

My hope is that this helps resolve a lot of the questions like "should this be a `requestClosure` or an `endpointClosure`" by saying that if the intention is to make a modification of a `URLRequest`, it should be a plugin--and maybe that plugin already exists! I think the question is whether or not those closures still would be necessary.

docs/Plugins.md
scottrhoyt: I'll take a stab at that.

BasThomas:
Hmm, should we also create some tests around `pluginsWithCompletion`?

Source/Plugin.swift
scottrhoyt: Not that I can see. `RequestType` is used to expose `Alamofire.Request`s functionality for adding `URLCredential`s. So if you eliminated it, `CredentialsPlugin` would need another path to adding credentials.

scottrhoyt:
The trivial case is tested by default. Testing the non-trivial case should be pretty straightforward. Just add a header to the request/response and then check for it in `willSendRequest` and the `completion` closure respectively. You could go further and test that the order of plugins is preserved, but I don't know if that is necessary unless we are documenting that assurance. I would have written the test, but it doesn't look like the test suite is in this repo?

BasThomas:
The tests are located in the Demo app, located [here](https://github.com/Moya/Moya/tree/master/Demo/Tests). :-)

Source/Moya+Internal.swift
```diff
@@ -12,25 +12,32 @@ public extension MoyaProvider {
         let stubBehavior = self.stubClosure(target)
         let cancellableToken = CancellableWrapper()
 
+        // Allow plugins to modify response
+        let pluginsWithCompletion: Moya.Completion = { result in
+            var result = result
+            self.plugins.forEach { result = $0.processResponse(result, target: target) }
```
ashfurrow: This _could_ be done with `reduce`, like:

```swift
let result = self.plugins.reduce(result, { resultSoFar, plugin in
  return plugin.processResponse(resultSoFar, target: target)
})
```

Or something similar. Not sure which reads better. Thoughts?

docs/Examples/AuthPlugin.md
```diff
@@ -0,0 +1,64 @@
...
+
+let source = TokenSource()
+let provider = MoyaProvider<Target>(
+  plugins: [
+    AuthPlugin(tokenClosure: { return source.token })
+  ]
+)
+
+source.token = "eyeAm.AJsoN.weBTOKen"
+```
```
ashfurrow: Documentation is :100: Thanks!

Changelog.md
```diff
@@ -2,6 +2,7 @@
 
 - **Breaking Change** Renamed `ReactiveCocoaMoyaProvider` to `ReactiveSwiftMoyaProvider`.
 - Renamed the `ReactiveCocoa` subspec to `ReactiveSwift`.
+- `PluginType` can now modify requests and responses through `prepareRequest` and `processResponse`
```
ashfurrow: Since we're adding to the `PluginType` protocol, won't this be a breaking change?

pedrovereza: Shouldn't be, as there is a default implementation for each function ([here](https://github.com/scottrhoyt/Moya/blob/1b7b08c2c1b405bc6099d1004119a36218a50489/Source/Plugin.swift#L24))

pedrovereza: To be fair: if we choose to rename the functions, then yes 😄

Source/Moya+Internal.swift
BasThomas: The `forEach` reads better for me, but not sure what others think. Would at least change the reduce to

```swift
let result = self.plugins.reduce(result) { resultSoFar, plugin in
  return plugin.processResponse(resultSoFar, target: target)
}
```

... losing the (by me) dreaded `})`.

BasThomas: Basically, I think both would be just fine.

scottrhoyt: I actually like `reduce` better because you can maintain immutability and also it's very clear that the `result` or `request` is being modified.

It might be slightly bulkier, but TBH `requestNormal` is pretty overweight as is 🙈, and I'd be more concerned about refactoring that to make it more readable and highlight the request/response lifecycle better. I'd at least write it like this:
```swift
let pluginsWithCompletion: Moya.Completion = {
  result in
  let processedResult = self.plugins.reduce(result) { $1.processResponse($0, target: target) }
  completion(processedResult)
}
```

scottrhoyt: I went ahead and made this change.

scottrhoyt:
I wrote some simple tests. Because I'm not that familiar with `Quick`, I put them where was the easiest for me which was where `MoyaProvider` with delayed stubs is being tested. It should probably go somewhere else since you don't need delayed stubs to test this and it's only increasing the test time unnecessarily, but there was already a `TestingPlugin` being used here. I'll leave the rest as an exercise for the reader.

docs/Plugins.md
scottrhoyt: @sunshinejr I tried updating the docs to be clearer. Let me know what you think.

scottrhoyt:
I think the biggest considerations here revolve around the API surface area that is being exposed. Once this is in place, it will be a source of future breaking changes that will need to be managed. Also, there is the question of whether or not `requestClosure` and `endpointClosure` should be deprecated or are still useful. I think there are more pros than cons with empowering plugins though.

sunshinejr:
@scottrhoyt love the changes to docs, good job! 🙌

ashfurrow:

PR #815, opened 2016-12-01
Swiftlint issue fix

BasThomas:


BasThomas:
Failed because of the missing summary, but I'd guess that is not necessary in this case.

BasThomas:
Ah, I [see](https://github.com/Moya/Moya/blob/master/Dangerfile) we can use `#trivial` to prevent this. Neat!

PR #825, opened 2016-12-06
Rename functions to comply with Swift 3 guidelines

BasThomas:
This renames some functions to better comply with the Swift 3 guidelines, as well as unifying the RxSwift and ReactiveSwift functions

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,9 @@
 # Next
 
 - **Breaking Change** Renamed `ReactiveCocoaMoyaProvider` to `ReactiveSwiftMoyaProvider`.
+- **Breaking Change** Renamed `filterStatusCodes(:)` to `filterStatusCodes(in:)`.
+- **Breaking Change** Renamed `request(token:)` to simply `request(:_)` (ReactiveSwift).
+- **Breaking Change** Renamed `notifyPluginsOfImpendingStub(request:)` to `notifyPluginsOfImpendingStub(for:)`.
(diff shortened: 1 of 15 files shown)


BasThomas:
Don't merge; let me write something in this mighty Changelog.

BasThomas:
Hmm, not sure about the renaming of `filterStatusCodes(:)` to `filterStatusCodes(in:)` quite sure myself, to be honest. I think `filterStatusCodes(0...9)` reads nicer than `filterStatusCodes(in: 0...9)`. Reading the first as "filter status codes zero through 9", which seems more readable. Thoughts?

AndrewSB:
FWIW I think "filter status codes zero through nine" is grammatically correct, and sounds better than "filter status codes in zero through nine"

pedrovereza:
I always thought of `0...9` as an interval ([Apple also calls it closed range](https://developer.apple.com/reference/swift/closedrange)). To me the function, as is, reads "filter status code interval 0 to 9".

But that's not a strong opinion, I'm ok with what makes more sense to us :wink:

ashfurrow:
Yeah, I think if we were to rename it to `filterStatusCodes(in:)`, I'd want to review the `filter` family of functions as a whole. For example, we also have `filterStatusCode(code: Int)`, why not `filter(statusCodes:)`, `filter(statusCode:)`, etc? I don't have an opinion either way, I think either works within the guidelines.

AndrewSB:
+1 for `filter(statusCodes:)`, `filter(statusCode:)`

PR #827, opened 2016-12-06
Cleanup closures #trivial

BasThomas:


ashfurrow:
Weird, I wouldn't think Danger would complain about that on trivial PRs. Might be worth a change.

PR #832, opened 2016-12-08
Make example code compile

BasThomas:
Also cleaned it up a bit.

(Yay, working with Moya for the first time 😬😬😬)


pedrovereza:
💯

PR #840, opened 2016-12-16
Various cleanups #trivial

BasThomas:


BasThomas:
Just saw that in the NetworkLoggerPlugin, there a non `is`-bool (verbose) that should be updated. Let's do that in a separate PR.

PR #841, opened 2016-12-17
Rename `verbose` to `isVerbose` in NetworkLoggerPlugin

BasThomas:


BasThomas:
Hmm...

```bash
Downloading CocoaPods master repo from cocoapods-specs S3 bucket...
Uncompressing CocoaPods master repo...
master/.git/objects/pack/pack-35326da53152e3c953d14b6383803365213400d7.pack: gzip decompression failed
tar: Error exit delayed from previous errors.
Download from S3 failed, cleaning up and falling back to standard checkout...

curl -sS https://cocoapods-specs.circleci.com/fetch-cocoapods-repo-from-s3.sh | bash returned exit code 1

Action failed: curl -sS https://cocoapods-specs.circleci.com/fetch-cocoapods-repo-from-s3.sh | bash
```

Let me restart the build

BasThomas:
Hmm, interesting. I am not sure. Should we use `verbose` in the initializer, but leaving it `isVerbose` internally? Does that make sense?

BasThomas:
Why are these public anyway? They are `let`s, and always configured in the init. Is there a reason they should be accessible?

pedrovereza:
Having it as `isVerbose` internally seems fine. In fact, [Carthage does it](https://github.com/Carthage/Carthage/blob/2808725a6c0cb7ae714b0cad28aeaf97d89064db/Source/carthage/Update.swift#L19), but you set this mode by using `--vebose` not  `--isVerbose`.

BasThomas:
Agreed 👍. Changed the constructor back to use `verbose` instead of `isVerbose`. @Moya/contributors, any idea as to [this](https://github.com/Moya/Moya/pull/841#issuecomment-267766014)?

> Why are these (read: `cURL` and `isVerbose`) public anyway? They are `let`s, and always configured in the init. Is there a reason they should be accessible?

If that's cleared up this should be good to go.

ashfurrow:
The properties are public because they're passed in as parameters to the initializer, and I [still adhere to API design principles](http://mattgemmell.com/api-design/) that specify all initializer parameters should be accessible after the fact. But that's just me – it's totally open for discussion!

BasThomas:
Yep, great! There's not really a problem keeping them public, so let's keep it this way. 👍

PR #842, opened 2016-12-17
Update dependencies

BasThomas:


BasThomas:
Yay Danger! :D

scottrhoyt:
After this was merged, when I run `pod install` in the Demo directory, I get this:

screen shot 2016-12-30 at 10 28 29 am

screen shot 2016-12-30 at 10 35 22 am

I'm using the beta of the new cocoapods release because I was having issues with the current release. I think we just need a minor change to the Podfile to support this.

BasThomas:
Strange; I don't think I have run into any issues myself doing this. 🤔

scottrhoyt:
Yeah, I don't know what's going on here. After playing around a bit more, I think it is quite possible it was on my local machine. Running a `pod update ReactiveSwift` and not committing the changes seems to have solved it for me. This is likely some difference between cocoapods 1.1.1 and 1.2.0-beta3.

PR #843, opened 2016-12-17
Update Demo's Podfile.lock after releasing

BasThomas:


docs/Releasing.md
```diff
@@ -23,5 +23,6 @@ To make a release:
 What you'll need to do manually afterwards:
 
 1. Update the Readme CocoaPods instructions to use the release you just made public. This might also include changes to the subspecs' instructions in the Readme.
+1. Update the Demo's `Podfile.lock` by running `pod install` in the Demo directory.
```
sunshinejr: 1. -> 2.

BasThomas: Seems like the list works anyway (using `1`s). I did the same as in the other lists, which are also using just `1`s.

PR #847, opened 2016-12-17
Update currently supported Swift version in Readme

BasThomas:


pedrovereza:
Can this be removed? 🤔 We already have that nice Swift -> Moya versions table

BasThomas:
Agree, that's much better.

PR #851, opened 2016-12-20
Update Carthage to use ReactiveSwift 1.0.0-rc.2

aschuch:


BasThomas:
Hey, I already updated dependencies in #842, which has yet to be merged. Might be better to update that (`rc.1` to `rc.2`) and update everything at once? :)

aschuch:
@BasThomas Oh I must have missed that. Yup that would be best I think.

BasThomas:
Closing this, then. :)

PR #860, opened 2016-12-24
Switch to Void over () for function returns.

scottrhoyt:
This is to clean up the warnings generated by SwiftLint.


AndrewSB:
Could someone explain what this change does?
From my understanding, `Void` is just a `typealias` for `()`, what is swiftlint's complaint?

scottrhoyt:
Yeah, there is nothing functional here. It's purely stylistic. I prefer Void, but if others would like to stay with (), we can silence these warnings by disabling this rule in .swiftlint.yml.

AndrewSB:
Which rule catches functions that return `()`?

scottrhoyt:
`void_return`

Issue #836, opened 2016-12-14
How to change baseURL at runtime?


piqiu21:
Hi,

I have a project where URL for a server is not known during compile time, only first server in chain is known, other servers URLs are given as a response from the first one. Also every server has its alternative URL in case of failure, so they must be switched on demand.

Here is how I tried to solve this, my provider:
```swift    
private lazy var provider: RxMoyaProvider<DataRequests> = {
        return RxMoyaProvider<DataRequests>(endpointClosure: {(target: DataRequests) -> Endpoint<DataRequests> in
            let server = URL(string: self.currentServer == .primary ? self.server.primary : self.server.secondary)!
            let url = server.appendingPathComponent(target.path).absoluteString
            return Endpoint(url: url, sampleResponseClosure: {.networkResponse(200, target.sampleData)}, method: target.method, parameters: target.parameters)
        }, manager: self.connectionManager)
    }()
```
and here is my request:
```swift
    public func dataModel() -> Observable<VeraDataModel>{
        return requestDataModel()
            .catchError{_ in
                self.currentServer = self.currentServer.changeActiveServer()    //change server and try again
                return self.requestDataModel()
        }
    }
    
    private func requestDataModel() -> Observable<VeraDataModel>{
        return self.provider
            .request(DataRequests.dataModel)
            .debug()
            .mapObject(VeraDataModel.self)
    }
``` 
So first requirement is solved by creating a new class instance (containing code I mentioned) every time I got an information about new server. 

The second one is more tricky. As far sa I understand Rx (and its async nature) since this line:
```swift
self.currentServer = self.currentServer.changeActiveServer()    //change server and try again
```
is introducing side effects it will be a problem when multiple dataModel() calls will occur (race condition)

Does anybody have simmilar problems?

scottrhoyt:
There are a couple of things to unpack here. First of all, currently Moya's support for executing simple requests against a single `URL` is pretty light. There are some new features suggested in #861 that would make this easier, but we are still deciding whether or not they agree with Moya's philosophy. If you want to include this in your own code, I'd suggest looking at how `SingleURLTarget` is implemented and possibly include that in your code.

Without a `SingleURLTarget`, I would recommend using Alamofire instead for the dynamic `URL` requests.

[...]

PR #861, opened 2016-12-24
Default TargetType Property Implementations And SingleURLTarget

scottrhoyt:
Not 100% sure this is a great idea, but hopefully this PR can facilitate a discussion. I'll add a CHANGELOG entry and update the docs if we agree on this direction.

### Problem

Currently, if a Moya user wants to make a simple GET request to a new URL (not represented by an existing `TargetType`) they have a couple of options: 

They can create a new `TargetType` which will allow them to keep a consistent networking layer as well as have access to `RxMoyaProvider`, `ReactiveMoyaProvider`, easy response mapping, and 3rd party extensions. However, there is a good amount of boilerplate needed for a simple request and a new type needs to be created even for one-off requests.

Alternatively, they could utilize `Alamofire` or `URLSession` directly, but this would introduce inconsistency into their networking layer and they wouldn't have access to easy reactive extensions and response mapping.

### Solution

Provide default implementations of all `TargetType` properties to represent the most commonly used combination of request settings: a GET request with no parameters.

Additionally, provide a trivial `TargetType` implementation (`SingleURLTarget`) so that users can easily create a `TargetType` that represents a GET request to a single URL with no parameters.

### Usage

```swift
let url = URL(string: "http://api.com")!

// Example 1
MoyaProvider().request(SingleURLTarget(url: url)) {
    result in
    // process result
}

// Example 2
let target = SingleURLTarget(url: url)
MoyaProvider().request(target) {
    result in
    // process result
}

// Example 3
let provider = MoyaProvider<SingleURLTarget>()
provider.request(target) {
    result in
    // process result
}
```

### Tradeoffs

* A reduction in conformance errors that are generated in compilation. These errors are often a source of documentation of available behavior.
* Explicit implementation of all behavior is self documenting and may be easier for developers to reason about.

These tradeoffs can be mitigated by improving the documentation for `TargetType`.

### Alternatives Considered

Provide `SingleURLTarget` without providing default implementations for other `TargetType`s. This would be a good compromise between servicing the specific use case of a simple target without reducing the self-documenting nature of implementing `TargetType`. However, there are other common use cases where other properties like `parameters`, `method`, `task`, or `sampleData` can result in unnecessary boilerplate.

### DIFF
--- a/Demo/Demo.xcodeproj/project.pbxproj
+++ b/Demo/Demo.xcodeproj/project.pbxproj
@@ -71,6 +71,9 @@
 		3B4D21BC1E16075C00E2AB87 /* MultipartFormDataSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3B4D21BB1E16075C00E2AB87 /* MultipartFormDataSpec.swift */; };
 		3B4D21BD1E16075C00E2AB87 /* MultipartFormDataSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3B4D21BB1E16075C00E2AB87 /* MultipartFormDataSpec.swift */; };
 		3B4D21BE1E16075C00E2AB87 /* MultipartFormDataSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3B4D21BB1E16075C00E2AB87 /* MultipartFormDataSpec.swift */; };
+		3B151C521E0F006A00BAF065 /* TargetSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3B151C511E0F006A00BAF065 /* TargetSpec.swift */; };
+		3B151C531E0F006A00BAF065 /* TargetSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3B151C511E0F006A00BAF065 /* TargetSpec.swift */; };
+		3B151C541E0F006A00BAF065 /* TargetSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3B151C511E0F006A00BAF065 /* TargetSpec.swift */; };
 		4ACBAFF102F6CE19E4A569DE /* Pods_DemoMultiTarget.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 6F71370BEE9C086CDD6D71B8 /* Pods_DemoMultiTarget.framework */; };
 		5EC197E61A1BB16D00F4DFD4 /* AppDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5EC197E51A1BB16D00F4DFD4 /* AppDelegate.swift */; };
 		5EC197E81A1BB16D00F4DFD4 /* ViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5EC197E71A1BB16D00F4DFD4 /* ViewController.swift */; };
@@ -121,6 +124,7 @@
 		292BF354BC027FDF50F146C1 /* Pods-DemoMultiTarget.debug.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = "Pods-DemoMultiTarget.debug.xcconfig"; path = "Pods/Target Support Files/Pods-DemoMultiTarget/Pods-DemoMultiTarget.debug.xcconfig"; sourceTree = "<group>"; };
 		3BA460501E15EA4700511D6A /* MultiTargetSpec.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = MultiTargetSpec.swift; sourceTree = "<group>"; };
 		3B4D21BB1E16075C00E2AB87 /* MultipartFormDataSpec.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = MultipartFormDataSpec.swift; sourceTree = "<group>"; };
+		3B151C511E0F006A00BAF065 /* TargetSpec.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = TargetSpec.swift; sourceTree = "<group>"; };
 		4BF06D5D3DD24C44F9301FB5 /* Pods-Demo.release.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = "Pods-Demo.release.xcconfig"; path = "Pods/Target Support Files/Pods-Demo/Pods-Demo.release.xcconfig"; sourceTree = "<group>"; };
 		5E0646561BE7999C00E900DC /* Moya.podspec */ = {isa = PBXFileReference; lastKnownFileType = text; name = Moya.podspec; path = ../Moya.podspec; sourceTree = "<group>"; };
 		5E0646571BE799C500E900DC /* License.md */ = {isa = PBXFileReference; lastKnownFileType = net.daringfireball.markdown; name = License.md; path = ../License.md; sourceTree = "<group>"; };
@@ -201,6 +205,7 @@
 				067D496D1C0D3886005BBA79 /* MoyaProviderSpec.swift */,
 				3BA460501E15EA4700511D6A /* MultiTargetSpec.swift */,
 				3B4D21BB1E16075C00E2AB87 /* MultipartFormDataSpec.swift */,
+				3B151C511E0F006A00BAF065 /* TargetSpec.swift */,
 				067D496E1C0D3886005BBA79 /* NetworkLoggerPluginSpec.swift */,
 				067D496F1C0D3886005BBA79 /* Observable+MoyaSpec.swift */,
 				067D49711C0D3886005BBA79 /* ReactiveSwiftMoyaProviderTests.swift */,
@@ -774,6 +779,7 @@
 				3B4D21BD1E16075C00E2AB87 /* MultipartFormDataSpec.swift in Sources */,
 				067D49981C0D3886005BBA79 /* RxSwiftMoyaProviderTests.swift in Sources */,
 				1D9CF9A91D4BD594004A95B7 /* MethodSpec.swift in Sources */,
+				3B151C531E0F006A00BAF065 /* TargetSpec.swift in Sources */,
 				067D498C1C0D3886005BBA79 /* NetworkLoggerPluginSpec.swift in Sources */,
 				067D49861C0D3886005BBA79 /* MoyaProviderIntegrationTests.swift in Sources */,
 				067D49801C0D3886005BBA79 /* ErrorTests.swift in Sources */,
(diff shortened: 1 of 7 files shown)


BasThomas:
I was thinking... should we default everything for `TargetType` as well, or just for your suggested `SingleURLTarget`? I think the latter might be better.

Also, a separate PR for the documentation for `TargetType` would also definitely welcome!

scottrhoyt:
Yeah, there are pros and cons to both approaches.

On one hand, you can see even from the sample `TargetType`s in TestHelpers.swift, defaulting for all `TargetType`s helps clean up a decent amount of boilerplate. I like how this makes simple API implementations even more readable.

On the other hand, the lack of default implementation is a powerful means for discovering functionality and ensuring intended behavior.

AndrewSB:
Hmmm

@scottrhoyt I feel as if you're totally right when you say

> lack of default implementation is a powerful means for discovering functionality and ensuring intended behavior

I thought the problem you've noticed is better serviced by something like `NSURLSession` or `Alamofire`. but then I looked around a bit at Moya's documentation, and realized that Moya is supposed to be a completely unified networking layer. If youre using Moya, there should be no need to use Alamofire.
diagram

So now I'm not totally sure whether or not Moya should integrate your solution. Can you share the code you were working with that led to you seeing the need of a default implementation? So I can see where you're coming from?
The way I've been using Moya is to model all of my API calls. So I look at my API's swagger, and then I convert each operation into a Moya enum case. Moya, for me, has become a way to document how I'm using the provided API. In that paradigm, being able to create a one-off API call has no value, I'd rather have the API call documented in my provider's enum

scottrhoyt:
I don't have any personal code to share, unfortunately. What got me thinking about this is when I was thinking through a good strategy for #836. In that case I thought writing all the typical `TargetType` boilerplate to utilize Moya in a simple request was a bit unnecessary, yet using Alamofire felt like going against the idea of Moya in the first place.

That made me realize you'd be likely to run into similar scenarios when utilizing 3rd party APIs in a light way with Moya. Some examples of this might be adding weather information to an app, using an image API like Flickr, or accessing a Google location API. In these cases you are unlikely to need to fully specify `TargetType`s properties. You may need nothing more than `baseURL`, `path`, and possibly `parameters`.

Also, I find myself breaking up my API client code into multiple `TargetType`s to keep them from becoming monolithic (anything more than a handful of cases with several associated values each and a `TargetType` starts to get less readable IMO). In these cases, I often wind up with `TargetType`s that don't need `method`, `task`, or even `parameters` to be configurable. So I do wind up repeating a decent amount of boilerplate.

Finally, with the introduction of `validate` and now `parameterEncoding` with default values, the case for not providing defaults for the other properties felt a little more diluted to me.

But like said, I definitely see the other side here so I was hoping this proposal would bring out some other opinions.

scottrhoyt:
If you do a [search](https://github.com/search?utf8=✓&q=language%3Aswift+-filename%3ATarget.swift+TargetType+baseURL+path+method&type=Code&ref=searchresults) for how `TargetType` is used in public GitHub repos, you can see there is a high percentage of cases where the suggested defaults would reduce the amount of code needed. `method`, `task`, `parameters`, and `sampleData` are commonly `get`, `request`, `nil`, and `Data()` respectively. I would think there is some bias in this sample because more complicated API client code is probably more likely to be in a private repository, but it's still an interesting data point.

scottrhoyt:
@Moya/contributors Any more thoughts on this?

AndrewSB:
@scottrhoyt what about a subclass of TargetType? `OpinionatedTargetType` or something that gives you nice defaults?
That way one can continue to take advantage of 
> lack of default implementation is a powerful means for discovering functionality and ensuring intended behavior

while removing the need to specify defaults in the simple cases

scottrhoyt:
Thanks for the suggestion @AndrewSB! Do you mean a new protocol conforming to `TargetType` and adding default implementations (a sub-protocol, if you will)? I see where you are coming from on that.

It's certainly an option, but at first blush I'm not in favor. That is primarily because I think making the API surface area larger works counter to the goal of this PR--which is to make Moya easier to use for simple cases. An `OpinionatedTargetType` would require users to read more documentation to confidently use Moya, and that's really the same problem as is introduced by just doing the defaults on `TargetType` except you don't get all the benefits.

At it's heart, this is a philosophical decision for Moya about convention or configuration. I favor convention, and if you look at the defaults provided when you execute `Alamofire.request` or use `URLSession` to process a `URLRequest`, there are similar approaches. If the goal of Moya is to provide a unified networking layer, then I feel we should endeavor to make simple tasks as easy to accomplish with Moya as with the two technologies it replaces. By not doing so, we encourage just a new type of fragmentation of networking code. Even if this causes the self-documenting nature of `TargetType` to not be maximized, it will still be pretty clear to my eyes, and better and more organized than writing an ad-hoc networking layer on top of Alamofire or `URLSession`. Users will still have the option to fully specify the `TargetType` protocol for maximum clarity.

But I really think we need more than just the two of us to weigh in on this, since it is a decision central to Moya's philosophy.

AndrewSB:
Yeah @scottrhoyt, I see what you're saying. There are pros and cons, and its definitely central.

Lets wait for some other people to chime in before we do anything here

DeepMissea:
I think `SingleURLTarget` is useless. but the `TargetType` default extension is a good change.

petester42:
You can already do this by using the baseUrl and path properties. Just set the baseUrl to be the URL and the path to be nothing for the URL case.

colinta:
Since we're collecting opinions, I'll share mine.  This is not to say I feel strongly one way or the other about this, so just take this as me throwing another idea into the hat.

TL;DR: I think some defaults could be helpful (`method`, `parameters` in particular), but honestly I prefer the current more verbose style, because it leaves nothing up to the imagination.  i.e. for a new developer, that had no experience with Moya, looking at the `TargetType` implementation, having everything explicit is great.  We're really just talking about removing `default: return .get` from the bottom of a switch statement.

-------------

It looks like `SingleURLTarget` would make it easy to send a request to a server/endpoint that is *not* enumerated in a `TargetType`.  I do think you have a valid argument about Moya being a network abstraction layer, but I actually *don't* see it going so far as to replace the functionality of Alamofire - and Alamofire is still a dependency of Moya, so we're already using it no matter what.  The first example you gave above could be accomplished with Alamofire easily, and is exactly what Alamofire is built for.

Real world example: In the Ello app we often get image URLs from the Ello API, and we use Alamofire to download those to prime the image cache.  Building a Moya `TargetType` for this purpose feels heavy handed - we're wrapping a wrapper at that point.  Then again "reaching through" Moya to use Alamofire is, I think, what people are trying to avoid.  To me, it's the right tool for the job.  I see Moya as an *API Abstraction* layer.  We hope, in the future, to pull out the `ElloAPI` Moya code, so that if developers want to access our API they can easily do so using Moya.  Or, if nothing else, it provides a concise programmatic reference to our API!

If anyone here is using Moya to abstract an API *but also* wants to use it for *all* network requests, then I think you have a different view of what Moya is designed for than I do, and so we'll certainly disagree on this feature/direction.

The thing that I'm most concerned on is the default `sampleData`.  I think that having a default value for this would make it *way too easy* to skip this feature (and it's already easy enough).  Since support for testing is so high on the list of Moya's original features, it pains me to push it any further down the priority list.

scottrhoyt:
If we are going to advocate that simple API requests still be handled with Alamofire/`URLSession` then I think we should change the image at the top of the README to this:

diagram

As for `sampleData`, even as an ardent unit tester, I think there is a difference between first-class support of unit testing and first-class forcing people to write testing code. We all know very few projects get started with full test suites in place. If tests do eventually get added, it is often done at a later date (i.e. not much TDD going on). The amount of Moya code that starts off with

```swift
var sampleData: Data {
  return Data()
}
```

considerably outnumbers the amount of code that is using `sampleData` in it's intended fashion. [This search](https://github.com/search?utf8=✓&q=language%3Aswift+-filename%3ATarget.swift+TargetType+baseURL+path+method&type=Code&ref=searchresults) backs that up at least for public repos. By adding even a little bit of friction to selecting Moya as your networking layer because of it's verbosity and forcing of unit testing code into your app, we just encourage people to go a different route that will be harder for them to add unit tests once they get to the point that they realize they need it. While it is admirable to say that we can change testing etiquette by forcing people to implement `sampleData`, the reality is that it's still just a--confusing for anyone that isn't familiar with Moya--`Data()` away from being ignored and that is what most people are doing. Changing testing behavior will require a bigger impetus than a single networking library, but every single networking library should still support it in some way. Frankly I think Moya should go further to support testing by providing an easy way for test data to be stored in fixtures because if you want to use real returns from even a moderate-complexity API, the `TargetType` code quickly gets bloated back to unreadable status.

I started off thinking that the verbosity of `TargetType` was a positive because of it's self documenting nature. But I think I have come around to see it differently. When I look at this:

```swift
enum SimpleApi: TargetType {
  case newsfeed, messages

  var baseURL: URL { return URL(string: "http://api.com")! }

  var path: String {
    switch self {
    case .newfeed:
      return "newsfeed"
    case .messages:
      return "inbox"
  }
}

MoyaProvider<SimpleApi>.request(.newsfeed) { //... }
```
  
It is pretty clear to me that what is being executed is a simple GET request with no parameters. That's the same as if I were to see this:

```swift
Alamofire.request("http://api.com/newsfeed") { // ... }
```

That is because pretty much every way that I interact with HTTP (Alamofire, `URLRequest`, curl, wget, Postman, my web browser) defaults to a GET request with no parameters. And the reality is that once your API becomes complicated enough to really worry about whether API specifications are clear in the Moya code, you will have overridden most of the defaults and gotten right back to where we are today, except you'll be glad that you used Moya for even the simple things because now you don't have to go back and figure out how to write unit tests for all the times you fell back to Alamofire.

scottrhoyt:
@BasThomas I think we definitely need that documentation if we are going to provide default values. If not, then the documentation is pretty trivial--but probably still a good idea to have the whole surface area of your public API documented. I'll open up a separate PR with some trivial documentation.

colinta:
I think you make a good point with the `SimpleApi` example.  You're right, it's safe to assume that the requests would be sent with no parameters via a GET request.  I thought I said as much in my first comment - that having those defaults makes sense to me, *even though* I prefer the verbosity.  Maybe my wording was unclear, but that's what I was going for.

I think it's glib to argue that *any* use of Alamofire, or some other network/request framework than Moya, is some egregious misuse of the Moya paradigm.  Abstractions are always leaky, and the use cases you are touting (I'm referring specifically to the `SingleURLTarget` idea) are exactly where Moya leaks.  You're proposing that we expand Moya's surface area to cover these requests, and I'm arguing against it.  But it wouldn't burn my house down if we added this feature.

As for the sampleData, it's a concern, but I'm not planting my flag here.  I don't care one way or the other if people use Moya or not, but you can't force good habits, I certainly agree with you there.

scottrhoyt:
Do any @Moya/contributors know why Danger might be failing here:

screen shot 2016-12-29 at 9 16 45 pm

BasThomas:
I *think* it's the ["Any changes to library code need a summary in the Changelog."](https://github.com/Moya/Moya/pull/861#issuecomment-269254867) maybe?

scottrhoyt:
I'd love to pick a direction on this one way or another soon. Does anyone else have any opinions on whether or not this is a good fit for Moya?

If we decide to not go this direction, it's not a big deal to me. Other users and I can still add these defaults without much hassle. We can even add a suggestion in the docs. However, I do think it aligns with Moya's stated goal:

> So the basic idea of Moya is that we want some network abstraction layer that sufficiently encapsulates actually calling Alamofire directly. It should be simple enough that common things are easy, but comprehensive enough that complicated things are also easy.

If we decide against defaulting, we might want to revisit whether or not `parameterEncoding` (and possibly `validate`) should have a default before the 8.0.0 release.

ashfurrow:
Hard to say, is defer to whatever is least surprising for users, but that's certainly up for debate in this case. It doesn't sound like there's a tonne of push for it, so it might make sense to hold off for now at least.

scottrhoyt:
Alright. I'm going to leave this PR open for the moment so that others that might be gone for the holidays can consider it when they get back. It can still be done post 8.0.0 since it won't be a breaking change. 

I think we can make the case for providing a default for `validate` since it's suppressing Alamofire behavior and we try to not leak Alamofire where possible. However, I'll open a PR to remove the default for `parameterEncoding` and include that default in this PR. That way we are consistent on the current approach and can decide on providing defaults for `TargetType` as a whole.

scottrhoyt:
If no-one has any further thoughts on including this, I will go ahead and close this.

PR #871, opened 2016-12-29
TargetType.swift refactor #trivial

scottrhoyt:
There are a few things in TargetType.swift that I think there might be better locations for. Here are the changes:

* Move the `Method` typealias to the Moya+Alamofire.swift file with the other Alamofire type aliases.
* Extract MultipartFormData to it's own file.
* Move `StubBehavior` to Moya.swift since it has more to do with `MoyaProvider` than `TargetType`.
* Move the extension on `Method` to Moya+Internal.swift where it is used. This could possible also go in Moya+Alamofire.swift.

Let me know what you guys think about this.

### DIFF
--- a/Source/Moya+Internal.swift
+++ b/Source/Moya+Internal.swift
@@ -6,12 +6,9 @@ import Result
 extension Method {
     public var supportsMultipart: Bool {
         switch self {
-        case .post,
-             .put,
-             .patch,
-             .connect:
+        case .post, .put, .patch, .connect:
             return true
-        default:
+        case .get, .delete, .head, .options, .trace:
             return false
         }
     }
--- a/Source/Moya+Internal.swift
+++ b/Source/Moya+Internal.swift
@@ -1,6 +1,24 @@
 import Foundation
 import Result
 
+// MARK: - Method
+
+extension Method {
+    public var supportsMultipart: Bool {
+        switch self {
+        case .post,
+             .put,
+             .patch,
+             .connect:
+            return true
+        default:
+            return false
+        }
+    }
+}
+
+// MARK: - MoyaProvider
+
 /// Internal extension to keep the inner-workings outside the main Moya.swift file.
 public extension MoyaProvider {
     // Yup, we're disabling these. The function is complicated, but breaking it apart requires a large effort.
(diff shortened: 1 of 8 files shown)


scottrhoyt:
Hmm... that's odd. Not quite sure why this would reduce test coverage by such a significant amount.

BasThomas:
I don't really think a changelog entry is needed for a refactor. :)

scottrhoyt:
Haha. Yeah, I didn't think so either. But I was a bit unsure because of Danger calling the build failed. I know I could add a #trivial tag to the PR, but this didn't really feel trivial either. Think I should go that route?

BasThomas:
Hmm, added #trivial in this case just to make Danger happy. Oh well.

Source/Moya+Internal.swift
```diff
@@ -1,6 +1,24 @@
 import Foundation
 import Result
 
+// MARK: - Method
+
+extension Method {
+    public var supportsMultipart: Bool {
```
AndrewSB: should we make this an exhaustive switch instead?

AndrewSB:
@scottrhoyt: thank you for doing this 😄

Changelog.md
```diff
@@ -3,6 +3,7 @@
 - `TargetType` can specify its `ParameterEncoding`. The Default is `URLEncoding`.
 - Added documentation for `TargetType` and associated data structures.
 - Re-add `MultiTarget` to project.
+- Refactor TargetType.swift
```
scottrhoyt: Great. Once we settle on changes I will revert that commit and rebase.

scottrhoyt:
No problem! Anyone have any ideas on the test coverage?

Source/Moya+Internal.swift
scottrhoyt: I like that idea.

scottrhoyt: Done!

scottrhoyt:
Ok, so I think I understand a little more about the drop in test coverage. For some reason, [CodeCov is registering](https://codecov.io/gh/Moya/Moya/src/sh_targettype_refactor/Source/MultipartFormData.swift) 285 missed lines for MultipartData.swift even though the file only has 32 lines with 5 lines of executable code. I wonder if it's possible that there is some confusion between Moya's `MultipartFormData` and Alamofire's `MultipartFormData`.

PR #874, opened 2016-12-30
Fix inaccurate Codecov reporting. #trivial

scottrhoyt:
While trying to figure out some of the codecov problems from #871, I realized that we were incorrectly getting a lot of coverage credit for the testing code itself. I added everything in the Demo and Carthage directories to be ignored.

Unfortunately this means a hit to the coverage metric. 😢

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -16,8 +16,8 @@ dependencies:
 
 test:
   pre:
-  override: 
+  override:
   - rake test && rake test:carthage
-  post: 
+  post:
     - bundle exec danger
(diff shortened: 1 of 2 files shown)


scottrhoyt:
I am also including a fix to the issue seen in #871 where coverage was reduced due to a file name collision with Alamofire.

scottrhoyt:
Looks like the fix worked, so this somewhat mitigates the loss in coverage from not including testing code in the coverage metrics.

PR #877, opened 2016-12-30
Rename Moya to MoyaProvider. #trivial

scottrhoyt:
Also renaming the associated files with extensions on MoyaProvider.


scottrhoyt:
If we do this, we should probably also rename Moya+RxSwift and Moya+ReactiveSwift. I will do that in a further PR if this gets merged.

PR #878, opened 2016-12-30
Fix issue with pod install and new pre-release version of ReactiveSwift #trivial

scottrhoyt:
This fixed the issue mentioned in #842.


scottrhoyt:
Going to close this unless it can be reproduced.

Issue #866, opened 2016-12-29
How do I provide sample data for each endpoint if the sample response is big?


SaketGupte1mg:
I have large sample response for each endpoint in my app. However, I am afraid that if I directly paste the response inside each switch case my code will start looking ugly. I was wondering what are the other was to do this:

Please find below part of my switch statement with sample response for the smallest case.

```swift
 var sampleData: Data {
    switch self {
    case .movieDetails(_):
      return "{\"adult\": false,\"backdrop_path\": \"/wSJPjqp2AZWQ6REaqkMuXsCIs64.jpg\",\"belongs_to_collection\": null,\"budget\": 63000000,\"genres\": [{\"id\": 18,\"name\": \"Drama\"}],\"homepage\": \"http://www.foxmovies.com/movies/fight-club\",\"id\": 550,\"imdb_id\": \"tt0137523\",\"original_language\": \"en\",\"original_title\": \"Fight Club\",\"overview\": \"A ticking-time-bomb insomniac and a slippery soap salesman channel primal male aggression into a shocking new form of therapy. Their concept catches on, with underground \"fight clubs\" forming in every town, until an eccentric gets in the way and ignites an out-of-control spiral toward oblivion.\",\"popularity\": 6.868575,\"poster_path\": \"/adw6Lq9FiC9zjYEpOqfq03ituwp.jpg\",\"production_companies\": [{\"name\": \"Regency Enterprises\",\"id\": 508},{\"name\": \"Fox 2000 Pictures\",\"id\": 711},{\"name\": \"Taurus Film\",\"id\": 20555},{\"name\": \"Linson Films\",\"id\": 54050},{\"name\": \"Atman Entertainment\",\"id\": 54051},{\"name\": \"Knickerbocker Films\",\"id\": 54052}],\"production_countries\": [{\"iso_3166_1\": \"DE\",\"name\": \"Germany\"},{\"iso_3166_1\": \"US\",\"name\": \"United States of America\"}],\"release_date\": \"1999-10-14\",\"revenue\": 100853753,\"runtime\": 139,\"spoken_languages\": [{\"iso_639_1\": \"en\",\"name\": \"English\"}],\"status\": \"Released\",\"tagline\": \"How much can you know about yourself if you've never been in a fight?\",\"title\": \"Fight Club\",\"video\": false,\"vote_average\": 8.1,\"vote_count\": 5863}".data(using: String.Encoding.utf8)!
    case .similarMovies(_):
```

[...]

Issue #600, opened 2016-09-04
Load stubs from files


tomj:
Hey there - I'm probably missing something suuuper obvious but is there an idiomatic Moya way for me to load stubs using .json local files?

Or should I just roll something myself and give an escaped string to `sampleData`.

Cheers! 😀

vzsg:
I don't know about being idiomatic, but you could try this approach:

```swift
var sampleData: NSData {
    let name: String

    switch self {
    case .Search:
        name = "sample_search"
        break
    }

    guard let path = NSBundle.mainBundle().pathForResource(name, ofType: "json"),
        data = NSData(contentsOfFile: path) else {
            return NSData()
    }

    return data
}
```

tomj:
Aaaand I just realised that was exactly the way to go.  Just wanted to make sure there wasn't some nice way I could give a filename to `sampleData` rather than doing that myself.  I think the escaped JSON string in the README example for `sampleData` made me wonder whether I had to provide some kind of escaped variant of the data also.

Thanks for your help either way - appreciate it 😃

ashfurrow:
@tomj How do you think we could improve our documentation around this? Any suggestions?

tomj:
@ashfurrow hope #603 is up to scratch 😄

ashfurrow:
Looks great!

Issue #425, opened 2016-03-12
Must stubbed responses be included in the main target?


hirad:
It seems Moya is encouraging the stubbed responses to be part of the production app (this seems to be the case in [Eidolon](https://github.com/artsy/eidolon) as well).

I'm wondering why this is and if there's a way not include unnecessary json files in the production app? 

I thought about having a `stubbedResponseProvider` that would simply return empty `NSData`s in production but figured I'd try to understand why Moya is encouraging the other way first.

Thanks!

ashfurrow:
This is a great question, thanks! It's not something I've personally had an issue with, but I can totally see the concerns. 

So, the rationale behind this decision is to make sure that people _have_ to define sample date to make testing easier / encourage testing. There's no way to define a requirement _only_ in a test target, so this was our solution. 

But it's not ideal for everyone. An empty `NSData()` would do the trick, but what if you wanted to use this for testing? Empty `NSData()` wouldn't work then. In that case, I'd take the JSON out of the Xcode target (so they're not shipped with the app) and instead of using `NSBundle` to load the JSON, use some other way to reference the Xcode. FBSnapshotTestCase does this using either environment variables:

```objc
[NSProcessInfo processInfo].environment[@"ENV_VAR_NAME"]
```

or by using `NSBundle`'s `resourcePath`:

```objc
[[NSBundle bundleForClass:self.class].resourcePath stringByAppendingPathComponent:@"FolderWithJSON"]
```

That way the code is loading the files directly for disk and not through `NSBundle`. This won't work if you're running tests on the simulator, though, so that's one potential downside.

Let me know what you think – I'd like to know what you think an optimum solution would look like, and how we could improve our documentation around this. Thanks again for opening the issue!

hirad:
Thanks for the great response!

I asked this question because I was trying to keep test code/data completely out of the main code, so anything I suggest may not fit well with the "encourage testing" goal.

My initial hunch on this was that it has dependency injection written all over it. We need something to provide stubbed data. That can be empty in production, but must be available for tests - therefore, different dependencies.

Keeping in mind that I haven't *actually* used Moya yet (so I don't know how this meshes with it), my first idea was this: 

```swift
struct MyAPI {
    enum Target {
        case GetUser, GetFiles
    }
    
    private let target: Target
    private let sampleDataProvider: (Target -> NSData)
    
    private static func DefaultDataProvider(t: Target) -> NSData {
        return NSData()
    }
    
[...]

PR #881, opened 2016-12-31
Stubbed responses from files for TargetType

scottrhoyt:
This is meant to address an issue that I have come across in using `sampleData` to return responses of moderate complexity: `TargetType`s quickly get bloated and harder to read. This works against one of the main use cases for Moya: better organization of client API code. This has also popped up in a number of issues/PRs (e.g. #866, #603, #600, etc.).

The solution provided here is to write a helper function in `TargetType` (`stubbedResponse`) to load stubbed responses from a file. The intended usage is:

```swift
var sampleData: Data {
  switch self {
  case .users:
    return stubbedResponse("Users")
  case .profile:
    return stubbedResponse("Profile")
  }
}
```

This is accomplished by assuming that the file exists in the main bundle and that it has a "json" extension, but both of those default values can be overridden at the call site. If the data cannot be successfully loaded, the function will trap in debug mode (via `assert`) or just return an empty data in release config.

I'd love to get feedback in general, but here are some specific things:

* What do you think about the function name? I also considered `bundledResponse` to allow room for it to be differentiated from other ways of finding the stubbed response (e.g. via file URL so that one can avoid bundling test data with a production app).
* What do you think about the `assert()` behavior? I like this because devs will know right away in testing that the stubbed response can't be loaded as opposed to needing to debug a failing test because of a typo or the file being in a different bundle. However, even though it should be fine I am nervous letting an `assert` get into a production app because of high consequence if for some reason it doesn't work correctly. I'm also not sure if this will just hang a CI build or crash the test and cause it to fail. It is also untestable to my knowledge. The alternative I considered was just printing a warning message and returning an empty `Data` always. This is lower risk but possibly harder to debug. It's probably the way to go.

Another thing to consider is if we go further to encourage that test data be stored in files, we bring up some of the concerns in #425. We might want to consider alternatives to encouraging testing that don't require users to bundle production apps with test data/code. I'm a big believer of keeping testing and dev code/data out of production apps. I think it is less clean and the risk of unintended consequences is too high. It would, however, require a change in philosophy from Moya forcing testing to supporting testing. This was also talked about quite a bit in #861.

*Note: I'll add docs and CHANGELOG after incorporating feedback*


scottrhoyt:
I went ahead an renamed the function to `stubbedResponses(forFileNamed:)`. What do you guys think?

scottrhoyt:
Oops. Messed up a rebase it looks like. Let me retry.

scottrhoyt:
Looks good now.

BasThomas:
FWIW, I can't really think of a *better* function name. About the `assert`: I think this is the best option here.

BasThomas:
> I'm also not sure if this will just hang a CI build or crash the test and cause it to fail.

... but we should definitely test this before doing so.

scottrhoyt:
I'm pretty sure it will behave the same way as a `fatalError()` but I will open up a test branch here and throw an `assert` in to see what happens.

scottrhoyt:
According to [this failed build](https://circleci.com/gh/Moya/Moya/1159?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) it looks like it behaves the way we want it to.

BasThomas:
Hmm, does `sampleResponse` make sense for the function name?

justinmakaila:
Maybe it's just me, but does this really belong in Moya? It seems like it's just sugar that could be implemented on a per-client basis. As OP said, this is only really relevant for testing, and this is stuff that should be coming from your network layer, rather than Moya itself.

scottrhoyt:
@justinmakaila I agree that point should be discussed. We should consider whether or not Moya needs a rethink on how it handles testing. I think it might.

However, if the current approach holds, then we advocate that test data should be provided by `TargetType`s. Since it's not ideal to provide most real world test data from code, we should provide the tools necessary to load it from files. Sure, it might be pretty trivial stuff, but at least we could update our documentation and/or demos to show how it's likely to look in the real world instead of counting on users to come to this same conclusion.

scottrhoyt:
@BasThomas, I'm pretty indifferent to the two names. I still slightly prefer omitting the first argument label, but that's just my personal style.

justinmakaila:
I don't think it's on us to provide tools for loading the sample data in an opinionated format.

For instance, in my most recent networking library, I have this ((pardon the space formatting...):
```
  var sampleData: Data {
    switch self {
      // ...
        case .verify:
            return try! JSON([
                "isValid": true
            ]).rawData()
        case .graphOperation:
            return try! JSON([
                "data": [:],
                "errors": []
            ]).rawData()
        case .checkProgramCode:
            return try! JSON([
                "isValid": true
            ]).rawData()
  }
}
``` 
I think we're headed down a slippery slope if we're just adding sugar.

The end user of Moya is the one implementing `TargetType`, and should be deciding where and how to load their data for their stubbed responses. If they're not building with tests in mind, they're going to put `return Data()`, and if they are, they probably have some custom way that they're going to serve their responses anyway.

scottrhoyt:
I don't disagree with you. However, I just think our viewpoint might be in disagreement with the stated goal of Moya:

> Treats test stubs as first-class citizens so unit testing is super-easy.

By forcing `TargetType`s to provide a `sampleData`, it seems we are already on the slippery slope of recommending testing patterns. This PR goes not much further really, as it's just providing an optional tool and there is nothing to say you can't still do it the old way. 

On the flip side, judging by the issues that I referenced above, some users do run into the problem of our documented pattern not going far enough to support testing with real world responses from moderate complexity APIs. The goal of this PR was to prevent all those users from having to write this method themselves, not represent my personal feeling on how testing should be done.

Personally, I think tests data should be in the test project and supplied by the test objects, not bundled with the main app. But that approach has it's own drawbacks, so I respect that Moya is trying to do it differently.

justinmakaila:
But isn't providing `sampleData` and built in mocks/stubbing enough to make it a first class citizen?

API resource models are literally providing their `Data` representation.

I'm just afraid that we'll start adding something like this while everyone has divergent ideas will put us in the same place we were when people started writing JSON extensions and saying "oh well 90% of services deliver data in this form, why don't we build in JSON handling with this provider?".

scottrhoyt:
That is fair. To me this is less divergent and more a natural extension of the tools we provide for testing, but I totally understand the pushback for going even a bit further in this direction as it's further promoting a specific testing pattern that doesn't sit well with everyone... including sometimes myself.

This is code that I personally use to reduce the pain of managing complex responses for sample data, but it's not a one-size-fits-all approach. Issues led me to believe others were having the same problems, so I offered my help. We don't have to include it though.

If you had some other ideas on how we could make testing more open and flexible to all approaches (including keeping test data in test bundles) while still keeping it first class, I would love to see that PR come through. 👍

BasThomas:
I think the least we can do here (especially since I think @justinmakaila makes a fair point), is to include this snippet in the docs. Just as we did with some of the String extensions. This then is also a reference point for future issues. :)

scottrhoyt:
The consensus here looks like this isn't a good idea, so I will go ahead and close this.

fjcaetano:
Hey @scottrhoyt, I'm also using a stubbed files approach and also wouldn't like to have these json files in the main bundle to be distributed with the app. Did you manage to add them to the test target without having to change the whole project membership?

PR #882, opened 2016-12-31
Remove default for TargetType.parameterEncoding

scottrhoyt:
As discussed in #861, this PR removes the default implementation for `TargetType`'s `parameterEncoding` to be consistent with current approach for `TargetType`. The default will be reintroduced as part of #861.


scottrhoyt:
Should be good to go now.

BasThomas:
Oh, didn't notice those were missing too. Just saw the GitHub one and fixed it within GitHub. 🙈

Teamwork. 🙌

PR #883, opened 2017-01-01
Rename Error to MoyaError #definitelynottrivial

scottrhoyt:
So this one is a pretty big breaking change, and that could be a decent reason to not do it. I figure we should consider it before 8.0.0 though.

There are a number of reasons I think we should do this though:
  * More idiomatic
  * Fixes ambiguities in code completion and semantics with `Swift.Error`
  * No more verbose than needing to namespace `Moya.Error` everywhere
  * Helps to resolve a class of issues where a failure to namespace `Moya.Error` leads to conformance errors or unintended behavior
  * Eliminates the need to namespace `Swift.Error`

Basically, it has been my experience that overloading Swift standard library type names causes extraneous 👿 for maintainers and users alike. But I'm curious if others feel that same way enough to justify the big break.

To ease the change I added a type alias to `Moya.Error` and marked it with an unavailable attribute that offers a fix-it to `MoyaError`. And if we do this, I'll volunteer to submit some PRs to Moya extensions to fix them.


AndrewSB:
For what it's worth, I prefer `Moya.Error` to `MoyaError`. 

Having either `MoyaError` or `Moya.Error` be more idiomatic isn't an objective measure, it's just a marker of familiarity with.
Your comment about the confusion between `Swift.Error` and `Moya.Error` is real though, it does make `Error` ambiguous in some cases, and forces one to namespace the error. But, I feel as if that makes sense, you have your module, each with its own kind of error (`Swift.Error`, `Moya.Error`, or `MyApplication.Error`), and each of those define an error (that's what I have, at least).
Then in my `catchError`, I usually switch to pattern match between those three.

It wouldn't be the end of the world if we made this change, I'd be ok with it, but I think having `MyModule.Error` instead of `MyModuleError` is a paradigm shift and I'd like to encourage it for swift. A few years from now I think it would be awesome if the former is the default, idiomatic naming option for Errors in our language 😄

scottrhoyt:
I can understand that line of reasoning @AndrewSB. I think it would be a reasonable way to go if Swift and SourceKit had stronger support for namespaces. Unfortunately that's not the case yet.

The only objective thing I can say about what is idiomatic or not is that I just scanned through the 10 most starred Swift libraries and I couldn't find any other examples of publicly overloading `Swift.Error`. So I think it would be fair to say that most users familiar with these libraries wouldn't expect it. That's not to say that doing something unexpected in a software library is bad, but there should be some good reasons for it.

Also important to crafting a well written library is consistent reasoning. I think we all can agree it wouldn't be a great idea for us to overload `Swift.String` or `Swift.URL`. If that's not the case, then I can share some tearful stories of days lost dealing with a library that overloaded `Swift.String` (the punchline is that I wrote some declarations as `let a: String` when they unexpectedly needed to be `let a: Swift.String` 😭). So if overloading the standard library type names is bad in general, then we need to ask why should `Swift.Error` be the special case? Again, it's not to say having a special case is bad, but it should come with some good reasoning.

Paradigm shifts also aren't bad. But I think they are only good when they make users lives better or easier.

I don't want to rock the boat too much here. I just think if we want to shift the paradigm to a special case that is unexpected, it should come with solid reasoning and make users' lives better. The reasoning for `Moya.Error` doesn't seem robust and I think it actually might makes users' lives a little more difficult--even in writing about this PR we have to explicitly namespace a swift standard library type .

I think your idea of every module exposing a `Module.Error` has a lot of appeal to it. If we could get everyone on board with that, having such a standard convention might be all the reason we need. But realistically I don't think it would work because not every library can cleanly fold all of their errors into a single `Module.Error`. For example, a library might be clearer and cleaner with a `JSONError`, `HTTPError`, and `CacheError`. So I'm not convinced that this would work as a widely-accepted standard either.

Sorry for the wall of text. I just think that Moya is so well respected that it helps craft standards, so we need to really think through how we make use of that.

BasThomas:
As we've seen with [Result](https://github.com/Moya/Moya/pull/885#issuecomment-269935129), I think it might not be a bad idea to go ahead with this and prevent namespacing issues, as well as making room for the use of more than one "Error type".

ashfurrow:
Yeah, I like `Moya.Error` from an idealogical perspective but pragmatically I think `MoyaError` makes more sense. Another solution could be to put the `Moya.Error` type inside the `MoyaProvider` class, since that class is what generates the errors. I'm fine with whatever we decide though, thanks for bringing up the discussion 👍

scottrhoyt:
No problem! Thanks for entertaining the idea!

I think embedding it into `MoyaProvider` is an interesting solution, yet I see a number of drawbacks:

* At ~50 lines (without much documentation), folding Error.swift into `MoyaProvider` will be a significant expansion.
* There are other types than `MoyaProvider` that need access to `Moya.Error` (e.g. `PluginType` and `Response`). So embedding it will cause those structures to need more namespacing and is probably a code smell that it shouldn't be embedded.
* It might not be straightforward for people where to find `Error` when browsing the source.

ashfurrow:
Cool, let's rename it `MoyaError` then. Like I said, it's not ideologically as awesome but we've got to live/code in the real world 😄

PR #884, opened 2017-01-01
LocalizedError for Moya.Error

petripartanen:
With this pull request Moya.Error conforms to LocalizedError thus providing more meaningful error descriptions. Also removed unused Moya.Error enumeration case.

Before
print(imageMappingError.localizedDescription): The operation couldn’t be completed. (Moya.Error error 0.)
print(underlyingError.localizedDescription): The operation couldn’t be completed. (Moya.Error error 5.)

After
print(imageMappingError.localizedDescription): Failed to map data to an Image
print(underlyingError.localizedDescription): Response status code was unacceptable: 404.

### DIFF
--- a/Demo/Tests/NetworkLoggerPluginSpec.swift
+++ b/Demo/Tests/NetworkLoggerPluginSpec.swift
@@ -2,6 +2,7 @@ import Quick
 import Nimble
 import Moya
 import Result
+import enum Alamofire.AFError
 
 final class NetworkLoggerPluginSpec: QuickSpec {
     override func spec() {
@@ -84,11 +85,12 @@ final class NetworkLoggerPluginSpec: QuickSpec {
         }
 
         it("outputs an empty response message") {
-            let response = Response(statusCode: 200, data: "cool body".data(using: .utf8)!, response: nil)
-            let result: Result<Moya.Response, Moya.Error> = .failure(Moya.Error.data(response))
-
+            let emptyResponseError = AFError.responseSerializationFailed(reason: .inputDataNil)
+            
+            let result: Result<Moya.Response, Moya.Error> = .failure(.underlying(emptyResponseError))
+            
             plugin.didReceive(result, target: GitHub.zen)
-
+            
             expect(log).to( contain("Response: Received empty network response for zen.") )
         }
 
--- a/Demo/Tests/NetworkLoggerPluginSpec.swift
+++ b/Demo/Tests/NetworkLoggerPluginSpec.swift
@@ -87,7 +87,7 @@ final class NetworkLoggerPluginSpec: QuickSpec {
         it("outputs an empty response message") {
             let emptyResponseError = AFError.responseSerializationFailed(reason: .inputDataNil)
             let result: Result<Moya.Response, Moya.Error> = .failure(.underlying(emptyResponseError))
-            
+
             plugin.didReceive(result, target: GitHub.zen)
 
             expect(log).to( contain("Response: Received empty network response for zen.") )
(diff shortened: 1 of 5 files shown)


Source/Error.swift
```diff
@@ -18,9 +17,30 @@ public extension Moya.Error {
...
+        case .imageMapping:
+            return "Failed to map data to an Image"
+        case .jsonMapping:
+            return "Failed to map data to a JSON"
+        case .stringMapping:
+            return "Failed to map data to a String"
+        case .statusCode:
+            return "Status code didn't fall within the given range"
+        case .requestMapping:
+            return "Failed to map Endpoint to a URLRequest"
```
scottrhoyt: Sorry for the little nitpick, but would you mind adding periods to these descriptions? Otherwise, they look great.

scottrhoyt:
FYI, the test code can be found in the Demo dir. Do a `pod install` and open up the Demo workspace to access and run the test code.

petripartanen:
Thanks for the encouraging feedback. Didn't notice the tests before CircleCI complained. Working on the requested improvements.

petripartanen:
codecov/patch coverage looks intimidating and i'm slightly tempted to write a test or two just to bump it up but doubt if the tests would provide any real value here.

Demo/Tests/NetworkLoggerPluginSpec.swift
```diff
@@ -2,6 +2,7 @@ import Quick
 import Nimble
 import Moya
 import Result
+import enum Alamofire.AFError
```
AndrewSB: woah, you can do this? That's incredible

ashfurrow:

Demo/Tests/NetworkLoggerPluginSpec.swift
BasThomas: Yep. You can even import specific functions :-)

petripartanen:
In my opinion we can close the issue and consider it completed. The following is bikeshedding and of little importance. 

Having the following behaviours for Moya.Error and enforcing them using unit tests wouldn't be completely nonsense.

* Moya.Error enumaration cases should return non-nil errorDescription.
* Enumeration cases that have associated error, for example .underlying(Swift.Error), should include associatedError.errorDescription in the returned errorDescription.

As for test implementation I would like it to guide the programmer to the relevant test cases either by 1) failing tests and / or by 2) compiler error when a new error enumeration case is added. If these requirements are not met the test cases go unnoticed too easily.

1) I couldn't think of easy way to test future enumeration cases as Swift doesn't have convenient way to enumerate enums. I don't want to add `static let allValues = [.imageMapping, ...]` only for tests.
2) Compiler error could be archived by using switch statement, something along lines:

```
describe("error description") {
    it("should return description") {
        expectNonNilErrorDescriptionFor(error: Moya.Error.imageMapping(response))
        ...
        expectNonNilErrorDescriptionFor(error: Moya.Error.underlying(NSError.init()))
    }
}

fileprivate func expectNonNilErrorDescriptionFor(error: Moya.Error) {
    switch error {
    case .imageMapping,
         .jsonMapping,
         ....
         .requestMapping:
        expect(error.errorDescription).to(beNil())
    }
}
```

Above implementation is not satisfying as it fills goals only partially. The implementation does guide programmer to relevant file by triggering compiler error and required adding a new case to the switch but there's no guarantee that the programmer would remember to add a new call to `expectNonNilErrorDescriptionFor` in `it("should return description")`.

Summa summarum: couldn't figure out meaningful test cases with elegant and future-proof implementation that didn't require further modifying Moya.Error enum.

PR #885, opened 2017-01-02
Migrate Moya to a swift package manager compatible layout

AndrewSB:
Picks off where #698 left off. Closes #870.

My concerns:

- [x] **Test to make sure the project is still CocoaPods compatible** I'm not familiar with Cocoapods, so the new layout is done to Carthage & SPM best practices, but it would be awesome to have a cocoapods person test it and make sure everything still woks :)
- [x] **Update README**: I'm sure some of the instructions will have changed, locations of Demos & Tests, pulling down dependencies for Demos, etc.
- [x] **Update CI for new test locations**: The tests are currently failing because the tests CircleCI is expecting are no longer in the same. I also have them setup to pull dependencies down with Carthage instead, so we need to set that up. @BasThomas: I remember you offering to take a look

Future enhancements:
- Use `swift test` once #643 is merged
- [x] Implement the Carthage caching strategy @scottrhoyt & I discussed in #870

PR #891, opened 2017-01-03
Consolidate Demos into Moya.xcodeproj

AndrewSB:
This Pull Request has been created for @Moya/contributors to discuss the pros and cons of a 2 xcodeproj vs 1 xcodeproj project structure.

Currently, we have two xcodeproj's in Moya. One, `Moya.xcodeproj` contains our source code and tests, and the other `Demo/Demo.xcodeproj` contains two example projects (`Demo` and `DemoMultiTarget`).

Here's what I think (and what was gathered from the conversation in #885)

### Pros of going one-project
1. Better discoverability of the demo projects - they're right there when you open Moya.xcodeproj and you don't have to dig for them
1. Simplicity of just having one xcodeproj

### Cons of going one-project
1. Having an example/demo project in separate sub folder and project is in line with CocoaPods best practices and most libraries do it this way
2. A higher potential for *.xcodeproj's merge conflicts. Keeping separate projects will allow people to contribute to Demo code and core library code simultaneously with less risk that the Carthage-driving project will get corrupted.

Its also worth noting that making this change will move us to will have the effect of making Moya more of a carthage-first library instead of cocoapods-first library, since the Demo code will by default be working off of a carthage integration and not a cocoapods integration


ashfurrow:
Good notes describing the change – my inclination is to move Moya towards being an SPM-first library, does this PR make sense in that context?

AndrewSB:
@ashfurrow I feel the same way, Carthage affords itself to SPM, and the project structure between the two are almost interchangeable.

Merging this PR would get us closer to that point. The lingering difference to make us truly SPM-first, would be 
1. To finish up #643
2. to use SPM to pull down dependencies ourselves, which we currently aren't doing, mostly because we don't have #643 merged, but also because SPM - as it is now - doesn't build `.frameworks`, it compiles to an executable (`*.build`). Here's what Result looks like compiled, for example 

screen shot 2017-01-02 at 4 59 32 pm

To be honest, I'm not sure how people are using SPM compiled dependencies for iOS apps, I've seen people that have dragged the `*.swift` files that SPM fetched into their projects, but not the `.build` files.
I don't think SPM is at that point yet, but when it is, we should be ready to go 😄

scottrhoyt:
IMO Both directory/project structures are compatible with carthage and SPM. We can make carthage work with either structure and SPM doesn't really care about xcodeproj's. It is driven off of the sub directory structure under Sources and how that translates into modules.

This change will move us more out of line with CocoaPods standards as mentioned. And we should consider the consequences to testing the cocoapods integration and Demo project build. Changing documentation to instruct users to do a `carthage bootstrap` to try out the Demo project will also make Moya more carthage-first.

AndrewSB:
@scottrhoyt you're right, both layouts are compatible with all 3 package managers. I've been using the word compatible when what I really mean is in the spirit of, which is definitely a matter of opinion, and not right and wrong.

I'm also biased since I've never really used CocoaPods (I started programming a couple months before Swift came out, and other than iOS hello world esque apps, I've always used Carthage). So the upside to have `Demo` in it's own directory is lost on me

AndrewSB:
Thank you for the reviews. Do you want to take a shot at those changes @scottrhoyt? Otherwise I can take care of them some time tomorrow 

What do you guys think? Should we wait a couple days to see what the other contributors think?

As it stands now, this PR is only a matter of preference, it has no functional benefit. Perhaps in the future, when we're using the SPM to generate our xcodeproj, it might. But theres a chance that SPM will change their implementation in the interim, and having only 1 xcodeproj will not matter.

scottrhoyt:
Yeah, I totally get that. I haven't been a CocoaPods user in a while. It's been a change for me to work on a library that has--until now--been cocoapods-first.

For me the advantage has lees to do with directory structure and more to do with merge conflicts. The rule of thumb is to break up the XML-based files (e.g. xcodeproj's and storyboards) into smaller files to avoid the minimize the conflict resolution mess.

AndrewSB:
Gotcha, I do the same thing with my storyboards (the project Im working on now has 9). I hadn't thought of/heard of splitting up xcodeproj's to reduce merge conflicts. I'll look into that

scottrhoyt:
@AndrewSB, I can take a shot at those in the next couple of days. I have a couple of other things I am working on at the moment. I also need a bit of a break from making CI changes. Iterating on builds that take 10-20 minutes gets exhausting 😩. So if you'd like to take a stab first, then go for it and shoot me any questions you might have.

One of the things to decide is what the testing strategy looks like here. Previously we were testing that the Demo project built using CocoaPods (implicitly validating our pod). 

With this scheme, if we want to replace that functionality and/or do a carthage integration test as well, then we will need to write some integration tests as you suggested. Such tests probably involve generating a new xcode project for the Demo code on the fly, but there might be something more clever that can be done.

scottrhoyt:
I might also consider removing the "PRO" of 

> Swift package manager expects there to be one xcodeproj in the root of the project.

... because I don't think SPM cares about Xcode projects at all. You can have zero or many. It's just going to build what you tell it to in the directory structure and `Package.swift`. People less familiar with SPM might think this is getting us closer to SPM compatibility when we already are compatible and this is more of a stylistic/project philosophy choice.

AndrewSB:
Haha, I totally understand. Thank you for all you did speeding up CI, its noticeable 😄 

Yeah, we do need to rethink our package manager integration tests. 
Lets prioritize the 8.0 release while we wait for a couple more people to give us their thoughts, meanwhile thinking of how to better test.
I'm on a flight on Wednesday, so I can write those tests and validate locally while I don't have internet, and then we can finish this PR up towards the end of the week. I think thats enough time for everyone to give their thoughts.

You're right, I'll move that from the pros 😄

AndrewSB:
TODO

- [ ] Update docs for new Demo project location & Carthage instead of CocoaPods instructions
- [ ] Update circle.yml to pass, moving the Demo project expectation from `Demo/Demo.xcodeproj` to `Moya.xcodeproj`
- [ ] Write (hopefully parallelized) tests on CI to validate CocoaPods, Carthage & SPM validity (lets make this is **separate PR**)

scottrhoyt:
Currently we test the integrity of our CocoaPods integration by compiling the Demo project. By moving the Demo project to the main project, it will no longer be testing any integrations because it will use the local project instead. So if we don't add another integration test to this PR, it will be a regression.

Maybe a good way to break this up into PRs without regressing is to replace the integration test with another integration test and then open up a new PR that is concerned with the other two integration tests and parallelizing them. What do you think about that strategy?

scottrhoyt:
If we want to go SPM-first, we might also consider resolving #894.

sunshinejr:
After changing structure to SPM-first, changing project to single target, I think we can now think about consolidating Demos. How do you feel about it, @AndrewSB?

AndrewSB:
That sounds great! I can finish this up (probably) early next month, but anyone else who's motivated to make this happen should feel free to take charge 😊

sunshinejr:
Is someone up for doing this one? Would be awesome to get it started again. cc @Moya/contributors

AndrewSB:
on it

PR #893, opened 2017-01-03
Add AccessTokenPlugin

scottrhoyt:
Following through with the promise of #813, this PR adds a new `PluginType` called `AccessTokenPlugin`. This plugin can be added to a `MoyaProvider` to add authorization headers of the form:

```
Authorization: Bearer <token>
```

By default, this plugin will add the auth header to all `TargetType`s. However, if you want to control this behavior, you can have your `TargetType` conform to `AccessTokenAuthorizable` and provide `var shouldAuthorize: Bool { get }`.

### DIFF
--- a/Sources/Moya/Plugins/AccessTokenPlugin.swift
+++ b/Sources/Moya/Plugins/AccessTokenPlugin.swift
@@ -3,23 +3,53 @@ import Result
 
 // MARK: - AccessTokenAuthorizable
 
+/// A protocol for controlling the behavior of `AccessTokenPlugin`.
 public protocol AccessTokenAuthorizable {
+
+    /// Declares whether or not `AccessTokenPlugin` should add an authorization header
+    /// to requests.
     var shouldAuthorize: Bool { get }
 }
 
 // MARK: - AccessTokenPlugin
 
+/** 
+ A plugin for adding bearer-type authorization headers to requests. Example:
+ 
+ ```
+ Authorization: Bearer <token>
+ ```
+ 
+ - Note: By default, requests to all `TargetType`s will receive this header. You can control this
+   behvaior by conforming to `AccessTokenAuthorizable`.
+*/
 public struct AccessTokenPlugin: PluginType {
+
+    /// The access token to be applied in the header.
     public let token: String
 
     private var authVal: String {
         return "Bearer " + token
     }
 
+    /**
+     Initialize a new `AccessTokenPlugin`.
+     
+     - parameters:
+       - token: The token to be applied in the pattern `Authorization: Bearer <token>`
+    */
     public init(token: String) {
         self.token = token
     }
 
+    /**
+     Prepare a request by adding an authorization header if necessary.
+     
+     - parameters:
+       - request: The request to modify.
+       - target: The target of the request.
+     - returns: The modified `URLRequest`.
+    */
     public func prepare(_ request: URLRequest, target: TargetType) -> URLRequest {
         if let authorizable = target as? AccessTokenAuthorizable, !authorizable.shouldAuthorize {
             return request
--- a/Sources/Moya/Plugins/AccessTokenPlugin.swift
+++ b/Sources/Moya/Plugins/AccessTokenPlugin.swift
@@ -13,13 +13,13 @@ public protocol AccessTokenAuthorizable {
 
 // MARK: - AccessTokenPlugin
 
-/** 
+/**
  A plugin for adding bearer-type authorization headers to requests. Example:
- 
+
  ```
  Authorization: Bearer <token>
  ```
- 
+
  - Note: By default, requests to all `TargetType`s will receive this header. You can control this
    behvaior by conforming to `AccessTokenAuthorizable`.
 */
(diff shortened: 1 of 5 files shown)


Sources/Moya/Plugins/AccessTokenPlugin.swift
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     Prepare a request by adding an authorization header if necessary.
+     
+     - parameters:
+       - request: The request to modify.
+       - target: The target of the request.
+     - returns: The modified `URLRequest`.
+    */
+    public func prepare(_ request: URLRequest, target: TargetType) -> URLRequest {
+        if let authorizable = target as? AccessTokenAuthorizable, !authorizable.shouldAuthorize {
```
BasThomas: Elsewhere in Moya, we use `== false` (as this is clearer?), maybe want to do that here as well?

docs/Authentication.md
```diff
@@ -33,6 +33,52 @@ let provider = MoyaProvider<YourAPI>(plugins: [CredentialsPlugin { target -> URL
...
+To make adding these authorization headers easier, Moya comes with an
+`AccessTokenPlugin`. It can be added to your provider like this:
+
+```swift
+let token = "eyeAm.AJsoN.weBTOKen"
+let authPlugin = AccessTokenPlugin(token: token)
+let provider = MoyaProvider<YourAPI>(plugins: [authPlugin])
+```
+
+By default, `AccessTokenPlugin` will add the authorization header to requests to
```
BasThomas: `[..] will add the authorization header to all requests of the TargetType.`

docs/Authentication.md
```diff
@@ -33,6 +33,52 @@ let provider = MoyaProvider<YourAPI>(plugins: [CredentialsPlugin { target -> URL
...
+any `TargetType`. If you want to control this behavior, you can have your
+`TargetType` conform to `AccessTokenAuthorizable` and use `shouldAuthorize` to
+specify the behavior:
```
BasThomas: This might need an example of how to use this. Also, why the boilerplate from the user's perspective?

Sources/Moya/Plugins/AccessTokenPlugin.swift
scottrhoyt: Sure, I can make that change.

docs/Authentication.md
scottrhoyt: I'll change that grammar.

docs/Authentication.md
scottrhoyt: What follows is an example of how to use this. Which boiler plate are you referring to?

docs/Authentication.md
scottrhoyt: Let me know if the update is clearer.

Sources/Moya/Plugins/AccessTokenPlugin.swift
scottrhoyt: Done.

docs/Authentication.md
BasThomas: I don't see how this would work at the call site. According to the tests you wrote, it seems like you add the `needsAuthentication` in your `TargetType` init:

```swift
let target = TestTarget(shouldAuthorize: true)
```

BasThomas: What was trowing me off-guard as well, was your example of using `case needsAuthorization` and `case doesntNeedAuthorization`, although these are obviously two endpoints. Don't know why I didn't see that; that was what seemed to "duplicate" / boilerplate, but obviously doesn't.

Does that make any sense? 😆

scottrhoyt: `shouldAuthorize` is implemented in the same way as `TargetType.validate` or `TargetType.parameters`. Normally you'd implement it as a switch on the `enum`. In my trivial test case, I made a `struct` that conforms to `TargetType` and `AccessTokenAuthorizable` and it takes an initialization parameter to indicate how is should respond to `shouldAuthenticate`. This is just to write more concise tests. This isn't the way you'd normally use it. In fact I think most people will be fine with the default behavior and not conform to `AccessTokenAuthorizable`. Does that make sense to you?

scottrhoyt: I changed the endpoints in the example to make that more clear as well.

BasThomas: Ah, that makes total sense. 👍

BasThomas: `targetThatNeedsAuthorization` should be `endpointThatNeedsAuthorization`, right?

scottrhoyt: I just used the name that was already present in the docs under "Basic HTTP Authentication". Do you think both should be renamed?

scottrhoyt: I think target makes sense though, because it is an instance of a `TargetType` and not of an `Endpoint`.

BasThomas: 👍

rlam3:
This looks great. But how would one handle the access token when it is expired in the request? Would there be a way for us to set a specific url to refresh the token?

PR #899, opened 2017-01-04
Add documentation. #trivial

scottrhoyt:
This adds some documentation around `Endpoint`, `Image`, `Response`, and `MultiTarget`.

### DIFF
--- a/Sources/Moya/Response.swift
+++ b/Sources/Moya/Response.swift
@@ -34,7 +34,13 @@ public final class Response: CustomDebugStringConvertible, Equatable {
 
 public extension Response {
 
-    /// Filters out responses that don't fall within the given range, generating errors when others are encountered.
+    /**
+     Returns the `Response` if the `statusCode` falls within the specified range.
+
+     - parameters:
+        - statusCodes: The range of acceptible status codes.
+     - throws: `MoyaError.statusCode` when others are encountered.
+    */
     public func filter(statusCodes: ClosedRange<Int>) throws -> Response {
         guard statusCodes.contains(statusCode) else {
             throw MoyaError.statusCode(self)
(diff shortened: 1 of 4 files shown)


Sources/Moya/Response.swift
```diff
@@ -38,14 +42,17 @@ public extension Response {
         return self
     }
 
+    /// Filters out responses of a specific status code, generating errors when others are encountered.
```
BasThomas: These just return `an Error`, right? Not multiple. Might want to make that more clear.

BasThomas: And maybe make it "Filters out the responses of the specified status code [..]"

scottrhoyt: Yeah, I was just going off of the preexisting documentation for consistency. But I will change that as well.

BasThomas:
We might want to unify the header docs at one point (only `///` or your approach with `\**`), but lets leave that for another time.

scottrhoyt:
Thus far my approach is to use `///` for single line documentation comments and `/**` for multiline documentation comments, but I agree we should make sure it is consistent. I was having trouble getting the markdown (e.g. `- parameters:` to correctly register on multiline with `///`.

BasThomas:
You might have to use `- parameter name: name does thing` per parameter with `///`.

scottrhoyt:
Ah, that might be it then. In that case, I think:

```swift
/**
 function

 - parameters:
    - x: Is an x
    - y: Is a y
*/
```

might be a touch more human readable than

```swift
/// function
///
/// parameter x: Is an x
/// parameter y: Is a y
```

But that's just personal preference. I could go either way.

BasThomas:
I agree, but I think I like `///` above `/**`. 😛

PR #901, opened 2017-01-04
Update Readme and Demo's Podfile.lock for 8.0.0

BasThomas:


Readme.md
```diff
@@ -112,7 +112,7 @@ pod 'Moya/RxSwift'
 # or
 
 pod 'Moya/ReactiveSwift'
-pod 'ReactiveSwift', '1.0.0-alpha.4'
+pod 'ReactiveSwift', '1.0.0'
```
BasThomas: Is this step still needed? If so, why?

ashfurrow: It is not, good catch.

PR #808, opened 2016-11-29
Add a Danger rule to check for (simultaneous) changes in CocoaPods / Carthage files

BasThomas:
Fixes #796.

Can we somehow check if this works as intended? I couldn't really find a way to do so.

PR #902, opened 2017-01-04
Updates for SPM.

ashfurrow:
I got some feedback about our use of SPM, so I'm updating our package description and readme to reflect those changes. I might not have interpreted the feedback correctly, so feedback would be 👍 Specifically:

- SPM ignores directories that start with `.`.
- Our sources are stored in `Sources` so we don't need to exclude `Demo` explicitly.
- `.Target(name: "Moya")` can be replaced with `"Moya"` for succinctness. This is a stylistic choice that's up for discussion.

Also updated our Swift version support matrix to reflect 8.0's final release (🎉)

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -85,7 +85,9 @@ let package = Package(
       dependencies: ["Moya"]
     )
   ],
-  dependencies: ["Moya"]
+  dependencies: [
+    .Package(url: "https://github.com/Moya/Moya", majorVersion: 8)
+  ]
 )
 ```
 
--- a/Readme.md
+++ b/Readme.md
@@ -79,12 +79,6 @@ import PackageDescription
 
 let package = Package(
   name: "MyApp",
-  targets: [
-    Target(
-      name: "MyApp",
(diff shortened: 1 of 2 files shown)


Readme.md
```diff
@@ -82,7 +82,7 @@ let package = Package(
   targets: [
     Target(
```
scottrhoyt: For a single target project, I think we can get even simpler with our recommended SPM install and do this:

```swift
import PackageDescription

let package = Package(
  name: "MyApp",
  dependencies: [
    .Package(url: "https://github.com/Moya/Moya", majorVersion: 8)
  ]
)
```

That should be all the majority of people need. You don't need `Target`s unless you have a multi-target package (maybe that will be more common when you can use SPM to build Apps). Also the dependencies on a `Target` level are usually used to specify inter-Target dependencies and not external package dependencies like Moya.

BasThomas:
Hmm, we might want to add a Danger rule to keep CP+Carthage+SPM in sync (as we did in #808).

PR #905, opened 2017-01-04
Providers that are deallocated never completes

sadiq81:
Added a check to ensure that the provider is not deallocated, if it is then it sends OnError event.

### DIFF
--- a/Sources/RxMoya/RxMoyaProvider.swift
+++ b/Sources/RxMoya/RxMoyaProvider.swift
@@ -21,10 +21,10 @@ open class RxMoyaProvider<Target>: MoyaProvider<Target> where Target: TargetType
 
         // Creates an observable that starts a request each time it's subscribed to.
         return Observable.create { [weak self] observer in
-            guard let self = self else {
+            guard let strongSelf = self else {
                 observer.onError(RxError.disposed(object: self))
             }
-            let cancellableToken = self?.request(token) { result in
+            let cancellableToken = strongSelf.request(token) { result in
                 switch result {
                 case let .success(response):
                     observer.onNext(response)
@@ -35,7 +35,7 @@ open class RxMoyaProvider<Target>: MoyaProvider<Target> where Target: TargetType
             }
 
             return Disposables.create {
-                cancellableToken?.cancel()
+                cancellableToken.cancel()
             }
         }
     }
--- a/Sources/RxMoya/RxMoyaProvider.swift
+++ b/Sources/RxMoya/RxMoyaProvider.swift
@@ -20,11 +20,8 @@ open class RxMoyaProvider<Target>: MoyaProvider<Target> where Target: TargetType
     open func request(_ token: Target) -> Observable<Response> {
 
         // Creates an observable that starts a request each time it's subscribed to.
-        return Observable.create { [weak self] observer in
-            guard let strongSelf = self else {
-                observer.onError(RxError.disposed(object: self?))
-            }
-            let cancellableToken = strongSelf.request(token) { result in
+        return Observable.create { observer in
+            let cancellableToken = self.request(token) { result in
                 switch result {
                 case let .success(response):
                     observer.onNext(response)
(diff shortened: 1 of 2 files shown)


sadiq81:
Some discussion on the problem http://stackoverflow.com/questions/41433337/observable-returned-from-function-never-sends-onnext/41474121#41474121

scottrhoyt:
Hmm. That's interesting. I'm thinking about what should be the correct behavior here.

Maybe the Observable should strongly capture the `RxMoyaProvider`? It seems like the ability to make the request should be tied to the lifecycle of the `Observable`, not of the provider. Any other thoughts @Moya/contributors?

ashfurrow:
This PR makes sense to me, we should probably have been doing this from the start 👍 @sadiq81 could you add a changelog entry for us? Thanks for the PR!

Sources/RxMoya/RxMoyaProvider.swift
```diff
@@ -21,7 +21,10 @@ open class RxMoyaProvider<Target>: MoyaProvider<Target> where Target: TargetType
 
         // Creates an observable that starts a request each time it's subscribed to.
         return Observable.create { [weak self] observer in
```
scottrhoyt: What about killing the `[weak self]` here and just strongly capturing. Then we can kill the `guard`. There doesn't appear to be a risk of a retain cycle here, so the only other reason to do it this way would be if we *want* to use the release of an `RxMoyaProvider` to error out observables it has created. I think we want those `Observable`s to be valid until they are disposed, right?

sadiq81: I think that would be the expected behavior, i produced the provider inside a method since i only needed it once. I can see someone else also doing the same thing.

scottrhoyt: What do you think @ashfurrow? I think you did a lot of the early work here. What makes more sense to you: Having the `Observable` be usable regardless of what happens with the `RxMoyaProvider` that generated it or emitting an error if that provider gets released? Maybe we can also think about how that would be consistent with behavior of `MoyaProvider`. If you release a `MoyaProvider` will inflight requests still be finished?

Also, to note, if we adopt the pattern of emitting an error on release, I think users might still be able to work around this by retaining the provider inside of a closure retained by the `Observable` (e.g. `do` or `subscribe`). A bit hacky though...

ashfurrow: I'm leaning towards having the provider stay weakly captured and producing an error if its been deallocated. I don't think it's the responsibility of the `request()` function to retain the provider, that falls on whoever is using it. And if the provider gets deallocated before the request completes, an error makes sense. As you said, users can work around this if they need to, so it provides the most flexibility.

scottrhoyt: Here is another potential solution: We could support both behaviors.

In the instance method `request` we can weakly retain the provider assuming the user is handling that. We will emit the error if it was deallocated.

Then we can create a class method on `RxMoyaProvider` that will retain the provider in the request. Then we can still support workflows where users want ad-hoc requests out of the box. Something like:

```swift
RxMoyaProvider<MyApi>.request(/* ... */)
```

scottrhoyt: That second method would mean we'd need to take all the initializer args to `RxMoyaProvider` in the class method. OR maybe we make this an instance method too of a different name?

I suppose both of these things can be left as exercises for the user though. Do we know what happens to an inflight request if a regular `MoyaProvider` is deallocated?

ashfurrow: I can't recall, could someone check in on that and get back to us? Really swamped right now 😬

ashfurrow:
@sadiq81 Thanks a lot for the pull request! Would you mind adding a note to the changelog?

sadiq81:
Sure, how?

ashfurrow:
@sadiq81 In your master branch, add a line to `Changelog.md` at the top that describes the bug you fixed. Commit it and push it up to add it to this PR 👍

AndrewSB:
I think I agree with @scottrhoyt, it doesn't seem best practice to strongly capture the provider when making a request.
The provider should be held on by some other object, and the response should only be acknowledged if the provider is still listening for it.

This may be sort of nitpicky though. Merging this would make the common case simpler (why is my request never responding), but could potentially create memory leaks for more complicated implementations (I set my `provider = nil`, but I can see that it hasn't been deallocated, wut)

ashfurrow:
Aye. I'm leaning towards merging this, the common case makes sense to optimize for. Let's give it one more day to discuss objections, but I think we've covered everything now.

ashfurrow:

PR #907, opened 2017-01-06
Update Carthage install instructions

bjarkehs:
With #563 it changes how Moya is used with Carthage. It seems that the importing conflicts are resolved, but the current wording can also be confusing, since it seems to suggest you only link either `RxMoya` or `ReactiveMoya` if you use either of these.

Granted I can have misunderstood how Moya should be used with Carthage.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -114,9 +114,8 @@ Then run `pod install`.
 
 Carthage users can point to this repository and use whichever
 generated framework they'd like, `Moya`, `RxMoya`, or `ReactiveMoya`.
-The full Moya framework is bundled in each of those frameworks;
-importing more than one framework in a single file will result in
-ambiguous lookups at compile time.
+When linking frameworks remember to always link `Moya` along with `RxMoya` and 
+`ReactiveMoya`, if you use any of these.
 
 ```
 github "Moya/Moya"


BasThomas:
Hmm, this means we should use:

```
github "Moya/Moya"
```

for Moya,

```
github "Moya/Moya"
github "Moya/RxMoya"
```

for RxMoya, and

```
github "Moya/Moya"
github "Moya/ReactiveMoya"
```

.. right? If that's the case, we might want to add these examples as we did with CocoaPods, to make it even more clear.

bjarkehs:
That would require having `ReactiveMoya` and `RxMoya` being located at those location and not in `Moya/Moya`.

I do think there's something wrong with the current setup, maybe not this particular thing, but I am experiencing build issues on a CI. I should probably create an issue on that.

scottrhoyt:
@BasThomas, in order to do that we would need to pull those out in to separate repos.

ashfurrow:
We've had that discussion before, and we can revisit it. The argument for keeping them all in one place is that it's far easier to maintain. But I'm not doing much maintenance nowadays, y'all are, so we can split them up if that's what the team thinks is right.

scottrhoyt:
It's not an bad idea. At least from the perspective that `RxSwift` and `ReactiveSwift` are heavyweights and allowing our carthage users the choice of whether or not to include them in their dependency graph could be nice. It does create a whole set of maintenance headaches though. 😬

ashfurrow:
Aye, that's the tradeoff: it's easier for Carthage users but harder for us. Moya's benefited from a tonne of community contributions, and the easier we can make that the better. But I don't like shifting the burden to users either. This could be something to discuss in a Hangout or Skype call, let me know and I can organize it.

scottrhoyt:
I think I'm partial to leaving them as is for the time being. A lot has changed in the structure of this project and we just release 8.0.0. This would mean 9.0.0 right away, and more maintenance. Let's hope that Rx and Reactive continue to provide reliable framework downloads and that we get ABI stability sooner rather than later to ease the burden on the users.

I'm open to changing that opinion though and would be good to do a call as well.

justinmakaila:
As a Carthage user, I'm very fond of switching into two separate repos, but as @ashfurrow mentioned, there's an impending shift towards the native package manager. Any work spent breaking up the repos is likely to conclude with SPM becoming widely adopted.

I'm in favor of keeping it as is, unless there was some tricky configuring we could do with `Cartfile`s in separate repositories.

i.e.:
Moya/ReactiveMoya hosts a repo with a Cartfile that specifies reliance on a specific branch of Moya, which includes a `Cartfile` for those dependencies.

Cartfile in Moya/ReactiveMoya
```
github "Moya/Moya" "reactive-moya"
```

Cartfile in Moya/Moya on "reactive-mota"
```
github "ReactiveCocoa/ReactiveSwift" ~> 1.0
/// ... The remainder of Moya's dependencies
```

scottrhoyt:
Hmm. With regards to the move towards SPM, I just want to make sure we don't put the cart before the horse. There are currently severe limitations to using SPM in building an App for Apple platforms. From what I've read, the story *may* be different come Swift 4.0 in late 2017. Am I missing something? Because it seems like we're the better part of a year and 100,000s of collective builds away from being able to consider being SPM-first. Is there some news I'm not aware of in terms of the SPM roadmap?

BTW, I am guilty of hijacking this PR. Which looks good to me!  😉

justinmakaila:
No, we're on the same page. This has been on the table for the better part of a year and a half while I've been contributing.

The only real pain point comes from waiting on Carthage to build dependencies that aren't in use. That can be mitigated (in part) by distributing binary frameworks, but that's always dependent on the build system used at the point of distribution (nothing like waiting 30 minutes for your dependencies to build only to have it fail because a binary distribution was compiled with Swift 3.0 instead of Swift 3.0.2).

it's a tough problem, because it's only relevant to our Carthage users. CocoaPods users have had this ability pretty much since the libraries inception. Any changes that are made for Carthage users can't effect CP users, I learned that the hard way when I came on. One of my first in-depth PRs broke CP support while attempting to address this issue.

We totally hijacked the thread. I think the PR looks good, and is required instruction. I'm good to merge in, and we can open an issue to further this discussion... in fact, I think there already is one somewhere...

ashfurrow:

BasThomas:
I *am* open for providing multiple repo's to make the users' lives easier. I think SPM will still take a while to catch on, and in the meantime, I think we should consider this. Would gladly join a call to discuss the implications.

AndrewSB:
I feel like keeping them together is the easiest thing to do right now. I've felt bullish on SPM for over 6 months now, but I do know using it is still a ways out.

I'm down for a call as well, but I think we should let this sit until someone else brings it up (unless one of you feel very strongly about making the change to multiple repos). Hopefully then we'll have a more informed perspective on the possibilities of SPM

Issue #830, opened 2016-12-08
A MoyaProvider variant which does not have a generic <Target> parameter?


AndrewSB:
@manas-chaudhari mentioned something interesting here: https://github.com/Moya/Moya/issues/823#issuecomment-265458718

Let's use this issue to discuss whether we can implement what he was trying to do 😄

AndrewSB:
closing this for now. @manas-chaudhari: please do re-open this if you're interested in getting this added to Moya

manas-chaudhari:
Yes, I am interested in getting this added to Moya. Let me summarize the points.

Current implementation of `MoyaProvider` has a template variable `<Target>` at the class level. Because of this, one instance of `MoyaProvider` can work with only one type of target. 

Using `MultiTarget`, we have created a wrapper so that a single instance can be used for different inner `TargetType`s. This works when there are no associated types involved in enums.

In order to extend Moya's functionality, it is sometimes necessary to use associated types by extending the `TargetType` protocol. One use case is to abstract deserialization, so that MoyaProvider can return an `Observable<ModelType>`, where `ModelType` is the associatedtype. This has been discussed in https://github.com/Moya/Moya/issues/823#issuecomment-265458718.

The idea is to build a `MoyaProvider` variant which does not have any template variable at the class level, but instead at the `request` method level. 

```func request<T: TargetType>(request: T, completion: Moya.Completion)```

This would allow basic usage and also allow Moya consumers to add extensions like:

```func requestDecoded<T: DecodableTargetType>(request: T, completion: (_ result: Result<T.ResultType, Moya.Error>) -> ())```
or simply
```func requestDecoded<T: DecodableTargetType>(request: T) -> Observable<T.ResultType>```

I don't have access to reopen the issue. @AndrewSB Can you do this.

AndrewSB:
Sounds interesting @manas-chaudhari, I'm having a little trouble understanding your example though, are you saying that if I wanted to request my user model, for example, I'd do `provider.request<UserModel>(.me, completion...)`? In my opinion, that doesn't add much more to `provider.request(.me).map(UserModel.init)`.

Would you like to make a PR to implement your idea? I think starting out with amending the docs would be the easiest for me to understand exactly what you're suggesting, and for the contributors to talk about (see https://github.com/Moya/Moya/pull/762 for an example). If we can get enough people saying they like your idea, we would be able to merge this in 😄

manas-chaudhari:
[...]

PR #910, opened 2017-01-09
[WIP] MoyaProvider variant without generic parameter

manas-chaudhari:
This is based on #830. I have updated only the docs to provide a usecase for why this is required.
I strongly think that this will enable end-users to extend Moya's functionality in different ways which are not possible right now. 

Would love to have feedback on whether this makes sense.
I am not sure where exactly this documentation belongs. Fundamentally, this feature provides a MoyaProvider that can be used with different targets. Hence, I added it to `MultiTarget`.

I would like to start working on the implementation as soon as this requirement is validated.

### DIFF
--- a/docs/Examples/MultiTarget.md
+++ b/docs/Examples/MultiTarget.md
@@ -35,3 +35,98 @@ want to see this API in action, check out our
 targets: one of them is `Demo`, which uses the basic form of Moya, and the
 second one is `DemoMultiTarget`, which uses the modified version with usage of
 `MultiTarget`.
+
+
+## Multiple targets when using `associatedtype`
+
+Using Moya enables you to statically verify the arguments when invoking a
+network request. You might want to extend Moya's `TargetType` to verify your
+custom types. One use case is to have the `request` method return deserialized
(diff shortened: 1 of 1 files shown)


manas-chaudhari:
My opinion is that if the end-user prefers strong types over grouping, he should have that choice.

Earlier, I was expecting `MultiMoyaProvider` implementation to get complicated, hence I wanted this to be included in the core.
However, I tried to implement this and it has turned out to be very simple. Thus, with respect to implementation, I am fine with building it as an extension.

```swift 
class MultiMoyaProvider: MoyaProvider<MultiTarget> { }
```

The key deliverable here is the documentation. My only concern now is whether an extension which contains only documentation makes sense.

ashfurrow:
> The key deliverable here is the documentation.

Yup, this PR is only docs, which can only really help. I'm going to merge and we can refine later if we add first-class support into the library. Thanks again!

ashfurrow:

AndrewSB:
Cool! I'm happy the community saw value in this. @manas-chaudhari, we'll look for your next PR which implements the documentation we've added 😄

PR #914, opened 2017-01-11
Fixes CI

ashfurrow:


scottrhoyt:
@ashfurrow I picked up a nice script the other day from some fastlane discussion that launches the latest SDK on a device simulator by name instead of UUID. We've seen a decrease in failed builds since putting it in place. Let me know if you think something like this would be useful for us here:

```bash
#!/bin/bash

if [ "$1" = "" ]; then
    echo "Usage: launch-ios-simulator.sh <device name>"
    exit 1;
fi

device="$1"
printf "Closing any open instances of the iphone simulator...\n"
killall "Simulator" || true
printf "Determining latest iOS simulator...\n"
latest_ios=$(xcodebuild -showsdks | grep -Eo "iphonesimulator(.+)" | tail -1)
latest_ios=${latest_ios##iphonesimulator}
printf "Detected latest iOS simulator version: ${latest_ios}\n"
printf "Pre-Launching iphone simulator for ${device} (${latest_ios})\n"
simulator_id=$(xcrun instruments -s | grep -Eo "${device} \(${latest_ios}\) \[.*\]" | grep -Eo "\[.*\]" | sed "s/^\[\(.*\)\]$/\1/")
open -b com.apple.iphonesimulator --args -CurrentDeviceUDID $simulator_id

RETVALUE=$?
if [ "$RETVALUE" != "0" ]; then
   printf "Something went wrong when attempting to launch the simulator for ${device} (${latest_ios})\n"
   exit 1;
fi

printf "Simulator launched for ${device} (${latest_ios})\n"
```

ashfurrow:
@scottrhoyt Neat! Looks like it could easily be adapted into a Rake task, do you think that makes sense? I can take a stab at that later today.

ashfurrow:
Actually, it might make more sense to merge this for now and open an issue to import that into the Rake task later, pretty busy this afternoon 😬 Could you open that issue for us @scottrhoyt ?

scottrhoyt:
Yep!

Issue #915, opened 2017-01-11
Create Rake Task To Determine Simulator UUIDs


scottrhoyt:
Hard coding simulator UUIDs has been a consistent source of intermittent CI failures. Perhaps we can introduce a rake task or shell script to dynamically look up UUIDs and launch the simulator if needed. Here is a script that we picked up from some helpful fastlane users that is a good reference for this:

```bash
#!/bin/bash

if [ "$1" = "" ]; then
    echo "Usage: launch-ios-simulator.sh <device name>"
    exit 1;
fi

device="$1"
printf "Closing any open instances of the iphone simulator...\n"
killall "Simulator" || true
printf "Determining latest iOS simulator...\n"
latest_ios=$(xcodebuild -showsdks | grep -Eo "iphonesimulator(.+)" | tail -1)
latest_ios=${latest_ios##iphonesimulator}
printf "Detected latest iOS simulator version: ${latest_ios}\n"
printf "Pre-Launching iphone simulator for ${device} (${latest_ios})\n"
simulator_id=$(xcrun instruments -s | grep -Eo "${device} \(${latest_ios}\) \[.*\]" | grep -Eo "\[.*\]" | sed "s/^\[\(.*\)\]$/\1/")
open -b com.apple.iphonesimulator --args -CurrentDeviceUDID $simulator_id

RETVALUE=$?
if [ "$RETVALUE" != "0" ]; then
   printf "Something went wrong when attempting to launch the simulator for ${device} (${latest_ios})\n"
   exit 1;
fi

printf "Simulator launched for ${device} (${latest_ios})\n"
```

BasThomas:
References #914

scottrhoyt:
I don't think this is relevant anymore after #922

PR #922, opened 2017-01-16
Replaces usage of device UUID with device name

### LINKED ISSUE #757: Revert CI change once Circle is fixed
Revert https://github.com/Moya/Moya/pull/753 once Circle has fixed its duplicate simulators issue.

pedrovereza:
Fixes #757

The issue on Circle CI [was fixed](https://discuss.circleci.com/t/os-x-build-image-update-xcode-8-2/9355) :tada:

I had to make a few changes to include not only the device name but also the OS version. Not sure if I took the best approach, suggestions are welcome :wink:

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -46,26 +46,31 @@ def sdks
   }
 end
 
-# See: https://discuss.circleci.com/t/multiple-ios-simulators-available-for-single-os-device-configurations/7854
 def devices
   return {
-    ios: ENV['CIRCLECI'] ? "id='00E25E25-4920-42E1-8CC6-360AEF34C5CE'" : "name='iPhone 6s'",
-    macos: "arch='x86_64'",
-    tvos: ENV['CIRCLECI'] ? "id='31F99277-CC9E-409E-A257-7D1907DECB03'" : "name='Apple TV 1080p'"
+    ios: "OS=#{device_os[:ios]},name=#{device_names[:ios]}",
+    macos: "arch=x86_64",
+    tvos: "OS=#{device_os[:tvos]},name=#{device_names[:tvos]}"
   }
 end
 
-# See: https://discuss.circleci.com/t/xcode-exit-code-65/4284/13
-def uuids
+def device_names
   return {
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -46,35 +46,40 @@ def sdks
...
+    tvos: "Apple TV 1080p"
   }
 end
 
-def open_simulator_and_sleep(uuid)
-  return if uuid.nil? # Don't need a sleep on macOS because it runs first.
-  sh "xcrun instruments -w '#{uuid}' || sleep 15"
+def device_os
+  return {
+    ios: "10.2",
```
scottrhoyt: Unless we want to ensure compatibility with specific version, is there an option to do OS `latest`?

pedrovereza: Oh, I didn't know about `latest`. Seems to be a good choice for the `xcodebuild` call.
Is there something similar for `instruments -w`? 

Actually: Do we really need the call to Instruments? 🤔

scottrhoyt: You can check #915 for an idea on how to launch the latest iOS simulator. Not sure if we still need the instruments call.

scottrhoyt:
Another idea for how to make this run a little smoother is to swap out the `build test` [here](https://github.com/Moya/Moya/pull/922/files#diff-52c976fc38ed2b4e3b1192f8a8e24cffR117) for a `build-for-testing test`. I have noticed better results that way as it delays the call to open the simulator until the test target has already been built. Therefore you are less likely to have a timeout issue.

Rakefile
pedrovereza: I tried using `latest` in another branch just for testing, didn't work on Circle CI:
https://circleci.com/gh/Moya/Moya/1191

Did I miss something?

petester42:
For iPhones using the name might not always work. If the simulator has a watch tied to it the OS and the name will not be unique. You can have more than one. I'm guessing this works for iPhone 6s because the watches are now tied to an iPhone 7 simulator in circle. 

Let's just keep that in mind if something breaks in the future.

PR #930, opened 2017-01-19
Added a progress parameter to the reactive providers 

AvdLee:
This pull requests adds a progress parameter to the reactive providers to track progress changes of upload & download tasks. Currently it's impossible to track these changes and to use the same Moya provider for download & upload tasks.

### DIFF
--- a/Sources/RxMoya/RxMoyaProvider.swift
+++ b/Sources/RxMoya/RxMoyaProvider.swift
@@ -21,8 +21,7 @@ open class RxMoyaProvider<Target>: MoyaProvider<Target> where Target: TargetType
 
         // Creates an observable that starts a request each time it's subscribed to.
         return Observable.create { observer in
-            let cancellableToken = self?.request(token, queue: nil, progress: progress, completion:
-                { (result) in
+            let cancellableToken = self.request(token, queue: nil, progress: progress, completion: { (result) in
                 switch result {
                 case let .success(response):
                     observer.onNext(response)
--- a/Sources/RxMoya/RxMoyaProvider.swift
+++ b/Sources/RxMoya/RxMoyaProvider.swift
@@ -17,11 +17,12 @@ open class RxMoyaProvider<Target>: MoyaProvider<Target> where Target: TargetType
     }
 
     /// Designated request-making method.
-    open func request(_ token: Target) -> Observable<Response> {
+    open func request(_ token: Target, progress: ProgressBlock? = nil) -> Observable<Response> {
 
         // Creates an observable that starts a request each time it's subscribed to.
         return Observable.create { observer in
-            let cancellableToken = self.request(token) { result in
+            let cancellableToken = self?.request(token, queue: nil, progress: progress, completion:
+                { (result) in
                 switch result {
                 case let .success(response):
                     observer.onNext(response)
@@ -29,7 +30,7 @@ open class RxMoyaProvider<Target>: MoyaProvider<Target> where Target: TargetType
                 case let .failure(error):
                     observer.onError(error)
                 }
-            }
+            })
 
             return Disposables.create {
                 cancellableToken.cancel()
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,7 @@
 # Next
 
 - Fixed an issue where `RxMoyaProvider` never sends `next` or errors if it's disposed before a subscription is made.
+- Added a progress parameter to the reactive providers to track progress changes of upload & download tasks.
 
 # 8.0.0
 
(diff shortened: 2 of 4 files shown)


Sources/RxMoya/RxMoyaProvider.swift
```diff
@@ -17,19 +17,20 @@ open class RxMoyaProvider<Target>: MoyaProvider<Target> where Target: TargetType
     }
 
     /// Designated request-making method.
-    open func request(_ token: Target) -> Observable<Response> {
+    open func request(_ token: Target, progress: ProgressBlock? = nil) -> Observable<Response> {
 
         // Creates an observable that starts a request each time it's subscribed to.
         return Observable.create { observer in
-            let cancellableToken = self.request(token) { result in
+            let cancellableToken = self?.request(token, queue: nil, progress: progress, completion:
```
pedrovereza: This should be `self.request`:wink:

AvdLee:
Ah indeed, I've fixed that. Also updated for the SwiftLint issue.

BasThomas:
Two tests seem to be failing:
```
a_subsclassed_reactive_provider_that_tracks_cancellation_with_delayed_stubs__cancels_network_request_when_subscription_is_cancelled, expected to be true, got <false>

provider_with_SignalProducer__a_subsclassed_reactive_provider_that_tracks_cancellation_with_delayed_stubs__cancels_network_request_when_subscription_is_cancelled, expected to be true, got <false>
```

AvdLee:
Hmm, you're right. It's strange though, as it's executing exactly the same request method, only with the progress parameter added.

I'll look into it further tomorrow!

AvdLee:
Ok, I've fixed the tests, everything seems to be OK now. The only thing I want to discuss here, is that another solution is already available for progress tracking:

```swift
func requestWithProgress(token: Target) -> SignalProducer<ProgressResponse, MoyaError>
```

However, this doesn't really fit in our JSON parsing functionalities and needs some more code to write on our side to make it easy to track progress.

With this pull request we're introducing another way of tracking progress, which is quite more easy to use and feels simpler.

How do you guys think about this?

ashfurrow:
The implementation in the PR makes sense to me.

scottrhoyt:
I am all for any suggestions to make the progress handling more straight forward in the reactive providers. However, I am not sure this is the right direction to take. A goal of the observer pattern is to replace the callback pattern. So, to have our reactive providers use elements of both the observer pattern and the callback pattern seems muddled to me. This would encourage the propagation of both observables and callbacks in the users' code.

Perhaps we could refine the current approach with `ProgressResponse`? Instead of it being a `struct` with a `Response?` and a `Progress?`, it could be:

```swift
enum ProgressResponse {
  case progress(double)
  case response(response)
}
```

That would simplify handling of progress/response emissions. In fact, we could go further and provide reactive operators like `onlyProgress()` and `onlyResponse()` that could flatmap the `Observable<ProgressResponse>` into a `Observable<Double>` and `Observable<Response>`, respectively. This would go further to support separating the flow of progress and response without needing to introduce inconsistent patterns.

leoneparise:
We already have a `ProgressResponse`

```swift
public struct ProgressResponse {
    public let response: Response?
    public let progressObject: Progress?

    public init(progress: Progress? = nil, response: Response? = nil) {
        self.progressObject = progress
        self.response = response
    }

    public var progress: Double {
        return progressObject?.fractionCompleted ?? 1.0
    }

    public var completed: Bool {
        return progress == 1.0 && response != nil
    }
}
```

a method called `requestWithProgress` for both reactive extensions and two Observable operators to filter progress and response.

```swift
extension ObservableType where E == ProgressResponse {
    /**
     Filter completed progress response and maps to actual response
     - returns: response associated with ProgressResponse object
     */
    public func filterCompleted() -> Observable<Response> {
        return self
            .filter { $0.completed }
            .flatMap { progress -> Observable<Response> in
                // Just a formatlity to satisfy the compiler (completed progresses have responses).
                switch progress.response {
                case .some(let response): return .just(response)
                case .none: return .empty()
                }
            }
    }

    /**
     Filter progress events of current ProgressResponse
     - returns: observable of progress events
     */
    public func filterProgress() -> Observable<Double> {
        return self.filter { !$0.completed }.map { $0.progress }
    }
}
```

I don't think this PR will add anything that our current implementation is not handling.

scottrhoyt:
Oops @leoneparise. I totally missed that we had the `filterCompleted()` and `filterProgress()` operators already. Thanks for pointing that out! I'm not sure the naming is unambiguous though (e.g. RxOptional's `filterNil()` takes optionals out of the observable, we follow the opposite semantics).

leoneparise:
No problem @scottrhoyt! Yeah, we can change the method names like `progress` and `response` to be more readable. But, in my head, this PR only adds complexity. A progress closure inside a observable? I can't agree with that.

AvdLee:
Good discussion, I agree with the above. We'll look further into implementing it using the suggested methods. Thanks!

Issue #792, opened 2016-11-24
Document release process


ashfurrow:
We release using a [rake task](https://github.com/Moya/Moya/blob/b0fe5de3e4cfefa20d1dd8c26ccfa8bb0ba4184e/Rakefile#L135) that's invoked using `rake release["VERSION"]`. But there are some dependencies, like having bundler installed, running a `bundle install`, and using a `netrc` file to authenticate with GitHub. 

@AvdLee Let me know what troubles you run into.

AvdLee:
So in this case I was releasing `8.0.0-beta.5` using the `rake release` method.

Everything went fine, I only had to do a `bundle install` as noted before I could use the `rake` method. 

The podspec however didn't update automatically, so the `pod trunk push` noted a duplicate entry of `beta 4`. Furthermore, I encountered an error with the beta release of `ReactiveSwift`, so the podspec didn't validate as well.

After updating the podspec manually I was able to create a new pod release, but it would be great if the rake file does a bump here as well. Not sure if this is possible with beta versioning?

ashfurrow:
Interesting, the Rakefile _should_ do that [here](https://github.com/Moya/Moya/blob/b0fe5de3e4cfefa20d1dd8c26ccfa8bb0ba4184e/Rakefile#L141-L145) but the regex won't work if the existing podspec is a prerelease (it looks for `"X.Y.Z"` and misses `"X.Y.Z-beta.a"`). 

I'll take a shot at a PR adding documentation steps to release, unless you beat me to it :wink:

BasThomas:
The [Changelog](https://github.com/Moya/Moya/blob/master/Changelog.md) seems to not be completely correct right now either. I see beta 5 duplicated.

pedrovereza:
@ashfurrow It seems that we only got a tag created, I think we should make a release out of it:
![image](https://cloud.githubusercontent.com/assets/1916041/20718051/ad265f54-b63e-11e6-95e0-70e7a2e9cc91.png)

ashfurrow:
Sounds good 👍 That can sometimes happen when the script fails near the end. Can you take care of that?

pedrovereza:
Sure, will do :wink:

BasThomas:
Updated the previous releases to include the changes from `Changelog.md` as well. 👍

pedrovereza:
[...]

PR #941, opened 2017-01-31
Updates regex to accept X.Y.Z-beta.W versions

### LINKED ISSUE #811: Rake release not including Changelog notes
Found in #792.

It seems that `rake release` is not including Changelog notes in the releases, even though [it should be](https://github.com/Moya/Moya/blob/b0fe5de3e4cfefa20d1dd8c26ccfa8bb0ba4184e/Rakefile#L174).

We only noticed this in versions `8.0.0-beta.x`.

pedrovereza:
Fixes #811 

@ashfurrow [gave us the answer](https://github.com/Moya/Moya/issues/792#issuecomment-262812613) in #792: the semver regex should accept an optional `-beta.a`.

I didn't hardcode it `beta` to give us the flexibility to go with `alpha`, `rc` or whatever makes sense in future versions.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -157,12 +157,12 @@ desc 'Release a version, specified as an argument.'
 task :release, :version do |task, args|
   version = args[:version]
   # Needs a X.Y.Z-text format.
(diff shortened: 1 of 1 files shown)


BasThomas:
Nice! 🙌

Issue #946, opened 2017-02-01
Release 8.0.2


pedrovereza:
Latest changes in how we pin dependency versions fixes #942 (see #944).

This change is probably useful for a lot of projects, which justifies releasing a new patch.

pedrovereza:
And it's [done](https://github.com/Moya/Moya/releases/tag/8.0.2) 🙌

BasThomas:
It's pretty ironic we're saying this now in the README:

> For Moya, use the following entry in your Podfile:
>
>```ruby
>pod 'Moya', '8.0.2'
>```

Instead of 

```ruby
pod 'Moya', '~> 8.0'
```

😜

ashfurrow:
Good point @BasThomas. Technically we shouldn't be specifying anything at all and just let the `Podfile.lock` file do it's thing. Someone want to update?

PR #947, opened 2017-02-01
Update installation instructions (version pinning)

BasThomas:
As discussed in #946

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -90,22 +90,21 @@ let package = Package(
 For Moya, use the following entry in your Podfile:
 
 ```rb
-pod 'Moya', '8.0.2'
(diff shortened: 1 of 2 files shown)


BasThomas:
By the way, how do the Reactive extensions work with SwiftPM?

pedrovereza:
I think we can keep it as `pod 'Moya'` (we always want people to be using the latest version, I feel)

BasThomas:
Yeah, that might be even better. Does SwiftPM work like that as well?

BasThomas:
It seems like we need to at least provide a `majorVersion` [there](https://github.com/apple/swift-package-manager/blob/master/Documentation/Reference.md#package-dependency).

PR #954, opened 2017-02-07
Changes POST example in Basic.md to send parameters as JSON

pedrovereza:
The POST example in `Basic.md` currently sends the parameters encoded in the URL. I don't think this is the most common way of sending parameters in a POST request 🤔  

I believe the usual case is to send parameters in the request body as JSON. Having this in the most basic example may be more helpful to people using Moya for the first time.

### DIFF
--- a/docs/Examples/Basic.md
+++ b/docs/Examples/Basic.md
@@ -10,6 +10,7 @@ enum MyService {
     case zen
     case showUser(id: Int)
     case createUser(firstName: String, lastName: String)
+    case updateUser(id:Int, firstName: String, lastName: String)
     case showAccounts
 }
 ```
@@ -25,7 +26,7 @@ extension MyService: TargetType {
         switch self {
         case .zen:
             return "/zen"
-        case .showUser(let id):
+        case .showUser(let id), .updateUser(let id, _, _):
             return "/users/\(id)"
         case .createUser(_, _):
             return "/users"
@@ -37,7 +38,7 @@ extension MyService: TargetType {
         switch self {
         case .zen, .showUser, .showAccounts:
             return .get
-        case .createUser:
+        case .createUser, .updateUser:
             return .post
         }
     }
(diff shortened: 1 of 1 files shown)


vzsg:
Why not both?

pedrovereza:
@vzsg Good question! 

What are the most basic (and common) cases that we should cover in this example? We don't have a GET with URL query parameters as well 🤔

sunshinejr:
This example you've added is good, but we could also add the basic `URLEncoding` below/above the example with `JSONEncoding`, imho.

pedrovereza:
@sunshinejr You mean adding just a note on how the POST request would look like if `URLEncoding` was used instead?

sunshinejr:
I was thinking about doing one more request with different type of encoding, maybe just below the one with user, and then make a comment how would it look like. But your suggestion works as well, make a comment how could it look like with `URLEncoding`.

pedrovereza:
@sunshinejr Doing another request feels more complete 👍 

Added a new endpoint to update a user passing the parameters in the URL, let me know what you think (cc. @vzsg)

docs/Examples/Basic.md
```diff
@@ -25,32 +26,38 @@ extension MyService: TargetType {
...
-        case .createUser(let firstName, let lastName):
+        case .createUser(let firstName, let lastName), .updateUser(_, let firstName, let lastName):
             return ["first_name": firstName, "last_name": lastName]
         }
     }
+
     var parameterEncoding: ParameterEncoding {
-        return URLEncoding.default
+        switch self {
+        case .zen, .showUser, .showAccounts, .updateUser(_, _, _):
```
sunshinejr: This is a minor one, but I think we can strip these `(_, _, _)` when we don't need any of the parameters in there.

BasThomas: You can :)

pedrovereza: Yep! Good catch

pedrovereza:
Should be good now 😄

PR #957, opened 2017-02-10
Add explicit dependency on Result.

scottrhoyt:
We shouldn't rely on this being there from ReactiveSwift.


BasThomas:
We should add Result to SwiftPM as well.

Issue #950, opened 2017-02-03
MoyaProvider generic function for better completion handler usability when both the return type and method of deserialization are known


mikesanderson:
Moya encapsulates everything necessary to make a call, but on the return requires clients to parse and cast `Result<Moya.Response, MoyaError>` objects, which inevitably clutters the completion block and leaks network concerns into the client. From #948 as @scottrhoyt characterized (without endorsing) it, the goal is "Provide better completion handler usability for MoyaProvider when both the return type and method of deserialization are known".  

This can be accomplished adding the following (queue, progress parameters removed for clarity) generic function to MoyaProvider: 

```swift
/// Generic request-making method.
@discardableResult
open func request<T>(_ target: Target, parser: @escaping (Moya.Response) throws -> T, completion: @escaping (Result<T, MoyaError>) -> Void) -> Cancellable {
    return request(target, queue: queue, progress: progress) { (result) in
        do {
            let response = try result.dematerialize()
            let value = try parser(response)
            completion(Result(value: value))
        } catch let error as MoyaError {
            completion(Result(error: error))
        } catch {
            completion(Result(error: MoyaError.underlying(error)))
        }
    }
}
``` 
It's straightforward, it calls the parser on the response and returns the value (or any error) to the completion. The generic function enforces that the type returned by the parser matches the parameter of the completion. (Side note, this is a great example of why to avoid optional sentinels: `T` can be an optional where `nil` is valid value, so only `throws` indicates an error.) 

So the call site becomes: 
```swift
func downloadRepositories(_ username: String) {
    GitHubProvider.request(.userRepositories(username), parser: Parsers.parseNSArray) { result in
        switch result {
        case let .success(json):
            self.repos = json
            self.tableView.reloadData()
        case let .failure(error):
            guard let error = error as? CustomStringConvertible else {
                break
            }
            self.showAlert("GitHub Fetch", message: error.description)
        }
    }
}
```
It may seem like a small distinction, moving the parser from something invoked in the completion to something passed to the request method; the call site still needs to know how the response will be parsed. The main benefit is it declutters the completion blocks, removing any knowledge or handling of response objects. A similar approach would work with the reactive versions. It also seems more in line with the goals of encapsulation to handle these things upstream.

BasThomas:
Heads-up in the meantime - you can highlight the syntax in the code snippets with ```swift at the start. :)

scottrhoyt:
Interested to hear what other @Moya/contributors think about this.

[...]

PR #960, opened 2017-02-12
In Demo, change completion for GitHub.userRepositories to use dematerialize()

mike-sand:
This creates a much cleaner completion block by using the Result method `dematerialize()` in the same do-try-catch block as the response mapping. Adds a `mapNSArray()` extension to Response in GitHubAPI.swift to support this and as an example. Also an example of how to create a error message from an error with default.

See #950 for more discussion.


ashfurrow:

PR #963, opened 2017-02-13
Add ablity to use class func to initialize in subclass

kemchenj:
For some reason, I need to use both `MoyaProvider` and `RxMoyaProvider` in project

I try to use a class function to initialize `MoyaProvider` and `RxMoyaProvider` with default plugins

```swift
fileprivate let defaultPlugins = [PluginOne(), PluginTwo()]

extension MoyaProvider {
    // Return `Self` here because the generic thing
    // Also because I want to return `RxMoyaProvider` when using RxMoyaProvider
    class func withDefaultPlugins() -> Self {
        return Self.init(plugins: defaultPlugins)
    }
}
```

But the complier raise an exception `Constructing an object of class type 'Self' with a metatype value must use a 'required' initializer`

So I add make it `required`, get some trouble with `ReactiveSwiftMoyaProvider`, and I figure out a kind of dirty way to solve it. 

The only problem is initialize like this, will get some trouble(ambiguous use of `init`)

```swift
ReactiveSwiftMoyaProvider(endpointClosure: endpointClosure,
                          requestClosure: requestClosure,
                          stubClosure: stubClosure,
                          manager: manager,
                          plugins: plugins,
                          trackInflights: trackInflights)
```

Do you guys have some idea about this?

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -3,6 +3,7 @@
 # 8.0.2
 
 - Changed dependency pinning to rely only on major versions.
+- Add `required` to `MoyaProvider` initializer (Fix an issue when initialize `MoyaProvider` in class function.)
 
 # 8.0.1
 
@@ -399,3 +400,5 @@ structure and more.
 # 0.1
 
 - Initial release.
+
+


pedrovereza:
Hey, would something like this work for you?

```swift
extension MoyaProvider {
    class func withDefaultPlugins() -> MoyaProvider<Target> {
        return MoyaProvider<Target>(plugins: [NetworkLoggerPlugin(verbose: true)])
    }
}
```

Usage:

```swift
let provider = MoyaProvider<MyService>.withDefaultPlugins()
```

kemchenj:
@pedrovereza 

I've try this way in the beginning. But when you code like this:

```swift
let rxProvider = RxMoyaProvider<MyService>.withDefaultPlugins()
```

`rxProvider`'s type is `MoyaProvider<MyService>`, not `RxMoyaProvider<MyService>`, so I use `Self` as the type return in `withDefaultPlugins` method

pedrovereza:
@kemchenj Oh, I see.

So, would moving the extension to `RxMoyaProvider` be enough?
```swift
extension RxMoyaProvider {
    class func withDefaultPlugins() -> RxMoyaProvider<Target> {
        return RxMoyaProvider<Target>(plugins: [NetworkLoggerPlugin(verbose: true)])
    }
}
```
Or do you need `withDefaultPlugin` available for both `MoyaProvider` and `RxMoyaProvider`?

*Not saying that your changes are good/bad, just trying to understand the use case* 😄

kemchenj:
@pedrofjfmartins for both `MoyaProvider` and `RxMoyaProvider`,

Sources/ReactiveMoya/ReactiveSwiftMoyaProvider.swift
```diff
@@ -6,16 +6,36 @@ import Moya
...
-    private let stubScheduler: DateSchedulerProtocol?
+    private var stubScheduler: DateSchedulerProtocol?
     /// Initializes a reactive provider.
     public init(endpointClosure: @escaping EndpointClosure = MoyaProvider.defaultEndpointMapping,
                 requestClosure: @escaping RequestClosure = MoyaProvider.defaultRequestMapping,
                 stubClosure: @escaping StubClosure = MoyaProvider.neverStub,
                 manager: Manager = ReactiveSwiftMoyaProvider<Target>.defaultAlamofireManager(),
-                plugins: [PluginType] = [], stubScheduler: DateSchedulerProtocol? = nil,
+                plugins: [PluginType] = [],
+                stubScheduler: DateSchedulerProtocol? = nil,
```
sunshinejr: This one parameter would have been without default value. Otherwise we have similar declarations, because if you don't need to specify this parameter, you have 2 identical initializers in most of the cases. Thus:
```swift
stubScheduler: DateSchedulerProtocol? = nil
```
-> 
```swift
stubScheduler: DateSchedulerProtocol?
```

Sources/ReactiveMoya/ReactiveSwiftMoyaProvider.swift
```diff
@@ -6,16 +6,36 @@ import Moya
...
+                  trackInflights: trackInflights)
+    }
+
+    public required init(endpointClosure: @escaping EndpointClosure,
+                         requestClosure: @escaping RequestClosure,
+                         stubClosure: @escaping StubClosure,
+                         manager: Manager,
+                         plugins: [PluginType],
+                         trackInflights: Bool) {
+        super.init(endpointClosure: endpointClosure,
```
sunshinejr: you need to store `nil` to `stubScheduler` before `super.init()`

Cartfile.resolved
```diff
@@ -1,7 +1,7 @@
 github "Alamofire/Alamofire" "4.3.0"
 github "Quick/Nimble" "v6.0.1"
 github "AliSoftware/OHHTTPStubs" "57feceaabf333e72b2c637dfba6c13a7a5c49619"
-github "Quick/Quick" "v1.0.0"
```
sunshinejr: can we please remove this change from the PR?

kemchenj:
@sunshinejr Actually I think there is not need to subclass `MoyaProvider`. Instead of make init required, maybe just make `MoyaProvider` final.

Then let `MoyaProvider` conform to `ReactiveCompatible` and ``, then we could change our code like this:

```swift
let provider = MoyaProvider<GitHubAPI>()

// RxSwift
provider.rx.request(target)

// ReactiveCocoa
provider.reactive.request(target)
```

sunshinejr:
@kemchenj no problem, don't worry about that. About your idea with conforming to protocols: I love it. Would be up for doing the PR with this idea implemented? :)

About `final` on `MoyaProvider` - I'm not sure about that. People may rely on subclassing it.

BasThomas:
I also really like `.rx` and `.reactive`!

sunshinejr:
Okay, @BasThomas took a stab at it, I'm gonna close this one then. Thank you so much @kemchenj for you work, we really appreciate it! 🙇

Issue #818, opened 2016-12-01
Discuss: Use a single target for all platforms?


AndrewSB:
Hey all! I was reading https://blog.pusher.com/what-i-learned-building-the-pusherswift-framework/, and one of the first things the author goes into is how the framework they built uses only one target for all platforms. 

Currently Moya has 4 targets for each 3 of our frameworks (12 total), and then 3 for tests.
<img width="175" alt="screen shot 2016-12-01 at 2 39 57 pm" src="https://cloud.githubusercontent.com/assets/3814772/20815764/11cd4af8-b7d4-11e6-9d3c-7a1d801ddc48.png">

I'm not sure how the author was able to make a universal target, but maybe we could do something similar?

petester42:
I know how. I've done it for other projects. It's fairly simple. There are couple things not really mentioned but this article covers mostly how to do this http://promisekit.org/news/2016/08/Multiplatform-Single-Scheme-Xcode-Projects/.

AndrewSB:
@petester42 that looks really cool, it would be awesome to have Moya vend through one target as well.
I could take a shot in a PR, but since you've done this before it may be easier for you to do it? Let me know what you think.

petester42:
Ya sure I could do it. I was creating a script that would automatically do this but I haven't finished it. I could do it manually, it wouldn't take too long.

AndrewSB:
💯

A script that would do this automatically for any project?

BasThomas:
That... that'd be awesome 😱

aaroncrespo:
How would this work with bundleID collisions with say a watchOS extension.

petester42:
This more for frameworks than actual applications. It's not really useful for application targets

aaroncrespo:
frameworks still have bundleID's though.

AndrewSB:
@aaroncrespo can you give a concrete example? 
I think the issue with bundle ID collisions disappears when you have one universal target, since there is only one target - it can't collide with something that no longer exist.

*sorry if I misunderstood your comment* 😅

aaroncrespo:
It's more a question maybe we weren't loading/linking the framework correctly across watch/extension/app.

petester42:
I have no idea about watchOS stuff.

AndrewSB:
@petester42 I think we should go ahead with this, and fix watchOS issues if/when they come up. When you create a PR, can you close this issue?

Thanks 😄

AndrewSB:
[...]

PR #967, opened 2017-02-15
Convert project to single target for all platforms

petester42:
Based on the discussion in #818 I have finally converted `Moya.xcodeproj` to use a single target for all platforms. The idea comes from the people over at [PromiseKit](http://promisekit.org/news/2016/08/Multiplatform-Single-Scheme-Xcode-Projects/).

I've been working on scripting the ideas from the post above using [`xcodeproj`](https://github.com/CocoaPods/Xcodeproj) and the results can be seen in the tool I created called [framework-generate](https://github.com/petester42/framework-generate). I used `framework-generate` to generate a new `Moya.xcodeproj` file by using a specification that can be found [here](https://gist.github.com/petester42/7ad0312dd1c1dcff780f2f95936200ea). 

I wasn't sure if I should add a dependency to Moya for `framework-generate` and the specification file. So let me know what everyone thinks about the change to single target for all platforms and also adding `framework-generate` as a dependency.


Moya.xcodeproj/project.pbxproj
```diff
@@ -7,2600 +7,945 @@
...
+		1BE7510B98D58C8AF4157077 /* Plugin.swift */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = sourcecode.swift; name = Plugin.swift; path = Sources/Moya/Plugin.swift; sourceTree = "<group>"; };
+		2785AE3C7632CB6FF6259B7E /* Moya+Alamofire.swift */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = sourcecode.swift; name = "Moya+Alamofire.swift"; path = "Sources/Moya/Moya+Alamofire.swift"; sourceTree = "<group>"; };
+		2CA5A9D68C54FB5BE42CF0B7 /* ReactiveMoyaAvailability.swift */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = sourcecode.swift; name = ReactiveMoyaAvailability.swift; path = Sources/ReactiveMoya/ReactiveMoyaAvailability.swift; sourceTree = "<group>"; };
+		319406C47A9FD34466719193 /* MultiTarget.swift */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = sourcecode.swift; name = MultiTarget.swift; path = Sources/Moya/MultiTarget.swift; sourceTree = "<group>"; };
+		321988FCC5EB35D72D53BDD7 /* Moya.framework */ = {isa = PBXFileReference; explicitFileT
```
aaroncrespo: I think anyone on a more recent version of Xcode will see prompts to update their project settings.

aaroncrespo: what do you think about pulling this out into a versioned bash script? 
calling with `$(SRCROOT)/scripts/copy_frameworks.sh`

aaroncrespo:
Looks like the length to the diff is causing my comments to be misplaced.

Moya.xcodeproj/project.pbxproj
aaroncrespo:
```diff
+				LastSwiftUpdateCheck = 0730;
```

aaroncrespo:
```diff
+			shellScript = "#!/bin/sh\n\ncase \"$PLATFORM_NAME\" in\n    macosx) plat=Mac;;\n    iphone*) plat=iOS;;\n    watch*) plat=watchOS;;\n    appletv*) plat=tvOS;;\n    *) echo \"error: Unknown PLATFORM_NAME: $PLATFORM_NAME\"; exit 1;;\nesac\n\nfor (( n = 0; n < SCRIPT_INPUT_FILE_COUNT; n++ )); do\n    VAR=SCRIPT_INPUT_FILE_$n\n    framework=$(basename \"${!VAR}\")\n    export SCRIPT_INPUT_FILE_$n=\"$SRCROOT\"/Carthage/Build/$plat/\"$framework\"\ndone\n\n/usr/local/bin/carthage copy-frameworks || exit\n\nfor (( n = 0; n < SCRIPT_INPUT_FILE_COUNT; n++ )); do\n    VAR=SCRIPT_INPUT_FILE_$n\n    source=${!VAR}.dSYM\n    dest=${BUILT_PRODUCTS_DIR}/$(basename \"$source\")\n    ditto \"$source\" \"$dest\" || exit\ndone";
```

petester42: I was using Xcode 8.2.1 to test the project was generated properly and didn't see any. The "update project settings" actually checked for properties of the project. I've already handled up to Xcode 8.2 and maybe 8.3 settings.

petester42: Ya that's possible. I'll add an option to my tool and regenerate the project.

aaroncrespo: Interested to hear more about the tool you used to generate a project like this...

petester42:
I linked it in the pull request. Here it is again https://github.com/petester42/framework-generate feedback is welcome.

aaroncrespo:
One thing I found missing that is a nice to have is that the Test Scheme is not set in the Library's Scheme as a Test Target. `⌘ + U` will build the library and run the tests with the Library selected if it is set. Otherwise the test scheme will need to be selected for `⌘ + U` to work.

I also think code coverage is turned off (not sure this is tracked).

scottrhoyt:
@aaroncrespo @petester42 , code coverage is being tracked. The shortcut for testing is nice to have. Anyway to bring that back?

petester42:
Adding the test command to the target is easy. I thought I had added that already. I'll check.

petester42:
Ok so the following should be addressed:
- `⌘ + U` for the targets
- Code coverage is now enabled again
- Carthage framework copy script is now part of the source control.

petester42:
I don't really know what to do about the danger errors. Any ideas since I don't think my change has those errors.

scottrhoyt:
First error is just that you need to put an entry in the `CHANGELOG`.

Second error is probably getting confused with the new project structure, thinking dependencies have changed the libraries are no longer in the "Linked Libraries" section but instead are found via a platform specific framework search path.

First warning won't go away because of the # of lines changed in the PR, that's fine.
Second warning is a reminder to add some documentation for big new features, but this doesn't warrant that.

So basically just worry about the `CHANGELOG`.

scottrhoyt:
Now, in general, I love this idea and want to see it happen. But I have to confess I'm a novice with this sort of conversion.

Some of my questions:
 * Will this work with both Carthage and Cocoapods? And have both been tested?
 * Does this method work with Carthage prebuilt binaries incase we want to introduce that for Moya?
 * Are there any other downsides to this?
 * This won't be breaking for anyone will it?

AndrewSB:
@scottrhoyt: I'm not an expert, but I think it **should** just make maintaining the Moya.xcodeproj a lot easier.

> Will this work with both Carthage and Cocoapods? And have both been tested?

It works fine with Cocoapods & Carthage ([PromiseKit](https://github.com/mxcl/PromiseKit/tree/swift-3.0) has one target and they support both + SPM). 

> Does this method work with Carthage prebuilt binaries incase we want to introduce that for Moya?

I'm not sure about this one. I think it should - since all we're doing is merging the targets, which should just mean we need to build the one target for each device type - but we should test before merging.

> Are there any other downsides to this?

Just that what we're doing isn't officially supported by Apple. http://promisekit.org/news/2016/08/Multiplatform-Single-Scheme-Xcode-Projects/ is a great resource on this idea

> This won't be breaking for anyone will it?

Not to my knowledge

scottrhoyt:
Thanks for your thoughts @AndrewSB! I definitely like the maintenance reducing-nature of this. I'd like to see us get a prebuilt binary out (opened #983 to address), but if that is incompatible with this, I think I would rather have this. Especially since build caching and "only rebuild on change" behavior is in the near term Carthage pipeline, so prebuilt frameworks will lose a certain amount of their benefit.

AndrewSB:
@petester42 I'd love to get this merged in. Do we just have those three issues left?

> ⌘ + U for the targets
> Code coverage is now enabled again
> Carthage framework copy script is now part of the source control.

petester42:
@scottrhoyt

> - Will this work with both Carthage and Cocoapods? And have both been tested?
Yes this works with both. Cocoapods doesn't use the xcode project so no problem there. I've used this same thing for other projects and Carthage works too. Just tested it to make sure and it works.

> - Does this method work with Carthage prebuilt binaries incase we want to introduce that for Moya?
Yup. Just tried it and it works too.

> - Are there any other downsides to this?
For a maintainer I don't see any downside. Might be a little odd for people who aren't familiar with it.

> - This won't be breaking for anyone will it?
I don't see why it would.

@AndrewSB
> ⌘ + U for the targets
Code coverage is now enabled again
Carthage framework copy script is now part of the source control.
All this has been taken care of already.

AndrewSB:
So after rebasing this & fixing CI, this should be ready to go. 

If no one else has taken a stab at resolving conflicts by tomorrow, I'll give it a shot and add some contributors as reviewers for this before we merge in 😄

AndrewSB:
I rebased, but we were still having a conflict, so I merged Podfile.lock. Going to get this CI passing and then we should be good to go 😄

AndrewSB:
Apparently something is wrong with Danger, but it looks like a nonsensical error. The Dangerfile on this branch matches the one on `master`. Can a @Moya/contributors who's familiar with Danger try to debug this?

orta:
considering the scope of your changes, I think it's fine to accept that the Danger rules may not be fine for this run

Changelog.md
```diff
@@ -1,4 +1,13 @@
 # Next
+- Move project to using a single target for all platforms
+
+# 8.0.3
```
pedrovereza: The entries in the changelog are incorrect. I think we only need the one under `Next`.

AndrewSB: yeah, I'm not sure how that happened. Thanks for catching it @pedrovereza

AndrewSB:
Thank you everyone for your work! I'm really excited that we got this merged in 😁

petester42:
👍

PR #968, opened 2017-02-15
Response with destinationURL

### LINKED ISSUE #959: Response should have destinationURL
I believe that Moya's Response should have a property called `destinationURL` much like Alamofire's response.

I'm trying to migrate a piece of code from `Alamofire.download` to the `.download` task and I've noticed that the response object does not have the destination path to where the downloaded file has been moved to.

I'm not sure if `temporaryURL` should also exist in the Response, but I believe it's a no-brainer.

fjcaetano:
fix #959

### DIFF
--- a/Sources/Moya/Endpoint.swift
+++ b/Sources/Moya/Endpoint.swift
@@ -4,11 +4,11 @@ import Alamofire
 /// Used for stubbing responses.
 public enum EndpointSampleResponse {
 
-    /// The network returned a response, including status code and data.
-    case networkResponse(Int, Data)
+    /// The network returned a response, including status code, data and destination path of any downloaded data.
+    case networkResponse(Int, Data, URL?)
 
     /// The network returned response which can be fully customized.
-    case response(HTTPURLResponse, Data)
+    case response(HTTPURLResponse, Data, URL?)
 
     /// The network failed to send the request, or failed to retrieve a response (eg a timeout).
     case networkError(NSError)
--- a/Sources/Moya/Moya+Alamofire.swift
+++ b/Sources/Moya/Moya+Alamofire.swift
@@ -65,7 +65,7 @@ public final class CancellableToken: Cancellable, CustomDebugStringConvertible {
 
 }
 
-internal typealias RequestableCompletion = (HTTPURLResponse?, URLRequest?, Data?, Swift.Error?) -> Void
+internal typealias RequestableCompletion = (HTTPURLResponse?, URLRequest?, Data?, URL?, Swift.Error?) -> Void
 
 internal protocol Requestable {
     func response(queue: DispatchQueue?, completionHandler: @escaping RequestableCompletion) -> Self
@@ -74,7 +74,7 @@ internal protocol Requestable {
 extension DataRequest: Requestable {
     internal func response(queue: DispatchQueue?, completionHandler: @escaping RequestableCompletion) -> Self {
         return response(queue: queue) { handler  in
-            completionHandler(handler.response, handler.request, handler.data, handler.error)
+            completionHandler(handler.response, handler.request, handler.data, nil, handler.error)
         }
     }
 }
@@ -82,7 +82,7 @@ extension DataRequest: Requestable {
 extension DownloadRequest: Requestable {
     internal func response(queue: DispatchQueue?, completionHandler: @escaping RequestableCompletion) -> Self {
         return response(queue: queue) { handler  in
-            completionHandler(handler.response, handler.request, nil, handler.error)
+            completionHandler(handler.response, handler.request, nil, handler.destinationURL, handler.error)
         }
     }
 }
--- a/Sources/Moya/MoyaProvider+Defaults.swift
+++ b/Sources/Moya/MoyaProvider+Defaults.swift
@@ -7,7 +7,7 @@ public extension MoyaProvider {
         let url = target.baseURL.appendingPathComponent(target.path).absoluteString
         return Endpoint(
             url: url,
-            sampleResponseClosure: { .networkResponse(200, target.sampleData) },
+            sampleResponseClosure: { .networkResponse(200, target.sampleData, nil) },
             method: target.method,
             parameters: target.parameters,
             parameterEncoding: target.parameterEncoding
--- a/Sources/Moya/MoyaProvider+Internal.swift
+++ b/Sources/Moya/MoyaProvider+Internal.swift
@@ -131,12 +131,12 @@ public extension MoyaProvider {
             }
 
             switch endpoint.sampleResponseClosure() {
-            case .networkResponse(let statusCode, let data):
-                let response = Moya.Response(statusCode: statusCode, data: data, request: request, response: nil)
+            case .networkResponse(let statusCode, let data, let url):
+                let response = Moya.Response(statusCode: statusCode, data: data, request: request, response: nil, destinationURL: url)
                 plugins.forEach { $0.didReceive(.success(response), target: target) }
                 completion(.success(response))
-            case .response(let customResponse, let data):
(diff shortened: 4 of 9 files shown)


fjcaetano:
If there are no objections to this PR, I can fix the issues Danger raised

scottrhoyt:
Thanks for the PR @fjcaetano! I'm going to pull the discussion into this PR.

Before I address some of the questions you have in #959, can you walk me through the use case you are trying to solve that doesn't lend itself to using `DownloadDestination` to declare where you want your file downloaded to?

fjcaetano:
`DownloadDestination` is a closure whose responsibility belongs to the Target. It's scope should be closed to within the Target.

`destinationURL` should be a property of `Response` because it is likely that that path is required at response time not at request time. Something like this:

``` swift
/// Downloads a URL file and returns its local path
func download(_ url: URL) -> Observable<URL> {
  return provider.request(.download(url))
    .map { (response: Response) -> URL in
      guard let url = response.destinationURL else {
        throw NSError(code: .noData)
      }
      return url    
  }
}
```
(We're using RxSwift, btw)

This could be accomplished by using `DownloadDestination` if we pass the closure as a stored value of the Target enum, but I think it would be outside the correct scope and would just look wrong.

scottrhoyt:
What do you think about having your `TargetType` case take a parameter on where to store the resulting download? Then you can construct the `DownloadDestination` in the `task` computed property. That way you would declaratively tell the request where to store the downloaded file and capture that location in your response closure (which is likely near the call site that you supplied the destination URL).

fjcaetano:
That may make sense for some cases where you'd like to have the file stored in a specific location known beforehand, but I don't know if it would be the general case. Nevertheless, it would cause the URL construction to be outside of the `TargetType` responsibilities, which I believe to be incorrect.  Also, what would happen if one passes an invalid URL or a path without `write` access? Would the request fail *after* download?

Keeping `DownloadDestination` as part of the `TargetType` and returning the `destinationURL` with the `Response` wouldn't keep you from creating a `TargetType` case that could receive the path in which to store the file, so these aren't exclusive.

Speaking of our case, our app downloads a few things from S3 and throws it all in the Documents directory, so it falls under the use case I gave you. As I said, I don't know if this is the general case, but I believe it is one that makes the solution I proposed to be reasonable.

scottrhoyt:
Yeah, I don't think what you are proposing is unreasonable at all @fjcaetano. I just want to explore what is possible currently before considering expanding the API surface area. Thanks for walking me through your use case.

I don't think constructing the destination `URL` for the downloaded file outside of `TargetType` is necessarily a problem. In some cases it might be the right way to do it. You could certainly still construct the `URL` inside of `TargetType` as well, but having the capability to pass the destination `URL` to `TargetType` is a more general solution.

So in your use case, how do you plan on specifying that the files should be downloaded to the Documents directory? And how do you specify what the file should be named in the documents directory? In other words, how do you figure out the `URL` that would eventually be in the proposed `destinationURL`?

fjcaetano:
We're using the `download` task to mimic Apple's On-Demand Resources (we had a bad experience using it), so our app knows which files to download. It is not user-defined. This is our `DownloadDestination` closure:
``` swift
fileprivate func downloadDestination(_ temporaryURL: URL, _ response: HTTPURLResponse) -> (URL, DownloadRequest.DownloadOptions) {
        let directoryURLs = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask)
        var path = temporaryURL
        
        if !directoryURLs.isEmpty {
            path = directoryURLs[0].appendingPathComponent(response.suggestedFilename!)
        }
        
        return (path, .removePreviousFile)
    }
```
As you can see, we keep the downloaded file name and replace it, if necessary. We could figure out the name by deconstructing the given URL, if we were required to give the `destinationURL` beforehand, but it is unnecessary in our case.

scottrhoyt:
Ah, I see. Thanks for sharing.

So, I do think that the question of whether or not to include a `destinationUrl` needs further consideration, and I would love to get the opinions of other @Moya/contributors.

On the other hand, for your particular use case, I might recommend a different architecture. Because the location of these files needs to be deterministic, I think that recommends specifying the destination `URL` declaratively even more. Having access to this `URL` would serve useful for other things like checking whether or not the file has already been downloaded before downloading it again. You could extract this from the asset `URL` as you say, or you could just construct the remote `URL`s differently. I imagine something like this:

```swift
fileprivate static let assetDir: URL = {
  let directoryURLs = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask)
  return directoryURLs.first ?? NSTemporaryDirectory()
}()

enum Assets: TargetType {
  case star
  case checkmark

  var baseURL: URL { /* ... */ }

  var assetName: String {
    switch self {
    case .star: return "star.png"
    case .checkmark: return "checkmark.png"
    }
  }

  var path: String {
    return "/assets/" + assetName
  }

  var localLocation: URL {
    return assetDir.appendingPathComponent(assetName)
  }

  var downloadDestination: DownloadDestination {
    return { _, _ in return (self.localLocation, .removePreviousFile) }
  }

  var task: Task {
    return .download(downloadDestination)
  }

  /*
    Rest of TargetType
  */
}

class AssetDownloader {
  let provider = MoyaProvider<Assets>()

  init() { }

  func load(asset: Asset, completion: ((Result<URL, MoyaError>) -> Void)?) {
    if FileManager.default.fileExists(atPath: asset.localLocation.path) {
      completion?(.success(asset.localLocation))
      return
    }

    provider.request(asset) { result in 
      switch result {
      case .success:
        completion?(.success(asset.localLocation))
      case let .failure(error):
        return completion?(.failure(error))
      }
    }
  }
}
```

fjcaetano:
You are right. When you mentioned the verification to see if the file already exists, it sounded familiar because I knew we already did something like that and, in fact, we do. Except it's on another part of the code.

So I've reached an impasse, because we build the `destinationURL` to check if the file exists, but, in the other hand, our downloader doesn't have that check, because we assume that if the method is called, it must download the file whether or not it exists.

I guess we could refactor our service layer to include this verification and the URL build.

scottrhoyt:
@fjcaetano 👍. I think that refactor sounds like a good idea because then you can centralize the responsibilities of managing assets into a single object--and I like that separation of concerns. I could see it also being beneficial to you if decide to do asset preloading, batch downloading, etc. It could really be a nice replacement for ODR.

Let's keep this PR open though and see if we can get more feedback on this for other use cases, and thanks again for your contribution!

xliangTrackVia:
@fjcaetano I am currently running into this exact scenario within our application. I need a destination URL, or some mechanism, to retrieve the Data instance from the downloaded request. Currently, the Moya.Response object for a download operation returns nil for Response.data. I believe this is because the extension does not set the Data value on response complete:

```swift
// In Moya+Alamofire.swift
extension DownloadRequest: Requestable {
    internal func response(queue: DispatchQueue?, completionHandler: @escaping RequestableCompletion) -> Self {
        return response(queue: queue) { handler  in
            // *** Return nil data here even though the handler has a reference to the destination URL
            completionHandler(handler.response, handler.request, nil, handler.error)
        }
    }
}
```
I am not sure when this PR will be approved, but it is certainly something that our application can use. Thanks again for tackling this!

fjcaetano:
@xliangTrackVia doesn't scott's suggestion, of providing the destinationURL with the enum, solve it for you?

sunshinejr:
I'm gonna close this one due inactivity. Thank you for taking your time for this PR though, @fjcaetano. 

To give my opinion about this one, I'm not sure `destinationURL`, which is `nil` most of the time (for normal `request` & sample data), is the best way to implement it. Maybe an enum property that would be having special metadata based on the task type would be better? Not sure. What i know for sure is that this example with assets by @scottrhoyt should be added to our examples.

sedwo:
Hi @xliangTrackVia , Did you ever solve this?

PR #979, opened 2017-02-23
Fixed `reversedPrint` arguments for output

jessearmand:
Related to #612, it fixed an issue where `reversedPrint` is printing an array of `items` which will ignore the separator and terminator that has been set. Without this, pretty print (e.g. with JSON pretty print) is broken. 

* Variadic argument needs to be enumerated and its element printed
* Renamed seperator to separator

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,7 @@
 # Next
 
+- Fixed `reversedPrint` arguments for output.
+
 # 8.0.2
 
 - Changed dependency pinning to rely only on major versions.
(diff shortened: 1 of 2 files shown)


jessearmand:
Also as shown in the [Danger](http://danger.systems/), I'm not sure whether it's a good idea to update the changelog every time there's a change. 

Because the `CHANGELOG.md` is updated for each  tagged version, not on every pull request. So, I don't know what to do here.

scottrhoyt:
@jessearmand you can tag your PRs with `#trivial` to not trigger this Danger warning. However, I think a nice fix like this should be in the Changelog. If you look at the document, the top heading is "Next". This is the space for changes the currently only exist on `master`. Currently, there are none, but you can go ahead and make the first entry!

jessearmand:
@scottrhoyt Thanks for clarification. I updated the changelog.

scottrhoyt:
Thanks! And thanks for the contribution!

ashfurrow:

Issue #975, opened 2017-02-20
How to handle HTTP errors when using Alamofire automatic validation?


mrtj:
I am using Moya 8.0.2 configured with OAuth2 authentication handling with Alamofire RequestRetrier as suggested in the [docs](https://github.com/Moya/Moya/blob/master/docs/Examples/Advanced.md), outlined in [this](https://github.com/Alamofire/Alamofire/issues/1097) Alamofire issue and implemented by [p2/OAuth2](https://github.com/p2/OAuth2/wiki/Alamofire-4). 

To make it working in Moya, I implemented the `validate` getter in my `Target` returning `true` for the requests that need validation.

Everything works like a charm, the installed Alamofire RequestRetrier intercepts the 401 Unauthorized responses, fires the OAuth flow and then resumes the original request signing it. My problem is that this configuration, in particular that Moya issues the `validate()` request to Alamofire without the possibility of configuring it, shades the response body from the server for non-2xx status codes. For example when issuing a request to a non-existing resource, the automatic validation that I need for the authorization makes failing also this request, returning a `underlying(Alamofire.AFError.responseValidationFailed(Alamofire.AFError.ResponseValidationFailureReason.unacceptableStatusCode(404)))` error but I have no access to the original response from the server.

Are there any ways to selectively validate the Alamofire responses or otherwise make the non-401 HTTP errors succeed the request?

BasThomas:
Maybe @scottrhoyt can help you with this one.

scottrhoyt:
@mrtj unfortunately I don't have any easy answers for you on this one. Using Alamofire's validation takes a lot of control out of Moya's hands. The way I see it, you have a couple of options, but they aren't simple.

1. Modify Alamofire's validation error to also expose the response. This could be via filling an issue/PR with them or creating your own fork.
2. Modify Moya's `TargetType.validate` property to return an optional integer range to validate with. A `nil` value indicates no validation while a range is forwarded to Alamofire's `validate(statusCode:)` that would go around [this section](https://github.com/Moya/Moya/blob/master/Sources/Moya/MoyaProvider+Internal.swift#L214). This would require you to fork Moya.
3. Build request retrying functionality into Moya.

Option 1 would probably be valuable to the wider Alamorefire community, but I think it might be hard to get a breaking change like that through in any reasonable timeline for you. You would also have to maintain separate forks of both Alamofire and Moya.

[...]

PR #986, opened 2017-02-26
get response and error when use Alamofire validation

yanamura3:
I tried to solve this problem.
https://github.com/Moya/Moya/issues/975

This change does not affect to current behavior, so I pull request to master branch.

## solved problem
when use Alamofire validation, we cannot get response in failure result.

## how to solved
check the Target is using "validate" in convertResponseToResult()

### DIFF
--- a/Sources/Moya/Moya+Alamofire.swift
+++ b/Sources/Moya/Moya+Alamofire.swift
@@ -86,3 +86,22 @@ extension DownloadRequest: Requestable {
         }
     }
 }
+
+extension Swift.Error {
+    internal func convertToMoyaError(response: Moya.Response?) -> MoyaError {
+        if let afError = self as? AFError {
+            switch afError {
+            case .responseValidationFailed:
+                if let response = response {
+                    return MoyaError.statusCode(response)
+                } else {
+                    return MoyaError.underlying(self)
+                }
+            case .invalidURL, .parameterEncodingFailed, .multipartEncodingFailed, .responseSerializationFailed:
+                return MoyaError.underlying(self)
+            }
+        } else {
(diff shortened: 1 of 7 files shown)


scottrhoyt:
Thanks for the contribution @yanamura3!

I have to think about this one for a second. Something feels a little off about transforming an `AFError` into a `MoyaError` like this. I don't know that it is a bad approach, it just feels unique as to how we handle other underlying errors. Also, having it depend on `validate` and possibly not doing the transform if a response doesn't exist feel a bit too non-deterministic.

On the other hand, this is a clever way to get the changes necessary for #975! It is technically a breaking change, though. If someone wrote code that looks for the a `AFError. responseValidationFailed` embedded in a `MoyaError.underyling`, their code will break with this update. Worse, it would probably be silent--no compiler error.

In the light that it would be a breaking change, I'm leaning towards it being better to just add an optional `Response` value to `MoyaError.underyling`, but that's not a strong opinion.

Would love to hear what other @Moya/contributors think of this approach.

yanamura3:
@scottrhoyt  Thanks for comments!

You are right, if someone handling responseValidationFailed, my change will be a breaking change..
> It is technically a breaking change, though. If someone wrote code that looks for the a AFError. responseValidationFailed embedded in a MoyaError.underyling, their code will break with this update. Worse, it would probably be silent--no compiler error.

I think my approach is not best solution ( just patch solution ).
It's may be better to solve #975 next major version.

gperdomor:
@scottrhoyt check  #993 please, i add the response to `MoyaError.underyling`

gperdomor:
@yanamura3 this was solved in #993 :D

Issue #1011, opened 2017-03-15
Support RxSwift's Single


andreyfel:
Since RxSwift released version 3.3 which adds support for Single observable, it would be nice if Moya supports this concept.

So, RxMoyaProvider should be able to return Single<> Instead of Observable<>, because ideologically it is more correct, because we emit only one response.

BasThomas:
Feel free to take a stab at implementing this! Would this be a breaking change or would/should it coexist with the `Observable`-implementation?

damienrambout:
It would probably be a breaking change since `request(_ token: Target)` would now return a `Single<Response>` (instead of `Observable<Response>`).
Otherwise, I think `Single` is pretty much the same as a regular `Observable`.

AndrewSB:
`Single<>` sounds incredible! I brought something similar up as an issue a few months ago, I'm so excited they implemented something that solved the same problem.

I'll take a shot at this 😄

AndrewSB:
Implemented in #1020

PR #1013, opened 2017-03-15
Merge master into 9.0.0 dev

BasThomas:

PR #993, opened 2017-03-01
[Review] Add response to moya underlaying error

gperdomor:
I tried to solve #975

PR #1016, opened 2017-03-16
update RxMoyaProvider.request to return a Single

AndrewSB:
Implements #1011, and changes the signature of RxMoyaProvider.request to return a `Single<Response>` instead of an `Observable<Response>`.

cc/ @andreyfel @damienrambout: would love both of you to sign off on this before it's merged in


BasThomas:
This is breaking and should go in `9.0.0` and not master, right?

BasThomas:
Also, some Cocoapods package versions don't match up with Cartfile anymore.

andreyfel:
I think it would be better to create new Provider, like RxSingleMoyaProvider to make this change non-breaking. In this way we will support both Observable and Single. And it will be easier to migrate to the Single.

AndrewSB:
@BasThomas: In its current state, it is breaking. @BasThomas can you match the Cocoapods package versions that I missed?

@andreyfel: I don't know if adding another choice to a very simple module is a smart idea long-term. I feel as though we should either stick with Observable and not make this change, or embrace RxSwift's new semantics and publish Single. But not try to inelegantly solve both
After understanding the change, the easy path to migrate to the new RxMoya syntax will simply be for a consumer to extend `RxMoyaProvider` and supply a method that returns an `Observable<Response>`, which `.asObservable()`'s the new `RxMoyaProvider.request`

gperdomor:
This PR revert the changes introduced in #993... This can't be merge yet

pedrovereza:
I think we should first deal with #1013 (which is also reverting #993)

gperdomor:
@AndrewSB please rebase

pedrovereza:
@AndrewSB can you rebase against `9.0.0-dev` and reapply your changes? (See #1017) :wink:

AndrewSB:
All rebased! cc/ @gperdomor can you approve?

I need to fix a couple tests to get CI passing, working on them soon 😄

gperdomor:
#993 is still reverted in this PR

pedrovereza:
@gperdomor @AndrewSB Hey, I fixed the branch by cherry-picking all changes on top of `9.0.0-dev`: https://github.com/Moya/Moya/commits/observable-single-9.0

We can:

1 - Open a new PR from that branch
2 - Push force to `observable-single`, meaning everyone will have to delete the branch locally and fetch again (but we keep this PR)

I'm ok either way 😄

gperdomor:
Option 1 is more cleaner i guess

justinmakaila:
I'm not a user of RxSwift, but just cruising through their API (you know, trying to form an educated opinion), it looks like `Single<MoyaResponse>` IS the most analogous structure to represent a network request, unless the request is repeating, in which case, `Observable` would make sense. In the case of a repeating request however, it should be on the developer to format their signal in that way.

Realistically, I don't think we should not embrace this change just because our API already utilizes `Observable`, but rather note it as a breaking change and release.

Forcing people to make changes to existing code bases to implement the best practices from one of our dependencies for a new version should be expected, and as @AndrewSB mentioned, you can just add `asObservable()` if you're a client who wants to continue to use `Observable`.

AndrewSB:
Thank you for sharing your thoughts @justinmakaila!

@pedrovereza: I'm lazy, and I don't see the harm in #2, but feel free to go ahead with whichever one makes you more comfortable

pedrovereza:
@BasThomas @AndrewSB @gperdomor @justinmakaila @andreyfel 
Hey I created #1020 to avoid more branch issues. Please continue the discussion in that PR and use branch `observable-single-9.0` for new additions. I'll delete `observable-single`to avoid confusions :wink:

AndrewSB:
Thanks Pedro 😊

PR #1020, opened 2017-03-22
Observable single 9.0

pedrovereza:
Continues #1016 as we had some issues merging the original branch with `9.0.0-dev`


pedrovereza:
@AndrewSB Is it worth adding in the changelog that we bumped the minimum version of RxSwift to 3.3? 🤔

BasThomas:
Yeah, I guess that wouldn't hurt.

BasThomas:
I went ahead and added it.

pedrovereza:
@BasThomas Thanks! 👍

BasThomas:
Ugh, that Swiftlint issue... what's their reasoning behind that? I think `let _ =` is better than `!= nil`. Thoughts?

pedrovereza:
Doing `let _ =` is just a way to check if an `Optional` value is `nil` or not. I personally prefer `!= nil` because the check is obvious.

This is not a strong opinion so I'm ok with whichever we agree to use :wink:

BasThomas:
🤔. I think both are equally obvious, and like `let _ =` for looking cleaner (imho).

BasThomas:
Let's deal with it later. Don't think its that important right now. FWIW, the Swiftlint rule is `unused_optional_binding`.

BasThomas:
Once again thanks for fixing the merging issues @pedrovereza! 🎉

AndrewSB:
Yeah, thanks a bunch @Pedrovereza!

scottrhoyt:
@pedrovereza thanks for this! I think maybe this extension also has to be modified? https://github.com/Moya/Moya/blob/master/Sources/RxMoya/Observable%2BResponse.swift

pedrovereza:
@scottrhoyt I think you're right 🤔 

@AndrewSB Did we miss the extension @scottrhoyt pointed out or is there a reason to keep returning `Observable`?

AndrewSB:
@scottrhoyt: thanks for catching that!

Here's what I think: the functions in Observable+Response are extensions on any type of `Observable<Response>`. I think it's totally valid to have a producer that sends multiple `.next` events to an `Observable<Response>` (for example, I could have a request processor, which is of type `Observable<Response>` and I pass every response that goes through my client into it).

I thought it best not to assume that _every_ `Observable<Response>` should be constrained to `Single`, and that's a choice a consumer of this library should make, by doing `.asSingle()` after using one of the operators

scottrhoyt:
Ok I understand that intention. Should we duplicate this as an extension on `PrimitiveType` too so that users can still do something like:

```swift
provider.request(.myRepos)
  .filterSuccessfulStatusCodes()
  .subscribe()
  .disposed(by: disposeBag)
```

Or would the recommended approach be to do:

```swift
provider.request(.myRepos)
  .asObservable()
  .filterSuccessfulStatusCodes()
  .asSingle()
  .subscribe()
  .disposed(by: disposeBag)
```

?

AndrewSB:
I see your pain point. We should also probably extend Single to have the same functions. I wonder if theres a common superclass/protocol we could extend? I'm not sure how the Rx project does it.

I'll look into doing a common extension tomorrow, but feel free to give it a shot as well 😄

AndrewSB:
Oh, just noticed that you know what the superclass is - `PrimitiveType`

Yes, lets make them functions on that

PR #1027, opened 2017-03-24
Fixes ReactiveSwift deprecation warning

pedrovereza:
Fixes #1018 

Our podspec claims we can take any version of ReactiveSwift equal or greater than 1.0. 
ReactiveSwift deprecated `DateSchedulerProtocol` in 1.1, which caused warnings when linting our pod.

Output of `pod lib lint`:

```
    - WARN  | [Moya/ReactiveCocoa,Moya/ReactiveSwift] xcodebuild:  Moya/Sources/ReactiveMoya/ReactiveSwiftMoyaProvider.swift:9:32: warning: 'DateSchedulerProtocol' is deprecated: renamed to 'DateScheduler'
    - NOTE  | [Moya/ReactiveCocoa,Moya/ReactiveSwift] xcodebuild:  Moya/Sources/ReactiveMoya/ReactiveSwiftMoyaProvider.swift:9:32: note: use 'DateScheduler' instead
    - WARN  | [Moya/ReactiveCocoa,Moya/ReactiveSwift] xcodebuild:  Moya/Sources/ReactiveMoya/ReactiveSwiftMoyaProvider.swift:15:60: warning: 'DateSchedulerProtocol' is deprecated: renamed to 'DateScheduler'
    - NOTE  | [Moya/ReactiveCocoa,Moya/ReactiveSwift] xcodebuild:  Moya/Sources/ReactiveMoya/ReactiveSwiftMoyaProvider.swift:15:60: note: use 'DateScheduler' instead
````

This PR replaces the usage of `DateSchedulerProtocol` with `DateScheduler` and bumps the minimum version of ReactiveSwift to 1.1

I wasn't sure if this requires a change in `Package.swift`, as I feel that adding a `minorVersion` would pin the dependency to that minor (as in, not as a minimum version, but that version exactly). Is that correct? 🤔

PR #1031, opened 2017-03-28
[WIP] Single scheme

### LINKED ISSUE #818: Discuss: Use a single target for all platforms?

gperdomor:
- Convert project to single schema target.
- Update deps for better swift 3.1 support.

Fix #818


BasThomas:
What does this do over #967?

gperdomor:
@BasThomas theoretically not much, the purpose is the same, but in this case i converted manually, without script.

Also i updated the dependencies for better compatibility with Swift 3.1 and remove some deprecation warnings... Demo was updated too.

This PR is target to 9.0.0

pedrovereza:
I think we can take the deprecation warning in master, as it's not a breaking change. We actually have #1027 to do it :wink:

I also think that converting to a single scheme is not a breaking change and we should take it in master (which makes me think that we should continue #967 instead).

The dependencies update is good, but is it a breaking change? If not, I'd vote for merging into master so we can release it before 9.0

gperdomor:
Closed in favor of #967

PR #1038, opened 2017-03-31
Fixes fourth-tier header in Advanced.md

pedrovereza:
The parser seems to need that space between the header mark and the actual text 🤔


BasThomas:
Yeah - I think this changed? I've been seeing this in a couple of repositories over the last week. Thanks though!

PR #1053, opened 2017-04-15
Only append path to URL when path is not empty

### LINKED ISSUE #1049: BaseURL request problem.
Moya : 8.0.3
Language : Swift 3.1

### Let's imagine that you have a link that looks like that:

```swift
let urlString = "https://google.com/123/somepath?X-ABC-Asd=123"
```

### When I do my request using `Alamofire` :

```swift
Alamofire.upload(fileURL, to: URL(string: urlString)!, method: .put, headers: nil).response {
    response in
    print(response)
}
```

it all seems fine(don't look at status code):

screen shot 2017-04-12 at 14 14 40

### but when I do the same request using Moya like that:

```swift
enum MyApi {
    case upload
}
extension MyApi : TargetType {
    var baseURL: URL {
        let urlString = "https://google.com/123/somepath?X-ABC-Asd=123"
        return URL(string: urlString)!        
    }

    var path: String {
        return ""
    }
    
    ...
}

provider.request(MyApi.upload)
```
The created request URL is different that in `Alamofire` request.

screen shot 2017-04-12 at 14 20 49

## The difference is that before `?` character moya inserts the `/` char. 

https://google.com/123/somepath?X-ABC-Asd=123
https://google.com/123/somepath/?X-ABC-Asd=123

Could you help me and tell me what I'm doing wrong?

pedrovereza:
Fixes #1049 

When a `TargetType`'s `path` is empty, `URL.appendingPathComponent` may introduce trailing `/`, which may not be wanted.

From the [docs](https://developer.apple.com/reference/foundation/url/1780239-appendingpathcomponent): 
> This function performs a file system operation to determine if the path component is a directory. If so, it will append a trailing /. If you know in advance that the path component is a directory or not, then use func appendingPathComponent(_:isDirectory:).

Using `appendingPathComponent(_:isDirectory:)` doesn't seem to solve the problem (note `somepath/?`)

```swift
import Foundation

let urlString = "https://google.com/123/somepath?X-ABC-Asd=123"
let url = URL(string: urlString)!

url.absoluteString //"https://google.com/123/somepath?X-ABC-Asd=123"
url.appendingPathComponent("").absoluteString //"https://google.com/123/somepath/?X-ABC-Asd=123"
url.appendingPathComponent("", isDirectory: false).absoluteString //"https://google.com/123/somepath/?X-ABC-Asd=123"
```

I think that having an empty `path` is not usual, but as seen in #1049 it may be necessary. These changes shouldn't cause side effects for non-empty `path`s.

### DIFF
--- a/Sources/Moya/MoyaProvider+Defaults.swift
+++ b/Sources/Moya/MoyaProvider+Defaults.swift
@@ -31,7 +31,7 @@ public extension MoyaProvider {
     }
 
     private final class func url(for target: Target) -> URL {
-        if (target.path.isEmpty) {
+        if target.path.isEmpty {
             return target.baseURL
         }
 
--- a/Sources/Moya/MoyaProvider+Defaults.swift
+++ b/Sources/Moya/MoyaProvider+Defaults.swift
@@ -4,9 +4,8 @@ import Alamofire
 /// These functions are default mappings to `MoyaProvider`'s properties: endpoints, requests, manager, etc.
 public extension MoyaProvider {
     public final class func defaultEndpointMapping(for target: Target) -> Endpoint<Target> {
-        let url = target.baseURL.appendingPathComponent(target.path).absoluteString
         return Endpoint(
-            url: url,
+            url: url(for: target).absoluteString,
             sampleResponseClosure: { .networkResponse(200, target.sampleData) },
             method: target.method,
             parameters: target.parameters,
@@ -29,5 +28,13 @@ public extension MoyaProvider {
         let manager = Manager(configuration: configuration)
         manager.startRequestsImmediately = false
         return manager
+    }
+
+    private final class func url(for target: Target) -> URL {
+        if (target.path.isEmpty) {
+            return target.baseURL
+        }
+
+        return target.baseURL.appendingPathComponent(target.path)
     }
 }
--- a/Sources/Moya/MoyaProvider+Defaults.swift
+++ b/Sources/Moya/MoyaProvider+Defaults.swift
@@ -30,6 +30,9 @@ public extension MoyaProvider {
         return manager
     }
 
+    // When a TargetType's path is empty, URL.appendingPathComponent may introduce trailing /, which may not be wanted in some cases
+    // See: https://github.com/Moya/Moya/pull/1053
+    // And: https://github.com/Moya/Moya/issues/1049
     private final class func url(for target: Target) -> URL {
         if target.path.isEmpty {
             return target.baseURL
--- a/Tests/MoyaProviderSpec.swift
+++ b/Tests/MoyaProviderSpec.swift
@@ -611,6 +611,23 @@ class MoyaProviderSpec: QuickSpec {
                 expect(dataString) == "sample data"
             }
         }
+
+        describe("a target with empty path") {
+            struct PathlessAPI: TargetType {
+                let baseURL = URL(string: "http://example.com/123/somepath?X-ABC-Asd=123")!
+                let path = ""
+                let method = Moya.Method.get
+                let parameters: [String: Any]? = ["key": "value"]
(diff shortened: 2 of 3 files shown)


Sources/Moya/MoyaProvider+Defaults.swift
```diff
@@ -30,4 +29,12 @@ public extension MoyaProvider {
         manager.startRequestsImmediately = false
         return manager
     }
+
+    private final class func url(for target: Target) -> URL {
```
AndrewSB: Let's add a comment here with the link to the issue and a rationale for why this is being done

Tests/MoyaProviderSpec.swift
```diff
@@ -611,6 +611,23 @@ class MoyaProviderSpec: QuickSpec {
...
+                let baseURL = URL(string: "http://example.com/123/somepath?X-ABC-Asd=123")!
+                let path = ""
+                let method = Moya.Method.get
+                let parameters: [String: Any]? = ["key": "value"]
+                let parameterEncoding: ParameterEncoding = URLEncoding.default
+                let task = Task.request
+                let sampleData = "sample data".data(using: .utf8)!
+            }
+
+            it("uses the base url unchanged") {
```
AndrewSB: In this test case as well. I think a link to the issue would go a long way 😄

Tests/MoyaProviderSpec.swift
```diff
@@ -611,6 +611,23 @@ class MoyaProviderSpec: QuickSpec {
...
+                let endpoint = MoyaProvider.defaultEndpointMapping(for: PathlessAPI())
+                expect(endpoint.url) == "http://example.com/123/somepath?X-ABC-Asd=123"
```
AndrewSB: I'm not 100% sure about the syntax here. Shouldn't you use a matcher? `.toEqual` or something?

AndrewSB: I could be wrong though

AndrewSB:
awesome job @pedrovereza 😄

Tests/MoyaProviderSpec.swift
sunshinejr: The syntax is correct, [Nimble allows operators for equivalence/comparison](https://github.com/Quick/Nimble#operator-overloads).

Sources/Moya/MoyaProvider+Defaults.swift
pedrovereza: Done! 👍

Tests/MoyaProviderSpec.swift
pedrovereza: Done! 👍

Tests/MoyaProviderSpec.swift
pedrovereza: Hm, I see that both ways are used in the file 🤔  The one I used as reference was [this test](https://github.com/Moya/Moya/blob/master/Tests/MoyaProviderSpec.swift#L553)

BasThomas:
Wouldn't this be a breaking change?

pedrovereza:
@BasThomas Feels like a bug fix to me. I think the expectation when `path` is empty is that the `baseURL` will be used unchanged (which was not happening).

Do you feel it's a breaking change? If so, why?

BasThomas:
Yes, it's a bugfix, but the "buggy" behavior may be (?) expected.

pedrovereza:
I see, you're saying that someone may be relying on this buggy behavior for something. But I'd say there are probably workarounds to avoid the issue 🤔 

I'd still not call it a breaking change as the API contract has not changed, and the current behavior seems odd and probably is not expected.

@AndrewSB @sunshinejr thoughts on that?

AndrewSB:
I'd say not breaking. It's a bug fix

BasThomas:
Sure, let's treat this as a bugfix then. 😃

Issue #1048, opened 2017-04-11
Rx/ReactiveMoya does not expose queue parameter


haritowa:
Hello, i am a little confused about [this](https://github.com/Moya/Moya/blob/9.0.0-dev/Sources/RxMoya/RxMoyaProvider.swift#L20) method(and it counterpart from ReactiveMoyaProvider). I know, I can later observe on any scheduler I want, but why I can't avoid unnecessary calls on main thread(which I can do with plain Moya api)? If it is not by design, I would like to make tiny contribution 😸

AndrewSB:
Hey @haritowa: I'm not sure I totally understand your question, but it sounds like youre looking for something similar to #1032. If that's the case, we have #762, and we'd love your help 😄

haritowa:
@AndrewSB yeah, sorry, 2am is not best time to create issue descriptions 😅 I'll try again:
As far as i know, by default Alamofire invokes all callbacks on Main queue, but you can customize this behavior via queue parameter. Moya [expose](https://github.com/Moya/Moya/blob/9.0.0-dev/Sources/Moya/MoyaProvider.swift#L85) this parameter, but RxMoyaProvider [don't](https://github.com/Moya/Moya/blob/9.0.0-dev/Sources/RxMoya/RxMoyaProvider.swift#L20). As a result, we can't avoid Main thread.

Only `self.request` [invocation](https://github.com/Moya/Moya/blob/9.0.0-dev/Sources/RxMoya/RxMoyaProvider.swift#L24) can be customized via subscribeOn, but [completion handler](https://github.com/Moya/Moya/blob/9.0.0-dev/Sources/RxMoya/RxMoyaProvider.swift#L25-L31) will be invoked on Main thread. To avoid future computation on main thread we can observe chain on another Scheduler, but we can't bypass main thread callback and unnecessary dispatches.

I look through https://github.com/Moya/Moya/pull/762 Are you sure that pass scheduler(instead of Queue) is the best idea? I believe this is not Rx-side work, and we should just pass queue to [this](https://github.com/Moya/Moya/blob/9.0.0-dev/Sources/Moya/MoyaProvider.swift#L85) method. I don't understand why you are propose to capture queue on provider creation (`RxMoyaProvider<GitHub>(queue: DispatchQueue.global(.utility))`) instead of pass it on each observable creation(this [method](https://github.com/Moya/Moya/blob/9.0.0-dev/Sources/RxMoya/RxMoyaProvider.swift#L20))
I propose:
[...]

Issue #1032, opened 2017-03-28
Scheduler issues with RxSwift


mr:
Using `Moya` 8.0.1 and `RxSwift` 3.1.0

I'm trying to make sure that `Moya` requests run on a background thread, because currently my UI gets locked up while doing requests. I've been trying to use a `ConcurrentDispatchQueueScheduler`, but invariably the `Moya` requests switch back to the main thread. Example:

```swift
MyAPI.instance.getIndex()
    .subscribeOn(ConcurrentDispatchQueueScheduler(qos: .background))
    .observeOn(MainScheduler.asyncInstance)
    .flatMap { index in /* Do some UI stuff here */ }
    .subscribe(onNext: { uiResults in /* Do something with UI results */ })
```

My `getIndex` function is just a simple

```swift
func getIndex() {
    return self.provider.request(.getIndex)
}
```

Putting a breakpoint inside the `Observable.create` part of the `RxMoyaProvider.request` function always shows `com.apple.main-thread`. This is true even if I remove the `observeOn` from above. Also any subsequent `map`s or `flatMap`s all run on `com.apple.main-thread`. Normally I would expect these to run on the `ConcurrentDispatchQueueScheduler`. This happens if I am not using `Moya`.

Any way to not lock up my UI while using `Moya`?

AndrewSB:
So happy you've made this issue! This is a feature we were looking to implement for RxMoya. Here's the pull request: https://github.com/Moya/Moya/pull/762

Let me know if you have any other questions, or if you're interested in implementing the functionality we're missing 😄

AndrewSB:
Let's continue discussion for this on #762, closing this for now

PR #762, opened 2016-11-09
Add an optional response thread param for provider

AndrewSB:
This is based of the conversation on #423, If anyone else wants to take a shot at this, they are more than welcome to! I'll implement it sometime soon, if not 😄 

I added some documentation on threading to the Readme, not sure if it should go there or in it's own file in `docs/`

TODO:
- [x] Update the docs
- [x] Do a final check

PR #1055, opened 2017-04-16
Add optional queue parameter to Reactive providers

haritowa:
Hello, i implemented callback queue propagation for reactive providers(which address https://github.com/Moya/Moya/issues/1032 https://github.com/Moya/Moya/issues/1048 https://github.com/Moya/Moya/pull/762). I have same complains, that should be resolved before merge.

## Queue vs Scheduler
I saw several suggestions to use Schedulers instead of queue for RxSwift(and whatever abstraction ReactiveSwift has), but I’m still believe this queue is not what rx provider responsible for, this is Alamofire threading stuff, users may be confused why they should pass scheduler instead of subscribe/observe on(and maybe this moment should be reflected at method comment).

  We can’t use `SchedulerType`, case we need DispatchQueue, which unnecessary for abstract scheduler, so we have to constraint which scheduler type user can pass(so, instead of reuse his scheduler, user may be forced to create new just to pass queue).

RxSwift do several optimizations for Schedulers(and i’m believe ReactiveSwift too), for example it “converts” Concurrent queue to serial(creating new serial queue with old queue as target), which queue should we use in this case(and even, can we still access concurrent queue)? Not sure we can rely on internal queue in future "optimizations" 😄 

Everything previously said was my own thoughts, but the fact Schedulers configuration is internal(which means we can’t access it outside of module and 3 paragraphs above has no sense) indicates that Krunoslav Zaher don’t want us to use internal queue :)

## Comments
Maybe we should add reacher comments to method, explicitly indicating that we nor observe, nor subscribe on this queue, we just pass it to Alamofire as a callback queue. If user really want subscription(request method call) to run on queue - he should subscribe on caller side.  User may create provider with queue and forget about it, which leads to unexpected element generation queue. In current implementation there is no way to pass nil queue to Alamofire, if provider has default queue, and I don’t know is it okay.

## Other stuff
As far as I understand, I should update Readme(as at [this](https://github.com/Moya/Moya/pull/762) PR). Maybe I shouldn't use `??` operator? Seems it can affect compilation time, so maybe we should sacrifice readability in favour of compilation speed?


haritowa:
I'll fix lint issues later, when there will be no other reason to reject PR(case code can change during PR :))

Sources/RxMoya/RxMoyaProvider.swift
```diff
@@ -6,22 +6,35 @@ import Moya
 
 /// Subclass of MoyaProvider that returns Observable instances when requests are made. Much better than using completion closures.
 open class RxMoyaProvider<Target>: MoyaProvider<Target> where Target: TargetType {
+    
+    /// Propagated to Alamofire as callback queue. If nil - main queue will be used.
+    fileprivate let queue: DispatchQueue?
+    
     /// Initializes a reactive provider.
-    override public init(endpointClosure: @escaping EndpointClosure = MoyaProvider.defaultEndpointMapping,
+    public init(endpointClosure: @escaping EndpointClosure = MoyaProvider.defaultEndpointMapping,
```
AndrewSB: why is override no longer needed here? is it because you added the `queue` param? If so, should we not also add the `queue` param to the regular callback `MoyaProvider`?

AndrewSB:
Let's get the linting issues fixed, maybe write some tests. 

Then we can merge this into `optional-response-thread-param`, and continue the discussion on which queue/scheduler primitive to use in #762

Sources/RxMoya/RxMoyaProvider.swift
haritowa: Maybe, I'm not sure 'case by some reason, queue wasn't in init. 
By the way, maybe we should create second init(return previous init, w\o queue and deprecate it? If anyone has reactive providers subclasses with overridden init - their code would not compile, sounds bad)

haritowa:
> I think it would be smart to implement some more tests as well though. Are you familiar with testing in ReactiveSwift/RxSwift?

Yep, for some degree, but only with RxTest/RxBlocking(that means I can't write ReactiveSwift tests w/o future investigation). I'll cover this commit functionality(check callback queue) within the next few days

Sources/RxMoya/RxMoyaProvider.swift
AndrewSB: I don't think we'd need to deprecate the old signature, if we have a default value,`nil`, for the `queue` param, then it would be a non-breaking change. So it should just be safe to change the `MoyaProvider` initializer 😄

AndrewSB:
Cool, it would be awesome to get those RxSwift tests then! Maybe someone else can add them for ReactiveSwift later.

Things pending before we merge to `optional-response-thread-param`
- [ ] write tests for RxSwift
- [ ] fix the overriding initializer and change the `MoyaProvider` initializer to take a queue as well

Sources/RxMoya/RxMoyaProvider.swift
haritowa: Sorry, i'm afraid you didn't get me. You are right, old init calls will not break after edit(case queue param is optional), but any RxSwift/ReactiveSwift provider descendants overridden init will not compile, until user remove `override`. Let me show you what i'm talking about:
#### This is Moya in current state with two user created provider descendants
This is Moya in current Moya state with two user created provider descendants

#### This is Moya state, after adding queue param(note that new param has default value)
screen shot 2017-04-19 at 12 22 07 pm

#### And finally that is was i proposed in previous commit
screen shot 2017-04-19 at 12 22 52 pm

Adding queue param to `MoyaProvider` will do even more mess, if we will stick to replace instead of deprecation

AndrewSB: Oh!! I see, interesting behavior. Thank you for explaining it with examples! I totally get it now

Let's just make a breaking change and target 9.0.0 instead? What are your thoughts? Do you think deprecating is better practice?

haritowa: If we can choose between shift breaking change to major upgrade or generate boilerplate with deprecating init - i prefer first approach :)

AndrewSB: Yes, I think adding it to all 3 is the best approach, so we have feature parity (as much as possible) between all three.
We also might be able to remove all the `queue` logic from our Reactive and Rx subclasses by instead implementing it in MoyaProvider and then doing super.request instead of self.request in places like https://github.com/Moya/Moya/pull/1055/files#diff-ffd4342ff685cb2dff7f9c3cf77cef22R38

Let's do the deprecate approach, and then maybe in 9.0.0 I can remove the deprecated function so we keep our codebase clean (I can make that PR after we're done with this one)

justinmakaila:
So I'm trying to understand... this is the queue that will be used by Alamofire for callbacks as a hook back into Moya, whereas the remaining Scheduler property is in place for delivery of whatever reactive abstraction you use?

AndrewSB:
That's correct @justinmakaila. So then if you're using the vanilla `MoyaProvider`, you pass a thread. But if you're using an `RxMoyaProvider`, you'd pass in the Rx abstraction of a thread - a scheduler, and whatever the equivalent is for ReactiveSwift as well.

For rationale and example usage, check out #762

haritowa:
As far as I remember, you can't get queue from a scheduler. In addition, you can't pass any Scheduler as a callback queue provider for Alamofire

AndrewSB:
Let's re-examine that assumption: that you can't get a queue from a scheduler.
Schedulers internally use DispatchQueues to do their work ([here's](https://github.com/ReactiveX/RxSwift/blob/master/RxSwift/Schedulers/MainScheduler.swift#L23) how the MainScheduler uses a queue).

Let's make an issue on RxSwift's repo asking if theres any way to grab it, since right now, it is private

haritowa:
It's not private, it's internal. You can implement scheduler in whatever way you want: you should just adopt one protocol, which has nothing common with GCD. So, for instance, we need scheduler which stick with one thread, we can't get it's queue 'case it doesn't use any. There is sync and async versions for main scheduler(and again, we are talking about Schedulers, provided out of the box), but Alamofire wouldn't respect this distinction.
As I said earlier: we don't use this queue for any Rx-related work, we just pass to non-rx world. We don't even guarantee, that each element will be emitted on this scheduler(because we has dependency on Alamofire implementation). If user really want to use scheduler - he can just `observeOn` it any rx-chain he want, but he can't avoid dispatching to main queue for now. 
Anyway, I remember about tests and moving queue to MoyaProvider :)

justinmakaila:
Yeah I agree. The queue passed to Alamofire doesn't necessarily have to be the same one as our Reactive `Scheduler`, all that really matters is that we provide this init option so people can move their request building/handling off of the main queue and into a dedicated background queue (if desired).

haritowa:
Hmm, any clue why tests failed? I can't even open log. By the way, i added tests for `MoyaProvider` `RxMoyaProvider` and moved queue to `MoyaProvider`. Deprecate approach doesn't work as i expected(swift can't choose proper overload)

AndrewSB:
@haritowa: yeah, CI failed because there's no changelog for this. We're working on making that more clear to contributors.

For now, can you just add an entry to the changelog, and then we can see how the tests do?

haritowa:
@AndrewSB Oh i see, i was afraid you just give up on this PR :) Ok, i'll update changelog

AndrewSB:
I got really busy with work, sorry about that @haritowa!

haritowa:
@AndrewSB It's ok, I'm busy on weekdays too :)

haritowa:
@AndrewSB Aaaaand... it's done :)

sunshinejr:
I'm wondering if the parameter name `queue` isn't too vague and e.g. `callbackQueue` wouldn't help with recognizing for what exactly is this queue used. Thoughts?

haritowa:
@sunshinejr I took this name from original request method, but I'm agree, callbackQueue would be much better(I believe "queue" parameter name misleading even in request method)

sunshinejr:
Cool! What do you guys think @AndrewSB, @justinmakaila?

BasThomas:
Yes, I think `callbackQueue` is a better name. It takes away any ambiguity.

@haritowa, do you think we can improve the `queue` parameter in the `request` method as well?

haritowa:
@BasThomas Yep, I think so. In method context `queue` is less confusing, but still does not provide any clue about its mission

BasThomas:
What do you think is a better option?

haritowa:
I prefer slightly more verbose `callbackQueue`

pedrovereza:
The CI failure is legit, seems like `RxSwiftMoyaProviderTests`needs a second review 😄

AndrewSB:
that was broken by me -- I'll fix it now. Sorry!

sunshinejr:
Seems like @AndrewSB fixed the build! Things to do before the merge: 
- Fix the **Chanelog.md** conflicts (basically add the new line to **Next** column instead of **8.0.4**, because **8.0.4** was released)
- Change the name of `queue` to `callbackQueue` (wherever we can)

Does that seem right? 😄

AndrewSB:
yup, that does! Then we can figure out docs and usage instructions in #762

sunshinejr:
@haritowa Are these two points, I wrote above, something you have time to do in the near future? If not, I might try to help :)

pedrovereza:
@sunshinejr @AndrewSB @haritowa I fixed the conflict in the changelog (used Github web interface to do it, that's why the commit message is bad 😕 )

Does it make sense to merge this PR into `optional-response-thread-param` and take the `queue` renaming (and docs) in #762? 🤔

sunshinejr:
I think so 👍 Probably in another PR tho, that would point to `optional-response-thread-param`. But I'm in.

AndrewSB:
Yeah, lets merge this into `optional-response-thread-param`, and then do another PR also targeted to `optional-response-thread-param` that handles the renaming and docs 👌

ashfurrow:

PR #1065, opened 2017-04-25
Added NetClient to Readme.md

alexruperez:
[NetClient](https://github.com/intelygenz/NetClient-iOS) has a subspec (NetClient/Moya) to make NetRequest compatible with your awesome abstraction layer, so I add it to the "Community Extensions" section of the Readme.md


bestwnh:
Just my own opinion: I don't think it's an `extensions` for Moya, it's just a library `Compatible with Moya`. I don't think it's a right place for it. But good work anyway.

AndrewSB:
What does NetClient do? I took a look at the README, but I had trouble seeing why I'd use it over just Alamofire or just Moya. Maybe a section saying "Why would someone use this over just Alamofire"

It seems like a nice piece of work though 👍

alexruperez:
@AndrewSB NetClient is another network abstraction layer, by default it uses URLSession behind but also has an implementation using Alamofire in a subspec. It has syntactic sugar, async/synch execution, interceptors... Maybe someone wants to use the power of the NetRequest builder pattern to build the request and after all launch the request using Moya, that's this pull request being.

I have pending improve the readme, we are starting the launch. Pull request would be greatly appreciated!

Thanks for your support. 😃

sunshinejr:
I also think this is not an extension to Moya, so unfortunately I'm gonna close this PR. 

However, it doesn't mean we can't include your library in our Readme. Maybe we need to rethink how we promote libraries/extension that work with Moya? For now we have only extensions, how about libraries, projects, abstractions? For instance, [Quick uses wiki](https://github.com/Quick/Quick/wiki), and other libraries may have some ideas for this kind of promoting.

I'm gonna open an issue for that, we'd love it if you could help us with that matter, @alexruperez. Thanks again for the nice work!

Issue #1058, opened 2017-04-19
Adding headers by default in target


jayesh15111988:
I have seen an example of code where Provider can add headers to request with endpoint closure

```swift
let endpointClosure = { (target: MyTarget) -> Endpoint<MyTarget> in
    let defaultEndpoint = MoyaProvider.defaultEndpointMapping(for: target)

    // Sign all non-authenticating requests
    switch target {
    case .authenticate:
        return defaultEndpoint
    default:
        return defaultEndpoint.adding(newHTTPHeaderFields: ["AUTHENTICATION_TOKEN": GlobalAppStorage.authToken])
    }
}
let provider = MoyaProvider<GitHub>(endpointClosure: endpointClosure)
```

However, it looks to me endpoint closure is repeated if we are creating provider in multiple places. I was wondering if there is any way to add headers to request in the same way we do add parameters, request type, endpoints etc. For example, in the GitHub struct, it would look like this,

```swift
extension GitHub: TargetType {
    var headers: [String: String] {
        switch self {
            case .repositories:
                return ["auth_header": "2312dasd"]
            case .users:
                return ["user_agent": "my_user_agent"]
        }
    }
}
```

In that way any module sending a request won't explicitly have to add an endpoint closure every time.

jayesh15111988:
I see this has already been raised in https://github.com/Moya/Moya/issues/1046. Should I just close this issue then?

pedrovereza:
@jayesh15111988 Yes, let's continue the discussion in #1046 :wink:

Issue #1046, opened 2017-04-05
Add a property `headers` to TargetType?


EvilNOP:
I think it's better adding a `headers` property to TargetType protocol, it seems like the following:
```swift
var headers: [String : String]?
```

Say that there is an authentication token we need for case `isAuthenticationRequired`, we have to provide an endpoint when initialize `MoyaProvider`.

```swift
let endpointClosure = { (target: MyTarget) -> Endpoint<MyTarget> in
    let defaultEndpoint = MoyaProvider.defaultEndpointMapping(for: target)

    switch target {
    case .isAuthenticationRequired:
        return defaultEndpoint.adding(newHTTPHeaderFields: ["AUTHENTICATION_TOKEN": GlobalAppStorage.authToken])
    default:
        return defaultEndpoint.adding
    }
}

let provider = MoyaProvider<GitHub>(endpointClosure: endpointClosure)
```

Having the `headers` property set, I would't have to write the endpoint, I just need to write the code in MyTarget:
```swift
var headers: [String : String]? = {
    switch self {
    case .isAuthenticationRequired:
        return  ["AUTHENTICATION_TOKEN" : GlobalAppStorage.authToken]
    default:
        return nil
    }
}
```

pedrovereza:
@EvilNOP Thanks for raising the idea 🙌  I've created #1067 with an initial implementation for `headers` inside `TargetType`. Feel free to review it and let us know what you think 😉

pedrovereza:
Implemented in #1067

Issue #1003, opened 2017-03-08
How can I add headers to RxMoyaProvider ? 


jorgelp94:
Hi I've been using Moya with RxSwift but I need to add some http headers to the request and I've been looking through the rxswift documentation but I can't find this. Can anyone help me with this issue please?

gperdomor:
@jorgelp94 you can use a endpoint closure. Check this `https://github.com/Moya/Moya/blob/master/docs/Endpoints.md`

> You can add parameters or HTTP header fields in this closure. For example, we may wish to set our application name in the HTTP header fields for server-side analytics.

```swift
let endpointClosure = { (target: MyTarget) -> Endpoint<MyTarget> in
    let defaultEndpoint = MoyaProvider.defaultEndpointMapping(for: target)
    return defaultEndpoint.adding(newHTTPHeaderFields: ["APP_NAME": "MY_AWESOME_APP"])
}
let provider = MoyaProvider<GitHub>(endpointClosure: endpointClosure)
```

jorgelp94:
@gperdomor thanks! I'll look into it

Issue #510, opened 2016-06-16
Custom/Default HTTP headers


mrfarukturgut:
Hi. I am trying to build network layer for my app with moya, but i have this issue. I have like 25-30 endpoints and i need a custom header for almost all of them. (headers which containing token etc.). 
Can I set a default header or return header for all endpoints from TargetType enumaration? 

Or do I have to create a endpoint closure for all of them?

colinta:
Just one, for sure!  Here's how we do it in the Ello app:

###### ElloProvider.swift
```swift
    public static func endpointClosure(target: ElloAPI) -> Endpoint<ElloAPI> {
        let sampleResponseClosure = { return EndpointSampleResponse.NetworkResponse(200, target.sampleData) }

        let method = target.method
        let parameters = target.parameters
        let endpoint = Endpoint<ElloAPI>(URL: url(target), sampleResponseClosure: sampleResponseClosure, method: method, parameters: parameters, parameterEncoding: target.encoding)
        return endpoint.endpointByAddingHTTPHeaderFields(target.headers())
    }
```

###### ElloAPI.swift
```swift
    public func headers() -> [String: String] {
        var assigned: [String: String] = ["Accept": "application/json", "Accept-Language": "", "Content-Type": "application/json"]

        if self.requiresAnyToken {
            assigned += [
                "Authorization": AuthToken().tokenWithBearer ?? "",
            ]
        }
```

colinta:
[ElloProvider](https://github.com/ello/ello-ios/blob/master/Sources/Networking/ElloProvider.swift#L32)

[ElloApi](https://github.com/ello/ello-ios/blob/master/Sources/Networking/ElloAPI.swift#L455)

mrfarukturgut:
Thank you! I think there would be no problem, if I examine your library and get some inspirations :)

I am following Ello as much as I can by the way. You are doing great work! May the success be with you 👍

justinmakaila:
I'm going to close this issue. @mrfarukturgut, feel free to open it again if you still have questions!

PR #1067, opened 2017-04-26
Adds property headers to TargetType

pedrovereza:
Implements #1046 (and #1058).

How to add header fields to a request is a question that shows up in issues every once in a while (e.g. #510, #1003, #208). The answers always mention a custom `endpointClosure` to add the fields to the header request. In fact, [this is suggested in the docs as well](https://github.com/Moya/Moya/blob/master/docs/Endpoints.md#from-target-to-endpoint).

However, having the header fields specified in the `TargetType` makes the whole process easier (and more obvious). Which raises a couple of questions: 

1. Is there any reason for us not to specify header information inside `TargetType`?
2. Is the usage of headers not as common as I think it is and it's better to leave the `endpointClosure` as the official way? Assuming most people don't need to specify headers at all, adding it to `TargetType` just forces an implementation of headers that would return `nil`. 

If we decide to go forward with the change, we still need to:

- [x] Update Changelog
- [x] Update documentation

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -7,6 +7,7 @@
 - **Breaking Change** Updated `RxMoyaProvider.request` to return a [`Single<Request>`](https://github.com/ReactiveX/RxSwift/pull/1123).
 - **Breaking Change** Changed `Moya.Response`'s `response`to use an `HTTPURLResponse` instead of a `URLResponse`.
 - Updated the `RxSwift` version requirement to `3.3`.
+- **Breaking Change** Added `headers` to `TargetType`.
 
 # 8.0.3
 
--- a/Demo/Shared/GiphyAPI.swift
+++ b/Demo/Shared/GiphyAPI.swift
@@ -42,6 +42,10 @@ extension Giphy: TargetType {
             return "{\"data\":{\"id\":\"your_new_gif_id\"},\"meta\":{\"status\":200,\"msg\":\"OK\"}}".data(using: String.Encoding.utf8)!
         }
     }
+
+    public var headers: [String: String]? {
+        return nil
+    }
 }
 
 func animatedBirdData() -> Data {
--- a/Demo/Shared/GitHubAPI.swift
+++ b/Demo/Shared/GitHubAPI.swift
@@ -76,6 +76,9 @@ extension GitHub: TargetType {
             return "[{\"name\": \"Repo Name\"}]".data(using: String.Encoding.utf8)!
         }
     }
+    public var headers: [String: String]? {
+        return nil
+    }
 }
 
 public func url(_ route: TargetType) -> String {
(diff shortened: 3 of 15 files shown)


pedrovereza:
@Moya/contributors Let me know what you think :wink:

justinmakaila:
Loving it, seeing as I've implemented this like, 8x in different projects. Should we add a default of `nil` to the `TargetType`? I think that would make sense, given that most requests currently don't implement this.

pedrovereza:
@justinmakaila Yea, having it defaulting to `nil` makes sense to me. I actually tried that at first, but was having some trouble overriding the value in the tests. Probably missed something, will give it another try 😄

BasThomas:
🤔, I remember that we actually didn't want to implemented this ourselves.

justinmakaila:
@BasThomas I'd be interested in re-hashing that discussion here; I've had to implement this myself a handful of times and it's particularly annoying, given that MOST private APIs will require some sort of authentication and it seems a little ridiculous to have the `Endpoint` initializer be considerate of state, given that most of the state is already incorporated in `TargetType`.

pedrovereza:
@justinmakaila I think @BasThomas was talking about whether or not we should default `headers` to `nil` 😄 

@BasThomas I think the discussion happened in #861

justinmakaila:
@pedrovereza @BasThomas Ahh, sorry, I misunderstood. I remember the no defaults conversation... I mean, I guess this will get wrapped up in 9.0, so breaking changes are legit... I just got excited and wanted to start using it immediately :)

AndrewSB:
@pedrovereza: love this change, I've also overrode the endpointClosure multiple times for HTTP auth stuff. so I think adding it is a great change

I am still against a default implementation though, I think not forcing consumers of Moya to implement any part of the protocol obfuscates the simplicity of creating a provider.
Even if you aren't planning to use headers (or sampleData, as was mentioned in #861), I think a consumer should still have to explicitly decide what should be done

pedrovereza:
Great, so I think we all agree having no default is better 🙌

justinmakaila:
👍🏻

AndrewSB:
why has danger been failing so often on PRs?

pedrovereza:
@AndrewSB We configured it to [fail on missing changelog entries](https://github.com/Moya/Moya/blob/master/Dangerfile#L18)

AndrewSB:
@pedroverrza: can we set danger to have a more descriptive failure?
Or add it as a separate integration, not through circle ci? So danger &
tests can run and be seen independent of one another?

ashfurrow:
It's possible to run Danger on a separate CI, but Danger also posts a comment using the @MoyaBot with a description of what went wrong (it updates that comment on subsequent runs). Customizing the error message is totally doable, let me know if I can help.

pedrovereza:
I think I made all the updates needed in the docs. Can someone double check? :wink:

Issue #1078, opened 2017-05-10
Promoting awesome work around Moya.


sunshinejr:
Hey all! Recently there was a PR #1065, where @alexruperez wanted to let the world know his library is compatible with Moya. Personally, I think this is awesome, but unfortunately it was listed against "Community Extensions" in our Readme. There are many awesome projects around Moya, which may or may not be "an extension" (which can also be arbitrary). Can we do something about it, @Moya/contributors? My few propositions:

1. Leave the sections as it is, but accept everything related to Moya.
2. Change the name of the section to some more generic one and accept PRs accordingly. 
3. Make a new doc with everything around Moya (extensions/libraries compatible with Moya/abstractions/projects(?)) and link it from Readme.
4. Maybe use wiki [like Quick does it](https://github.com/Quick/Quick/wiki)?
5. Let's just stay with what we've got, no need for a change.

I think the goal is to make people know that we ❤️ all the work around Moya and make them feel appreciated in some way or another. The question is how to do it? 🤔

BasThomas:
Great idea, thanks for some initial thoughts @sunshinejr!

1. Seems OK, but might grow a bit big long term.
2. Seems even better than `1.`, without any downsides.
3. This seems like a great option to me. Easily expandable, and when linking from the Readme, it's also easily accessible.
4. I have a feeling wikis are not often used nor seen. I can't remember the last time I've looked a wiki for anything else than documentation (for reference, I used it when browsing [Freddy](https://github.com/bignerdranch/Freddy/wiki) documentation).
5. I think the other options are all better, so let's change things for the better. 😃

pedrovereza:
Thanks for starting the discussion @sunshinejr 🙌 

I vote for 3.

devxoul:
+1 for 3. `docs/Community` or `docs/Projects` seems great :tada:

alexruperez:
Thanks @sunshinejr! 😃

BenEmdon:
I like the idea of 3! 🙌

bchrobot:
+1 for 3 and for `docs/Community` 😃

justinmakaila:
I'm in for 3

sunshinejr:
Great, seems like another doc is a way to go :) Thank you all for the feedback! Gonna tackle that one later today.

BasThomas:
Let me know if I can help in any way :)

orta:
Might be time to make a website?

sunshinejr:
[...]

PR #1082, opened 2017-05-12
[Docs] Community projects.

sunshinejr:
As per #1078. Name **Community.md** is taken in our root directory so I didn't want to pick the same (the distinction might be key). Right now I've chosen **CommunityProjects.md** but I'm open to suggestions. 

Also I wasn't sure how can we word/point from **Readme.md** to **CommunityProjects.md**, so if you have any suggestions, please also let me know.

I've also added NetClient project from @alexruperez as a first Library 🎉 

Cheers!

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -252,21 +252,7 @@ responses.
 Community Extensions
 --------------------
 
-Moya has a great community around it and some people have created some very helpful extensions.
-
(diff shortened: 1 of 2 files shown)


pedrovereza:
I think this is a good first step, we can take improvements (if any) in PRs.

Thanks for the work @sunshinejr!

Issue #1072, opened 2017-05-04
Adding Moya to Swift Source Compatibility Test Suite


aksswami:
I really like using Moya in daily basis.
It's probably a good idea, to add Moya to "Swift Source Compatibility Test Suite".
https://swift.org/blog/swift-source-compatibility-test-suite
https://swift.org/source-compatibility/#current-list-of-projects
Alamofire, ObjectMapper are already in the list.

AndrewSB:
yup, that would be awesome! Do you want try to make a pull request against apple's repository?

aksswami:
Sure, I will try to make a PR against apple's repo in a day or two.

BasThomas:
Thanks for looking into this, @aksswami! 😀

aksswami:
@BasThomas I have create a PR [Moya Swift compat](https://github.com/apple/swift-source-compat-suite/pull/35). But I am getting error in `./check Moya` step. Not able to get much information from the error logs also. I have added comment for asking some help from [Swift Compat](https://github.com/apple/swift-source-compat-suite) team.

BasThomas:
This has been merged on Apple's side. @aksswami, thank you so much for picking this up!

sunshinejr:
@BasThomas so it was, in fact, this `.git` suffix bug? Maybe we should let the world know somehow, this was found by an accident 🤔

BasThomas:
I think so? Maybe better to ask over at https://github.com/apple/swift-source-compat-suite/pull/35; @lplarson fixed it (by updating just the hash as far as I understand, so I think the `.git` suffix did the trick.

sunshinejr:
Huh, now where should we raise an issue, on SPM or SPM-compat? 🤔

BasThomas:
Good question. I'd just start from the Compat PR, and go from there?

BasThomas:
Or going straight to the mailing list is also an option.

sunshinejr:
Right, gonna just comment on the PR what was wrong in our project then. :+1:

PR #1093, opened 2017-05-20
[SPM] Git suffix + .gitignore

sunshinejr:
In this PR:

1. Add .git suffixes. Rationale: [Stencil#84](https://github.com/kylef/Stencil/pull/84) and [Stencil#101](https://github.com/kylef/Stencil/issues/101). Basically really similar situation in our case. `ReactiveSwift` has `Result` as a dependency with `.git` suffix in `Package.swift`, and we also have `Result` dependency - but without `.git` in `Package.swift`. It might also resolve our problem in [SPM-compat](https://github.com/apple/swift-source-compat-suite/pull/35).
2. Add `Package.pins` to `.gitignore`, as per [SPM docs](https://github.com/apple/swift-package-manager/blob/master/Documentation/Usage.md#automatic-pinning).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@ Demo/Pods
 # Swift Package Manager
 .build
 Packages
+Package.pins
--- a/Package.swift
+++ b/Package.swift
@@ -16,10 +16,10 @@ let package = Package(
         )
     ],
     dependencies: [
-        .Package(url: "https://github.com/Alamofire/Alamofire", majorVersion: 4),
-        .Package(url: "https://github.com/ReactiveCocoa/ReactiveSwift", majorVersion: 1),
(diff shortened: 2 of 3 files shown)


BasThomas:
Nice! Could this also fix #1072 / https://github.com/apple/swift-source-compat-suite/pull/35? Seems like this might have been the issue there.

sunshinejr:
It _might_, we will have to test it once this is merged.

BasThomas:
Should we also add the git suffix to the Readme SPM instructions for Moya itself?

BasThomas:
Other than that, this LGTM.

sunshinejr:
Oh, right! Thanks, I'm on it. 👍

sunshinejr:
Done 👍

Issue #1071, opened 2017-05-04
[RFC]: extend MoyaBot to automatically follow up with issues & PRs where we're waiting for author response


AndrewSB:
I recently saw @KrauseFx's talk on scaling open source communities (great talk, I'd recommend it to everyone who's doing anything at all in open source). https://news.realm.io/news/tryswift-felix-krause-scaling-open-source-communities-github-management

One of my big take aways was fastlane's fastlane-bot. Which would automatically follow up with issues and close them if there was no response.
Right now, a couple of contributors have been manually doing that, but I think it would be awesome if we could add some of fastlane-bot's functionality to our bot so we don't have to politely close stale issues

Any thoughts from other contributors? Does anyone want to take a shot at implementing that for MoyaBot?

BasThomas:
I really like this - though looking at fastlane-bot's [repository](https://github.com/fastlane/issue-bot), I have no idea how it would work implementing it. Does anybody have an idea, or should we go ahead and file an issue there?

sunshinejr:
We can wait for [Peril](https://github.com/danger/peril) from @orta. It looks really promising. It is not fully stable yet, but it can already post on issues AFAIK.

AndrewSB:
Peril does look really nice, let's wait for that to be usable

pedrovereza:
Just came across this: https://github.com/probot/stale 🤔

sunshinejr:
This one is really cool! I think we can try to do it. What do you think, @Moya/contributors?

BasThomas:
Sure, this would be really nice! Re: MoyaBot: I think you should ping @ashfurrow to ask if that's possible. :)

orta:
We just merged it in on CocoaPods - https://github.com/CocoaPods/CocoaPods/pull/6745 - I can't really compete with GitHub on resources for this kind of project, so I wouldn't wait on peril

sunshinejr:
Ye, and this one is looking cool. The question is if we want/can use MoyaBot for that. Also, we would need someone to do a PR with config. The values we will probably discuss in a PR (if they are off).

AndrewSB:
😆
I'm happy this is happening!

sunshinejr:
I was talking with Ash and we've figured that MoyaBot would need a write access so it would be better to use different bot or someone else's access token. Thus, it might be better just to install this stale bot and go from that. I'm gonna do a PR with some default values for config so we can discuss it there 👍

BasThomas:
[...]

PR #1101, opened 2017-05-25
[StaleBot] Add first version of stale.yml config.

sunshinejr:
Rationale: #1071. 

~~(I'm gonna comment in the PR my opinions, bear with me for a sec :D)~~
**Edit:** Finished my feedback, I'd love to hear your opinion on this!

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -25,7 +25,7 @@ unmarkComment: false
 
 # Comment to post when closing a stale Issue or Pull Request. Set to `false` to disable
 closeComment: >
-  This issue has been auto-closed because there hasn't been any activity for a few months.
+  This issue has been auto-closed because there hasn't been any activity for a 74 days.
   However, we really appreciate your contribution, so thank you for that! 🙏
   Also, feel free to [open a new issue](https://github.com/Moya/Moya/issues/new) if you still experience this problem :+1:
 
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,7 +1,7 @@
 # Configuration for probot-stale - https://github.com/probot/stale
 
 # Number of days of inactivity before an Issue or Pull Request becomes stale
-daysUntilStale: 60
+daysUntilStale: 45
 
 # Number of days of inactivity before a stale Issue or Pull Request is closed
 daysUntilClose: 14
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -9,23 +9,25 @@ daysUntilClose: 14
 # Issues or Pull Requests with these labels will never be considered stale
 exemptLabels:
   - "enhancement"
+  - "confirmed bug"
+  - "chore"
 
 # Label to use when marking as stale
-staleLabel: "awaiting input"
+staleLabel: "no recent activity"
 
 # Comment to post when marking as stale. Set to `false` to disable
 markComment: >
-  This issue has been automatically marked as stale because it has not had
-  recent activity. It will be closed if no further activity occurs. Thank you
-  for your contributions.
+  This issue has been marked as stale because it has not had
+  recent activity. It will be closed if no further activity occurs. 
 
 # Comment to post when removing the stale label. Set to `false` to disable
 unmarkComment: false
 
 # Comment to post when closing a stale Issue or Pull Request. Set to `false` to disable
 closeComment: >
-  This issue will be auto-closed because there hasn't been any activity for a few months.
-  Feel free to [open a new one](https://github.com/Moya/Moya/issues/new) if you still experience this problem :+1:
+  This issue has been auto-closed because there hasn't been any activity for a few months.
+  However, we really appreciate your contribution, so thank you for that! 🙏
+  Also, feel free to [open a new issue](https://github.com/Moya/Moya/issues/new) if you still experience this problem :+1:
 
 # Limit to only `issues`
 only: issues
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,31 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 60
+
+# Number of days of inactivity before a stale Issue or Pull Request is closed
+daysUntilClose: 14
+
+# Issues or Pull Requests with these labels will never be considered stale
+exemptLabels:
+  - "enhancement"
+
+# Label to use when marking as stale
+staleLabel: "awaiting input"
+
+# Comment to post when marking as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed if no further activity occurs. Thank you
+  for your contributions.
+
+# Comment to post when removing the stale label. Set to `false` to disable
+unmarkComment: false
+
+# Comment to post when closing a stale Issue or Pull Request. Set to `false` to disable
+closeComment: >
+  This issue will be auto-closed because there hasn't been any activity for a few months.
+  Feel free to [open a new one](https://github.com/Moya/Moya/issues/new) if you still experience this problem :+1:
+
+# Limit to only `issues`
+only: issues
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,31 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 60
```
sunshinejr: In my opinion this is the most reasonable - one month is too short, 3 months kinda too long?

.github/stale.yml
```diff
@@ -0,0 +1,31 @@
...
+
+# Number of days of inactivity before a stale Issue or Pull Request is closed
+daysUntilClose: 14
```
sunshinejr: On the other hand, if we introduce this bot now, we might want to give more time for users to respond to issues. Thus, making me change the default of 7 days to 14 for days until close. I think we may adjust the values as we go, but at the beginning it might be better to start off high.

.github/stale.yml
```diff
@@ -0,0 +1,31 @@
...
+
+# Issues or Pull Requests with these labels will never be considered stale
+exemptLabels:
+  - "enhancement"
```
sunshinejr: Wasn't sure if this should be the only one white-listed labels, thoughts?

.github/stale.yml
```diff
@@ -0,0 +1,31 @@
...
+
+# Label to use when marking as stale
+staleLabel: "awaiting input"
```
sunshinejr: And also wasn't sure with naming in this one, took the default here.

.github/stale.yml
```diff
@@ -0,0 +1,31 @@
...
+
+# Comment to post when marking as stale. Set to `false` to disable
+markComment: >
```
sunshinejr: This is also default comment, someone with better wording skills might want to take a stab at this one :D

.github/stale.yml
```diff
@@ -0,0 +1,31 @@
...
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed if no further activity occurs. Thank you
+  for your contributions.
+
+# Comment to post when removing the stale label. Set to `false` to disable
+unmarkComment: false
```
sunshinejr: I think it's meaningless to have a comment that says "label was removed", since it will be seen after any comment appears in the issue. This is just my opinion, though.

.github/stale.yml
```diff
@@ -0,0 +1,31 @@
...
+
+# Comment to post when closing a stale Issue or Pull Request. Set to `false` to disable
+closeComment: >
+  This issue will be auto-closed because there hasn't been any activity for a few months.
+  Feel free to [open a new one](https://github.com/Moya/Moya/issues/new) if you still experience this problem :+1:
```
sunshinejr: Here, I was contemplating if suggestion to create a new issue will be better than reopening current one. I might like latter, but I'd love to hear your opinions.

.github/stale.yml
```diff
@@ -0,0 +1,31 @@
...
+
+# Limit to only `issues`
+only: issues
```
sunshinejr: Also, wasn't sure here, but I've figured, that we don't have a problem with PRs right now, but more with Issues.

BasThomas:
Maybe not for now, but I was thinking: can we lock issues with x comments after being closed?

sunshinejr:
Looking at the [lib file](https://github.com/probot/stale/blob/master/lib/stale.js), I think there is no `lock()` method and any other that are needed for this to happen. Although I love the idea and we can try to add an Issue or PR if someone is more skilled in .js than me 😅

.github/stale.yml
pedrovereza: We may want to include `chore` as well 🤔

.github/stale.yml
pedrovereza: Would it better the specify the exact number of days we waited for a response?

.github/stale.yml
AndrewSB: as well as `confirmed bug` those often take a while to be resolved

.github/stale.yml
AndrewSB: LGTM

.github/stale.yml
AndrewSB: I think it would be, and also add a sentence saying something along the lines of "please open a new issue to further discuss the issue, if desired"

AndrewSB:
@BasThomas read my mind. Locking would be awesome, let's create that issue 😄

BasThomas:
GitHub apparently doesn't support locking (yet)... https://github.com/probot/stale/issues/16#issuecomment-291862121

BasThomas:
... but there seems to be a plugin that *does* work? I can't really find out how this would work: https://github.com/lee-dohm/lock.

BasThomas:
Thanks for all this work though, @sunshinejr! This seems like a great addition. 🎉

.github/stale.yml
BasThomas: I'm not sure on reopening vs opening a new one. On wording: "This issue has been auto-closed [..]" I think would better reflect the action.

.github/stale.yml
BasThomas: What about just "stale"? I can't really think of anything better, although it seems like something's just on the tip of my tongue. "awaiting input" seems... off.

BasThomas: "no recent activity" maybe?

.github/stale.yml
BasThomas: "This issue has been marked as stale because it has not had any recent activity. If no further activity occurs, it will be closed. Please feel free to open a new issue / Please feel free to reopen this issue if needed."

At least the "automatically" seems redundant. Also, for issues, "Thank you for you contributions" seems strange.

BasThomas: It would be nice to thank them for taking the time to file an issue, though.

.github/stale.yml
pedrovereza: I'm ok with reopening, creating a new one would probably be just duplication.

AndrewSB: hmm, we wouldn't be able to lock the issue then.
Should we keep locking as a manual thing, not something we do to stale issues?

I'm just afraid of someone responding to a closed issue later, and one of us having to pay attention to that and re-opening the issue.
I feel as if that goes against what we're trying to automate away with this bot

.github/stale.yml
sunshinejr: I would agree if we had that much Issues, but we don't, really. On the first page we have an issue from December last year, so in my opinion 10/30 will be too short.

.github/stale.yml
sunshinejr: I agree! Both `chore` & `confirmed bug` are a great additions.

.github/stale.yml
sunshinejr: I really like `no recent activity`. Gonna change to that one!

sunshinejr:
I've updated the PR:
- Added `confirmed bug` & `chore` in exempt labels.
- Renamed stale label to `no recent acitivity`
- Changed the comments for marking & closing.

Let me know what do you guys think :)

.github/stale.yml
sunshinejr: +1 to creating a new one just because of locking, so let's stay with what we've got ^^

.github/stale.yml
AndrewSB: That's true, but if I think of issues that are opened now, (to me) it feels like an issue without a response for a month feels stale

I would love to hear what other people think - @BasThomas especially (since you handle closing so many of our current stale issues 😋)

Totally fine with 60 otherwise, Im just trying to make sure our bot is useful

sunshinejr: So, how are we feeling about this one? @AndrewSB, @pedrovereza, @BasThomas

BasThomas: Let's just go with 45 in that case and adjust if needed?

pedrovereza: I'm ok going with what we have now. We probably won't get to the optimal number before we see the bot running and judge how useful it's being :wink:

AndrewSB: I'm down to go with 45 and then adjust in the future

sunshinejr:
Should be good to go 👍

PR #1103, opened 2017-05-25
[Tests] Add separate SwiftLint config & clean up.

sunshinejr:
Last few PRs were to the Tests directory and thus our MoyaBot reminded me that we need to take care of this directory with SwiftLint as well. Of course, these are test cases, so the rules in this directory might been a little bit loosen up. That's why I've added another config file with less strict rules in there. If you don't agree with (not) having some rule on the list, let me know. 

I've also fixed most of the errors, but there are 2 left. These are about file length & type length, but I believe we don't need to exclude these rules. I'm more keen to refactor a bit `MoyaProviderSpec`, because there is a room for that.


Tests/.swiftlint.yml
```diff
@@ -0,0 +1,12 @@
+disabled_rules:
+  - line_length
+  - type_name
+  - valid_docs
+  - force_unwrapping
+  - force_cast
+  - force_try
+  - function_body_length
```
sunshinejr: I've decided to add this rule because it's tedious to comment every `spec()` func in the tests.

Tests/.swiftlint.yml
```diff
@@ -0,0 +1,12 @@
...
+  - cyclomatic_complexity
```
sunshinejr: And this one because we have many workaround that we need to do for testing, so we would also have a lot of comments disabling it.

PR #1105, opened 2017-05-26
[Documentation] Update Releasing.md instructions

sunshinejr:
So I was doing a release today and found out that the instructions weren't as clear as I thought it would be. I've added some examples, note for `ZSH` users & format it a little bit. 

[Preview](https://github.com/Moya/Moya/blob/releasing-patch/docs/Releasing.md)

### DIFF
--- a/docs/Releasing.md
+++ b/docs/Releasing.md
@@ -3,6 +3,8 @@ Releasing
 
 (_Note: This document is a reference for people with push access to Moya and to [CocoaPods](https://cocoapods.org/pods/Moya)._)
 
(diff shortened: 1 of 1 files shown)


BasThomas:
🤔, never saw this, but it seems like we're using `----` style and `#` style headers all around the place. I can open  a PR soon changing these all to the `#` notation? That, at least to me, is a bit cleaner. Thoughts?

sunshinejr:
Yeah, good idea. Let's make it consistent :)

PR #1106, opened 2017-05-26
Use consistent heading styles, fix indentation + spelling

BasThomas:


docs/Authentication.md
```diff
@@ -23,18 +21,17 @@ which is usually not necessary. This might be a better idea:
 
 ```swift
 let provider = MoyaProvider<YourAPI>(plugins: [CredentialsPlugin { target -> URLCredential? in
-    switch target {
-      case .targetThatNeedsAuthentication:
-        return URLCredential(user: "user", password: "passwd", persistence: .none)
-      default:
-        return nil
+        switch target {
```
sunshinejr: shouldn't the indentation here be smaller?

sunshinejr:
This is great 👍 Also, can we maybe make a SwiftLint rule to avoid `=====` & `------` and use `#` in .md files? 🤔

docs/Authentication.md
BasThomas: Wasn't sure here - as it is a closure, with the `])` on the last line, I thought this would be better than two `}`'s on the same space vertically. 🤔

sunshinejr: Oh, okay, let's just leave it as it is then 👍

BasThomas:
Let's open a separate one for that. 😃 I'd need to dive into SwiftLint for that though, so if you'd like to do that, feel free.

BasThomas:
Does SwiftLint actually work with non-`.swift` files? 😮

sunshinejr:
Haha, probably not, I've meant Danger but well, that happened 😅

BasThomas:
Hah, of course. Let me think of a nice way to check.

PR #1112, opened 2017-06-04
Update readme regarding json fixture files

pawurb:
Hello. I could not get the json fixture file working for specs when following the example provided in readme. I updated readme to use the approach that worked for me. Looks like `Data(base64Encoded: path)` only encodes the String path into data, and does not actually read json data from the file.


ashfurrow:

Issue #1119, opened 2017-06-12
How do I create a path with a '?' within the string?


pete183:
Moya Version - 8.0.5

When I create a string for my path like:

```
"/id/?access_token=ACCESSTOKEN"
```

The '?' gets converted to '%3F' in the url which is sent and the server can't read this. Is there a way to send the '?' rather than the encoded unicode?

**What I want to send as the URL:**

```
https://api.test.co.uk/v1-dev/test/3?access_token=e02fkjhbjb4faf363jn0bcbeca0fad4aff
```

**What is actually sent as the URL:**

```
Moya_Logger: [12/06/2017 11:31:20] Request: https://api.test.co.uk/v1-dev/test/3%3Faccess_token=e02fkjhbjb4faf363jn0bcbeca0fad4aff
```

BasThomas:
What's your `parameterEncoding`?

haritowa:
Just use params instead of embedding em on your own

pete183:
@BasThomas The parameterEncoding is 

> URLEncoding.default

@haritowa The server needs the access token and limits to be in the url. Would using params allow for this?

pedrovereza:
@pete183 Yes, `URLEncoding.default` will put your parameters in the url :wink:

pete183:
@pedrovereza @haritowa Thanks for the help, it works now!

sunshinejr:
Just to add to it: `URLEncoding.default` will behave differently in `GET` and `POST` requests. In `GET` method you will, in fact, get parameters in URL, @pete183. In `POST` on the other hand, you will get parameters in body. You can specify if you want it all the time in the URL or body using different properties of `URLEncoding` - more on the matter [here](https://github.com/Alamofire/Alamofire/blob/master/README.md#parameter-encoding).

pedrovereza:
@sunshinejr Please check #1120 😅

sunshinejr:
We don't have a method exclusively for this behavior, but you can use custom `ParameterEncoding`, that would decide whether the parameter should be in body/url/header. You can find instructions on creating your own encoding [here](https://github.com/Alamofire/Alamofire/blob/master/README.md#custom-encoding).

**Edit:** Now that I think about that, that would be a good addition to our examples. If you end up doing your own `ParameterEncoding`, we would love to have it added to our examples directory.

pete183:
@sunshinejr 

I think something like this would work:

```
public var parameters: [String: Any]? {
[...]

PR #1120, opened 2017-06-12
[Documentation] Fix example on how to send params in URL on POST

pedrovereza:
When doing a POST request, we should use `URLEncoding.queryString` instead of `URLEncoding.default` (as explained by @sunshinejr in [#1119](https://github.com/Moya/Moya/issues/1119#issuecomment-307929162))

### DIFF
--- a/docs/Examples/Basic.md
+++ b/docs/Examples/Basic.md
@@ -52,9 +52,9 @@ extension MyService: TargetType {
     var parameterEncoding: ParameterEncoding {
         switch self {
         case .zen, .showUser, .showAccounts:
-            return URLEncoding.default // Send parameters in URL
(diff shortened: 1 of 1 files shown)


docs/Examples/Basic.md
```diff
@@ -49,11 +49,12 @@ extension MyService: TargetType {
...
         }
     }
-
     var parameterEncoding: ParameterEncoding {
         switch self {
-        case .zen, .showUser, .showAccounts, .updateUser:
+        case .zen, .showUser, .showAccounts:
             return URLEncoding.default // Send parameters in URL
+        case .updateUser:
+            return URLEncoding.queryString // Send parameters in URL
```
sunshinejr: Let's specify in comment that it would **always** send parameters in URL here, and above will depend on a http method (`GET`, `HEAD` and `DELETE` being in URL, and in body otherwise)

pedrovereza: done :wink:

PR #1127, opened 2017-06-15
Enable "Allow app extension API only" flag

### LINKED ISSUE #1110: Support app extension API only
Hi, 

I want use Moya in my framework with **Allow app extension API only** flag enabled, but I have a Xcode warning that says: **Linking against dylib not safe for use in application extension**. 
I think this is due to the fact that Moya haven't app extension flag enabled.

How can I solve? 
Thank you for your help

BasThomas:
Fixes #1110.

Was there a reason this was not enabled before?


Changelog.md
```diff
@@ -1,6 +1,11 @@
 # Next
 
+- **Breaking Change** Renamed all occurrences of `queue` to `callbackQueue`.
+- Add optional callback queue parameter to reactive providers.
+- Enabled the "Allow app extension API only" flag.
```
pedrovereza: The first two entries are meant for 9.0

BasThomas: Oh, oops. Started working on this and it seems like I did so from the wrong branch. I'll remove those.

Issue #1135, opened 2017-06-22
Native support for Encodable types in Swift 4


Dschee:
I'm currently writing an app to be **released after the official release of iOS 11** that'll only support iOS 11 onwards. So naturally I'm using the **Xcode 9** betas and I've also decided to write code in **Swift 4** from the beginning. What I try right now is to combine the great new possibilities with the **`Encodable` and `Decodable` protocols with Moya**. While there's #1118 for adding native support for the `Decodable` protocol, I'm currently **struggling with the `Encodable` protocol**.

My idea was to **encapsulate the request parameters into it's own type** (they are completely different than the response data, so it's `Encodable` only):

```swift
struct SignInRequest: Encodable {
    // MARK: - Sub Types
    struct AppInfo: Encodable {
        let clientName: String
        let clientVersion: String
        let platformName: String
        let platformVersion: String
        let pushToken: String?
    }

    // MARK: - Stored Instance Properties
    let appInfo: AppInfo
    let username: String
    let password: String
}
```

My API target looks like this:

```swift
enum MyTarget {
    case signIn(SignInRequest)
    case profile(String) // example for classical approach
}
```

Now when making `MyTarget` conform to `TargetType` I have to provide the `parameters` and `parameterEncoding` computed properties. But the `parameters` have a return type of `[String : Any]?` which I can't create using `JSONEncoder` and also the API of `ParameterEncoding` is not flexible enough to write an own type that would conform to it and use the new `Encodable` inference from Swift 4.

**What I need** would be something like this as a _replacement_ of `parameters` and `parameterEncoding`:

```swift
var requestData: RequestData? {
   switch self {
   case .signIn(let request):
       return .jsonEncodable(request)

   case .profile(let id):
       return .parameterEncoding(["id": id], URLEncoding.default)
   }
}
```

For this to work, we would need to add a `RequestData` type like so:

```swift
enum RequestData {
    case jsonEncodable(Encodable)
    case parameterEncoding([String: Any], ParameterEncoding?) // a default parameter encoding could be provided in the TargetType, too (with an empty default implementation)
}
```

What do you think? I think we should definitely support `Encodable` types as first class citizens as many people will use them instead of mappers. The only question is, how can we make the `TargetType` flexible enough to support them?

Any ideas around how I could get this work with Moya 8.0 would be appreciated as well (for the time until this issue is fixed). Or did I miss something and can use `Encodable` already somehow in a good way? Cause AFAIK Moya is intended to be as flexible as possible.

[...]

Issue #314, opened 2015-11-23
Multiple ParameterEncoding


shinyaohira:
I need to make a POST request with query string and json body. How can I pass each parameters to Moya?

ashfurrow:
Hey there! Great question. We cover a bit about parameter encoding [in the docs](https://github.com/Moya/Moya/blob/master/docs/Endpoints.md#from-target-to-endpoint-), but the short-and-simple answer is that you need to pass a custom `endpointClosure` when you create the provider. It takes a target, which you `switch` on to customize individual endpoints. 

The parameter encoding you're describing is pretty custom – query string in the URL and a JSON body – so you should use the `.Custom` parameter encoding type. It takes a closure that modifies the `NSURLRequest` and returns it. You can see the source [here](https://github.com/Moya/Moya/blob/731db29d6a68d350ef97ad059e881f361dc35e3f/Moya/Moya.swift#L38-L57). 

Like I said, it's a pretty custom thing so I'm afraid to say we don't have many examples on it yet. We'll help you figure it out, though! And once we do, it would be :100: if you wanted to submit a pull request that adds some more examples to the docs based on what your experience :cake:

shinyaohira:
Thanks, I'll try it!

ashfurrow:
Hey, just following up! Anything I can to do help?

shinyaohira:
Like this?

```swift
let compositeParameterEncodingClosure: (URLRequestConvertible, [String: AnyObject]?) -> (NSMutableURLRequest, NSError?) = { request, parameters in
    guard let parameters = parameters else {
        return (request.URLRequest, nil)
    }

    let (queryRequest, _) = ParameterEncoding.URL.encode(request, parameters: parameters["query"] as! [String: AnyObject])
    let (bodyRequest, _) = ParameterEncoding.JSON.encode(request, parameters: parameters["body"] as! [String: AnyObject])
    let compositeRequest = queryRequest.mutableCopy() as! NSMutableURLRequest
    compositeRequest.HTTPBody = bodyRequest.HTTPBody
    return (compositeRequest, nil)
}
```

ashfurrow:
Looks good to me! I must admit I've not tried this before, but it looks like it should work :+1:

shinyaohira:
Thanks!

gfpacheco:
The current version of this solution is:

```swift
[...]

PR #1136, opened 2017-06-24
[WIP] Support Encodable types in TargetType

Dschee:
This is an exploration of my idea explained in #1135 which fixes the same issue.

Unfortunately, I'm getting an error which renders this PR useless as it is right now:
> Cannot invoke 'encode' with an argument list of type '((Encodable))'

This error appears in places such as 
```swift
request.httpBody = try? JSONEncoder().encode(encodable)
```
since the `encodable` parameter is of a generic type (`Encodable`).

Does anybody have a good solution for preventing this issue? Then I can continue exploring this.

Note that I have removed all convenience methods for `parameters` and `parameterEncoding` as a first step. We could (and probably should) reintroduce them later. Also I have not yet included a solution for the side-requirement of fixing #314 (which we could tackle using the same approach as well as explained [here](https://github.com/Moya/Moya/issues/1135#issuecomment-310677003)). I think we should discuss that in a subsequent PR.

**TODOs:**
- [x] Explain the basic idea using code
- [ ] Make the code usable
- [ ] Update the Docs
- [ ] Update the Tests
- [ ] Write a migration guide
- [ ] Write a changelog entry

### DIFF
--- a/Sources/Moya/Endpoint.swift
+++ b/Sources/Moya/Endpoint.swift
@@ -21,29 +21,21 @@ open class Endpoint<Target> {
     open let url: String
     open let method: Moya.Method
     open let sampleResponseClosure: SampleResponseClosure
-    open let parameters: [String: Any]?
-    open let parameterEncoding: Moya.ParameterEncoding
+    open let requestData: RequestData?
     open let httpHeaderFields: [String: String]?
 
     /// Main initializer for `Endpoint`.
     public init(url: String,
                 sampleResponseClosure: @escaping SampleResponseClosure,
                 method: Moya.Method = Moya.Method.get,
-                parameters: [String: Any]? = nil,
-                parameterEncoding: Moya.ParameterEncoding = URLEncoding.default,
+                requestData: RequestData? = nil,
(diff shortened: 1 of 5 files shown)


sunshinejr:
I think we can close this one in favor of #1147.

Dschee:
For now, yes. But note that this included **real** support for the Encodable types, so we might want to add something like that later. But this, yes, this is now deprecated and should be closed. Let's wait what Alamofire does in order to support the `Encodable` type as September comes closer ...

sunshinejr:
Exactly what I was thinking 👍

Dschee:
I've created #1148 which is a simple reincarnation of this PR, but based on the approach we have introduced in #1147. I know we might not need it at all if Alamofire integrates it somehow themselves. But if they don't we might need it.

And I'd like learn why I get the compiler error in there and how to work around it anyways, so I've also created it because of technical reasons. I'd like to at least play around with that API and see if it works well from the consumer side. Maybe the Alamofire guys introduce support, but we don't like it? You never know ...

PR #1142, opened 2017-06-25
Add testing.md

### LINKED ISSUE #1141: Add testing instructions
To run tests, we need to install our dependencies via Carthage first (ref: https://github.com/Moya/Moya/pull/1140#issuecomment-310927240). Add this to the documentation.

BasThomas:
Closes #1141

### DIFF
--- a/docs/Development.md
+++ b/docs/Development.md
@@ -1,4 +1,6 @@
-# Testing
+# Development
+
+## Testing
 
 To run tests locally, you will need to download Moya's dependencies.
 To do so, run `carthage update --platform iOS` and take a nap, waiting for it to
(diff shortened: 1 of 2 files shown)


sunshinejr:
This is great idea, thank you, @BasThomas! Although, maybe rename it do something more verbose about it being in the "development" phase, rather than testing in Moya itself (`sampleData`, `stubbing` etc.)? Like `Development.md`? These are the steps that we can really use for development as well :D 

Also maybe we need to split documentation on usage of Moya and everything around it (like Community, Code of Conduct, Development, Releasing etc.)? 🤔

BasThomas:
Both seem like a great idea. I'll create another issue for your second point! :)

PR #1147, opened 2017-06-28
Flatten enums & parameters into extended Task type

Dschee:
This solution was dicussed starting here:
https://github.com/Moya/Moya/issues/1135#issuecomment-310913115

Specifically this PR does the following:
- [x] Remove `parameters` and `parameterEncoding` from everywhere
- [x] Extend `Task` with different types of request data (e.g. `data` & `compositeEncoded`)
- [x] Flatten the `DownloadType` and `UploadType` enums into the `Task` type
- [x] Use those new task cases to set url params & http body appropriately
- [x] Update Tests
- [x] Update Docs
- [x] Write a migration guide
- [x] Add Changelog entry
- [x] Handle the `TODO:` comments (exception handling) appropriately

This solves the most apparent part of #1135 and it probably also solves #314.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,6 +4,10 @@
 # 9.0.0-alpha.1
 
 - **Breaking Change** Added support to get the response (if any) from `MoyaError`.
+- **Breaking Change** Replaced `parameters` & `parameterEncoding` in `TargetType` with extended `Task` cases.
+- **Breaking Change** Flattened `UploadType` and `DownloadType` into Task cases.
+- **Breaking Change** Updated `RxMoyaProvider.request` to return a [`Single<Request>`](https://github.com/ReactiveX/RxSwift/pull/1123).
+- **Breaking Change** Changed `Moya.Response`'s `response`to use an `HTTPURLResponse` instead of a `URLResponse`.
 - **Breaking Change** Added `headers` to `TargetType`.
 - **Breaking Change** Updated `RxMoyaProvider.request` to return a [`Single<Request>`](https://github.com/ReactiveX/RxSwift/pull/1123).
 - **Breaking Change** Updated `Moya.Response`'s `response`to use an `HTTPURLResponse` instead of a `URLResponse`.
@@ -17,10 +21,6 @@
 - Add optional callback queue parameter to reactive providers.
 
 - Bumped minimum version of ReactiveSwift to 2.0.
-- **Breaking Change** Replaced `parameters` & `parameterEncoding` in `TargetType` with extended `Task` cases. Migration guide available in Readme.
-- **Breaking Change** Flattened `UploadType` and `DownloadType` into Task cases. Migration guide available in Readme.
-- **Breaking Change** Replaced `parameters` & `parameterEncoding` in `TargetType` with extended `Task` cases.
-- **Breaking Change** Flattened `UploadType` and `DownloadType` into Task cases.
 - Add optional callback queue parameter to reactive providers.
 - Enabled the "Allow app extension API only" flag.
 - Added an optional `requestDataFormatter`in `NetworkLoggerPlugin` to allow the client to interact with the request data before logging it.
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,8 +6,6 @@
 # 9.0.0-alpha.1
 
 - **Breaking Change** Added support to get the response (if any) from `MoyaError`.
-- **Breaking Change** Updated `RxMoyaProvider.request` to return a [`Single<Request>`](https://github.com/ReactiveX/RxSwift/pull/1123).
-- **Breaking Change** Changed `Moya.Response`'s `response`to use an `HTTPURLResponse` instead of a `URLResponse`.
 - **Breaking Change** Added `headers` to `TargetType`.
 - **Breaking Change** Updated `RxMoyaProvider.request` to return a [`Single<Request>`](https://github.com/ReactiveX/RxSwift/pull/1123).
 - **Breaking Change** Updated `Moya.Response`'s `response`to use an `HTTPURLResponse` instead of a `URLResponse`.
@@ -16,11 +14,6 @@
 - **Breaking Change** Removed parameter name in `requestWithProgress` for `ReactiveSwiftMoyaProvider`.
 - **Breaking Change** Removed deprecated in Moya 8.0.0: `Moya.Error`, `endpointByAddingParameters(parameters:)`, `endpointByAddingHttpHeaderFields(httpHeaderFields:)`, `endpointByAddingParameterEncoding(newParameterEncoding:)`, `endpointByAdding(parameters:httpHeaderFields:parameterEncoding)`, `StructTarget`, `filterStatusCodes(range:)`, `filterStatusCode(code:)`, `willSendRequest(request:target:)`, `didReceiveResponse(result:target:)`, `ReactiveCocoaMoyaProvider`, `ReactiveSwiftMoyaProvider.request(token:)`. 
 - Added optional callback queue parameter to reactive providers.
-- Updated the `RxSwift` version requirement to `3.3`.
-- Fixed a bug where you would have two response events in `requestWithProgress` method on ReactiveSwift module.
-- Bumped minimum version of ReactiveSwift to 2.0.
-- Enabled the "Allow app extension API only" flag.
-- Added an optional `requestDataFormatter`in `NetworkLoggerPlugin` to allow the client to interact with the request data before logging it.
 - Added public `URL(target:)` initializator that creates url from `TargetType`.
 - Added an optional `requestDataFormatter`in `NetworkLoggerPlugin` to allow the client to interact with the request data before logging it.
 - Updated minimum version of `RxSwift` to `3.3`.
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,11 +1,11 @@
 # Next
 - Added Swift 4.0 support.
+- **Breaking Change** Replaced `parameters` & `parameterEncoding` in `TargetType` with extended `Task` cases.
+- **Breaking Change** Flattened `UploadType` and `DownloadType` into `Task` cases.
 
 # 9.0.0-alpha.1
 
 - **Breaking Change** Added support to get the response (if any) from `MoyaError`.
-- **Breaking Change** Replaced `parameters` & `parameterEncoding` in `TargetType` with extended `Task` cases.
-- **Breaking Change** Flattened `UploadType` and `DownloadType` into `Task` cases.
 - **Breaking Change** Updated `RxMoyaProvider.request` to return a [`Single<Request>`](https://github.com/ReactiveX/RxSwift/pull/1123).
 - **Breaking Change** Changed `Moya.Response`'s `response`to use an `HTTPURLResponse` instead of a `URLResponse`.
 - **Breaking Change** Added `headers` to `TargetType`.
--- a/Changelog.md
+++ b/Changelog.md
@@ -18,7 +18,6 @@
 - Added optional callback queue parameter to reactive providers.
 - Updated the `RxSwift` version requirement to `3.3`.
 - Fixed a bug where you would have two response events in `requestWithProgress` method on ReactiveSwift module.
-- Add optional callback queue parameter to reactive providers.
 - Bumped minimum version of ReactiveSwift to 2.0.
 - Enabled the "Allow app extension API only" flag.
 - Added an optional `requestDataFormatter`in `NetworkLoggerPlugin` to allow the client to interact with the request data before logging it.
--- a/Changelog.md
+++ b/Changelog.md
@@ -12,6 +12,15 @@
 - **Breaking Change** Removed parameter name in `requestWithProgress` for `ReactiveSwiftMoyaProvider`.
 - **Breaking Change** Removed deprecated in Moya 8.0.0: `Moya.Error`, `endpointByAddingParameters(parameters:)`, `endpointByAddingHttpHeaderFields(httpHeaderFields:)`, `endpointByAddingParameterEncoding(newParameterEncoding:)`, `endpointByAdding(parameters:httpHeaderFields:parameterEncoding)`, `StructTarget`, `filterStatusCodes(range:)`, `filterStatusCode(code:)`, `willSendRequest(request:target:)`, `didReceiveResponse(result:target:)`, `ReactiveCocoaMoyaProvider`, `ReactiveSwiftMoyaProvider.request(token:)`. 
 - Added optional callback queue parameter to reactive providers.
+- Updated the `RxSwift` version requirement to `3.3`.
+- Fixed a bug where you would have two response events in `requestWithProgress` method on ReactiveSwift module.
+- Add optional callback queue parameter to reactive providers.
+
+- Bumped minimum version of ReactiveSwift to 2.0.
+- **Breaking Change** Replaced `parameters` & `parameterEncoding` in `TargetType` with extended `Task` cases. Migration guide available in Readme.
+- **Breaking Change** Flattened `UploadType` and `DownloadType` into Task cases. Migration guide available in Readme.
+- Enabled the "Allow app extension API only" flag.
+- Added an optional `requestDataFormatter`in `NetworkLoggerPlugin` to allow the client to interact with the request data before logging it.
 - Added public `URL(target:)` initializator that creates url from `TargetType`.
 - Added an optional `requestDataFormatter`in `NetworkLoggerPlugin` to allow the client to interact with the request data before logging it.
 - Updated minimum version of `RxSwift` to `3.3`.
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,7 +5,7 @@
 
 - **Breaking Change** Added support to get the response (if any) from `MoyaError`.
 - **Breaking Change** Replaced `parameters` & `parameterEncoding` in `TargetType` with extended `Task` cases.
-- **Breaking Change** Flattened `UploadType` and `DownloadType` into Task cases.
+- **Breaking Change** Flattened `UploadType` and `DownloadType` into `Task` cases.
 - **Breaking Change** Updated `RxMoyaProvider.request` to return a [`Single<Request>`](https://github.com/ReactiveX/RxSwift/pull/1123).
 - **Breaking Change** Changed `Moya.Response`'s `response`to use an `HTTPURLResponse` instead of a `URLResponse`.
 - **Breaking Change** Added `headers` to `TargetType`.
@@ -19,9 +19,7 @@
 - Updated the `RxSwift` version requirement to `3.3`.
 - Fixed a bug where you would have two response events in `requestWithProgress` method on ReactiveSwift module.
 - Add optional callback queue parameter to reactive providers.
-
 - Bumped minimum version of ReactiveSwift to 2.0.
-- Add optional callback queue parameter to reactive providers.
 - Enabled the "Allow app extension API only" flag.
 - Added an optional `requestDataFormatter`in `NetworkLoggerPlugin` to allow the client to interact with the request data before logging it.
 - Added public `URL(target:)` initializator that creates url from `TargetType`.
--- a/Changelog.md
+++ b/Changelog.md
@@ -19,6 +19,9 @@
 - Bumped minimum version of ReactiveSwift to 2.0.
 - **Breaking Change** Replaced `parameters` & `parameterEncoding` in `TargetType` with extended `Task` cases. Migration guide available in Readme.
 - **Breaking Change** Flattened `UploadType` and `DownloadType` into Task cases. Migration guide available in Readme.
+- **Breaking Change** Replaced `parameters` & `parameterEncoding` in `TargetType` with extended `Task` cases.
+- **Breaking Change** Flattened `UploadType` and `DownloadType` into Task cases.
+- Add optional callback queue parameter to reactive providers.
 - Enabled the "Allow app extension API only" flag.
 - Added an optional `requestDataFormatter`in `NetworkLoggerPlugin` to allow the client to interact with the request data before logging it.
 - Added public `URL(target:)` initializator that creates url from `TargetType`.
--- a/Readme.md
+++ b/Readme.md
@@ -244,6 +244,22 @@ for filtering out certain status codes. This means that you can place your code 
 handling API errors like 400's in the same places as code for handling invalid
 responses.
 
+## Migration Guides
+
+This project follows [Semantic Versioning](http://semver.org).
+
+Please follow the appropriate guide below when **upgrading to a new major version** of Moya (e.g. 8.0 -> 9.0).
+
+### Upgrade from 8.x to 9.x
+
+- Move the `parameters` and `parameterEncoding` to the `task` computed property by using the case `.requestParameters(parameters:,encoding:)`
+- Replace the task type `.request` with either `.requestPlain` (if you have no parameters) or `.requestParameters(parameters:,encoding:)`
+- There's no `parameters` and `parameterEncoding` on Endpoints any more (e.g. `addingParameters()`), provide them through the `task` on your target instead
+- To send URL encoded parameters AND body parameters, you can now use the task type `.requestCompositeParameters(urlParameters:,bodyParameters:,bodyEncoding:)`
+- Simplify occurences of task type `.download(.request(destination))` to `.downloadDestination(destination)`
+- Simplify occurences of task type `.upload(.file(url))` to `.uploadFile(url)`
+- Simplify occurences of task type `.upload(.multipart(data))` to `.uploadMultipart(data)`
+
 ## Community Projects
 
 [Moya has a great community around it and some people have created some very helpful extensions.](https://github.com/Moya/Moya/blob/master/docs/CommunityProjects.md)
--- a/Readme.md
+++ b/Readme.md
@@ -244,22 +244,6 @@ for filtering out certain status codes. This means that you can place your code 
 handling API errors like 400's in the same places as code for handling invalid
 responses.
 
-## Migration Guides
-
-This project follows [Semantic Versioning](http://semver.org).
-
-Please follow the appropriate guide below when **upgrading to a new major version** of Moya (e.g. 8.0 -> 9.0).
-
-### Upgrade from 8.x to 9.x
-
-- Move the `parameters` and `parameterEncoding` to the `task` computed property by using the case `.requestParameters(parameters:,encoding:)`
-- Replace the task type `.request` with either `.requestPlain` (if you have no parameters) or `.requestParameters(parameters:,encoding:)`
-- There's no `parameters` and `parameterEncoding` on Endpoints any more (e.g. `addingParameters()`), provide them through the `task` on your target instead
-- To send URL encoded parameters AND body parameters, you can now use the task type `.requestCompositeParameters(urlParameters:,bodyParameters:,bodyEncoding:)`
-- Simplify occurences of task type `.download(.request(destination))` to `.downloadDestination(destination)`
-- Simplify occurences of task type `.upload(.file(url))` to `.uploadFile(url)`
-- Simplify occurences of task type `.upload(.multipart(data))` to `.uploadMultipart(data)`
-
 ## Community Projects
 
 [Moya has a great community around it and some people have created some very helpful extensions.](https://github.com/Moya/Moya/blob/master/docs/CommunityProjects.md)
--- a/Sources/Moya/Endpoint.swift
+++ b/Sources/Moya/Endpoint.swift
@@ -72,6 +72,44 @@ extension Endpoint {
         request.httpMethod = method.rawValue
         request.allHTTPHeaderFields = httpHeaderFields
 
+        switch task {
+        case .requestPlain, .uploadFile, .uploadMultipart, .downloadDestination:
+            break
+
+        case .requestData(let data):
+            request.httpBody = data
+
+        case let .requestParameters(parameters: parameters, encoding: parameterEncoding):
+            do {
+                request = try parameterEncoding.encode(request, with: parameters)
+            } catch {
+                return nil
+            }
+
+        case let .requestCompositeData(urlParameters: urlParameters, bodyData: bodyData):
+            do {
+                request = try URLEncoding.default.encode(request, with: urlParameters)
+            } catch {
+                return nil
+            }
+            request.httpBody = bodyData
+
+        case let .requestCompositeParameters(urlParameters: urlParameters, bodyParameters: bodyParameters, bodyEncoding: bodyParameterEncoding):
+            do {
+                request = try URLEncoding.default.encode(request, with: urlParameters)
+                request = try bodyParameterEncoding.encode(request, with: bodyParameters)
+            } catch {
+                return nil
+            }
+
+        case let .downloadParameters(_, parameters: parameters, encoding: parameterEncoding):
+            do {
+                request = try parameterEncoding.encode(request, with: parameters)
+            } catch {
+                return nil
+            }
+        }
+
         return request
     }
 }
--- a/Sources/Moya/Endpoint.swift
+++ b/Sources/Moya/Endpoint.swift
@@ -82,7 +82,7 @@ extension Endpoint {
         case let .requestParameters(parameters: parameters, encoding: parameterEncoding):
             return try? parameterEncoding.encode(request, with: parameters)
 
-        case let .requestCompositeData(urlParameters: urlParameters, bodyData: bodyData):
+        case let .requestCompositeData(bodyData: bodyData, urlParameters: urlParameters):
             do {
                 request = try URLEncoding.default.encode(request, with: urlParameters)
             } catch {
@@ -90,7 +90,7 @@ extension Endpoint {
             }
             request.httpBody = bodyData
 
-        case let .requestCompositeParameters(urlParameters: urlParameters, bodyParameters: bodyParameters, bodyEncoding: bodyParameterEncoding):
+        case let .requestCompositeParameters(bodyParameters: bodyParameters, bodyEncoding: bodyParameterEncoding, urlParameters: urlParameters):
             do {
                 request = try URLEncoding.default.encode(request, with: urlParameters)
                 request = try bodyParameterEncoding.encode(request, with: bodyParameters)
@@ -98,7 +98,7 @@ extension Endpoint {
                 return nil
             }
 
-        case let .downloadParameters(_, parameters: parameters, encoding: parameterEncoding):
+        case let .downloadParameters(parameters: parameters, encoding: parameterEncoding, _):
             return try? parameterEncoding.encode(request, with: parameters)
         }
 
--- a/Sources/Moya/Endpoint.swift
+++ b/Sources/Moya/Endpoint.swift
@@ -75,27 +75,20 @@ extension Endpoint {
         switch task {
         case .requestPlain, .uploadFile, .uploadMultipart, .downloadDestination:
             return request
-
         case .requestData(let data):
             request.httpBody = data
             return request
-
         case let .requestParameters(parameters, parameterEncoding):
             return try? parameterEncoding.encode(request, with: parameters)
-
         case let .uploadCompositeMultipart(_, urlParameters):
             return try? URLEncoding(destination: .queryString).encode(request, with: urlParameters)
-
         case let .downloadParameters(parameters, parameterEncoding, _):
             return try? parameterEncoding.encode(request, with: parameters)
-
         case let .requestCompositeData(bodyData: bodyData, urlParameters: urlParameters):
             request.httpBody = bodyData
             return try? URLEncoding(destination: .queryString).encode(request, with: urlParameters)
-
         case let .requestCompositeParameters(bodyParameters: bodyParameters, bodyEncoding: bodyParameterEncoding, urlParameters: urlParameters):
             if bodyParameterEncoding is URLEncoding { fatalError("URLEncoding is disallowed as bodyEncoding.") }
-
             guard let bodyfulRequest = try? bodyParameterEncoding.encode(request, with: bodyParameters) else { return nil }
             return try? URLEncoding(destination: .queryString).encode(bodyfulRequest, with: urlParameters)
         }
--- a/Sources/Moya/Endpoint.swift
+++ b/Sources/Moya/Endpoint.swift
@@ -84,18 +84,18 @@ extension Endpoint {
             return try? parameterEncoding.encode(request, with: parameters)
 
         case let .uploadCompositeMultipart(_, urlParameters):
-            return try? URLEncoding.default.encode(request, with: urlParameters)
+            return try? URLEncoding(destination: .queryString).encode(request, with: urlParameters)
 
         case let .downloadParameters(parameters, parameterEncoding, _):
             return try? parameterEncoding.encode(request, with: parameters)
 
         case let .requestCompositeData(bodyData: bodyData, urlParameters: urlParameters):
             request.httpBody = bodyData
-            return try? URLEncoding.default.encode(request, with: urlParameters)
+            return try? URLEncoding(destination: .queryString).encode(request, with: urlParameters)
 
         case let .requestCompositeParameters(bodyParameters: bodyParameters, bodyEncoding: bodyParameterEncoding, urlParameters: urlParameters):
             guard let bodyfulRequest = try? bodyParameterEncoding.encode(request, with: bodyParameters) else { return nil }
-            return try? URLEncoding.default.encode(bodyfulRequest, with: urlParameters)
+            return try? URLEncoding(destination: .queryString).encode(bodyfulRequest, with: urlParameters)
         }
     }
 }
--- a/Sources/Moya/Endpoint.swift
+++ b/Sources/Moya/Endpoint.swift
@@ -83,11 +83,8 @@ extension Endpoint {
         case let .requestParameters(parameters, parameterEncoding):
             return try? parameterEncoding.encode(request, with: parameters)
 
-        case let .uploadFileParameters(parameters, parameterEncoding, _):
-            return try? parameterEncoding.encode(request, with: parameters)
-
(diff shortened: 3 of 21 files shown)


Dschee:
I have just integrated my fork into the project where I'm using the `Encodable` type. Here's how my target now looks:

```swift
import Foundation
import Moya

enum MyTarget {
    case createUser(CreateUserRequest)
    case signIn(SignInRequest)
}

extension MyTarget: TargetType {
    var baseURL: URL {
        return URL(string: "https://example.com/api/v1/")!
    }

    var path: String {
        switch self {
        case .createSession:
            return "user"

        case .createSessionProvision:
            return "session"
        }
    }

    var method: Moya.Method {
        return .post
    }

    var sampleData: Data {
        return Data()
    }

    var task: Task {
        switch self {
        case . createUser(let createUserRequest):
            return .request(.data(encode(createUserRequest)))

        case .signIn(let signInRequest):
            return .request(.data(encode(signInRequest)))
        }
    }

    func encode<T: Encodable>(_ encodable: T) -> Data {
        do {
            return try JSONEncoder().encode(encodable)
        } catch {
            let message = "Could not encode '\(type(of: encodable))' using JSONEncoder. Error: \(error)"
            log.error(message)
            fatalError(message)
        }
    }
}
```

It builds fine and I think it should work (haven't tested yet). But this is how the usage might look now. Just to have something to review. I'm not too happy about the many parentheses like in `.request(.data(encode(createUserRequest)))`, but I can live with it until we find a better solution.

sunshinejr:
I will go ahead and say we might not want to add it to 9.0.0 release since we have many breaking changes already (and few big ones). To make transition smoother we may just hold onto this one till after the release. Meanwhile, it would be awesome if it got some attention & feedback from @Moya/contributors.

ashfurrow:
I looked this over and it seems to make sense to me.

I agree with @sunshinejr that it gets a lot more verbose at the call site... definitely not great. I'm wondering if instead of associating the new `RequestDataType` with the request enum case, we can extend the number of enum cases? like `requestWithData`, `requestWithParameters` etc?

Dschee:
@ashfurrow Sure, why not. Think that makes sense. Just rebased onto the current master and added the requested change. The task computed property example from above now looks like this:

```swift
    var task: Task {
        switch self {
        case .createUser(let createUserRequest):
            let requestData = encode(createUserRequest)
            return .requestData(requestData)

        case .signIn(let signInRequest):
            let requestData = encode(signInRequest)
            return .requestData(requestData)
        }
    }
```

Dschee:
Shall I flatten the `UploadType` and `DownloadType` enums into the `Task` enum as well? It would be consistent, for sure. But I'm not sure cause there's no pressing issue with them and it would require more people to change their code when updating to a newer version, so I'd understand if we would keep it as is.

Dschee:
Any preferences on how to deal with the exceptions thrown in places like this:

```swift
do {
    preparedRequest = try parameterEncoding.encode(preparedRequest, with: parameters)
} catch {
    // TODO: Add exception handling here
}
```

@ashfurrow @sunshinejr ?

Sources/Moya/TargetType.swift
```diff
@@ -49,14 +50,26 @@ public enum UploadType {
...
 /// Represents an HTTP task.
 public enum Task {
 
-    /// A basic request task.
-    case request
+    /// A requests body set with data.
+    case requestData(Data)
+
+    /// A requests body set with parameters and encoding.
+    case requestEncoded(parameters: [String: Any], encoding: ParameterEncoding)
```
sunshinejr: maybe `request`/`requestParameters`? or `requestWithParameters`? I feel like `encoded` is not as a user-friendly name as `parameters` 😄 `requestWithParameters` reminds me of Swift 2.0 though 😞

Sources/Moya/TargetType.swift
```diff
@@ -13,11 +13,8 @@ public protocol TargetType {
     /// The HTTP method used in the request.
     var method: Moya.Method { get }
 
-    /// The parameters to be encoded in the request.
-    var parameters: [String: Any]? { get }
-
-    /// The method used for parameter encoding.
-    var parameterEncoding: ParameterEncoding { get }
+    /// The default parameterEncoding for the `.encoded` RequestDataType case.
```
sunshinejr: I feel like we don't actually need to add this to the core. We do not use it anywhere, and this can be easily added to the migration guide as a best practice to use. Or better yet, make an extension to `ParameterEncoding`:
```swift
extension ParameterEncoding. {
    static var default: JSONEncoding {
        return JSONEncoding.default
    }
}
```

This way it would be easily used in task:
```swift
return .request(parameters: [:], encoding: .default)
```

Sources/Moya/MoyaProvider+Internal.swift
```diff
@@ -83,18 +84,57 @@ public extension MoyaProvider {
             switch stubBehavior {
             case .never:
                 switch target.task {
-                case .request:
+                case .requestData(let data):
+                    preparedRequest.httpBody = data
+                    cancellableToken.innerCancellable = self.sendRequest(target, request: preparedRequest, queue: queue, progress: progress, completion: networkCompletion)
+
+                case let .requestEncoded(parameters: parameters, encoding: parameterEncoding):
+                    do {
```
sunshinejr: Usually this is done in an `Endpoint`, take a look [here](https://github.com/Moya/Moya/blob/master/Sources/Moya/MoyaProvider%2BDefaults.swift#L16-L22). If we cannot encode parameters, we throw a `MoyaError.requestMapping(_:)` error to the user.

sunshinejr:
I left you comments around most issues @Dschee. For flattening I agree we would need to flat upload & download types as well. I think we might do it in the same PR since naming convention in this one would be similar. 

Anyhow, I like the direction that this PR has taken. Keep it up :)

Sources/Moya/TargetType.swift
Dschee: Did as suggested. Two variations though. One: I used `ParameterEncoding` as the return type isntead of `JSONEncoding`. Two: I had to backtick `default` as I saw an error (in Xcode 9 beta 3).

Sources/Moya/TargetType.swift
Dschee: I've renamed to `requestParameters`, although I also considered `requestParameterEncoded` (tell me if you like that more, otherwise I'll keep it).

Sources/Moya/MoyaProvider+Internal.swift
Dschee: Sorry, I don't really understand how I could throw a `MoyaError.requestMapping` from within the `performNetworking` closure, also I have no idea how to move the code to the endpoint. How can I switch on the `target.task` from within the Endpoint class? Need some help with this ...

Dschee:
Sorry for letting you wait for two weeks, I've been on vacation this entire time. Now that I'm back I've had a look onto your feedback @sunshinejr and commented on them (two of three I fixed). Also I have flattened the upload and download types into the `Task` enum. Unfortunately I will need more help to handle the exceptions properly (the TODOs).

Other than that, is there anything that speaks against merging this except for the TODOs on the initial Post that are still open? If not, then I would proceed implementing the TODOs so we can merge this soon.

Sources/Moya/TargetType.swift
Dschee: Actually the extension didn't work correctly. I received the following error in Xcode 8.3:
`Static member 'default' cannot be used on protocol metatype 'ParameterEncoding.Protocol'`

I'm removing it now ... I think `JSONEncoding.default` and `URLEncoding.default` should be fine enough for now.

sunshinejr:
Hey @Dschee, just a heads-up. I'll try to get to it ASAP and edit this comment. Thank you for the constant work on this one. 👍

Dschee:
Okay, I have now worked through my TODOs. Except for the exception handling, I think this is now finished. I've removed the `[WIP]` from the title cause I think it's theoretically mergeable as it is, cause I can imagine the work on the exception handling could be decided upon later and the TODO's kept in code until then, as it is right now.

Dschee:
Just rebased onto the current master so things are updated (and tests can pass, hopefully).

Sources/Moya/Endpoint.swift
```diff
@@ -21,59 +21,29 @@ open class Endpoint<Target> {
...
                 sampleResponseClosure: @escaping SampleResponseClosure,
                 method: Moya.Method = Moya.Method.get,
-                parameters: [String: Any]? = nil,
-                parameterEncoding: Moya.ParameterEncoding = URLEncoding.default,
                 httpHeaderFields: [String: String]? = nil) {
 
         self.url = url
         self.sampleResponseClosure = sampleResponseClosure
         self.method = method
-        self.parameters = parameters
```
sunshinejr: I think that we can't just remove option to update `parameters`/`parameterEncoding` in the Endpoint closure. Yes, we don't have it a separate properties, but we can replace it with `task` - it would be even more customizable.

Sources/Moya/Endpoint.swift
```diff
@@ -99,7 +69,7 @@ extension Endpoint {
         request.httpMethod = method.rawValue
         request.allHTTPHeaderFields = httpHeaderFields
 
-        return try? parameterEncoding.encode(request, with: parameters)
+        return request
```
sunshinejr: This is a place where you would do a switch and prepare the request based on a task. We don't really want to do it in the core file since preparing the request based on an `Endpoint` is `Endpoint`'s responsibility. Another benefit is the fact that when we return nil there is an error thrown in basic request mapping and we don't even have to do a request ([source](https://github.com/Moya/Moya/blob/master/Sources/Moya/MoyaProvider%2BDefaults.swift#L16-L22)).

Sources/Moya/TargetType.swift
```diff
@@ -30,37 +24,42 @@ public protocol TargetType {
 }
 
 public extension TargetType {
+    var defaultParameterEncoding: ParameterEncoding {
```
sunshinejr: I think we can remove the `defaultParameterEncoding` from our codebase - I meant that someone can implement it in their own codebase if its convenient for them.

Sources/Moya/TargetType.swift
```diff
@@ -30,37 +24,42 @@ public protocol TargetType {
...
 
-    /// Upload "multipart/form-data"
-    case multipart([MultipartFormData])
-}
+    /// A requests body set with data.
+    case requestData(Data)
 
-/// Represents a type of download task.
-public enum DownloadType {
+    /// A requests body set with parameters and encoding.
```
sunshinejr: "A requests body set with encoded parameters"

Readme.md
```diff
@@ -242,6 +242,22 @@ for filtering out certain status codes. This means that you can place your code
 handling API errors like 400's in the same places as code for handling invalid
 responses.
 
+## Migration Guides
+
+This project follows [Semantic Versioning](http://semver.org).
+
+Please follow the appropriate guide below when **upgrading to a new major version** of Moya (e.g. 8.0 -> 9.0).
+
+### Upgrade from 8.x to 9.x
```
sunshinejr: I feel like this have its own file in the docs.

Sources/Moya/TargetType.swift
```diff
@@ -30,37 +24,42 @@ public protocol TargetType {
...
+
     var validate: Bool {
         return false
     }
 }
 
-/// Represents a type of upload task.
-public enum UploadType {
+/// Represents an HTTP task.
+public enum Task {
```
sunshinejr: we also need to properly test encoding the request from an endpoint

Sources/Moya/TargetType.swift
Dschee: Uh, I forgot to remove it (thought I actually did that). No problem with that.

Dschee: Fixed this.

Sources/Moya/TargetType.swift
Dschee: Fixed.

Readme.md
Dschee: Sure, I've moved it to its own `MigrationGuides.md` file within the `docs` folder. I think we should consider linking to the file when doing major releases within the changelog & release notes.

Sources/Moya/Endpoint.swift
Dschee: Ok, let me try ...

Dschee: I think I got this now. Note that there's no convenience `.addingParameters` or something. Don't know if we can provide a replacement with the task enum. We can't append all enum cases with each other (e.g. Data with Data), can we?

Sources/Moya/Endpoint.swift
Dschee: Ok, I see. Let me try ...

Dschee: I think I got this, too. Trying to clean up the internal MoyaProvider now ...

Dschee: Ok, that was easy. As you suggested, I'm now returning nil when an exception is thrown. This way I can even mark my last TODO as completed. Thanks for the feedback! 👍

Sources/Moya/TargetType.swift
Dschee: Can we do that in a subsequent PR so this one can be in the 9.0 release? Cause I won't have much time to work on this in the near future (I'm concentrating my free time onto [this project](https://github.com/Flinesoft/Raid-Meetup-for-Pokemon-GO) right now) and I feel like the logic of adding the task cases data to the request isn't overly complicated, so it should all work fine. Maybe except for some edge cases, but those would probably be related to the new cases (like `.composite...`) anyways ...

sunshinejr: Making this one into the 9.0.0 release would be cool, but we can't do that without testing it (this is one of the major things we changed in the whole core). We can do that in separate PR, but still I think merging this should happen with tests in it. 

I understand that you may be busy now, though. If that's the case, let me know and maybe someone can help with the tests and additional feedback on the code review (and rebase in the near future). There is still work to do, unfortunately, but we are close.

Sources/Moya/Endpoint.swift
```diff
@@ -99,7 +72,45 @@ extension Endpoint {
...
-        return try? parameterEncoding.encode(request, with: parameters)
+        switch task {
+        case .requestPlain, .uploadFile, .uploadMultipart, .downloadDestination:
+            break
+
+        case .requestData(let data):
+            request.httpBody = data
+
+        case let .requestParameters(parameters: parameters, encoding: parameterEncoding):
+            do {
```
sunshinejr: we can just do returns in cases I think:
```swift
return try? parameterEncoding.encode(request, with: parameters)
```

Sources/Moya/Endpoint.swift
Dschee: Sounds good. 👍

Sources/Moya/Endpoint.swift
Dschee: Good catch, updated the cases `requestParameters` and `downloadParameters` where this was possible. This didn't work in the composite cases though as they need to to multiple things with the request.

Dschee:
Worked through your comments again and rebased. Are you planning to add this to the current release after the alpha is out? Would be awesome to see this in the next version. :)

sunshinejr:
Thanks, @Dschee! I'd love to have it, but we are still without tests and I'm not the only one to decide such things. Would love to get some feedback on it from @Moya/contributors, since it is almost finished now.

Also, sorry that I didn't mentioned it earlier, but would you rebase it to `9.0.0-alpha.1` branch, please?

Sources/Moya/TargetType.swift
```diff
@@ -35,32 +29,33 @@ public extension TargetType {
...
-/// Represents a type of download task.
-public enum DownloadType {
+    /// A requests body set with encoded parameters.
+    case requestParameters(parameters: [String: Any], encoding: ParameterEncoding)
 
-    /// Download a file to a destination.
-    case request(DownloadDestination)
-}
+    /// A requests body set with data, combined with url parameters.
+    case requestCompositeData(urlParameters: [String: Any], bodyData: Data)
```
SD10: I think it would be better to swap the order of the `urlParameters` and `bodyData` associated values. The first associated value should match the last word of the enum case.

SD10:
I just barely got to skim this over but will take an in depth look this weekend. Some really cool work going on here! 😃

Sources/Moya/MoyaProvider+Internal.swift
```diff
@@ -83,18 +84,25 @@ public extension MoyaProvider {
             switch stubBehavior {
             case .never:
                 switch target.task {
```
SD10: A little trivial and unrelated to the PR - but can we pull this switch statement out into a `sendRequest` function? This way we don't have nested switch statements and can start cleaning up this behemoth of a method 😃

Sources/Moya/TargetType.swift
Dschee: Does 4a2bc0e57c745c97a817ea71e23ce50e09d0bf1a fix it correctly?

Sources/Moya/MoyaProvider+Internal.swift
Dschee: Not sure what you mean. There is already a `sendRequest` function, so I'm a little confused. Also I'm not sure why the we should pull only the `.never` stub behavior switch statement out into a function named `sendRequest`, I don't really understand how that's the same thing. Did you mean the entire `performNetworking` closure? I could extract that into a method named `performNetworking` to make the function shorter if that's what you mean.

Dschee: Okay, I think I understand your issue now. I've just extracted the inner switch statement to a method named `unstubbedRequest` in ab18763cfd7f70a3057b6c9af4c21871c380b21a. I hope this is what you were expecting.

Dschee:
@SD10 Just worked through your comments, looking forward for your feedback this weekend!

@sunshinejr No problem, just rebased onto `9.0.0-alpha.1`. 🎉

Sources/Moya/TargetType.swift
Dschee: I have rebased and worked through some more feedback from @SD10. I'd be happy to work through any other feedback this weekend, but regarding tests it would be great to have help there. I don't think I'll have the time to set them up in the coming weeks ...

Sources/Moya/Endpoint.swift
```diff
@@ -99,7 +72,37 @@ extension Endpoint {
...
+            break
+
+        case .requestData(let data):
+            request.httpBody = data
+
+        case let .requestParameters(parameters: parameters, encoding: parameterEncoding):
+            return try? parameterEncoding.encode(request, with: parameters)
+
+        case let .requestCompositeData(bodyData: bodyData, urlParameters: urlParameters):
+            do {
```
BasThomas: Nit, but why the `do-catch` instead of `try?` like in some other `case`s?

Dschee: Well, I can only use `try?` if the `request` is not needed after that line like in those other cases. But here for example the `request` type is needed to set the http body afterwards. Of course I could change the order here and set the http body first ... but on the composite parameters case I need to do two `try`s anyway so I simply kept the `do-catch`.

But since you seem not to like the `do-catch`s I've changed them up to this:

```swift
case let .requestCompositeData(bodyData: bodyData, urlParameters: urlParameters):
            request.httpBody = bodyData
            return try? URLEncoding.default.encode(request, with: urlParameters)

        case let .requestCompositeParameters(bodyParameters: bodyParameters, bodyEncoding: bodyParameterEncoding, urlParameters: urlParameters):
            guard let bodyfulRequest = try? bodyParameterEncoding.encode(request, with: bodyParameters) else { return nil }
            return try? URLEncoding.default.encode(bodyfulRequest, with: urlParameters)
```

I hope this is better.

BasThomas: I have nothing against `do-catch`, just noticed the inconsistency. :)

Dschee: Thanks for reviewing & approving anyways! 👍

Changelog.md
```diff
@@ -2,6 +2,8 @@
 
 - **Breaking Change** Renamed all occurrences of `queue` to `callbackQueue`.
 - **Breaking Change** Added support to get the response (if any) from `MoyaError`.
+- **Breaking Change** Replaced `parameters` & `parameterEncoding` in `TargetType` with extended `Task` cases.
+- **Breaking Change** Flattened `UploadType` and `DownloadType` into Task cases.
```
sunshinejr: this is minor, but you forgot `` around `Task`

Changelog.md
```diff
@@ -13,6 +15,7 @@
 - Add optional callback queue parameter to reactive providers.
 
 - Bumped minimum version of ReactiveSwift to 2.0.
+- Add optional callback queue parameter to reactive providers.
```
sunshinejr: this is probably leftover from rebasing, but we have this entry already (line 15)

Sources/Moya/MoyaProvider+Internal.swift
```diff
@@ -107,6 +97,28 @@ public extension MoyaProvider {
     // swiftlint:enable cyclomatic_complexity
     // swiftlint:enable function_body_length
 
+    private func unstubbedRequest(_ target: Target, request: URLRequest, callbackQueue: DispatchQueue?, progress: Moya.ProgressBlock?, completion: @escaping Moya.Completion) -> Cancellable {
```
sunshinejr: I wouldn't know what this function does without checking the source code, unfortunately. I'm thinking if we can create a function that would return cancellable, but also it would check the `stubBehavior` and base on that one. Like:
```swift
func abc(...) -> Cancellable {
    switch stubBehavior {
        ...
    }
}
```
so we could remove this logic in `performNetworking` and just assign the `Cancellable`:
```swift
cancellableToken.innerCancellable = abc(...)
```

it's responsibility would be to send a request based on types & return `Cancellable`. What do you think?

Sources/Moya/TargetType.swift
sunshinejr: I can help with tests, no problem with that 👍 I'll try to do it ASAP.

sunshinejr:
Also, could you check CI, @Dschee, please? Seems like it is failng.

Changelog.md
Dschee: Fixed.

Changelog.md
Dschee: Removed.

Sources/Moya/MoyaProvider+Internal.swift
Dschee: Sure, sounds good, better than what it is now. :) I've named the method `performRequest`. Hope it's more readable now.

Dschee:
@sunshinejr I worked through your comments, should be all fixed now.

Regarding the CI: I'm running the tests locally but seems like I did not see the demo apps. :)
Those were not building and I actually saw that in one of them there was a multipart upload that also required parameters to be sent along. Since this could also happen on uploading a file, I've added two more cases: `uploadFileParameters` and `uploadMultipartParameters`.

Sources/Moya/TargetType.swift
```diff
@@ -38,32 +32,39 @@ public extension TargetType {
...
+    /// A file upload task with extra parameters using the given encoding.
+    case uploadFileParameters(parameters: [String: Any], encoding: ParameterEncoding, URL)
+
+    /// A "multipart/form-data" upload task.
+    case uploadMultipart([MultipartFormData])
 
-    /// A basic request task.
-    case request
+    /// A "multipart/form-data" upload task with extra parameters using the given encoding.
+    case uploadMultipartParameters(parameters: [String: Any], encoding: ParameterEncoding, [MultipartFormData])
```
sunshinejr: Correct me if I'm wrong, but if you want to add parameters to the multipart request, you want to add them using `MultipartFormData`, not `parameters` property (at least in Alamofire). You may want to add parameters to the URL, though, so that would be an exception. But I'm not sure if we should allow more encodings, like JSON, too.

Sources/Moya/TargetType.swift
```diff
@@ -38,32 +32,39 @@ public extension TargetType {
...
-/// Represents an HTTP task.
-public enum Task {
+    /// A requests body set with encoded parameters combined with url parameters.
+    case requestCompositeParameters(bodyParameters: [String: Any], bodyEncoding: ParameterEncoding, urlParameters: [String: Any])
+
+    /// A file upload task.
+    case uploadFile(URL)
+
+    /// A file upload task with extra parameters using the given encoding.
+    case uploadFileParameters(parameters: [String: Any], encoding: ParameterEncoding, URL)
```
sunshinejr: Also, not sure about that one. If you want to upload a file with parameters, you would just use `uploadMultipart`.

Sources/Moya/TargetType.swift
Dschee: I have no idea to be honest, never uploaded files and added parameters. But probably you're right, so I've changed this to be `uploadCompositeMultipart` and only have `urlParameters` as parameter.

Sources/Moya/TargetType.swift
Dschee: Sounds good to me, reverting this.

Sources/Moya/Endpoint.swift
```diff
@@ -99,7 +72,31 @@ extension Endpoint {
...
-        return try? parameterEncoding.encode(request, with: parameters)
+        switch task {
+        case .requestPlain, .uploadFile, .uploadMultipart, .downloadDestination:
+            return request
+
+        case .requestData(let data):
+            request.httpBody = data
+            return request
+
+        case let .requestParameters(parameters, parameterEncoding):
```
SD10: Awesome. This looks great with the `do/catch removed`.

SD10:
We still need to update `docs/Examples/Basic.md` and possibly more in the examples section. While the examples are not pure documentation, `Basic.md` is a very import reference file.

docs/Examples/Basic.md
```diff
@@ -140,20 +144,15 @@ public func url(route: TargetType) -> String {
...
 
 let endpointClosure = { (target: MyService) -> Endpoint<MyService> in
-    return Endpoint<MyService>(url: url(target), sampleResponseClosure: {.networkResponse(200, target.sampleData)}, method: target.method, parameters: target.parameters)
+    return Endpoint<MyService>(url: url(target), sampleResponseClosure: {.networkResponse(200, target.sampleData)}, method: target.method, task: target.task)
 }
 ```
 
 The block you provide will be invoked every time an API call is to be made. Its
 responsibility is to return an `Endpoint` instance configured for use by Moya.
-`parameters` is passed into this block to allow you to configure the `Endpoint`
```
Dschee: I did not really understand this sentence and I'm also not sure if it's needed with the new `task` type. So I removed it. I would need an explanation if this should be kept and updated.

Dschee:
@SD10 I've just rebased to master after #1181 was merged and updated the Basic example file. Note that there's one point I'm not sure about ([see here](https://github.com/Moya/Moya/pull/1147/files#r130229006)) – other than that I think the file should be ready.

Dschee:
@SD10 I've just checked all example files and found only two which need to be updated: `ArrayAsRootContainer.md` and `OptionalParameters.md`. I've updated the latter but I can't update the first as already explained in my previous comment https://github.com/Moya/Moya/pull/1147#issuecomment-316640623. I hope this is what you were expecting!

SD10:
I'm actually not sure about `ArrayAsRootContainer.md` but it looks like more of a specialty example, so if it wasn't updated before a release I wouldn't mind too much. My biggest concern was `Basic.md` considering it is our `#.1` most viewed documentation file. Thanks for taking care of this!

docs/Examples/Basic.md
SD10: This is confusing to me as well 🤔 I have no suggestions for a replacement.

Dschee:
FYI: I have just tried adding fix-its via `@available` lines in a reintroduced `MoyaAvailability.swift` file but I couldn't cover any of the cases this PR changes. Specifically my code looked as follows with the reason why it doesn't work as comments:

```swift
import Foundation

extension TargetType {
    @available(*, unavailable, message: "parameters has been removed in version 9.0.0, use new task case .requestParameters instead")
    public var parameters: [String: Any]? { fatalError() } // does not show up in a type conforming to TargetType

    @available(*, unavailable, message: "parameterEncoding has been removed in version 9.0.0, use new task case .requestParameters instead")
    public var parameterEncoding: ParameterEncoding { fatalError() } // does not show up in a type conforming to TargetType
}

extension Task {
    @available(*, unavailable, message: "upload has been removed in version 9.0.0, use new task cases .uploadFile, .uploadMultipart or .uploadCompositeMultipart instead")
    case upload // leads to an error: "Enum 'case' is not allowed outside of an enum"
}
```

SD10:
@Dschee I think what you've listed is fine. We don't need to add `unavailable` attributes to specific cases in the enum. The migration guide will cover this.

Changelog.md
```diff
@@ -3,6 +3,10 @@
 # 9.0.0-alpha.1
 
 - **Breaking Change** Added support to get the response (if any) from `MoyaError`.
+- **Breaking Change** Replaced `parameters` & `parameterEncoding` in `TargetType` with extended `Task` cases.
```
sunshinejr: all changes should go to the `Next` section

Dschee: Done.

Tests/MultiTargetSpec.swift
```diff
@@ -32,20 +30,28 @@ class MultiTargetSpec: QuickSpec {
             }
 
             it("uses correct parameters") {
-                expect(target.parameters?["key"] as? String) == "value"
-                expect(target.parameters?.count) == 1
+                if case let .requestParameters(parameters: parameters, encoding: _) = target.task {
+                    expect(parameters["key"] as? String) == "value"
+                    expect(parameters.count) == 1
+                } else {
+                    fail()
```
sunshinejr: We should add a description _why_ this test has failed.

Tests/MultiTargetSpec.swift
```diff
@@ -32,20 +30,28 @@ class MultiTargetSpec: QuickSpec {
...
+                }
             }
 
             it("uses correct parameter encoding.") {
-                expect(target.parameterEncoding is JSONEncoding) == true
+                if case let .requestParameters(parameters: _, encoding: parameterEncoding) = target.task {
+                    expect(parameterEncoding is JSONEncoding) == true
+                } else {
+                    fail()
```
sunshinejr: Same here.

docs/Examples/Basic.md
```diff
@@ -59,6 +59,16 @@ extension MyService: TargetType {
             return JSONEncoding.default // Send parameters as JSON in request body
         }
     }
+    var task: Task {
+        switch self {
+        case .zen, .showUser, .showAccounts: // Send no parameters
+            return .requestPlain
+        case .createUser(let firstName, let lastName): // Send parameters in URL
```
sunshinejr: In fact `URLEncoding.default` sends parameters in URL when possible, not always (see [#ParameterEncoding](https://github.com/Alamofire/Alamofire#parameter-encoding)).

Sources/Moya/TargetType.swift
```diff
@@ -38,32 +32,36 @@ public extension TargetType {
...
-    /// Download a file to a destination.
-    case request(DownloadDestination)
-}
+    /// A requests body set with data, combined with url parameters.
+    case requestCompositeData(bodyData: Data, urlParameters: [String: Any])
 
-/// Represents an HTTP task.
-public enum Task {
+    /// A requests body set with encoded parameters combined with url parameters.
+    case requestCompositeParameters(bodyParameters: [String: Any], bodyEncoding: ParameterEncoding, urlParameters: [String: Any])
```
Dschee: Although that is true, I don't think we can handle this properly. Best we could do would be to ensure that both parameters get merged in that case. But I think that if this is really used that way then it is a user side error cause it doesn't make sense to send url parameters twice.

Sources/Moya/TargetType.swift
```diff
@@ -38,32 +32,36 @@ public extension TargetType {
...
 
-    /// An upload task.
-    case upload(UploadType)
+    /// A file download task to a destination.
+    case downloadDestination(DownloadDestination)
 
-    /// A download task.
-    case download(DownloadType)
+    /// A file download task to a destination with extra parameters using the given encoding.
+    case downloadParameters(parameters: [String: Any], encoding: ParameterEncoding, DownloadDestination)
```
Dschee: Done. Called it `destination`.

Tests/MultiTargetSpec.swift
Dschee: Updated.

Tests/MultiTargetSpec.swift
Dschee: Updated.

docs/Examples/Basic.md
Dschee: Updated the Endpoint `urlRequest` cases to use `URLEncoding(destination: .queryString)` instead of `URLEncoding.default`.

Sources/Moya/TargetType.swift
sunshinejr: I'm more and more keen to remove this one case. It makes it so we still have undefined behavior. And undefined behavior is kinda what we wanted to avoid in the first place. Oh, and we can achieve this one using `requestCompositeData` quite easy.

sunshinejr: I'm more and more keen to remove this one case. It makes it so we still have undefined behavior. And undefined behavior is kinda what we wanted to avoid in the first place. Oh, and we can achieve this one using `requestCompositeData` quite easy. Any other @Moya/contributors have an opinion on that one? 🤔

SD10: Yeah, we can achieve this with `requestCompositeData`. It may not be as clean, but I think it's better than undefined behavior. In my experience, if it's possible for someone to make a mistake -- someone will. It could be as simple as using the wrong encoding w/ auto-complete.

I'm fine with removing it for now. We can start simple and then expand the API from there based on user feedback. Another suggestion would be to throw an error in this situation. The previous API returned nil and ignored encoding errors, but now we have a lot going on here.

Dschee: I'm actually against removing the case only because of one possible way of misusing it. I agree though that we should deal with that situation somehow to minimize any unexpected behavior: For that I think throwing a fatalError would be possible. What do you think @sunshinejr? @SD10 seemed to be okay with that.

Dschee: I've just updated the `requestCompositeParameters` case to include this check in f24ee454f9429035aed611ce2c1f19620b42f5e0:
```
if bodyParameterEncoding is URLEncoding { fatalError("URLEncoding is disallowed as bodyEncoding.") }
```

Changelog.md
```diff
@@ -1,8 +1,12 @@
 # Next
+- **Breaking Change** Replaced `parameters` & `parameterEncoding` in `TargetType` with extended `Task` cases.
+- **Breaking Change** Flattened `UploadType` and `DownloadType` into `Task` cases.
 
 # 9.0.0-alpha.1
 
 - **Breaking Change** Added support to get the response (if any) from `MoyaError`.
+- **Breaking Change** Updated `RxMoyaProvider.request` to return a [`Single<Request>`](https://github.com/ReactiveX/RxSwift/pull/1123).
```
Dschee: Fixed.

Sources/Moya/TargetType.swift
SD10: I was thinking of something more along the lines of making `urlRequest` a throwing method:
```swift 
func urlRequest() throws -> URLRequest
```
This way we can make use of the other encoding errors + add our own `MoyaError` cases.
I don't know how difficult this would be to implement without looking at the callers.

I'm not 100% against using a `fatalError` but I'll acknowledge I am a bit biased towards its use. It's certainly more explicit but it also feels like we're avoiding the problem here 😞

sunshinejr: Okay, we can start with failing on wrong encoding (although checking if this is a `URLEncoding` type might be not so efficient, you can still create your own encoding that does the same encoding as `URLEncoding`, just with a different type name 😄). We are using `fatalError` when checking if everything is correct with multipart [here](https://github.com/Moya/Moya/blob/master/Sources/Moya/MoyaProvider%2BInternal.swift#L91-L93) as well, but we might want to update it in the future to some more user-friendly error throwing. I think this is a decent start.

docs/Examples/Basic.md
sunshinejr: Great 👍

Sources/Moya/Endpoint.swift
```diff
@@ -99,7 +72,33 @@ extension Endpoint {
         request.httpMethod = method.rawValue
         request.allHTTPHeaderFields = httpHeaderFields
 
-        return try? parameterEncoding.encode(request, with: parameters)
+        switch task {
+        case .requestPlain, .uploadFile, .uploadMultipart, .downloadDestination:
+            return request
+
```
sunshinejr: This is a stylistic one: can we remove the empty lines in the code (separating cases, first line empty in a method body)? This is probably a lot nicer for many people (including you), but I don't think we want to change the code style right now, especially with so much code changes.

Tests/TestHelpers.swift
```diff
@@ -63,7 +55,7 @@ func url(_ route: TargetType) -> String {
 
 let failureEndpointClosure = { (target: GitHub) -> Endpoint<GitHub> in
     let error = NSError(domain: "com.moya.moyaerror", code: 0, userInfo: [NSLocalizedDescriptionKey: "Houston, we have a problem"])
-    return Endpoint<GitHub>(url: url(target), sampleResponseClosure: {.networkError(error)}, method: target.method, parameters: target.parameters)
+    return Endpoint<GitHub>(url: url(target), sampleResponseClosure: {.networkError(error)}, method: target.method)
```
sunshinejr: We should pass task in the `.init` as well here, because otherwise we will get a `.requestPlain` always.

docs/MigrationGuides.md
```diff
@@ -0,0 +1,15 @@
+# Migration Guides
+
+This project follows [Semantic Versioning](http://semver.org).
+
+Please follow the appropriate guide below when **upgrading to a new major version** of Moya (e.g. 8.0 -> 9.0).
+
+## Upgrade from 8.x to 9.x
+
+- Move the `parameters` and `parameterEncoding` to the `task` computed property by using the case `.requestParameters(parameters:,encoding:)`
```
sunshinejr: Another stylistic one, signatures are rather references as:
```swift
.requestParameters(parameters:encoding:)
```

instead of:
```swift
.requestParameters(parameters:,encoding:)
```

Sources/Moya/Endpoint.swift
Dschee: Sure, no problem. We can talk about code style somewhere else. I've removed the empty lines.

Tests/TestHelpers.swift
Dschee: Good catch, overlooked this. Added the task now.

docs/MigrationGuides.md
Dschee: Don't know what I was thinking back then, but I usually use it like that, too. So sorry for that, fixed it via 516764c503bc60b63cf59da49f3798f02ba365de. :)

Dschee:
Thanks @sunshinejr, I've worked through your comments and rebased onto the `9.0.0-dev` branch!

BasThomas:
Amazing work everyone, @Dschee especially! 💪

Dschee:
Thank you everyone for reviewing and going all the way with me. You're the best! 💪 👏 👍

T1ASH:
Thank you for your work on this, however I'm having issues getting it to work. I looked through this thread hoping to figure it out, but unfortunately I've come up empty. Is there an example or documentation on how to get .requestData to work in my Task?

sunshinejr:
Hey @T1ASH. If I understand your question correctly, you can do it as follows:

```swift
enum GitHub {
    case zen
}

extension GitHub: TargetType {
    ...
    var task: Task {
        let data = "test data".data(using: .utf8)
        return .requestData()
    }
    ...
}
```

Where, of course, your `data` is whatever you want to send in this endpoint. This will send the `data` request (meaning it will replace `httpBody` in the `URLRequest`) for `GitHub.zen` endpoint.

Let me know if this is what you wanted or I've completely missed 😄

pedrovereza:
@sunshinejr I Edited your reply to be `return .requestData(data)` instead of `return .requestData()` :wink:

sunshinejr:
Damn it! Thanks for the great catch, though, @pedrovereza!

felixsolorzano:
> I have just integrated my fork into the project where I'm using the `Encodable` type.
> 
> Here's how my target now looks:
> 
> ```swift
> import Foundation
> import Moya
> 
> enum MyTarget {
>     case createUser(CreateUserRequest)
>     case signIn(SignInRequest)
> }
> 
> // MARK: - TargetType Protocol Implementation
> extension MyTarget: TargetType {
>     var baseURL: URL {
>         return URL(string: "https://example.com/api/v1/")!
>     }
> 
>     var path: String {
>         switch self {
>         case .createUser:
>             return "user"
> 
>         case .signIn:
>             return "session"
>         }
>     }
> 
>     var method: Moya.Method {
>         return .post
>     }
> 
>     var sampleData: Data {
>         return Data()
>     }
> 
>     var task: Task {
>         switch self {
>         case .createUser(let createUserRequest):
>             let requestData = encode(createUserRequest)
>             return .request(.data(requestData))
> 
>         case .signIn(let signInRequest):
>             let requestData = encode(signInRequest)
>             return .request(.data(requestData))
>         }
>     }
> }
> 
> // MARK: - Helper Methods
> extension Tm5Target {
>     private func encode<T: Encodable>(_ encodable: T) -> Data {
>         do {
>             return try JSONEncoder().encode(encodable)
>         } catch {
>             fatalError("Could not encode '\(type(of: encodable))' using JSONEncoder. Error: \(error)")
>         }
>     }
> }
> ```
> 
> It builds fine and I think it should work (haven't tested yet). But this is how the usage might look now. Just to have something to **review**.

How can I use the encode extension? Was it added to the library or should I simple add it myself?

PR #1148, opened 2017-06-28
[WIP] Support Encodable types natively

Dschee:
This is a follow-up on #1136 and is based on the current state of #1147. We should not merge this until #1147 was merged and Swift 4 nears completion.

What I have done here is to implement the intention behind #1136 with the new approach we have taken in #1147. Although the code is like I would expect it to work, the Swift compiler (currently) seems not to be able to work with this code. I'm still getting the same error as in #1136, for example:

```swift
case let .request(.jsonEncodable(encodable, encoder)):
    preparedRequest.httpBody = encoder.encode(encodable) // ERROR: Cannot invoke 'encode' with an argument list of type '(Encodable)'
    cancellableToken.innerCancellable = self.sendRequest(target, request: preparedRequest, queue: queue, progress: progress, completion: networkCompletion)
```

If you know how we can work around this issue, then please let me know. I couldn't find a good alternative solution or quick fix for this.

Here's what is done in this PR:

- [x] Add `jsonEncodable` and `propertyListEncodable` cases to `RequestDataType`
- [x] Add `compositeJsonEncodable` and `compositePropertyListEncodable`, too
- [x] Integrate the new cases into the request creation logic

The following are TODOs:

- [ ] Fix the above mentioned compiler error (find alternative)
- [ ] Update Changelog
- [ ] Update Tests
- [ ] Update Documentation

### DIFF
--- a/Sources/Moya/Endpoint.swift
+++ b/Sources/Moya/Endpoint.swift
@@ -21,29 +21,18 @@ open class Endpoint<Target> {
     open let url: String
     open let method: Moya.Method
     open let sampleResponseClosure: SampleResponseClosure
-    open let parameters: [String: Any]?
-    open let parameterEncoding: Moya.ParameterEncoding
     open let httpHeaderFields: [String: String]?
 
     /// Main initializer for `Endpoint`.
     public init(url: String,
                 sampleResponseClosure: @escaping SampleResponseClosure,
                 method: Moya.Method = Moya.Method.get,
-                parameters: [String: Any]? = nil,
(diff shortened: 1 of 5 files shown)


sunshinejr:
Please look at my [comment here](https://github.com/Moya/Moya/pull/1147#issuecomment-312091964). Because of that and because we may have too many PRs opened at the same time and we want to make it easier for our contributors to find the ones that matter the most right, I'm gonna close it for now. We will continue the discussion later on.

PR #1151, opened 2017-06-29
Remove string percent encoding in NetworkLoggerPlugin

MrAsterisco:
At the moment, the NetworkLoggerPlugin, which is very useful to debug automatically encoded parameters, outputs the `httpBody` as a not-human-readable string (or, in other words, it outputs the string already encoded to be sent via HTTP).

I have modified the NetworkLoggerPlugin plugin to output a human-readable string to simplify debugging.

_Disclaimer: this is my first pull request to such a big project. If this is not the correct way to pull changes to your project, please just tell me and I'll take the necessary steps to follow your rules. I apologize in advance for any error that I might have made._

### DIFF
--- a/docs/Examples/Basic.md
+++ b/docs/Examples/Basic.md
@@ -70,8 +70,8 @@ extension MyService: TargetType {
             return "{\"id\": \(id), \"first_name\": \"\(firstName)\", \"last_name\": \"\(lastName)\"}".utf8Encoded
         case .showAccounts:
             // Provided you have a file named accounts.json in your bundle.
-            guard let path = Bundle.main.path(forResource: "accounts", ofType: "json"),
-                let data = Data(base64Encoded: path) else {
+            guard let url = Bundle.main.url(forResource: "accounts", withExtension: "json"),
+                let data = try? Data(contentsOf: url) else {
                     return Data()
             }
             return data


sunshinejr:
Hey @MrAsterisco, thank you so much for this PR. 🎉  This is a really nice idea, however, from the user perspective, I often need this plugin to tell me exactly what is the body of a request I send to the API. So for me it would be not-so-obvious that Moya is doing some kind of decoding under the hood. 

Maybe we can figure out a way that user can opt-in (like parameter?) for that feature. Do you have any idea how we can achieve that? 🤔

MrAsterisco:
Ok, @sunshinejr 

I've made some changes to allow an arbitrary and optional function to be passed to the constructor that can perform any kind of operation on the `httpBody`.

If nothing is passed, the behaviour is to fallback to your implementation.

The optional function takes the original `Data` object to let the client handle different types of encoding or anything that's not a string.

Sources/Moya/Plugins/NetworkLoggerPlugin.swift
```diff
@@ -78,8 +80,9 @@ private extension NetworkLoggerPlugin {
             output += [format(loggerId, date: date, identifier: "HTTP Request Method", message: httpMethod)]
         }
 
-        if let body = request?.httpBody, let stringOutput = String(data: body, encoding: .utf8), isVerbose {
-            output += [format(loggerId, date: date, identifier: "Request Body", message: stringOutput)]
+        if let body = request?.httpBody, isVerbose {
```
sunshinejr: This line below is kinda unused if we do have our own formatter. Also in case of lack of message (`nil`), we still print request body (which we didn't before). How about this little change to `if let`:
```swift
if let body = request?.httpBody, let stringOutput = requestDataFormatter?(body) ?? String(data: body, encoding: .utf8), isVerbose {...}
```
Let me know what do you think.

MrAsterisco: Yes, you're right. I think your fix looks good.

sunshinejr:
Looks good, @MrAsterisco! I've fixed some indentation in your PR, so now we need only 2 more things:
- [ ] Add `Changelog.md` entry
- [ ] Add a test case.

About `Changelog.md` entry, just need to add a new item in the `Next` column with the change you implemented. Example [here](https://github.com/Moya/Moya/pull/1020/files#diff-9b8fab691c00b9e5380b19ce882f3271) (it is not a breaking change, so without that **breaking change** label).

About test case, you can take a look how it is done for [formatting response data here](https://github.com/Moya/Moya/blob/master/Tests/NetworkLoggerPluginSpec.swift#L76-L85) and make something similar.

Let me know if it looks good for you. Also, do not hesitate to ask for help If you have any troubles with the above. We will try to assist with whatever you want.

MrAsterisco:
Ok, done!

sunshinejr:
Seems good! Indent is quite off though, could you take care of that? :)

MrAsterisco:
I'm sorry: what's your indentation style? This is what I see on Xcode:

screen shot 2017-06-30 at 18 11 07

BasThomas:
AFAIK it's 4 spaces. :)

PR #1140, opened 2017-06-25
Create Reactive providers via `rx` or `reactive`

BasThomas:
Supersedes #963. (cc @kemchenj)

Also fixes #1144.

PR #1153, opened 2017-06-29
[Feature] Migrate from multiple providers to Reactive implementations (`rx` & `reactive`).

sunshinejr:
Related #963. This is a reimplementation of #1140. 

To-do:
- [x] Update documentation.
- [ ] Tests.
- [x] Moya.xcodeproj fix.

To avoid code duplication I've created internal requests that are used by both reactive providers and new `rx.` & `reactive.` API. Although I'm not sure how we should go about testing this one. We still have tests for reactive providers that cover most of it I believe.

Please let me know what do you guys think!


BasThomas:
We might want to deprecate `RxMoyaProvider` and `ReactiveSwiftMoyaProvider` instead of removing them completely?

docs/Examples/ReactiveSwift.md
```diff
@@ -0,0 +1,23 @@
+# ReactiveSwift examples
+
+ReactiveSwift extension to Moya is added via `reactive` property of `MoyaProvider. As in normal setup,
```
sunshinejr: lack of ` after MoyaProvider breaks the whole paragraph :D

docs/Examples/RxSwift.md
```diff
@@ -1,16 +1,16 @@
 # RxSwift examples
 
-A `RxMoyaProvider` can be created much like a
-[`MoyaProvider`](../Providers.md) and can be used as follows:
+RxSwift extension to Moya is added via `rx` property of `MoyaProvider. As in normal setup,
```
sunshinejr: also lack of ` after MoyaProvider

ashfurrow:
Interesting idea – I like it a lot. My only hesitation is that that some users (including myself) subclass the existing reactive subclasses (subclassing `RxMoyaProvider`) to introduce custom behaviour within the `request` methods. This implementation wouldn't allow for the same mechanism to be used, so we'd need to find some alternative. 

My gut says composition would be a good idea (so, instead of subclassing a provider, we would recommended containing it within a sort of provider-provider). If that's acceptable, we would just want to make sure that we have a migration guide before releasing 9.0. Happy to assist with any of this.

sunshinejr:
This is a great idea. I'd love it if you could help with migration guide, @ashfurrow. Especially because we will only deprecate providers, without removing them, we might get more feedback on the change as well (and thus we could see what features are really used in Moya).

ashfurrow:
Well, [Eidolon](https://github.com/artsy/eidolon) has become a sort of go-to implementation for how to do Rx with Moya. Why don't I migrate that over this week, and we can refer to the actual PR from the migration docs. It'll likely be a few days before I get to this, but I'll follow up 👍

sunshinejr:
Great! Thank you, Ash 🎉

Sources/Moya/MoyaProvider.swift
```diff
@@ -25,8 +25,15 @@ public struct ProgressResponse {
     }
 }
 
+public protocol MoyaProviderProtocol: class {
```
AndrewSB: thoughts on `MoyaProviderProtocol` v `MoyaProviderType`? I've always preferred the latter, but I don't know if I've fallen behind the times ;p

docs/Examples/ReactiveSwift.md
AndrewSB: 😮

Sources/Moya/MoyaProvider.swift
sunshinejr: I don't mind either so I did a quick research what we use in the project: we use `...Type` & `...able` for protocols and I agree, I'll change to `MoyaProviderType` 👍

docs/Examples/ReactiveSwift.md
sunshinejr: Was doing my own CR and didn't want to make multiple commits with code review fixes, so waited till someone did it as well so I can cumulate 😄

sunshinejr:
Thanks guys for the review! The migration in Eidolon looks like was easier than expected - and that's awesome. I've committed some changes you guys requested - please take another look in your spare time 😄

sunshinejr:
Btw. anyone have an idea why is CI failing? 🤔

ashfurrow:
Hmm not sure what's going on. Can you test locally to see if it passes from within Xcode? From the command line with `rake test`?

Moya.xcodeproj/project.pbxproj
```diff
@@ -302,23 +288,20 @@
...
 			files = (
-				55D38822179474648606A08E /* RxMoya.h in Headers */,
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 		};
 		F4CA905CBCCD3A2D1D028529 /* Headers */ = {
 			isa = PBXHeadersBuildPhase;
 			buildActionMask = 2147483647;
 			files = (
-				42414D9A4174BE3C8A63A2D4 /* Moya.h in Headers */,
```
SD10: @sunshinejr CI fail is possibly related to this?
`no umbrella header found for target 'Moya', module map will not be generated`

SD10: I think if you click on the `.h` files and go to _File Inspector_ -> Check the respective box for Target Membership this should resolve the issue:
screen shot 2017-07-05 at 3 33 05 pm

sunshinejr:
Ok I think, I fixed it (it actually passed from the state it was stuck on 😄 ). There was a problem with project (I had to regenerate it using `framework-generate`. I'll bring that up in another issue) + problem with Carthage deps.

sunshinejr:
Thank you guys so much for the reviews and help!

BasThomas:
Thanks for picking this up, @sunshinejr!

PR #1157, opened 2017-07-03
Update stale.yml configuration

SD10:
screen shot 2017-07-03 at 2 14 49 pm

I'm not sure if the discoloration in the syntax highlighting is what is causing `stale` to not close issues. I've compared our `stale.yml` to other projects and this is the only difference.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -27,7 +27,7 @@ unmarkComment: false
 closeComment: >
   This issue has been auto-closed because there hasn't been any activity for 59 days.
   However, we really appreciate your contribution, so thank you for that! 🙏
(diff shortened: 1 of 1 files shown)


SD10:
Removing the emojis could work too. I think we will have to wait about an hour+ to see if Probot makes any changes.

BasThomas:
It might be good to open an issue over at Probot as well, to get them in the loop? https://github.com/probot/probot/issues

SD10:
@BasThomas Great idea, I just want to get a better handle on the problem before opening a premature issue. I'm not sure if it's the emojis or just any colon in the message body.

PR #1158, opened 2017-07-03
Remove emojis from stale.yml configuration

SD10:
@pedrovereza per your recommendation I've removed the emojis from the `stale.yml` configuration. I've been able to find other GitHub repos with the close comment feature working as of 3 days ago, but none containing emojis.


SD10:
I found [babel](https://github.com/babel/babel.github.io) is using emoji literals in their stale comments. 
Possibly writing 👍 as `:+1:` could have thrown it off? I'm not too convinced.
I was skeptical of all colons (like the one in the link) but [hubotio](https://github.com/hubotio/hubot-redis-brain) is using links in their comments. I'll have to sleep on this and see if any thing has taken effect by morning.

PR #1159, opened 2017-07-05
Removes markdown from stale.yml

pedrovereza:
Related to #1071, #1157 and #1158.

Stale-bot is still not closing stale issues. I'm guessing this could be related to having markdown in the close comment 🤔.

Also confirmed that having `closingComment: false` worked on a test repo (https://github.com/pedrovereza/stale-bot-test/issues/1)

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -27,7 +27,7 @@ unmarkComment: false
 closeComment: >
   This issue has been auto-closed because there hasn't been any activity for 59 days.
   However, we really appreciate your contribution, so thank you for that!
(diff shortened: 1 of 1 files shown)


SD10:
@pedrovereza I did find repos using both emojis and markdown inside of the `markComment`. None inside of the `closeComment` but I would assume the behavior would be similar.

[babel](https://github.com/babel/babel.github.io) uses emojis.
[hubotio](https://github.com/hubotio/hubot-redis-brain) uses markdown.

I created an issue [here](https://github.com/probot/stale/issues/43) in the stale repo asking for some troubleshooting support.

Nonetheless, let's give this a try. I think when updating the `stale.yml` file the changes should be applied immediately by `probot`.

PR #1160, opened 2017-07-05
Set probot closeComment to false in stale.yml configuration

SD10:
@pedrovereza What about setting `closeComment` to false to test if this problem is comment related or if we have another problem? I think the effects should be applied immediately if the PR is a success.


pedrovereza:
@SD10 sounds good to me, let's see if it works now 😅

PR #1161, opened 2017-07-05
Remove quotes around 'no recent activity' label in stale.yml

SD10:
So setting the `closeComment` variable to false had no effect. This leads me to believe we have a different problem. I took a look at the source and this is the method that gets the stale issues:

```javascript
  getStaleIssues() {
    const labels = [this.config.staleLabel].concat(this.config.exemptLabels);
    const query = labels.map(label => `-label:"${label}"`).join(' ');
    const days = this.config.days || this.config.daysUntilStale;
    return this.search(days, query);
  }
```

It looks like when it creates the `query` it wraps the label in quotes. Since the current `stale.yml` already provides these quotes I think it results in an invalid query. I suggest we remove them with this PR and see if it has any effect. Sorry for the PR spam 😓

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -13,7 +13,7 @@ exemptLabels:
   - "chore"
 
 # Label to use when marking as stale
-staleLabel: "no recent activity"
+staleLabel: no recent activity
 
 # Comment to post when marking as stale. Set to `false` to disable
 markComment: >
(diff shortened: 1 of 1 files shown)


BasThomas:
I think, at this point, we might be better off opening an issue over at Stalebot, instead of doing this hit-and-miss approach. 😐

SD10:
@BasThomas I agree, this was my last idea. I already have an open [issue](https://github.com/probot/stale/issues/43).

PR #1165, opened 2017-07-07
convert "if let" -> "switch case"

AndrewSB:
stylistic change, take it or leave it. Would love to  hear what your thoughts are @sunshinejr, to which one seems more readable

### DIFF
--- a/Sources/Moya/MoyaProvider.swift
+++ b/Sources/Moya/MoyaProvider.swift
@@ -103,12 +103,11 @@ open class MoyaProvider<Target: TargetType> {
         let stub: () -> Void = createStubFunction(cancellableToken, forTarget: target, withCompletion: completion, endpoint: endpoint, plugins: plugins, request: request)
         switch stubBehavior {
         case .immediate:
-            if let callbackQueue = callbackQueue {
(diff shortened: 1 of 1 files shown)


Sources/Moya/MoyaProvider.swift
```diff
@@ -103,12 +103,11 @@ open class MoyaProvider<Target: TargetType> {
...
-            if let callbackQueue = callbackQueue {
-                callbackQueue.async {
-                    stub()
-                }
-            } else {
+            switch callbackQueue {
+            case .none:
                 stub()
+            case .some(let callbackQueue):
+                callbackQueue.async(stub)
```
pedrovereza: `callbackQueue.async(execute: stub)` :wink:

sunshinejr:
This is a good one! But I think a comment would be sufficient, no need to make a PR out of this 😄 I'm gonna merge it and add alias to the method as @pedrovereza pointed out 👍

Sources/Moya/MoyaProvider.swift
AndrewSB: thanks @pedrovereza

PR #1172, opened 2017-07-13
Add BasicAuthPlugin

yar1vn:
This plugin adds support for Basic Authorization header using an API key.

### DIFF
--- a/Moya.xcodeproj/project.pbxproj
+++ b/Moya.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		0968C9681F182A5000B5C404 /* BasicAuthPlugin.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0968C9671F182A5000B5C404 /* BasicAuthPlugin.swift */; };
 		0A5523C9271A2C73FC9B8BAA /* SignalProducer+Response.swift in Sources */ = {isa = PBXBuildFile; fileRef = BE381F1B80067731317902BF /* SignalProducer+Response.swift */; };
 		0D7C15675C8987BDA99DBA21 /* CredentialsPlugin.swift in Sources */ = {isa = PBXBuildFile; fileRef = A0FCC5560603A2B95F717FA7 /* CredentialsPlugin.swift */; };
 		14C482A52841F0D4C51A5A8E /* NetworkLoggerPlugin.swift in Sources */ = {isa = PBXBuildFile; fileRef = C09089EAE9CE725A67B34035 /* NetworkLoggerPlugin.swift */; };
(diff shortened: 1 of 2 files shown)


sunshinejr:
Hey @yar1vn, thank you for the PR! Seems like a good idea of introducing Basic auth. Although, we might want to add this one to  our existing [AccessTokenPlugin](https://github.com/Moya/Moya/blob/master/Sources/Moya/Plugins/AccessTokenPlugin.swift). Because right now this plugin is not really flexible - you can't use both OAuth & Basic on different endpoints (sometimes it happens) - you either use it or not. 

I'm thinking about updating the `AccessTokenAuthorizable` protocol so it would look like:

```swift
enum AuthorizationType {
    case None
    case Basic
    case OAuth
}

public protocol AccessTokenAuthorizable {

    /// Declares whether or not `AccessTokenPlugin` should add an authorization header
    /// to requests. If it should, what type of authorization does this request need?
    var authorizationType: AuthorizationType { get }
}
```

And then it would be easy to append header based on the type specified:
```swift
    public func prepare(_ request: URLRequest, target: TargetType) -> URLRequest {
        guard let authorizable = target as? AccessTokenAuthorizable, authorizable.authorizationType != .None else {
            return request
        }

        var request = request
        switch authorizable.authorizationType {
            case .OAuth:
                request.addValue(authVal, forHTTPHeaderField: "Authorization")
            case .Basic:
                // add Basic header
        }
        return request
    }
```

This makes it really flexible to use and also this way we can really cover most of the use-cases in the networking layer with auths.

What do you think about it, @yar1vn?

yar1vn:
@sunshinejr seems like a good idea.
Not sure what would be the use case for `none` (lowercase in Swift 3). Also I only see the implementation for Bearer in `AccessTokenAuth`.
Anyway you seem to have implemented most of it in your comment so feel free to go ahead and include this code wherever it makes more sense. 

We (currently) don't have a need for OAuth in our project, and I'm also using [SugarMoya](https://github.com/devxoul/MoyaSugar) -> `httpHeaderFields` property to add Basic Auth instead of a plugin.
Note: SugarMoya has a really nice wrapper for Moya by using `Route`. You might want to consider adding it to Moya.

SD10:
@yar1vn I believe @sunshinejr was just trying to mock up a good way to add this functionality considering the state of Moya's current API. 

The code he provided in the comment is a rough draft and may not even compile due to the switch statement not being exhaustive. There may be other things to implement such as updating docs/tests.

Maintaining a project requires a large effort and we're always encouraging new contributors. Small contributions have a snowball effect and save maintainers a great deal of time. 

Would you be interested in moving this forward?

yar1vn:
@SD10 that explain why I couldn't see the implementation for OAuth. I thought I was looking at the wrong file.
As I mentioned, we are currently not using OAuth so I don't have any code from my project to contribute besides Basic Authentication. `AccessTokenPlugin` only implementers Bearer so I'm not sure it makes sense to add OAuth on top of Basic Authorization.

SD10:
@yar1vn Sorry I'm not an expert in this area. Could we just switch on the `authValue` property to provide a different value based on the case?

yar1vn:
@SD10 I'm definitely not an expert in OAuth - but the implementation is a lot more complicated than just adding another field to the http header. There are entire frameworks that handle OAuth.

sunshinejr:
What I meant was to update what we have with `Basic` authorization on top of it. We can name it `Bearer` type, instead of `OAuth`, to describe it properly. We would just send a `Bearer` or `Basic` token in the header, so it makes sense to name it that way. And also, as @SD10 mentioned, to that we would also need to add tests, `Changelog` entry (with a breaking change) and check docs/update them accordingly.

Would you be up for that one, @yar1vn?

yar1vn:
Let's do it. How do I setup an environment where I can build and run tests?

BasThomas:
Check out https://github.com/Moya/Moya/blob/master/docs/Development.md :)

sunshinejr:
Hey @yar1vn! Just a heads-up that we just released 9.0.0-alpha.1 version of Moya. We would be up for taking your PR in the next release, so if you'd have time to implement it in the near future it would be awesome 👍 Thanks for taking this one!

yar1vn:
Great. I'll find time as soon as I can take a break. Got a tight deadline on a new app 🙂

sunshinejr:
Sure, @yar1vn. Thanks for your response. 🐼  

Please let me know if you'll be able to do it this week, would be awesome! Otherwise we might just help you out since this is a breaking change and is a perfect fit for a new major release we are planning.

yar1vn:
Hmm, I wouldn't have time this week while I'ma t work. Possibly on the weekend but I wouldn't want to be the one blocking anything 😳

SD10:
This shouldn't take that long to make the changes to sources, tests, and docs.
@yar1vn If you end up not being able to get around to making the changes this weekend, would you like me to take care of it for you?

sunshinejr:
We don't want you to feel pressured to do this, @yar1vn, since nothing good happens from it. Thus, I think, that we should just take it from here and you can always contribute in the future when you have more free time. We will always have some tasks for starters, so don't worry about that :) Thanks for the initiative and your time spent on this one! 🙇 

@SD10 could you please create an issue and then we can move there and close this one? :)

sunshinejr:
Moving this to #1214.

PR #1173, opened 2017-07-15
[Documentation] Move threading docs to Threading.md & expand it a bit.

sunshinejr:
So this should be a final step to merge queues (#762). I've moved the docs from `Readme.md` to its own `Threading.md` + added examples for both normal usage and `ReactiveSwift`. 

Also, I've checked the docs but didn't found anything on `queues` I needed to change, so it should be all set 👍

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -242,24 +242,6 @@ for filtering out certain status codes. This means that you can place your code 
 handling API errors like 400's in the same places as code for handling invalid
 responses.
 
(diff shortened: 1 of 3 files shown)


SD10:
I like the content of `Threading.md`. Let's also add a link to it in the docs/README.md? Possibly right after the RxSwift & ReactiveSwift sections since it builds on them.

sunshinejr:
Good catch, @SD10! Added this one as well.

PR #1178, opened 2017-07-20
[General] Add public URL(target:) initializer.

sunshinejr:
This is a little gift, because I always need to implement my own function/extension that creates `URL` from `TargetType`. I figured that other people might find it useful too 😄

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -8,6 +8,7 @@
 - **Breaking Change** Deprecated `ReactiveSwiftMoyaProvider` and `RxSwiftMoyaProvider`. Use `MoyaProvider` with reactive properties now: `provider.reactive._`, `provider.rx._`. In case you were subclassing reactive providers, please take a look at [this PR from Eidolon](https://github.com/artsy/eidolon/pull/669). It covers migration from subclassing given providers, to usage by composition.
 - **Breaking Change** Removed deprecated in Moya 8.0.0: `Moya.Error`, `endpointByAddingParameters(parameters:)`, `endpointByAddingHttpHeaderFields(httpHeaderFields:)`, `endpointByAddingParameterEncoding(newParameterEncoding:)`, `endpointByAdding(parameters:httpHeaderFields:parameterEncoding)`, `StructTarget`, `filterStatusCodes(range:)`, `filterStatusCode(code:)`, `willSendRequest(request:target:)`, `didReceiveResponse(result:target:)`, `ReactiveCocoaMoyaProvider`, `ReactiveSwiftMoyaProvider.request(token:)`. 
 - Fixed a bug where you would have two response events in `requestWithProgress` method on ReactiveSwift module.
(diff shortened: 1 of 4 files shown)


BasThomas:
Would it not make more sense to make the `private` func public? Having this extension on `URL` feels a bit weird / hard to discover.

sunshinejr:
Hmm, you might be right and it might be hard to discover manually. But I think you will find it from the docs anyways. But I'm open for a change 😄

SD10:
I like the `URL` init over having to access a class method on `Endpoint`. It feels right.

I guess this raises the question of providing these kinds of utilities in Moya, considering I think we haven't extended Apple's frameworks thus far?

From [Targets.md](https://github.com/Moya/Moya/blob/master/docs/Targets.md):
```Swift
extension String {
    var urlEscaped: String {
        return addingPercentEncoding(withAllowedCharacters: .urlHostAllowed)!
    }
}
```
I believe there also use to be one like this: 
(which I'm thankful is removed because it confused me when starting out and not seeing the bottom)
```Swift
extension String {
    var utf8StringEncoding: Data {
        return data(using: String.Encoding.utf8)!
    }
}
```

Just my two cents, to say we discussed it in the case someone adds another extension and tries to point to this as justification.

sunshinejr:
Good point @SD10! I think that it might be a rare case that we extend Apple framework, but if we do, I feel like it has to be something unique to Moya. In this case, we are creating `URL` from a `TargetType`, which is only viable for `Moya` codebase. This way users shouldn't get confused since the type accepted by this `init` is important only for Moya users. In case of these other two extensions you showed, they may seem like standard library methods 😄

SD10:
> I think that it might be a rare case that we extend Apple framework, but if we do, I feel like it has to be something unique to Moya.

This sounds like a good precedent 👍. Where to document it? Endpoints.md?

Endpoints.md...

----

The first might resemble the following:
```diff
let endpointClosure = { (target: MyTarget) -> Endpoint<MyTarget> in
+   let url = URL(target: target).absoluteString
    return Endpoint(url: url, sampleResponseClosure: {.networkResponse(200, target.sampleData)}, method: target.method, parameters: target.parameters)
}
```

This is actually the default implementation Moya provides. If you need something custom, like if your API requires custom parameter mapping, or if you're creating a test provider that returns non-200 HTTP statuses in unit tests, this is where you would do it.

```diff
+ Notice the URL(target:) initializer, Moya provides this convenient extension to create a URL from any TargetType.
```

sunshinejr:
Yeah, I think `Endpoint` is the place where we use this extension and where you would need it for a basic implementation. Would you be up for updating the docs in another PR? :)

SD10:
Yeah @sunshinejr I'll take care of it tomorrow afternoon

sunshinejr:
👍 What do you think about it, @BasThomas? :)

BasThomas:
👍

Issue #1126, opened 2017-06-15
Provide sampleData in Provider instead of TargetType


tbaranes:
I'd like to bring back an old subject: #425, I know it has been closed but since there was a solution suggested, I think it could be worth a second though!

In the current state of Moya, everyone has to implement `sampleData` in the `TargetType`. It's great to make testing APIs easy, but I don't think it's right to have to embed the stubbed response directly in the app instead of having them only in the tests target.

I know there was a few potential workaround for this in #425, but they all comes with downsides: won't work anymore on simulators, breaks the target types basic flow...  In the end, I completely agree with Hirad's solution: adding a `sampleData` response to `StubClosure`, and removing `sampleData` from `TargetType`. 

One downside that Ash brought up: it won't encourage people to add tests, but I do think, if people doesn't want to test, they will just return `Data()`. So, in the end, I see more advantages of this implementation than the current one (except that it's introducing a breaking change).

Could it become a reality soon? 🙃

BasThomas:
When doing a search, it does seem like a lot of people just `return Data()`: https://github.com/search?q=%22var+sampledata%22+language:Swift+extension:.swift&type=Code&utf8=%E2%9C%93.

On it being a breaking change: we're releasing `9.0` soon, and with Swift 4 around the corner, there will probably be room to include this in `10.0` in the near future if we want to go this way.

SD10:
I completely agree with @tbaranes. I find the inclusion of testing code directly in the app unpalatable. As someone who is passionate about testing, I would like to have this separation of concerns.

I don't really believe testing is a practice that can be encouraged through an API. If one does not see the benefit of or appreciate writing tests, they are simply are not going to do so.

Furthermore, from the perspective of responsibility, I think it makes far more sense for `sampleData` to be the concern of the `Provider` as opposed to the `TargetType`.

BasThomas:
What do other @Moya/contributors think?

colinta:
I was, for a long time, of the "let's encourage testing" mindset, but these days I'm leaning more towards having a good separation of concerns, and hoping people choose to take the time to have stubbed responses (but it's up to them).

I think including it in the provider makes a ton of sense, I'm on board with this change.

ashfurrow:
Yup, makes sense to me.

Dschee:
[...]

PR #1183, opened 2017-07-22
Adds Vision document

### LINKED ISSUE #1081: Vision document / FAQ
Do you think we, @Moya/contributors, can create a Vision document? This would help us create some clear goals for the project as well as giving the community and users a better idea of what we expect of this project going forward.

What do you think? And more concrete: what should be in it?

FAQ: 
Separating subspecs into repositories
Model decoding: soon with Apple's APIs
https://github.com/Moya/Moya/issues/1126#issuecomment-312125039

ashfurrow:
This is a draft of the vision document requested in #1081 to help guide project direction. I want to make sure we cover everything, so please only consider this a starting point for the community to guide from here. I've also tried to include the history of the project to contextualize the goals.

@Moya/contributors your input would be welcome 🙇

Fixes #1081.


SD10:
Maybe add [#1126 comment](https://github.com/Moya/Moya/issues/1126#issuecomment-312125039):
- favor explicitly defined behavior over default implementations

There are also a couple other things mentioned in #1081. Such as object decoding.

This document heavily refers to `requests` but doesn't talk about the `responses`. 

Currently, I wouldn't say Moya is in the business of response handling -- aside from providing a few `Moya.Response` extensions. But what's the vision for this? Are we for or against this behavior?

BasThomas:
Two other small things:

- we might want to reference to the Vision document in (at least the) Readme.
- for consistency call the document `Vision.md` (instead of `VISION.md`) - or update all top level files to use uppercase.

ashfurrow:
Okay, cool. Good ideas everyone – I've made some changes and pushed.

mikesanderson:
Hey, I think this is great. Thanks @ashfurrow and @orta for founding this and staying involved. 

I think the Project Goals list could be easier to grasp by adjusting the order, grouping the goals starting with "provide..." together and ordering in a progression from requests to responses to customization and extension. I went ahead and committed the edit-- easier to show the result than describe it. Let me know. 

@orta the no caps / full stops gets me to, but I can't think of a solution that doesn't look worse. I think each has an implicit "Moya will..." or "First goal is to..." and that makes it work the way it is.

ashfurrow:
Okay, I've un-WIP'd this PR. Going to give a few extra days for @Moya/contributors to chime in.

pedrovereza:
I think the suggestion from @SD10 is valid:
> Maybe add #1126 comment:
> favor explicitly defined behavior over default implementations

He also raised a good question on what's our vision regarding responses in general:

> Currently, I wouldn't say Moya is in the business of response handling -- aside from providing a few Moya.Response extensions. But what's the vision for this? Are we for or against this behavior?

ashfurrow:
@pedrovereza makes sense, I'll amend for the first point. I added something for response handling already, let me know what you think @SD10.

pedrovereza:
@ashfurrow Sorry, I totally missed the response part on the document 🙈

PR #1192, opened 2017-07-30
Update docs for reactive changes

### LINKED ISSUE #1191: Update reactive docs.
In 9.0.0 release we deprecate both reactive providers. However, in our `Readme.md` we still encourage to use them. We need to update `Readme.md` to use classic `MoyaProvider` and both `provider.rx.request` & `provider.reactive.request` instead. 

We also forgot about updating [docs/RxSwift](https://github.com/Moya/Moya/blob/master/docs/RxSwift.md) and [docs/ReactiveSwift](https://github.com/Moya/Moya/blob/master/docs/ReactiveSwift.md). I think we don't need to remove the reactive providers usage in the docs (besides Readme), but we should label it with **DEPRECATED** sign. And add both requests usage (with and without progress) at the top of the file maybe.

When we cover all of the above, we can remove both providers docs from  examples ([docs/Examples/RxSwift](https://github.com/Moya/Moya/blob/master/docs/Examples/RxSwift.md), [docs/Examples/ReactiveSwift](https://github.com/Moya/Moya/blob/master/docs/Examples/ReactiveSwift.md)). Right now I feel they do not provide any value. We just need to make sure we got everything in the `docs` part and it should be alright.

BasThomas:
Fixes #1191.

Some points:

- I've used the normal initializers (`MoyaProvider<GitHub>()` and add the `.rx` and `.reactive` extensions before the request. This means the provider object is just a `MoyaProvider`. Is this the intended use, or should we create an Rx-provider like `MoyaProvider<GitHub>().rx`?
- I've not yet touched the examples.
- I did not any warnings for deprecation; this is due to the fact we have fix-its for this, and I don't see why it would be necessary to add this.
- The request progress has not been added yet either.


Readme.md
```diff
@@ -203,8 +203,8 @@ that you can start or bind or map or whatever you want to do. To handle errors,
 for instance, we could do the following:
 
 ```swift
-provider = ReactiveSwiftMoyaProvider<GitHub>()
-provider.request(.userProfile("ashfurrow")).start { event in
+provider = MoyaProvider<GitHub>()
```
sunshinejr: Yup, usage is 💯

BasThomas: Wasn't sure about it. This way you can more easily switch (but why?), if you use the suffix immediately you can't forget it by mistake.

BasThomas: Curious how many people will "forget" this and end up with unexpected results. :/

docs/ReactiveSwift.md
```diff
@@ -1,22 +1,22 @@
...
 
-Moya provides an optional `ReactiveSwiftMoyaProvider` subclass of
+Moya provides an optional `ReactiveSwift` implementation of
 `MoyaProvider` that does a few interesting things. Instead of
 calling the `request()` method and providing a callback closure
 to be executed when the request completes, we use `SignalProducer`s
 (`RACSignal`s are also available for those who need it).
 
-A `ReactiveSwiftMoyaProvider` can be created much like a
+A `ReactiveSwift`-based `MoyaProvider` can be created from a
```
sunshinejr: I'd change that a little bit. It feels a bit odd to me. Maybe something along the lines:
> To use reactive extensions you don't need any additional setup. Just use your `MoyaProvider` instance.

Readme.md
sunshinejr: Oh, this is probably more of a preference and I don't mind both - just the general assumption and code style is to rather preserve the original object and use extension when needed. Also, I wouldn't worry so much about forgetting to use `rx` or `reactive`, since these are really extensively used by people who use reactive programming with ReactiveSwift or RxSwift.

sunshinejr:
Thanks for doing this, @BasThomas! I agree that deprecation warnings may not be needed 👍 Would you be also up for completing all of the rest in this PR? I mean finishing the examples & progress?

BasThomas:
I can definitely remove the examples and update the text. Not sure what you would like to add regarding the progress?

docs/ReactiveSwift.md
BasThomas: I've updated this in both `ReactiveSwift.md` and `RxSwift.md` 👍

sunshinejr:
We can just mention that these exist, it doesn't need to be another big example :)

sunshinejr:
Would you be up for adding these in this PR @BasThomas or let's just add it later? :)

BasThomas:
I've opened #1203; lets merge this for now.

PR #1193, opened 2017-07-30
Update core to support Swift 4.

sunshinejr:
In this PR I've updated `NetworkLoggerPlugin` as it was the only reason Swift 4 compiler was having troubles with moya. I've also updated our CI to use Xcode 9. In the future we should look to support 2 builds - with Swift 3.2 & 4.0. 

Right now, with this changes, 9.0.0 release should support:
- Swift 3.1
- Swift 3.2
- Swift 4.0

I've tested it locally and it seems to be working correctly.

**Edit:** Oh, and also I've created a new **9.0.0-dev** branch.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,4 +1,5 @@
 # Next
+- Added Swift 4.0 support.
 
 # 9.0.0-alpha.1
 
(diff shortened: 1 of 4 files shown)


sunshinejr:
Sure! Good catch, @pedrovereza, PR updated :)

PR #1213, opened 2017-08-11
[RxSwift] Add operators to Single<Response>

sunshinejr:
This was discussed in #1020, but I think it wasn't added yet. I found this one out by playing with new API and I certainly would love these to be in the next release (having to add `asObservable()` and loose all the nice `Single` API would be really sad).

Also, I've read that @AndrewSB wanted to do a generic protocol of some sort, I also thought about it, but I don't think there is an easy/clean way to do that. `RxSwift` uses similar approach to this one in PR (see [here](https://github.com/ReactiveX/RxSwift/blob/1314805175263ad8141054f6be21f6bbef224d2c/RxSwift/Traits/PrimitiveSequence.swift#L464-L475) for instance)


AndrewSB:
👋

Thanks for doing this @sunshinejr, let's also remove the functions that are on `ObservableType`, since the ones you've added on `PrimitiveSequence` should work for both `Single`s and `Observable`s

sunshinejr:
Hmm, are you sure that it works for both @AndrewSB? I don't see it in code, but I might be missing something.

AndrewSB:
From my understanding, observables are primitivetypes, but I'm not sure
either

I'll take a look soon

AndrewSB:
https://github.com/ReactiveX/RxSwift/blob/master/RxSwift/Traits/PrimitiveSequence.swift#L60

It looks like PrimitiveSequence conforms to ObservableConvertibleType... so I was wrong, we need to keep the operators duped between Observable & Primitive :/ 

sais la vie

sunshinejr:
Oh, you are right. Maybe we will get something better in the future, doesn't seem too clean to me as well. 

Nonetheless, thanks for the CR, @AndrewSB ❤️

Issue #1214, opened 2017-08-11
Add Basic Auth Plugin


SD10:
### Add Basic Auth Plugin

As discussed in #1172, `AccessTokenAuthorizable` can be updated using an enum `AuthorizationType` with the cases `none`, `basic`, and `bearer` to determine the header value.

### TO DO:
- [x] Implementation
- [x] Testing
- [x] Update Documentation

sunshinejr:
This can be closed and move any conversation to #1215 where is the implementation of this one.

PR #1215, opened 2017-08-12
Add AuthorizationType with Basic header

SD10:
This is in favor of #1172 and related to #1214.

I've added the `AuthorizationType` enum, updated the existing tests, and added a new one for `.basic` authorization.

### Thoughts:
The old implementation of `AccessTokenAuthorizable` added an authorization header of "Bearer" to all target types by default. We need to determine which authorization type we're going to default to - if any. The current tests represent defaulting to the case `.none` -- no headers added. This would be a breaking change to the current and default expected behavior.

What do you guys think?


SD10:
It looks like the build error is from compiling the project in Xcode 8? Admittingly, I haven't been using Xcode 9 beta enough 😅

sunshinejr:
We've changed from 8.3.3 to 9.0.0 and the there is still cache built with 8.3.3. I re-triggered your build without cache, @SD10, and it seems fine :)

Sources/Moya/Plugins/AccessTokenPlugin.swift
```diff
@@ -28,15 +37,11 @@ public struct AccessTokenPlugin: PluginType {
...
-    }
-
     /**
      Initialize a new `AccessTokenPlugin`.
 
      - parameters:
-       - token: The token to be applied in the pattern `Authorization: Bearer <token>`
+       - token: The token to be applied in the pattern `Authorization: <AuthorizationType> <token>`
     */
     public init(token: String) {
```
sunshinejr: I was thinking also about this `token` here. Normally you don't really have your token when initializing the `MoyaProvider`. Could we use a closure here that will get us a token? Marking it with `autoclosure` should also allow people with token upfront to have a nicer migration. 

What do you think?

Sources/Moya/Plugins/AccessTokenPlugin.swift
```diff
@@ -51,12 +56,20 @@ public struct AccessTokenPlugin: PluginType {
      - returns: The modified `URLRequest`.
     */
     public func prepare(_ request: URLRequest, target: TargetType) -> URLRequest {
-        if let authorizable = target as? AccessTokenAuthorizable, authorizable.shouldAuthorize == false {
-            return request
-        }
+        guard let authorizable = target as? AccessTokenAuthorizable else { return request }
```
sunshinejr: Nice! Really a clean way of implementing it :)

Sources/Moya/Plugins/AccessTokenPlugin.swift
SD10: I'm fine with that, a closure would be far more flexible. Good idea regarding using `autoclosure` 😃

SD10: @sunshinejr Were you thinking of removing the `token` property and replacing it with a `tokenClosure` too?
```Swift
let tokenClosure: () -> String

public init(tokenClosure: @autoclosure () -> String) {
   self.tokenClosure = tokenClosure
}
```
Do we want this to be escaping?

sunshinejr: We probably need to since it will be used in later stages of the endpoint -> request cycle.

SD10:
@sunshinejr I made the requested `tokenClosure` changes. If anyone else has any other changes or feedback just let me know. Once the implementation is agreed on I'll:
- [ ] Update Docs
- [ ] Add a CHANGELOG entry
- [ ] Do you want me to add something in the MigrationGuide.md?

Sources/Moya/Plugins/AccessTokenPlugin.swift
```diff
@@ -6,40 +6,45 @@ import Result
...
 /**
- A plugin for adding bearer-type authorization headers to requests. Example:
+ A plugin for adding basic or bearer-type authorization headers to requests. Example:
 
  ```
  Authorization: Bearer <token>
+ Authorization: Basic <token>
  ```
 
  - Note: By default, requests to all `TargetType`s will receive this header. You can control this
```
SD10: This probably needs to be removed now

sunshinejr:
👍 Code-wise seems good to me :)

SD10:
@sunshinejr I updated the documentation and the `CHANGELOG.md`. Let me know how those look. 
Also, let me know if you want me to write something in `MigrationGuide.md`.

sunshinejr:
Yeah, I feel like we should add something to the `MigrationGuide`, especially if we have this file already created. I didn't closely check the PR, but from the brief overview it looks great 👍 

Also, now that I think about it, we should add few lines about migrating to new reactive APIs, but thats another issue.

SD10:
@sunshinejr Done. Sorry for all the push spam. My rebase attempt failed so I just took care of the CHANGELOG conflict with a commit.

sunshinejr:
No problem, @SD10 👍 Is this PR still `WIP`? If you feel like this is ready for a review, we can remove the prefix, so people can chime in.

About migration guide and reactive providers - you are correct. We should stop using reactive providers and use `rx` and `reactive` namespaces instead. We could also point to this [PR from Ash in Eidolon](https://github.com/artsy/eidolon/pull/669), that covers migration from subclassing to composition.

SD10:
@sunshinejr Yeah, I removed the WIP tag. I'll open a PR to work on the MigrationGuide for the reactive API sometime later today.

Sources/Moya/Plugins/AccessTokenPlugin.swift
```diff
@@ -51,12 +54,20 @@ public struct AccessTokenPlugin: PluginType {
...
-        }
+        guard let authorizable = target as? AccessTokenAuthorizable else { return request }
+
+        let authorizationType = authorizable.authorizationType
 
         var request = request
-        request.addValue(authVal, forHTTPHeaderField: "Authorization")
+
+        switch authorizationType {
+        case .basic:
```
pedrovereza: Since `.basic` and `.bearer` have the same implementation, maybe it'd be better to do `case .basic, .bearer:`?

Sources/Moya/Plugins/AccessTokenPlugin.swift
```diff
@@ -6,40 +6,43 @@ import Result
...
+    /// Represents the authorization header to use for requests.
+    var authorizationType: AuthorizationType { get }
+}
+
+// MARK: - AuthorizationType
+
+/// An enum representing the header to use with an `AccessTokenPlugin`
+public enum AuthorizationType: String {
+    case none
+    case basic = "Basic "
```
pedrovereza: The trailing space feels a bit weird. I'm assuming it's because of the formatting in:
```swift 
request.addValue(authorizationType.rawValue + tokenClosure(), forHTTPHeaderField: "Authorization")
```

If so, maybe we take the space out of the enum and properly combine `AuthorizationType` with the `tokenClosure` when adding the header?

SD10:
@pedrovereza Changes made. Thank you for jumping in and giving me your review! Those were some great suggestions 😃

PR #1218, opened 2017-08-15
[Project] Fix Single+Response owner to RxMoya.

### LINKED ISSUE #1217: RxSwift linked to 9.0.0-dev even if not needed
I'm using the most current commit of the 9.0.0-dev Branch and I'm not using any reactive libraries. Since updating the dependency from my own fork (commit 4bcac8144fd438b9b93c2a9907a8581dfd899278) to commit 0c25037dcf2a4213124fe0a423656b14ee0209d4 I'm seeing the following error when building:

```
dyld: Library not loaded: @rpath/RxSwift.framework/RxSwift
  Referenced from: /Users/JamitLabs/Library/Developer/CoreSimulator/Devices/98DE006C-DE6C-45D2-9A42-4440CFD6DA9C/data/Containers/Bundle/Application/F1A76703-C1FA-4236-AD1B-7167A09B72F6/TreasuryManager.app/Frameworks/Moya.framework/Moya
  Reason: image not found
```

I have not included the RxSwift library into my project, but I don't want to since I'm not using reactive Moya. The inclusion is done via the current Carthage version. I would expect that error to not appear so I don't have to include RxSwift like it always worked before.

sunshinejr:
This should fix #1217. By mistake, new file was added to `Moya` target rather than `RxMoya`. Can you confirm, @Dschee, that it works on this commit?


Dschee:
I can confirm that it works. Thank you for the quick fix @sunshinejr.

PR #1224, opened 2017-08-16
[SPM] Fix warnings when building Moya module.

sunshinejr:
`Supporting files` directory included prints a warning and I get an error from not including `Foundation` in these two files, so this should fix it 🤞 

Also, we might need to add compatibility checks on our CI soon.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -22,6 +22,7 @@ let package = Package(
         .Package(url: "https://github.com/antitypical/Result.git", majorVersion: 3)
     ],
     exclude: [
-        "Tests"
+        "Tests",
+        "Sources/Supporting Files"
     ]
 )
--- a/Sources/ReactiveMoya/ReactiveMoyaAvailability.swift
+++ b/Sources/ReactiveMoya/ReactiveMoyaAvailability.swift
@@ -1,6 +1,7 @@
 #if !COCOAPODS
     import Moya
 #endif
+import Foundation
 import ReactiveSwift
 
 /// Subclass of MoyaProvider that returns SignalProducer instances when requests are made. Much better than using completion closures.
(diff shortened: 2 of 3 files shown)


freak4pc:
Yikes 
> Module compiled with Swift 3.1 cannot be imported in Swift 3.2: /Users/distiller/Moya/Carthage/Build/tvOS/Alamofire.framework/Modules/Alamofire.swiftmodule/x86_64.swiftmodule

SD10:
Same error recently happened to me. I think @sunshinejr fixed it by clearing the build cache 🤔

sunshinejr:
This error is gonna haunt me forever. Not sure why I always need to rebuild without cache. How does it get this old cache? Anyone has an idea? 🤔

freak4pc:
First time i'm seeing CircleCI and can't say it feels very reliable thus far 🤔

freak4pc:
Seems like you can just add a post hook to manually clear the cache. But still super annoying: https://circleci.com/docs/1.0/how-cache-works/#clearing-cache

sunshinejr:
Clearing cache is not something I would pursue, since building dependencies each time might be too painful. I'm thinking if this is a temporary problem, since on master we have this old cache with 3.1, and on 9.0.0-dev we have new cache with 3.2. Maybe new PRs are getting cache from master and that's the problem? 🤔

freak4pc:
Yeah that makes total sense, since I imagine the entire project reuses some of the same caches/machines? I guess we can "wait the pain out" until 9.0.0 is out next month, right? :)

SD10:
Yeah that makes sense the docs updates / gardening of master matches the inconsistency of the error.
Next month?! 😱 We're so close!!

sunshinejr:
Yeah, let's just wait that out until we merge it for next version (which should be a lot faster than one month 😄)

PR #1237, opened 2017-08-23
`Progress` into the response when calling progress callback on completion

phimage:
The object is accessible, why not receive it.
At completion, some information into `Progress` object could be very useful

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -8,8 +8,8 @@
 - Added all the `filter`/`map` operators that were available for `Observable<Response>` to `Single<Response>` as well.
 - Added `AuthorizationType` to `AccessTokenAuthorizable` representing request headers of `.none`, `.basic`, and `.bearer`. 
 - Added tests for `Single<Response>` operators.
-- Fixed a bug where you weren't notified on progress callback for data request.
 - Added `Progress` object into the response when calling progress callback on completion.
+- Fixed a bug where you weren't notified on progress callback for data request.
 
 # 9.0.0-alpha.1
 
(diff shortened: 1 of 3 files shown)


Changelog.md
```diff
@@ -9,6 +9,7 @@
 - Added `AuthorizationType` to `AccessTokenAuthorizable` representing request headers of `.none`, `.basic`, and `.bearer`. 
 - Added tests for `Single<Response>` operators.
 - Fixed a bug where you weren't notified on progress callback for data request.
+- Added `Progress` object into the response when calling progress callback on completion.
```
sunshinejr: Can we add this one before `Fixed` entry? Just to keep it in parts `Added` -> `Fixed`...

Sources/Moya/MoyaProvider+Internal.swift
```diff
@@ -250,7 +250,18 @@ private extension MoyaProvider {
             let result = convertResponseToResult(response, request: request, data: data, error: error)
             // Inform all plugins about the response
             plugins.forEach { $0.didReceive(result, target: target) }
-            progressCompletion?(ProgressResponse(response: result.value))
+            if let progressCompletion = progressCompletion {
```
sunshinejr: Do you think we should add tests that check whether the last result have both progress & response?

phimage: We can add it to prevent behaviour change

phimage: Not a Nimble expert, so I do not find any sequence `containNil` predicate ( I try to call `contain` or `containElementSatisfying`)

I keep all progress objects into this array
```swift
var progressObjects: [Progress?] = []
```
then I test like this
```swift
 expect(progressObjects.filter{$0 != nil}.count) == 5
```
or if I test only the last
```swift
expect(progressObjects.last).toNot(beNil())
```

PR #1239, opened 2017-08-25
Version table update

### LINKED ISSUE #1186: Update master readme for Swift 3.2 and Swift 4.0
Current Moya Readme shows:

| Swift version | Moya version  |
| ------------- | ------------- |
| 3.X           | >= 8.0.0      |

* Can you double check that Moya 8.0.0 is OK for Swift 3.2 (on Xcode 9)? Otherwise, please specify the Moya branch/version for it
* Can you document the Moya branch/version for Swift 4?

BasThomas:
Removed Swift 2 and added Swift 4. :)

Closes #1186


Readme.md
```diff
@@ -52,15 +52,13 @@ ready for production use.
...
-Because of the many Swift versions Moya supports, it might be confusing to
-find the version of Moya that you need. Below is a table that shows which version of Moya
-you should use for your Swift version.
+Below is a table that shows which version of Moya you should use for
+your Swift version.
 
 | Swift version | Moya version  |
 | ------------- | ------------- |
-| 3.X           | >= 8.0.0      |
-| 2.3           | 7.0.2 - 7.0.4 |
```
sunshinejr: I wouldn't remove old Swift versions, especially 2.3 which is still used in some projects.

Readme.md
```diff
@@ -52,15 +52,13 @@ ready for production use.
...
-| 2.2           | <= 7.0.1      |
+| 4.X           | >= 9.0        |
```
sunshinejr: 4.X would be supported probably only for Moya/Core, not for Moya/RxSwift (dunno how about Moya/ReactiveSwift, but probably the same situation as with RxSwift). Not sure how to describe it in this table, though.

Readme.md
```diff
@@ -52,15 +52,13 @@ ready for production use.
...
+| 3.X           | 8.0.0 - 8.0.5 |
```
sunshinejr: 3.0 is probably not working from some of the last versions (like 8.0.3 and up maybe?). Not sure if all versions from 8.0 support Swift 3.1/3.2. It might be good idea to separate Swift 3.0, 3.1, 3.2 in this table.

Readme.md
SD10: Thank you, I was biting my tongue on this 😓

Readme.md
Coeur: That would be a shame for RxSwift. I suggest you reconsider supporting Swift 4.

BasThomas: Because of waiting for the dependencies to update to Swift 4? Or why?

Readme.md
BasThomas: 🤔, you might be right there. I wonder why the table wasn't updated when Swift 3.0 was dropped (if it was? Not sure)

Readme.md
sunshinejr: @BasThomas @Coeur This is because we can't support two major versions of reactive extensions libraries at once. We could support Swift 4 in our core with very little change in our codebase, so we did it. Of course we will create a new version with new major versions of reactive extensions libraries just after 9.0.0, where beta might be really soon available.

About the table, maybe we should split the Moya version column on Moya and Moya/RxSwift, Moya/ReactiveSwift.

sunshinejr:
Hey @BasThomas, how are we with the updates? 😉

BasThomas:
I'll take a look later today!

BasThomas:
@sunshinejr - is this what you meant? If not, I don't really know what should be done further... let me know :)

sunshinejr:
This is good, thanks, @BasThomas! What we should do next, I think, is that we should check whether `3.*` is true for all 3.0 Swift versions that we have in this table. So I've imagined that if there is a problem with 3.0, we would have 2 rows: one for 3.0 (with Moya versions that are working for Swift 3.0) and one for 3.1-3.2 Swift (with Moya versions that work for Swift 3.1-3.2). Does it make sense?

BasThomas:
That does make sense. I *think* that all versions should work fine with all `3.x` versions though. Why do you think they might not be?

sunshinejr:
Oh, I was pretty sure that there was some issue that not all version were working correctly. So maybe let's just double check if it works? Will you have time to do that maybe, @BasThomas?

BasThomas:
Unfortunately not. Think that would involve downloading old versions of Xcode as well, right? :(

sunshinejr:
I'm currently using Xcode 8.2.1 for 2.3 project, but it has probably 3.0.2 Swift and I can use 3.1.1 Swift toolchain there as well. On the other hand 3.2 & 4.0 Swift can be checked on Xcode 9.0 beta I think. This is unfortunately time consuming and I'm not sure when/if I can do it, so maybe someone with more free time could pick this one, @Moya/contributors?

sunshinejr:
FYI, I removed the milestone as the 9.0.0 is released already. We should try to get it merged ASAP, though!

SD10:
We still need some more work here on determining which versions of Moya are compatible with `Swift 3.0` and `Swift 3.1-3.2`

pedrovereza:
Support for Swift 3.0 started on `8.0.0`. According to the Changelog, [Swift 3.1 started to be supported on `8.0.3`](https://github.com/Moya/Moya/blob/master/Changelog.md#803). And I don't think we had to make changes for 3.2.

sunshinejr:
Just a friendly reminder, that we still need a volunteer to help out with this one! cc @Moya/contributors

sunshinejr:
Alright, I think we don't have to wait for this one until someone does the check. For now, we can just mention whatever we have figured out now: 4.0 is supported for Moya/Core & Moya/ReactiveSwift. Moya/RxSwift will be covered in Moya 10.0.

So, would you add the info in the `ReactiveSwift` column, @BasThomas? (the same as in the Moya/Core, >= 9.0.0). Also, I'm thinking that we don't need to remove Swift 2 row at all, maybe someone somewhere will need this info, it doesn't hurt to have it on.

pedrovereza:
@sunshinejr @BasThomas just made the update for ReactiveMoya 👍

sunshinejr:
If you are at it, can you bring back the Swift 2 row as well, @pedrovereza? 😄

pedrovereza:
Ok I think now I got it 😅

sunshinejr:
Cool, thanks, @pedrovereza & @BasThomas! Let's get this one in :)

PR #1241, opened 2017-08-25
[Tests] Add tests for creating urlRequest from updated task structure.

### LINKED ISSUE #1206: [Tests] Add tests for updated task type.
In #1147 we removed multiple lines of code from test target that was testing parameters and its encoding. Unfortunately we didn't add all the test cases for upgraded `Task` type. We should aim to fix that before 9.0.0 release.

_Note: I'm gonna assign myself to that one for now, but thing might change in the future (depends on my schedule), so better have this task in the Issues so we can track it and do not forget about it 😉._

sunshinejr:
Fixes #1206. These aren't perfect, but I think this is good enough for now. Should cover most of current behaviors if not all of them. 

**Edit:** Woah, first time above 80% code coverage in at least half a year! 🎉


Tests/EndpointSpec.swift
```diff
@@ -2,42 +2,230 @@ import Quick
...
+                    endpoint = endpoint.replacing(task: .requestCompositeParameters(bodyParameters: bodyParameters, bodyEncoding: encoding, urlParameters: urlParameters))
+                    request = endpoint.urlRequest
+                }
+
+                it("updated url") {
+                    let expectedUrl = endpoint.url + "?Harvey=Nemesis"
+                    expect(request.url?.absoluteString).to(equal(expectedUrl))
+                }
+
+                it("updated the request correcly") {
```
bart-g: correctly

Issue #1115, opened 2017-06-08
upload progress wrong


KieSun:
```swift
provider.request(Network.uploadPic(img: data), queue: nil, progress: { (response) in
                print(CGFloat(response.progress))
                
                view.progress = CGFloat(response.progress)
                
                if response.completed {
                    view.removeFromSuperview()
                }
                
            }, completion: { (result) in
```

I can upload image success. But progress is wrong.It`s always 1.0

sunshinejr:
Hey @KieSun. This is really strange. [We have tests around progress](https://github.com/Moya/Moya/blob/master/Tests/MoyaProviderSpec.swift#L700-L742), but if you found an edge case which our tests do not cover - please let us know with more detailed description. Even better if you could write a test for that. Right now, I can't really help you based on this information.

petester42:
It was a bug in previous versions. What version are you using?

phimage:
There is no test about upload progress, only download progress 

But I made an unit test on httpbin and we receive progress info
https://github.com/phimage/Moya/commit/765113a198d3058235a2197b9c5f9757cdc72aa5

BasThomas:
Hey, thanks @phimage! Can you open a PR so we can add that test? That'd be awesome!

phimage:
Yes I will do it
I have 3 commits, maybe 3 or 2 PR to do, and all about "progress", so potentially a lot of conflicts if I do not make it one after another.
First one is #1232, I must convert it for 9.0.0-dev

PR #1242, opened 2017-08-26
[Tests] Added a test for progress callback when doing upload task

phimage:
a simple test for upload using httpbin.org website

see #1115

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,10 +6,9 @@
 - **Breaking Change** `TargetTypes` no longer receive the `Authorization: Bearer <token>` header by default when using `AccessTokenPlugin`.
 - Added Swift 4.0 support.
 - Added all the `filter`/`map` operators that were available for `Observable<Response>` to `Single<Response>` as well.
-- Added `AuthorizationType` to `AccessTokenAuthorizable` representing request headers of `.none`, `.basic`, and `.bearer`. 
+- Added `AuthorizationType` to `AccessTokenAuthorizable` representing request headers of `.none`, `.basic`, and `.bearer`.
 - Added tests for `Single<Response>` operators.
 - Added `Progress` object into the response when calling progress callback on completion.
-- Added a test for progress callback when doing upload task.
 - Fixed a bug where you weren't notified on progress callback for data request.
 
 # 9.0.0-alpha.1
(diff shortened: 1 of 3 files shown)


Changelog.md
```diff
@@ -9,6 +9,7 @@
 - Added `AuthorizationType` to `AccessTokenAuthorizable` representing request headers of `.none`, `.basic`, and `.bearer`. 
 - Added tests for `Single<Response>` operators.
 - Added `Progress` object into the response when calling progress callback on completion.
+- Added a test for progress callback when doing upload task.
```
sunshinejr: Yeah, I think we can skip changelog entry here.

Tests/MoyaProviderSpec.swift
```diff
@@ -771,6 +771,48 @@ class MoyaProviderSpec: QuickSpec {
...
 
+        describe("a provider with upload progress tracking") {
+            var provider: MoyaProvider<HTTPBin>!
+            beforeEach {
+                provider = MoyaProvider<HTTPBin>()
+            }
+
+            it("tracks progress of request") {
+
+                let url = Bundle(for: MoyaProviderSpec.self).urlForImageResource("testImage.png")!
```
sunshinejr: Hmm, CI is failing on this one `Bundle.urlForImageResource("testImage.png")!`, could you check if it compiles for you locally?

SD10: I believe this is failing because I don't think`urlForImageResource` is available on iOS.

phimage: yes I test with first target only, mac os, my bad
fixed

Changelog.md
phimage: Line removed
I just let [atom](https://atom.io/) remove the useless space in changelog (two lines)

sunshinejr:
Thanks for the fixes, @phimage! Would you be up for cleaning the swiftlint warnings as well? Would be awesome.

phimage:
Any reason to not add swiftlint in project build phrase?

sunshinejr:
@phimage I think we just didn't think about it, but one thing to keep in mind is that Carthage needs special settings for that phase so that lint does work only for our development phase and not in users projects. For reference, you can look at [Quick#693](https://github.com/Quick/Quick/issues/693), [Quick#694](https://github.com/Quick/Quick/pull/694) and [Nimble#409](https://github.com/Quick/Nimble/pull/409)

If you want to take a stab at it, go ahead :)

PR #1243, opened 2017-08-26
[Tests] Add unit test about uploading multipart data

phimage:
I look at coverage
I do not see test on uploading multipart data, so I make one like I make the others
I do not know if there is a better place for this

I upload file and data to httpbin.org
And check the JSON data returned by the server.

### DIFF
--- a/Tests/MoyaProviderSpec.swift
+++ b/Tests/MoyaProviderSpec.swift
@@ -835,32 +835,8 @@ class MoyaProviderSpec: QuickSpec {
                     }
 
                     let progressCompletionClosure: Completion = { (result) in
-                        switch result {
-                        case .failure(let err):
+                        if case .failure(let err) = result {
                             error = err
-                        case .success(let response):
-                            do {
-                                let json = try response.mapJSON()
-                                if let json = json as? [String: Any] {
-                                    if let files = json["files"] as? [String: Any] {
-                                        expect(files["file"] ).toNot(beNil())
-                                    } else {
-                                        fail("No file received \(json)")
-                                    }
-                                    if let form = json["form"] as? [String: Any] {
-                                        expect(form["data"] ).toNot(beNil())
-                                        expect(form["data"] as? String) == string
-                                    } else {
-                                        fail("No form data received \(json)")
-                                    }
-                                } else {
-                                    fail("Wrong json format received \(json)")
-                                }
-                            } catch let moyaError as MoyaError {
-                                error = moyaError
-                            } catch {
-                                fail("Error must of type \(MoyaError.self): \(error):")
-                            }
                         }
                         done()
                     }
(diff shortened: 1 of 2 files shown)


phimage:
I see one think about multipart data
`MoyaProvider.performRequest` could do execute an `fatalError`

If I provide an empty array of multipart data. A test maybe could be done

Tests/MoyaProviderSpec.swift
```diff
@@ -811,6 +811,70 @@ class MoyaProviderSpec: QuickSpec {
...
+                        progressValues.append(progress.progress)
+                        completedValues.append(progress.completed)
+                    }
+
+                    let progressCompletionClosure: Completion = { (result) in
+                        switch result {
+                        case .failure(let err):
+                            error = err
+                        case .success(let response):
+                            do {
```
sunshinejr: Huh, this one I'm having trouble with (kinda a bit too complicated for me). Do you think you could simplify this one a little bit?

phimage: I can remove all code about "progress" but the test I think must be elsewhere
Just a test (or many tests in one séparantes file) about doing an upload with multipart data. 

I see anything else to simplify.

phimage: I can remove all code about "progress" but the test I think must be elsewhere
Just a test (or many tests in one séparantes file) about doing an upload with multipart data. 

I see anything else to simplify.

sunshinejr: Oh, I think I understand now. We are checking two things (correct me if I'm wrong):
1. Check if the progress is tracked correctly.
2. Check if the response is correct as well.

I think that we don't really have to test the response here (2.), as this is just probably sample data test and if it's returned correctly in the stubbed response.  If there is no such test yet (not sure now, would have to check), we could introduce it, but separately. 

What do you think?

phimage: Httpbin.org return a beautiful JSON output with all the form data that I post.

So I check that all multipart data I post, are in the JSON response.
I check that the content uploaded is exact ie. same string, have a file

Yes, separated tests could be done.

sunshinejr: Hmm, we should still split it, because in this test we check only for progress of the upload. In the next one we could check if the upload was correct (checking the JSON). Also we should somehow simplify this if/let hell.

sunshinejr:
Hey @phimage, just checking out if everything's fine. Let me know if you have time to split this one into two tests and simplify logic a little bit. If not we can just revisit it in the future, no problem with that 👍

phimage:
@sunshinejr 
I simplify the test by removing all the check on returned data by httpbin

I think another PR could be done later with some tests,
which upload some objects and check the result provided by httpbin

Issue #1247, opened 2017-08-30
Moya 9 Beta breaks parameters with all requests


Legoless:
Using Moya 9.0.0 Beta 1 version and since that, Moya will ignore and discard all parameters built in `TargetType` `task`. This is a breaking change, requiring to revert back to version 8.x.x, as it is useless for any API that requires parameters:

```
enum User {
    case register([String : Any])
    case ping
}

extension User : TargetType {
    var task: Task {
        switch self {
            case .register(let parameters):
                // All parameters here are ignored.
                return .requestParameters(parameters: parameters, encoding: JSONEncoding.default)
            default:
                return .requestPlain
        }
}
```

What is the solution for this? Exploring `MoyaProvider+Internal` shows that the only time Moya will call to `Task` type is at line 96:

```
switch target.task {
    case .requestPlain, .requestData, .requestParameters, .requestCompositeData, .requestCompositeParameters:
```

Since this task is a Swift Enum returning parameters, here they are completely ignored. I was unable to find any other code that would call to `.requestParameters` task type.

How to solve this?

SD10:
@Legoless The parameters get applied inside of the provider's `requestClosure`. 
It is set to [`MoyaProvider.defaultRequestMapping`](https://github.com/Moya/Moya/blob/master/Sources/Moya/MoyaProvider%2BDefaults.swift) by default. 
It does this by calling the [`urlRequest`](https://github.com/Moya/Moya/blob/master/Sources/Moya/Endpoint.swift) property on `Endpoint`.

Legoless:
I created the `Endpoint` manually, and apparently with old API, this was not required, but with new, default will always be `.requestPlain`. Might be a good idea to remove default parameter, since I just wasted 3 hours figuring this out, why it doesn't work. Since task is still called again from the internal code, it can quickly be confusing where this is applied. Issue can be closed. :) Sorry.

sunshinejr:
@Legoless this is a great idea. In fact, I think we should remove all default parameters from `Endpoint.init` as all of them might get confusing. Would you be up for a PR maybe? :)

It will be a breaking change, but I think this is for a good cause.

SD10:
I agree, let's remove them so no one else faces this issue. @Legoless Thanks for the feedback and sorry for the confusion!

SD10:
[...]

PR #1248, opened 2017-08-31
Propagate parameter encoding error for URLRequest mapping

SD10:
### Changes:

**Adds:**
`MoyaError.parameterEncoding(Swift.Error) case`

**Refactors:**
`urlRequest` -> `urlRequest()` (now throwing)

### Summary of Pull Request:

As discussed in #1247 I saw that it would be really easy to pass on the errors thrown when attempting to encode `URLRequest` parameters.

All that had to be done is convert `Endpoint`'s `urlRequest` property to a throwing function and pass the error to the `.failure` closure in `MoyaProvider.defaultRequestMapping`.

Personally, I'm not too fond of having to call `try` now when accessing an `Endpoint`'s `urlRequest`. Therefore, I mostly consider this "research" unless any of you think there's value in it.

### DIFF
--- a/Sources/Moya/Endpoint.swift
+++ b/Sources/Moya/Endpoint.swift
@@ -60,9 +60,9 @@ open class Endpoint<Target> {
     }
 }
 
-/// Extension for converting an `Endpoint` into an optional `URLRequest`.
+/// Extension for converting an `Endpoint` into a `URLRequest`.
 extension Endpoint {
-    /// Returns the `Endpoint` converted to a `URLRequest` if valid. Returns `nil` otherwise.
+    /// Returns the `Endpoint` converted to a `URLRequest` if valid. Throws an error otherwise.
     public func urlRequest() throws -> URLRequest {
         guard let requestURL = Foundation.URL(string: url) else {
             throw MoyaError.requestMapping(url)
--- a/Sources/Moya/Endpoint.swift
+++ b/Sources/Moya/Endpoint.swift
@@ -63,8 +63,10 @@ open class Endpoint<Target> {
 /// Extension for converting an `Endpoint` into an optional `URLRequest`.
 extension Endpoint {
     /// Returns the `Endpoint` converted to a `URLRequest` if valid. Returns `nil` otherwise.
-    public var urlRequest: URLRequest? {
-        guard let requestURL = Foundation.URL(string: url) else { return nil }
+    public func urlRequest() throws -> URLRequest {
+        guard let requestURL = Foundation.URL(string: url) else {
+            throw MoyaError.requestMapping(url)
+        }
 
         var request = URLRequest(url: requestURL)
         request.httpMethod = method.rawValue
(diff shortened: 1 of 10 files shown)


sunshinejr:
This is really cool, but I feel like this could be something we think about for version 10.0. We need to release 9.0 in a few. What do you think about it, @SD10?

SD10:
@sunshinejr I have no problem delaying this until v10.0 or even not going with this solution at all

sunshinejr:
Damn, GitHub did let me remove the 9.0.0-dev branch while PRs were targeted to it still. Reopening!

sunshinejr:
Hey @SD10, would you be up for finishing that one? Conflicts + Docs + Changelog is there to finish I believe? Other than that I think this looks good!

SD10:
@sunshinejr Yes, I will finish this up. I need about 2 days though

sunshinejr:
Great! Take your time 👍

SD10:
That's what I get for trying to resolve a merge conflict in nano 😃

sunshinejr:
@SD10 would you rebase & fix the conflicts? I've merged few other PRs in the meantime.

SD10:
@sunshinejr Sure thing. Should be good to go if resolving the conflicts went well and CI is all green. Unless anyone else has any input/review. Giving it another look over myself now

SD10:
Yes, I'd really appreciate it if another @Moya/contributors gave their input. The goal of this PR is to expose more information about the failure of parameter encoding for a request. 

I don't want to hear a ton of complaints about having to use `try` syntax when accessing the `urlRequest` for an `Endpoint` if this were to be merged.

Sources/Moya/Endpoint.swift
```diff
@@ -77,32 +79,35 @@ extension Endpoint {
...
     public static func == <T>(lhs: Endpoint<T>, rhs: Endpoint<T>) -> Bool {
-        if lhs.urlRequest != nil, rhs.urlRequest == nil { return false }
-        if lhs.urlRequest == nil, rhs.urlRequest != nil { return false }
-        if lhs.urlRequest == nil, rhs.urlRequest == nil { return lhs.hashValue == rhs.hashValue }
-        return (lhs.urlRequest == rhs.urlRequest)
+        let lhsRequest = try? lhs.urlRequest()
+        let rhsRequest = try? rhs.urlRequest()
+        if lhsRequest != nil, rhsRequest == nil { return false }
+        if lhsRequest == nil, rhsRequest != nil { return false }
+        if lhsRequest == nil, rhsRequest == nil { return lhs.hashValue == rhs.hashValue }
```
ashfurrow: This is totally fine, though I will point out that if two different Endpoints both fail to produce a `urlRequest`, then they would be "equal". I'm sure that's fine, especially since it's consistent with `hashValue`. The only consequences would be using an `Endpoint` as key in a dictionary, or in a `Set`. I think it's worth a comment, but otherwise this is totally 👍

Tests/EndpointSpec.swift
```diff
@@ -72,8 +72,8 @@ final class EndpointSpec: QuickSpec {
 
         it("returns a nil urlRequest for an invalid URL") {
             let badEndpoint = Endpoint<Empty>(url: "some invalid URL", sampleResponseClosure: { .networkResponse(200, Data()) }, method: .get, task: .requestPlain, httpHeaderFields: nil)
-
-            expect(badEndpoint.urlRequest).to(beNil())
+            let urlRequest = try? badEndpoint.urlRequest()
```
ashfurrow: This is good – can we also add a test that makes sure the error thrown is correct?

Sources/Moya/Endpoint.swift
SD10: If two different Endpoints both fail to produce a `urlRequest` it will compare both endpoints using the hashValue of the `url: String` property. 

So if two different Endpoints both fail to produce a `urlRequest` and they have the same `String` for the `url` -- they will be considered equal. I'll add a comment noting this behavior.

Tests/EndpointSpec.swift
SD10: Great idea 👍 Thanks for the input

SD10:
As per @ashfurrow 's feedback I added tests to make sure `urlRequest` is throwing the correct error.
 
I changed the implementation to throw the `.parameterEncoding` error directly from `urlRequest()` instead of wrapping it up in `defaultEndpointMapping`. This is because I was receiving an `Alamofire` error instead of a `MoyaError` through `urlRequest`. Nice how writing a test caught this 😉

Truthfully, it was extremely hard to write a test that caught an error -- it even made me question the value of this change. 

This error will only ever be triggered when `JSONSerialization` fails to serialize the parameters. I went hunting in `core-libs-foundation` to see what errors it throws -- but the majority of them aren't caught gracefully and come back as an `NSInvalidArgumentException`.

Tests/EndpointSpec.swift
```diff
@@ -221,10 +221,46 @@ final class EndpointSpec: QuickSpec {
...
+                    expect(recievedError).toNot(beNil())
+                    expect(recievedError).to(beOfSameErrorType(expectedError))
+                }
+            }
+
+            context("when parameter encoding is unsuccessful") {
+                it("throws a .parameterEncoding error") {
+
+                    // Non-serializable type to cause serialization error
+                    class InvalidParameter {}
```
ashfurrow: Nice.

ashfurrow:
Awesome, great work! Let's leave it a few days in case anyone else has input. 

@Moya/core-team I wonder if we should target a new `moya-10` branch with the PR, to keep master for any non-breaking fixes we can release as 9.0.x patches. Anyone have thoughts on a branching strategy?

pedrovereza:
@ashfurrow This PR is already targeting branch `10.0.0-dev` 😉

ashfurrow:
Oops 🙈

sunshinejr:
I was rebasing 10.0.0-dev again as we had documentation changes, so I rebased this one as well. Sorry for trouble.

ashfurrow:
Cool, looks good to merge on 🍏

sunshinejr:
Great! Thanks for all your hard work on that one, @SD10 😉

PR #1249, opened 2017-08-31
[WIP] Remove default arguments for Endpoint initializer

SD10:
As discussed in #1247, the fact that `Endpoint` has default values of `.get` & `.requestPlain` can be a point of confusion when users are creating their own Endpoints.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,4 +1,5 @@
 # Next
+- **Breaking Change** Remove all default arguments from `Endpoint` initializer.
 
 # 9.0.0-beta.1
 - **Breaking Change** Replaced `parameters` & `parameterEncoding` in `TargetType` with extended `Task` cases.
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,4 +1,6 @@
 # Next
+
+# 9.0.0-beta.1
 - **Breaking Change** Replaced `parameters` & `parameterEncoding` in `TargetType` with extended `Task` cases.
 - **Breaking Change** Flattened `UploadType` and `DownloadType` into `Task` cases.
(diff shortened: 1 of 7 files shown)


sunshinejr:
This is cool, thanks @SD10! Did you also check Demo project if it needs updating anywhere?

SD10:
@sunshinejr I meant to check the docs and forgot. I will shift this to WIP and take care of it. About the demo app -- I'm guilty of not even thinking about that. I will do this too 😅

sunshinejr:
No worries! We have 2 projects right now that we have to check whenever developing stuff so it happens often, to me as well. We should take care of consolidating them (#891) in the near future so it is easier for us.

sunshinejr:
Hmm, I thought about it for a while and I don't think there is time for another breaking change. What we can do is to remove default parameter from the `task` property and release it as a patch. This way we won't have a breaking change. Just need to confirm docs/tests/demo, but I believe not much (if any) will have to be changed. We can think about the the rest parameters in 10.0. 

What do you think, @SD10?

SD10:
@sunshinejr Sorry I didn't get a chance to follow up on this. I will do this in a couple hours from now. Why don't you want to do this breaking change? Am I holding things up?

sunshinejr:
First of all we would have to do some checks, then break semver and in the end probably test it a bit more. Having removed only task default property probably will be an easier one for 9.0.0 and we could follow this up in 10.0. I think we just need to release 9.0 in a few.

SD10:
@sunshinejr No problem, good thoughts. It's good to know when to draw the line. I noticed there are no changes in the demo app or the docs. We weren't using the default values in the docs. I opened up a fresh PR so I'm closing this one.

PR #1255, opened 2017-09-02
[Documentation][Examples] Add MultipartUpload.md

### LINKED ISSUE #1219: Add an example of multipart upload with parameters.
There were a few issues regarding multipart upload with parameters and how to do it (e.g. #1209). I feel like this is a good time to add an example to our [docs/Examples](https://github.com/Moya/Moya/tree/master/docs) directory, especially having a new release in place. What I'm thinking about is a document with one (or more) examples, but the one we need for sure is an image upload with additional parameters.

Let me know if something is not clear 👍

sunshinejr:
Fixes #1219.

### DIFF
--- a/docs/Examples/MultipartUpload.md
+++ b/docs/Examples/MultipartUpload.md
@@ -1,0 +1,1 @@
+# Multipart Upload
--- a/docs/Examples/MultipartUpload.md
+++ b/docs/Examples/MultipartUpload.md
@@ -1,1 +1,38 @@
 # Multipart upload
+
+Suppose we want to upload a GIF with additional parameters in one request. How to approach this problem? It depends if our parameters should be in body of the request or in the URL.
+
+## Setup
+
+Let's say we have our `TargetType` already setup - it's called `MyService`:
+
+```swift
+public enum MyService {
+    case uploadGif(Data, description: String)
+}
+```
+
+Here, our additional parameter is `description`, which is a `String`.
+
+## Parameters in body
+
+When we want to add parameters to the body of our request, we need to append them as a `MultipartFormData` in the `task` property:
+
+```swift
+extension MyService: TargetType {
+//...
+    public var task: Task {
+        switch self {
+        case let .uploadGif(data, description):
+            let gifData = MultipartFormData(provider: .data(data), name: "file", fileName: "gif.gif", mimeType: "image/gif")
+            let descriptionData = MultipartFormData(provider: .data(description.data(using: .utf8)!), name: "description")
+            let multipartData = [gifData, descriptionData]
+
+            return uploadMultipart(multipartData)
+        default:
+            return .requestPlain
+        }
+    }
+//...
+}
+```
--- a/docs/Examples/MultipartUpload.md
+++ b/docs/Examples/MultipartUpload.md
@@ -1,10 +1,10 @@
 # Multipart upload
 
-Suppose we want to upload a GIF with additional parameters in one request. How to approach this problem? It depends if our parameters should be in body of the request or in the URL.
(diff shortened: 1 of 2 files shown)


sunshinejr:
Could someone review it today, @Moya/contributors? Would be awesome that we'd have it on the release as we've changed the existing behavior a bit.

docs/Examples/MultipartUpload.md
```diff
@@ -0,0 +1,57 @@
+# Multipart upload
+
+Suppose we want to upload a GIF with additional parameters in one request. How to approach this problem? It depends if our parameters should be in body of the request or in the URL.
```
freak4pc: > How do we approach this problem? It depends if the parameters should be part of the body, or part of the query string

You could change to `part of the URL`, but query string is actually the more correct term IMO

docs/Examples/MultipartUpload.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Setup
+
+Let's say we have our `TargetType` already setup - it's called `MyService`:
```
freak4pc: Should be `set up` , and perhaps would be nicer to just do 

> Let's say we have a `MyService` service conforming to `TargetType`:

freak4pc:
Reviewing now as well, have a couple of notes

docs/Examples/MultipartUpload.md
```diff
@@ -0,0 +1,57 @@
...
+public enum MyService {
+    case uploadGif(Data, description: String)
+}
+```
+
+Here, our additional parameter is `description`, which is a `String`.
+
+## Parameters in body
+
+When we want to add parameters to the body of our request, we need to append them as a `MultipartFormData` in the `task` property:
```
freak4pc: I would change to something like
> When we want to perform a multipart request with a request body, we have to override the `task` property,  create a `MultiPartFormatData` for each of our parts and them return a ` .uploadMultiPart` task with the appended parts.

docs/Examples/MultipartUpload.md
sunshinejr: You are correct that `query string` is a better term, but it might not be more understandable vs just using URL term. But I agree with the rest of the changes here 👍

docs/Examples/MultipartUpload.md
SD10: We probably shouldn't use the word `override`

docs/Examples/MultipartUpload.md
```diff
@@ -16,7 +16,7 @@ Here, our additional parameter is `description`, which is a `String`.
 
 ## Parameters in body
 
-When we want to add parameters to the body of our request, we need to append them as a `MultipartFormData` in the `task` property:
+When we want to perform multipart upload request with additional parameters in request body, we have to create a `MultiPartFormatData` for each of our parts and them return a `.uploadMultipart(_:)` in the `task` property:
```
freak4pc: nit: 
perform *a* multipart
parameters in *the* request body
them => then

docs/Examples/MultipartUpload.md
sunshinejr: Exactly, but few good improvements there so I've added them as well!

sunshinejr:
I've updated the document with improvements you suggested, @freak4pc @SD10. Let me know if there is still something we can improve or merge it instead 😄

docs/Examples/MultipartUpload.md
sunshinejr: Done, good catches! 👍

freak4pc:
I'm all good. Fire at will! 💣 !

sunshinejr:
Thank you so much guys, it was really a quick one! 👏

PR #1256, opened 2017-09-02
[Documentation] Update docs for 9.0.0 release.

### LINKED ISSUE #1134: [Documentation] Fix RequestClosure usage in the docs.
We forgot to change the documentation around new `RequestClosure` in some places. We need to double-check where we are using it, but for sure we have to update it in [Authentication.md](https://github.com/Moya/Moya/blob/master/docs/Authentication.md#oauth). Transition is rather simple, you can take a look at our tests [here](https://github.com/Moya/Moya/blob/aa33a0466d7333516691b3e5212c5acb41990928/Tests/MoyaProviderSpec.swift#L262-L270) to see how it is used now. [This](https://github.com/Moya/Moya/blob/master/docs/Endpoints.md#request-mapping) can be useful in case of explaining how the closure works nowadays.

Also, I think this might be good idea to introduce #starter-task (stolen from IGListKit 😄, thanks guys!), so I've added the label here.

### LINKED ISSUE #1205: Check documentation for parameters/encoding/task migration.
We might need to check if our documentation is up-to-date (in branch 9.0.0-dev) with the changes to parameters/parameterEncoding/task from #1147. 

If someone is up for this task, please assign yourself so we won't dup our work. Cheers! 😉

sunshinejr:
So originally I wanted to fix #1205, but I've just fixed whatever I saw wrong with documentation. Worth noting is that I removed two examples: 
- `ArrayAsContainer` (which was using old ParameterEncoding from Alamofire 3.*)
- `SubclassingProvider` (which was suggesting to use subclassing on RxMoyaProvider, which we deprecated)

I will create issues for both in a few.

**Edit:** Now also fixes #1134.


sunshinejr:
Thanks guys ❤️

Issue #1240, opened 2017-08-25
Update SPM.


BasThomas:
The [`Package.swift`](https://github.com/Moya/Moya/blob/9.0.0-dev/Package.swift) will need to update its [API](https://github.com/apple/swift-package-manager/blob/swift-4.0-branch/Documentation/PackageDescriptionV4.md). Changes to the Package should also be reflected in the Readme.

sunshinejr:
Hey @BasThomas. Not sure, but does migrating to new SPM API means that we will drop Xcode 8/Swift 3.1 support? If yes, we probably need to postpone it to 10.0 release.

SD10:
@sunshinejr I was just looking into this because I had to familiarize myself with the new API for danger-swift. It looks like it should be backwards compatible according to [this](https://swift.org/blog/swift-package-manager-manifest-api-redesign/).

sunshinejr:
Hmm, it seems like we would have to specify `swift-tools-version:<specifier>`, which probably will be 4.0, and someone who is working on Xcode 8 could be missing Swift 4.0 tools. Also, I wasn't sure what is this `swift-tools-version` parameters, but swift.org got me covered:

> // The swift-tools-version declares the minimum version of Swift required to build this package.

josefdolezal:
Is this issue back in the game with Moya v10?

sunshinejr:
Yup, I think it its, @josefdolezal! PR should be up for grabs 😉

sunshinejr:
That's actually fixed already, so I'm closing this one.

PR #1266, opened 2017-09-05
Updates SPM manifest to V4 API

josefdolezal:
This PR implements new SPM manifest API for swift 4 as described in #1240. The changes are also reflected in the Readme.


josefdolezal:
Hey @pedrovereza, thank you for the feedback! I tried my best to match the dependencies version in package manifest changes. However, it looks like I missed the Result framework.

I can change the version back to 3.0 (same as CocoaPods) now and then update all PMs in separate PR, what do you think? The 3.2 version is however required by ReactiveSwift.

Package.swift
```diff
@@ -1,28 +1,35 @@
...
     name: "Moya",
+    products: [
+        .library(name: "Moya", targets: ["Moya"]),
+        .library(name: "ReactiveMoya", targets: ["ReactiveMoya"]),
+        .library(name: "RxMoya", targets: ["RxMoya"]),
+    ],
+    dependencies: [
+        .package(url: "https://github.com/Alamofire/Alamofire.git", .upToNextMajor(from: "4.1.0")),
+        .package(url: "https://github.com/antitypical/Result.git", .upToNextMajor(from: "3.2.0")),
+        .package(url: "https://github.com/ReactiveCocoa/ReactiveSwift.git", .upToNextMajor(from: "2.0.0")),
```
SD10: Should these be `.exact` versions as well?

josefdolezal: I guess we can change it to `.exact`, however it would kill the possibility of package updating using `swift package update`, right? The exact versions are defined in `Package.resolved`, which is equivalent to `Carthage.resolved`.

Package.swift
```diff
@@ -1,28 +1,35 @@
...
-        .Package(url: "https://github.com/antitypical/Result.git", majorVersion: 3)
+            dependencies: [
+                "Moya",
+                "RxSwift"]),
     ],
-    exclude: [
-        "Tests",
-        "Sources/Supporting Files"
-    ]
+    swiftLanguageVersions: [4]
```
sunshinejr: Should we also add Swift 3?

Readme.md
```diff
@@ -67,23 +67,37 @@ you should use for your Swift version.
...
+    ],
+    targets: [
+        .target(
+            name: "MyPackage",
+            dependencies: ["Moya", "ReactiveMoya"])
     ]
 )
 ```
 
+Note that as of Moya 10, SPM only works with Swift 4 and greater.
```
sunshinejr: I know that you need to have Swift 4 on your machine, but can you use Swift 3.2 anyways? 🤔

sunshinejr:
Hey @josefdolezal, just checking out if everything's fine. Let me know if there is something I can do to help with this one.

josefdolezal:
Hey @sunshinejr! Sorry for the delay, I was out for a few last days. I updated the package manifest to use same dependencies versions as CocoaPods and updated the Readme for SPM.

I also tried to find more informations about `swiftLanguageVersions`, however from the docs, I am not sure how exactly it works. If we require swift 4 toolchain, can we explicitly use swift 3 to build the library?

sunshinejr:
@josefdolezal I'm not sure either, unfortuntely. We will have to investigate that one.

pedrovereza:
@sunshinejr @josefdolezal Based on [the documentation](https://github.com/apple/swift-package-manager/blob/e9ff9a89ce1a36c29ed3b550151ebc2af15c6a04/Documentation/PackageDescriptionV4.md#swiftlanguageversions) and on [Swift 4 Compatibility](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/Compatibility.html) it seems safe to specify `swiftLanguageVersions = [3,4]`

> If we require swift 4 toolchain, can we explicitly use swift 3 to build the library?

From the docs I just linked to, it seems like we can:

> For e.g. if Swift language version is set to [3], both Swift 3 and 4 compilers will select '3'

But I may have missed something 😅

josefdolezal:
Ok seems reasonable! I have updated the manifest with support for Swift 3. Thank you all for your help with this one! 🎉

Issue #1273, opened 2017-09-08
How to use Moya with access token as url parameter?


dotprox:
Hello

I have the following format for a url to access data: 
```
https://api.sitename.com/method.name/?someparameter=somedata&access_token=token
```

How to use Moya with this token type?

BasThomas:
You can just add the `access_token` in the parameters.

dotprox:
Hello @BasThomas, thank you for your answer. Is it good idea pass token each service method? I mean, here is authentication plugin, can we modernise it for different token types?

BasThomas:
With the access token being a parameter, I think the best solution would be to add it via `parameters`; in this case the access token is not really different from any other parameter.

But it *is* a parameter that needs to be added to every request, so you might want to look into adding it to the Authentication Plugin, yes. Do you want to take a stab at implementing that?

dotprox:
@BasThomas thanks for the suggestion, I will try! 🙂

sunshinejr:
I'm closing this one in favor of #1283 as we discuss the issue further there.

PR #1279, opened 2017-09-12
Update AccessTokenPlugin to be able to pass token as url parameter

dotprox:
Some sites let you access to their api via access token which is passed as a parameter in the URL path. This commit allows to use AccessTokenPlugin also in this task. #1273

### DIFF
--- a/Sources/Moya/Plugins/AccessTokenPlugin.swift
+++ b/Sources/Moya/Plugins/AccessTokenPlugin.swift
@@ -12,7 +12,7 @@ public protocol AccessTokenAuthorizable {
 
 // MARK: - AuthorizationType
 
-/// An enum representing the header and url parameter to use with an `AccessTokenPlugin`
+/// An enum representing the header or url parameter to use with an `AccessTokenPlugin`
 public enum AuthorizationType: String {
     case none
     case basic = "Basic"
@@ -68,16 +68,19 @@ public struct AccessTokenPlugin: PluginType {
             request.addValue(authValue, forHTTPHeaderField: "Authorization")
           
         case .urlParameter:
-          var existingComponents = URLComponents(url: request.url!, resolvingAgainstBaseURL: true)
+          guard let requestUrl = request.url,
+            var existingComponents = URLComponents(url: requestUrl, resolvingAgainstBaseURL: true)
+            else { return request }
+          
           let tokenQueryItem = URLQueryItem(name: authorizationType.rawValue, value: tokenClosure())
           var updatedQueryItems = [tokenQueryItem]
           
-          if let queryItems = existingComponents?.queryItems {
-            queryItems.forEach { updatedQueryItems.append($0) }
+          if let currentQueryItems = existingComponents.queryItems {
+            updatedQueryItems.append(contentsOf: currentQueryItems)
           }
           
-          existingComponents?.queryItems = updatedQueryItems
-          request.url = existingComponents?.url
+          existingComponents.queryItems = updatedQueryItems
+          request.url = existingComponents.url
           
         case .none:
             break
--- a/Sources/Moya/Plugins/AccessTokenPlugin.swift
+++ b/Sources/Moya/Plugins/AccessTokenPlugin.swift
@@ -6,47 +6,49 @@ import Result
 /// A protocol for controlling the behavior of `AccessTokenPlugin`.
 public protocol AccessTokenAuthorizable {
 
-    /// Represents the authorization header to use for requests.
+    /// Represents the authorization header or url parameter to use for requests.
     var authorizationType: AuthorizationType { get }
 }
 
 // MARK: - AuthorizationType
 
-/// An enum representing the header to use with an `AccessTokenPlugin`
+/// An enum representing the header and url parameter to use with an `AccessTokenPlugin`
 public enum AuthorizationType: String {
     case none
     case basic = "Basic"
     case bearer = "Bearer"
+    case urlParameter = "access_token"
 }
 
 // MARK: - AccessTokenPlugin
 
 /**
- A plugin for adding basic or bearer-type authorization headers to requests. Example:
+ A plugin for adding basic or bearer-type authorization headers, or url parameter to requests. Example:
 
  ```
  Authorization: Bearer <token>
  Authorization: Basic <token>
+ access_token=<token>
  ```
 
 */
 public struct AccessTokenPlugin: PluginType {
 
-    /// A closure returning the access token to be applied in the header.
+    /// A closure returning the access token to be applied in the header or url parameter.
     public let tokenClosure: () -> String
 
     /**
      Initialize a new `AccessTokenPlugin`.
 
      - parameters:
-       - tokenClosure: A closure returning the token to be applied in the pattern `Authorization: <AuthorizationType> <token>`
+       - tokenClosure: A closure returning the token to be applied in the pattern `Authorization: <AuthorizationType> <token>` or `access_token=<token>`
     */
     public init(tokenClosure: @escaping @autoclosure () -> String) {
         self.tokenClosure = tokenClosure
     }
 
     /**
-     Prepare a request by adding an authorization header if necessary.
+     Prepare a request by adding an authorization header or url parameter if necessary.
 
      - parameters:
        - request: The request to modify.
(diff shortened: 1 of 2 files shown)


Sources/Moya/Plugins/AccessTokenPlugin.swift
```diff
@@ -64,6 +66,19 @@ public struct AccessTokenPlugin: PluginType {
         case .basic, .bearer:
             let authValue = authorizationType.rawValue + " " + tokenClosure()
             request.addValue(authValue, forHTTPHeaderField: "Authorization")
+          
+        case .urlParameter:
+          var existingComponents = URLComponents(url: request.url!, resolvingAgainstBaseURL: true)
```
BasThomas: Any reason for the force unwrap here? We might want to crash more gracefully (or skip) if the url is not valid.

dotprox: Yeh, I agree. Would be better if I will use optional binding for url, and then continue other logic?

SD10:
@dotprox Thank you for your contribution 🎉 I'm just curious could this not be done with a regular target type by adding `token_url` as a URL parameter?

BasThomas:
@SD10 see the discussion [here](https://github.com/Moya/Moya/issues/1273#issuecomment-328272433).

Sources/Moya/Plugins/AccessTokenPlugin.swift
```diff
@@ -6,47 +6,49 @@ import Result
...
 
-    /// Represents the authorization header to use for requests.
+    /// Represents the authorization header or url parameter to use for requests.
     var authorizationType: AuthorizationType { get }
 }
 
 // MARK: - AuthorizationType
 
-/// An enum representing the header to use with an `AccessTokenPlugin`
+/// An enum representing the header and url parameter to use with an `AccessTokenPlugin`
```
SD10: Very minor detail but "and url parameter" -> "or url parameter"

dotprox: Hello @SD10 , thank you for help. I will make some fixes at evening 👍

SD10:
@BasThomas Thank you!
@dotprox This will also require us to update the documentation around `AccessTokenPlugin` and its tests. If you need some help with that I'd be happy to assist.

Sources/Moya/Plugins/AccessTokenPlugin.swift
```diff
@@ -6,47 +6,49 @@ import Result
...
 public enum AuthorizationType: String {
     case none
```
AndrewSB: Unrelated to this PR: Why is there a `none` case? It doesn't seem to be doing anything. Being a plugin, the `none` case should just be modeled by not using the plugin. Am I missing something?

SD10: @AndrewSB Since `AccessTokenAuthorizable` is implemented by adding its conformance to an existing `TargetType`, there could be endpoints within that target that don't want auth

dotprox:
@SD10 @BasThomas I edited some logic. Please review the code when you have time, thank you.

Sources/Moya/Plugins/AccessTokenPlugin.swift
AndrewSB: Ahh, I see. Thanks for clearing that up 😄

Sources/Moya/Plugins/AccessTokenPlugin.swift
```diff
@@ -6,47 +6,49 @@ import Result
...
 
 // MARK: - AuthorizationType
 
-/// An enum representing the header to use with an `AccessTokenPlugin`
+/// An enum representing the header or url parameter to use with an `AccessTokenPlugin`
 public enum AuthorizationType: String {
     case none
     case basic = "Basic"
     case bearer = "Bearer"
+    case urlParameter = "access_token"
```
BasThomas: I am thinking this might not cover everything now. What if you have an API that uses `api_key` or `api_token` as an url parameter?

SD10: I had similar thoughts regarding how many cases this plugin would support. I was thinking about changing the enum to use associated values.
```Swift
enum AuthorizationType {
    case header(String)
    case urlParameter(String)
    case none
}
```

SD10: If you needed to use the header and urlParameters you would have to add the parameters in the TargetType still. The AccessTokenPlugin was designed around supplying header values.

BasThomas: I feel like it would make sense to support a way to add a specific parameter to every request in a provider though.

SD10: What if we had a `ParameterPlugin` or `HeaderPlugin` that would add a shared parameter or header across `TargetTypes`? Too awkward? I agree with the need for reuse.

dotprox: Hello guys, you suggested nice ideas. Can I try to implement case with argumented enum or additional plugins?

sunshinejr: I agree we could improve current plugin to allow using the token as a parameter (I had multiple occasions to use it once as a parameter and once as a header). 

How about adding the `parameter` name in initializer? We could also add `header` name if needed. Then the cases could be:
```swift

init(tokenClosure: ..., parameterName: ..., headerName: ...) { ... }

enum AuthorizationType {
    case basicInHeaders
    case bearerInHeaders
    case basicInParameters
    case bearerInParameters
    none
}
```

Just an idea, though.

SD10: @dotprox I don't know what's best to be honest and it's not for me to decide alone. I'm just trying to drive some discussion. I don't want to ask you to do A then decide B 😐 Maybe we should open an issue discussing reuse across multiple providers and see if anyone has other ideas/input

dotprox: Guys, what you think about combination of optional name for token and initialization with token key? If you need to use prefix, you can add specific name, otherwise you will use default names like `access_token` or `Authorization`.

```swift
init(tokenClosure: ..., parameterName: ... = "access_token", headerName: ... = "Authorization") { ... }

public enum AuthorizationType: String {
    case none
    case header(name: String?) // -> Authorization: <name?><token>
    case parameter(name: String?) // -> access_token=<name?><token>
}
```

inspired by @sunshinejr and @SD10 suggestions

SD10:
```Swift
public struct HeaderPlugin: PluginType {
    var headerValue: String
    var headerName: String
    var shouldApply: Bool
}
public struct ParameterPlugin: PluginType {
   var parameterValue: Any
   var parameterName: String
   var encoding: ParameterEncoding
   var shouldApply: Bool
}
```
This sort of removes the concept of auth but it is a way to share a specific header or parameter across multiple providers. @sunshinejr I like what you suggested but what feels awkward is when I don't want to deal with a `header` vs a `parameter` or vice versa. It would be nice to see this decoupled.

sunshinejr: Hmm, how would it handle having one call with `parameter` and one with `header` in one `TargetType`? Also, I agree that we could discuss it in another issue as this one will be buried fast.

SD10:
```Swift
protocol ParameterApplicable {
    var shouldApply: Bool
}
protocol HeaderApplicable {
    var shouldApply: Bool
}
```
The `TargetType` would need to conform to one or both of these then you just return `true` or `false` depending on the `case` of the target. I can open an issue a little bit later documenting what's been discussed

pedrovereza: :+1: On moving the discussion to an issue

SD10: Created an issue #1283 to summarize what has been discussed here

sunshinejr:
Hey @dotprox, thank you so much for your work here! We really appreciate the time you have given to support Moya. However, seems like we have a problem with how do we want to implement this one. @SD10 created an issue to discuss it - see #1283. We would love to get your feedback on it.

For now I will close this PR, though.

Issue #1233, opened 2017-08-22
Moya support for Swift 4 with RxSwift ?


neoD1:
I am using the following pods.  How to make RxSwift and Moya go hand in hand in Swift 4 ?

```swift
pod 'RxSwift', :git => 'https://github.com/ReactiveX/RxSwift.git', :branch => 'swift4.0'
pod 'RxCocoa', :git => 'https://github.com/ReactiveX/RxSwift.git', :branch => 'swift4.0'
pod 'Moya/RxSwift', :git => 'https://github.com/Moya/Moya.git', :branch => '9.0.0-dev'
```

[!] Unable to satisfy the following requirements:

- `RxSwift (from `https://github.com/ReactiveX/RxSwift.git`, branch `swift4.0`)` required by `Podfile`
- `RxSwift (= 4.0.0-alpha.1)` required by `Podfile.lock`
- `RxSwift (~> 4.0.0-alpha.1)` required by `RxCocoa (4.0.0-alpha.1)`
- `RxSwift (= 1.4)` required by `Moya/RxSwift (1.1.0)`

sunshinejr:
Hey @neoD1, thanks for the issue. I guess you are right - we cannot support Swift 4.0 with RxSwift and Moya 9.0. This is because we probably gonna release it before Swift 4.0 and RxSwift 4, and after the new releases we will update our extensions. 

Right now we offer Moya core working with Swift 4.0 or Moya with Swift 3.2 that should work correctly with RxSwift 3.3+. Sorry for the inconvenience! 

If you have any more questions, let me know.

aninhakrieger:
:'(

linbx08:
Wait for the day is long, hope Moya support for Swift 4 with RxSwift the truth will become soon .

sunshinejr:
Hey guys, @aninhakrieger, @linbx08, @neoD1! Just a heads-up that today we are planning to release version 9.0.0 and immediately after we will start with development branch for 10.0.0, which will include support for Swift 4.0 and reactive extensions. We will update this issue once we get something working.

linbx08:
Ah Fantastic! Thanks a lot

sunshinejr:
Hey again, @aninhakrieger, @linbx08, @neoD1! I've created a branch that is fully Swift 4.0.0-compatible! Branch name: `10.0.0-dev`. Let me know if it's working for you :)

linbx08:
Very thankful to you

neoD1:
[...]

PR #1291, opened 2017-09-17
Update Readme.md

Sepho:
Added info about Swift 4.0 support (Moya 9.0.0 and above)

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -58,6 +58,7 @@ you should use for your Swift version.
 
 | Swift version | Moya version  |
 | ------------- | ------------- |
+| 4.0           | >= 9.0.0      |
 | 3.X           | >= 8.0.0      |
 | 2.3           | 7.0.2 - 7.0.4 |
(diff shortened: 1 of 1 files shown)


pedrovereza:
Hey @Sepho, thanks for taking the time to contribute to Moya! 🎉 

While it's true that, in 9.0, Moya core components are compatible with Swift 4, our RxSwift component is not. We are still waiting on RxSwift 4.0 to be stable (see https://github.com/Moya/Moya/issues/1233#issuecomment-326912247).

If we're gonna claim that Moya 9.0 is compatible with Swift 4, we should add a note mentioning that it's doesn't include our `rx` extension.

On a side note, full support for Swift 4 should come in Moya 10 :wink:

pedrovereza:
Actually, turns out there is already a PR making the updates on the version table. I'm closing this one in favor of https://github.com/Moya/Moya/pull/1239.

Thanks again for taking the time, @Sepho 🙏

PR #1292, opened 2017-09-18
Consolidate Demos into Moya.xcodeproj

AndrewSB:
reimplements #891


AndrewSB:
Can someone test to see if the demos work? I only have Xcode 9 on my machine, so I cant currently build and run the project

sunshinejr:
Thank you so much for doing this, @AndrewSB! I will try to check that one more later, but seems like we need to update the [Demo building script](https://github.com/Moya/Moya/blob/master/Rakefile#L85) as well, as CI is failing.

AndrewSB:
@sunshinejr I tried to fix this, but I'm not sure why the Rakefile is now invalid. Here were my changes https://github.com/Moya/Moya/pull/1292/commits/c1fb4a00af632a8937deddcc58caa0861109ab21

Can you point me in the right direction?

pedrovereza:
@AndrewSB Seems like you deleted a block but forgot to remove the `end` ([here](https://github.com/Moya/Moya/blob/c1fb4a00af632a8937deddcc58caa0861109ab21/Rakefile#L170))

AndrewSB:
Thanks @pedrovereza 😄

sunshinejr:
@AndrewSB Oh, seems like you mistyped `Moya.xcworkspace` instead of `Moya.xcodeproj`. We would also need to change the build command to use project instead of workspace.

sunshinejr:
Hmm, any idea why the build is failing now? Maybe copy-frameworks from Carthage is needed?

sunshinejr:
@AndrewSB I think I resolved the failing build. First, our example GithubAPI wasn't updated to Moya 9. Second, framework search paths needed to be added. And the last one was importing `Result` & `Alamofire` in `GithubAPI.swift` as well. I don't understand why we have to import `Alamofire` when we just use our own type-aliased `ParameterEncoding`. And `Result` as well. Do we have to import these in a `CocoaPods` project? 🤔

sunshinejr:
Alright, it has been built and it's green. It should be ready for review. cc @petester42 @ashfurrow @pedrovereza @SD10 @BasThomas and all @Moya/contributors that want to help. This one is very important for newcomers as they might be confused as to what to do when developing a new feature/fix.

pedrovereza:
@BasThomas SwiftLint is not configured to check the Demo project, just [`/Sources` and `/Tests`](https://github.com/Moya/Moya/blob/master/.swiftlint.yml#L10)

sunshinejr:
@BasThomas are you up for removing these indents? Seems like the last one for merge :) (we can add linting afterwards)

BasThomas:
Done, feel free to merge once green!

sunshinejr:
Thanks so much everyone! ❤️

AndrewSB:
❤️ thank you for finishing this up @sunshinejr, and for the reviews everyone

@pedrovereza @BasThomas: Is there a reason we're not linting the demo project? Now that it's in the same xcodeproj it would be trivial to enable. Thoughts?

sunshinejr:
@AndrewSB yeah, good idea, if you have some spare time you can go for it :)

AndrewSB:
working on it now 👌

pedrovereza:
@AndrewSB no reason, I think we just never did it 😅

Issue #1267, opened 2017-09-06
Objects with RxMoyaProvider are no longer retained in 9.0


bhstrifork:
Hey,

I'm running into an interesting issue with Moya 9.0

I have a `Service` type for each kind of request that looks like:

```swift
struct UserService {
    let provider: MoyaProvider<MyAPI>
    
    init(provider: MoyaProvider<MyAPI>) {
        self.provider = provider
    }

    func login(usingUsername username: String, password: String) -> Single<String?> {
        return self.provider.rx.request(.login(username: username, password: password))
            .performMapping()
    }
}
```

This is then used like this from a singleton class:

```swift
let userService = UserService(provider: MoyaProvider<MyAPI>())
userService.performLogin(username: "foo", password: "bar").subscribe(onSuccess: {...}, onError: {...}
```

This used to work fine in Moya 8.x, where I was using a `RxMoyaProvider`. Any suggestions to workarounds or whether this should be fixed on Moya's side?

bhstrifork:
Forgot to add usage example of the service.

pedrovereza:
@bhstrifork Hey, could you clarify which behavior is different now? I can get a hint from the issue title, but I feel a more complete description would be helpful :wink:

bhstrifork:
Yeah, I should've been clearer.

In Moya 8.x I could use the `UserService` as shown above, and it would not get deallocated. Now in Moya 9.0 it is deallocated immediately, meaning my request either doesn't fire, or I never receive the response in my subscription. I was trying to find the issue discussing this already in Moya 8, but I can't seem to find it.

So basically I would like to have a function call as follows:

```swift
func login() {
  let userService = UserService(provider: MoyaProvider<MyAPI>())
  userService.performLogin(username: "foo", password: "bar")
    .subscribe(onSuccess: {...}, onError: {...}
}
```

And then maybe have it call a completion or similar. Or even just forward the request like this:

```swift
func login() -> Single<String?> {
  let userService = UserService(provider: MoyaProvider<MyAPI>())
  return userService.performLogin(username: "foo", password: "bar")
}
```

So the caller doesn't have to care about the `UserService`.

Does this clarify things?

AndrewSB:
I think the issue is the provider itself isn't retained, usually a singleton is defined somewhere outside with a `static let`. Are you familiar with how ARC works in Swift?

Here is your function with some annotations from me
```swift
func login() {
  // begin creating a UserService
[...]

PR #1294, opened 2017-09-19
Fixing potential issue with RxSwift request method

magi82:
Before anything else, thank you for this library. Recently, I attempted to network with the server using the `request` method of `RxSwift` Unfortunately, I received no response.

```
MoyaProvider<API>().rx.request(.version)
    .debug()
    .subscribe {
        print($0)
    }
    .disposed(by: self.disposeBag)
```

```
2017-09-19 16:34:12.833: SplashViewModel.swift:97 (setVersionInfo()) -> subscribed
```

I assume it was due to the fact that `Single<Response` has been captured as `weak` within the RxProvider’s `create` closure. I propose that it is better to have a potential retain cycle rather than having no response at all. Therefore, I’ve tried to circumvent the issue by having it similar to that of the `requestWithProgress` method which retains. Again, I appreciate your service for the community.

```
MoyaProvider<API>().rx.request(.version)
    .debug()
    .subscribe {
        print($0)
    }
    .disposed(by: self.disposeBag)
```

```
2017-09-19 16:36:50.421: SplashViewModel.swift:97 (setVersionInfo()) -> subscribed
2017-09-19 16:36:50.637: SplashViewModel.swift:97 (setVersionInfo()) -> Event next(Status Code: 200, Data Length: 281)
success(Status Code: 200, Data Length: 281)
2017-09-19 16:36:50.640: SplashViewModel.swift:97 (setVersionInfo()) -> Event completed
2017-09-19 16:36:50.640: SplashViewModel.swift:97 (setVersionInfo()) -> isDisposed
```


AndrewSB:
Hmm, I have a feeling you should be storing a strong reference to your provider somewhere instead. This reminds me of #1267, can you try that and see if it fixes your issue?

volodg:
@AndrewSB
1. Created ```Observable```/```Single``` should guarantee callback calling
2. Before ```request``` was storing reference to provider, so this changes incompatible with legacy code
3. ```rxRequestWithProgress``` is holding the strong reference, why it is ok for this method and bad for similar one: ```request```
4. It is normal when closure holds required info for call

AndrewSB:
I'm reading through [RxSwift's guide to disposing](https://github.com/ReactiveX/RxSwift/blob/master/Documentation/GettingStarted.md#disposing ) and, in my potentially uninformed mindset of things Rx, it seems to me like you have to hold onto your provider if you want your subscriptions to stay alive. 

Can we arrive at a consensus for that? @sunshinejr @pedrovereza @freak4pc what do you think?

Also, if that is the case, maybe we shouldn't be capturing `self` in [`rxRequestWithProgress`](https://github.com/Moya/Moya/blob/master/Sources/RxMoya/MoyaProvider%2BRx.swift#L53)

sunshinejr:
I believe we always guided to retain the provider to get the correct behavior. We will need to investigate why there is no response for a `Single` extension - there should be at least an error. Also we should check the consistency across both RxSwift request methods as well as both ReactiveSwift methods.

iwheelbuy:
Hello everybody.

@volodg have you tried [using operator](http://reactivex.io/documentation/operators/using.html) for holding logic? Your provider will live as long as it performs a request.

Small example of how I usually do it:

```
.flatMapLatest(observableCancel: observableCancel, observableFactory: { _ -> Observable<...> in
    return Observable
        .using(factoryApiService) { (service: Service) -> Observable<...> in
            return service.request()
        }
}

```
Where Service holds strong reference to Moya provider. This way I never store any provider and I also very often use different providers for each request

volodg:
@iwheelbuy thanks a lot, I did not know about it before, but nevertheless, even if a request was released there should be at least an error.

iwheelbuy:
@magi82 @AndrewSB @volodg @sunshinejr
There is one thing I've discovered. When you do it this way:
```
MoyaProvider<API>().rx.request(.version)...
```
The provider is deinited before the **rxRequest** is called

```
internal func rxRequest(_ token: Target, callbackQueue: DispatchQueue? = nil) -> Single<Response> {
    return Single.create { [weak self] single in
        ...
    }
}
```
So... There are two simple temp solutions:

**Solve it locally this way:**
```
let provider = MoyaProvider<API>()
provider.rx.request(.version)...
```
**Solve it inside Moya** (added defer)
```
internal extension MoyaProviderType {

    internal func rxRequest(_ token: Target, callbackQueue: DispatchQueue? = nil) -> Single<Response> {
        return Single.create { [weak self] single in
            defer {
                if self == nil {
                    single(.error(NSError.init(domain: "MoyaProviderType is nil", code: 100500, userInfo: nil)))
                }
            }
            let cancellableToken = self?.request(token, callbackQueue: callbackQueue, progress: nil) { result in
                switch result {
                case let .success(response):
                    single(.success(response))
                case let .failure(error):
                    single(.error(error))
                }
            }

            return Disposables.create {
                cancellableToken?.cancel()
            }
        }
    }
}
```

AndrewSB:
> even if a request was released there should be at least an error.
@volodg I disagree, I think Rx's philosophy is to do nothing when an Observable is disposed, the pattern doesn't recommend sending an error event before disposing.
I think you need to handle the disposed case just like the error case, if your request can be disposed under any circumstances

volodg:
@AndrewSB how to handle the disposed case when any callback is called?

sunshinejr:
@AndrewSB I agree in the terms of Rx, but the `Single` type is either a value or a failure, so if there is no value in it we can't really complete because there is no such event for this trait. From [docs](https://github.com/ReactiveX/RxSwift/blob/master/Documentation/Traits.md#single):
```
A Single is a variation of Observable that, instead of emitting a series of elements, is always guaranteed to emit either a single element or an error.
```

This makes me think that implementing an error there might be the right thing to do to not break the contract. What do you think?

AndrewSB:
I think we should do whatever RxSwift does whenever a `Single` created through `someObservable.asSingle()` gets disposed.
If that `Single` receives an error on disposal, lets do the same, else dispose without erring. 

@volodg do you want to try to see what happens and post your findings?

AndrewSB:
@sunshinejr: this may be a good candidate for hacktoberfest!

volodg:
@AndrewSB 
this example:
```Swift
    let stream = Observable<Int>
      .interval(1, scheduler: MainScheduler.instance)
      .timeout(1, scheduler: MainScheduler.instance)
    
    let singleStream = stream.asSingle().do(
    onNext: { _ in
      print("onNext")
    }, onError: { _ in
      print("onError")
    }, onSubscribe: {
      print("onSubscribe")
    }, onSubscribed: {
      print("onSubscribed")
    }, onDispose: {
      print("onDispose")
    })
    
    singleStream.subscribe().dispose()
```
prints:
```
onSubscribe
onSubscribed
onDispose
```

AndrewSB:
#1311 does The Right Thing and supersedes this

PR #1302, opened 2017-09-21
fix(rxRequest method)

volodg:
With weak ```self``` reference, ```Single``` stream will never complete


AndrewSB:
This dupes #1294, @volodg we'd love your input there instead!

PR #1311, opened 2017-09-26
Stop retaining self in rxReuestWithProgress

AndrewSB:
Fixes inconsitency in #1294

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,10 +6,12 @@
 - **Breaking Change** `Endpoint.init` so it doesn't have any default arguments (removing default argument `.get` for `method` parameter and `nil` for  `httpHeaderFields` parameter). [#1289](https://github.com/Moya/Moya/pull/1289) by [@sunshinejr](https://github.com/sunshinejr).
 - **Breaking Change** `NetworkActivityPlugin` so its `networkActivityClosure` has now `target: TargetType` argument in addition to `change: NetworkActivityChangeType`. [#1290](https://github.com/Moya/Moya/pull/1290) by [@sunshinejr](https://github.com/sunshinejr).
 - **Breaking Change** `Endpoint`'s `urlRequest` property to `urlRequest()` a throwing method. [#1248](https://github.com/Moya/Moya/pull/1248) by [@SD10](https://github.com/SD10).
-- `MoyaProvider+Rx` fix self-retaining bug in rxRequestWithProgress [#1311](https://github.com/Moya/Moya/pull/1311) by [@AndrewSB](https://github.com/AndrewSB).
 
 ### Removed
 - **Breaking Change** `RxMoyaProvider` and `ReactiveSwiftMoyaProvider`. [#1320](https://github.com/Moya/Moya/pull/1320) by [@SD10](https://github.com/SD10)
+
+### Fixed
+- `MoyaProvider+Rx` self-retaining bug in `requestWithProgress`. [#1311](https://github.com/Moya/Moya/pull/1311) by [@AndrewSB](https://github.com/AndrewSB).
 
 # 9.0.0
 - Removed default value for task from `Endpoint` initializer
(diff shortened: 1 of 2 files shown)


SD10:
@AndrewSB I grabbed the error from the build log so you don't need to go hunting 😃 
`/Users/distiller/Moya/Sources/RxMoya/MoyaProvider+Rx.swift:64:17: value of optional type 'Cancellable?' not unwrapped; did you mean to use '!' or '?'?`

Also, we have started to [formalize the CHANGELOG guidelines](https://github.com/Moya/contributors/blob/master/Changelog%20Guidelines.md). I noticed you added this under the **Changed** section but it could be better under **Fixed**.

Changelog.md
```diff
@@ -2,6 +2,7 @@
 ### Changed
 - **Breaking Change** `Endpoint.init` so it doesn't have any default arguments (removing default argument `.get` for `method` parameter and `nil` for  `httpHeaderFields` parameter). [#1289](https://github.com/Moya/Moya/pull/1289) by [@sunshinejr](https://github.com/sunshinejr).
 - **Breaking Change** `NetworkActivityPlugin` so its `networkActivityClosure` has now `target: TargetType` argument in addition to `change: NetworkActivityChangeType`. [#1290](https://github.com/Moya/Moya/pull/1290) by [@sunshinejr](https://github.com/sunshinejr).
+- `MoyaProvider+Rx` fix self-retaining bug in rxRequestWithProgress [#1311](https://github.com/Moya/Moya/pull/1311) by [@AndrewSB](https://github.com/AndrewSB).
```
sunshinejr: Can we add it in the new column, **Fixed**, instead of **Changed**? [Here](https://github.com/Moya/contributors/blob/master/Changelog%20Guidelines.md) is the guidelines we created for the Changelog.md.

Sources/RxMoya/MoyaProvider+Rx.swift
```diff
@@ -50,8 +50,8 @@ internal extension MoyaProviderType {
             }
         }
 
-        let response: Observable<ProgressResponse> = Observable.create { observer in
-            let cancellableToken = self.request(token, callbackQueue: callbackQueue, progress: progressBlock(observer)) { result in
+        let response: Observable<ProgressResponse> = Observable.create { [weak self] observer in
+            let cancellableToken = self?.request(token, callbackQueue: callbackQueue, progress: progressBlock(observer)) { result in
```
sunshinejr: You forgot to add `?` in `Dispsables.create { cancellableToken.cancel }` below 😄

volodg:
This solution contains a bug, example:
```Swift
    var stream = Observable<Void>.create { observer in
      return Disposables.create {
      }
    }
    
    stream.subscribe(onError: { _ in
      print("doOnError")
    }, onCompleted: {
      print("doOnCompleted")
    }, onDispose: {
      print("doOnDispose")
    })
```

No callback is called !!! Observable should be silent. If you are saying it is disposed when self is nil, call  onDispose callback

AndrewSB:
Sorry about the sloppy pull requests @SD10 @sunshinejr, both here and #1292. I'll work on testing my changes before submitting them so you guys don't have to play compiler for me -- I've been doing less Swift lately so it looks like my intuition is less on point than it used to be 😥

volodg:
@AndrewSB 
this example:
```Swift
    let singleStream = Observable<Int>
      .interval(1, scheduler: MainScheduler.instance)
      .timeout(1, scheduler: MainScheduler.instance)
      .asSingle()
      .do(onNext: { _ in
      print("onNext")
    }, onError: { _ in
      print("onError")
    }, onSubscribe: {
      print("onSubscribe")
    }, onSubscribed: {
      print("onSubscribed")
    }, onDispose: {
      print("onDispose")
    })
    
    singleStream.subscribe().dispose()
```
prints:
```
onSubscribe
onSubscribed
onDispose
```

So onDispose at least should be called, please fix

sunshinejr:
@volodg this PR focus rather on inconsistency than on fixing the `Single` behavior - let's focus at one thing at a time. @AndrewSB I think there is only a Changelog fix left to merge 👍

AndrewSB:
@volodg from my understanding, onDispose is already being called if the Single is disposed. in this comment https://github.com/Moya/Moya/pull/1294#issue-258728763 for example, the `.debug()` operator prints `isDisposed`.

@sunshinejr sounds good! Once I commit can you re-review before merge?

sunshinejr:
Sure thing 👍

sunshinejr:
@AndrewSB I don't see any changes to the `Changelog` 🤔 What I'm talking about is that comment about making it in the separate column **Fixed**, instead of having it in **Changed**.

sunshinejr:
I went ahead and fixed the Changelog. It should be good to go now. Please re-review @Moya/contributors.

sunshinejr:
@SD10 thank you!

PR #1315, opened 2017-09-26
Set indentWith and tabWidth in project settings

devxoul:
These settings help people to use same indentation in Moya.xcodeproj. These settings have higher priority than the indentation setting in Xcode.

screen shot 2017-09-26 at 9 17 00 pm

### DIFF
--- a/Moya.xcodeproj/project.pbxproj
+++ b/Moya.xcodeproj/project.pbxproj
@@ -264,7 +264,9 @@
 				9A58CD622B8FBB54907843BE /* Sources */,
 				1952C4A00E2EF89FFE965603 /* Tests */,
 			);
+			indentWidth = 4;
 			sourceTree = "<group>";
+			tabWidth = 4;
 		};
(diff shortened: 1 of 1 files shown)


sunshinejr:
Thanks, @devxoul! I'm wondering if we want/can do the same options for [framework-generate](https://github.com/petester42/framework-generate). We are using for generating the project so it might come handy later on. 

What do you think? cc @petester42

petester42:
I don’t think that’s a good idea. I want the tool to be just about targets and setting them up easier. I’d like the tool to maybe just read the project and update your targets based on the `FrameworkSpec` rather than recreate the project every time. I’ve seen it not work very well when trying to just migrate version.  In that case setting xcode project setting wouldn’t change when using the tool. 

As far as providing the ability to add this when we generate the project it should be doable without changing the tool at all. All targets should be already be exposing the xcodeproj objects. All that would need to be done is dig through their documentation and apply the setting in the `FrameworkSpec`

sunshinejr:
Makes sense @petester42. Alright, merging this one then!

PR #1320, opened 2017-09-27
[Removed] RxMoyaProvider and ReactiveSwiftMoyaProvider

SD10:
- [x] Removes the `RxMoyaProvider` and `ReactiveSwiftMoyaProvider` deprecated in `v9.0`. 
- [X] Changes MoyaProviderRx -> MoyaProviderRxSpec 😉
- [x] Add **Removed** Changelog entry


Tests/MoyaProvider+RxSpec.swift
```diff
@@ -6,7 +6,7 @@ import OHHTTPStubs
 @testable import Moya
 @testable import RxMoya
 
-final class MoyaProviderRx: QuickSpec {
+final class MoyaProviderRxSpec: QuickSpec {
```
sunshinejr: 👌

sunshinejr:
Thanks for the work, @SD10! One thing that I didn't say before is that with removal of reactive providers we don't need additional abstraction over reactive methods (we needed these as we shared code (e.g. `rx.request` and `RxMoyaProvider<>().request`). Would you like to remove the unnecessary abstraction as well? ( e.g. [here](https://github.com/Moya/Moya/blob/master/Sources/RxMoya/MoyaProvider%2BRx.swift#L9) and in ReactiveSwift extensions as well)

SD10:
@sunshinejr I'm a little confused about the refactoring that needs to be done to remove the two abstractions. It requires I prefix all `MoyaProvider` methods with `.rx.base` & `.reactive.base` then.

I tried moving the methods back into `extension Reactive where Base: MoyaProviderType` but have problems regarding capturing `self` on a non-class constrained type.

Feel free to take over if you'd like

sunshinejr:
@SD10 oh, let's just fix these linting issues and we can do the refactor in another PR, no worries!

Issue #1286, opened 2017-09-14
[CI] Don't run for documentation-only changes.


sunshinejr:
I found [an example](https://github.com/facebook/react/pull/2000/files#diff-354f30a63fb0907d4ad57269548329e3), but unfortunately it's for Travis so this is in `.travis.yml` config. We would have to migrate it to either Rakefile/Circle config or maybe even a new script.

As always, please assign yourself if you are working on it so we won't duplicate our work ✌️

Sherlouk:
Assign this one to me, gives me an excuse to look into Circle 😅

sunshinejr:
Sure thing, thanks @Sherlouk!

SD10:
Yeah, there's an issue requesting that feature [here](https://github.com/isaacs/github/issues/100). It's been 4 years already so I don't think it's coming anytime soon 😭 I really want it though

yeskunall:
I found [this](https://circleci.com/blog/circleci-hacks-automate-the-decision-to-skip-builds-using-a-git-hook/) article that shows how to skip CI builds if changes are in a certain directory by combining git hooks with a `.ciignore` (or any name) file. The only problem is that git hooks are local and are not copied over when a repo is cloned or forked, so I've hit a dead-end there 😓

I'm still looking for a way to achieve this in CircleCI tho 😅

sunshinejr:
@yeskunall you can check #1339 as there should be a working script that does that.

Issue #1194, opened 2017-07-30
[CI] Support multiple Swift versions.


sunshinejr:
Right now #1193 will update our Swift support to Swift 3.1, 3.2 and 4.0. While it is awesome, we should do some checks to make sure we do not break these unintentionally. We can do builds for all of them, but I think checking for 3.2 & 4.0 compatibility should be enough. What do you guys think?

BasThomas:
I think those two would be enough. Then in September we can drop Swift 3 tests.

sunshinejr:
**Update:** So I tried the new workflows API from CircleCI, but [seems like macOS is not yet supported there](https://discuss.circleci.com/t/can-one-do-macos-builds-in-circleci-2-0/11691/5). Thus, we probably need to fallback to rake.

sunshinejr:
Update: CircleCI 2 on macOS is in beta stages and for the last few days I was tweaking the script to at least run 3 jobs asynchronously - with [success](https://circleci.com/workflow-run/88dd0026-d33b-407b-a3bb-4e209012c6fb)! You can take a look at branch [workflows](https://github.com/Moya/Moya/tree/workflows) which is kinda in a WIP stage. I've managed to split our build process into 3 routes:
- build_demo (builds our Demo project, with CocoaPods)
- test_integration_carthage (tests Carthage integration)
- test (tests the project)

And after that there is a danger/codecov plugin that I didn't really test that well, because I would need a PR on top of that.

Also, all of these reduced our build time to around ~**12-13** minutes (from **20-25**)!

Now I just need to add additional asynchronous routes so it can test with Swift 3.2 & Swift 4 at the same time. I'm gonna update you guys here with the progress!

sunshinejr:
Got a working copy of both Swift 3.2 & 4.0 builds on [here](https://circleci.com/workflow-run/28a1ad8f-c9c8-406a-8fcf-267376b85b7e), really cool stuff in there. Might try to get skipping trivial builds if it's possible as well.

PR #1324, opened 2017-09-27
[WIP][CI] Add new script to exit on non-trivial changes (docs, markdown, etc)

Sherlouk:
Working towards #1286, created a new script which lists the file names of anything changed within the provided commit.

As per CircleCI's docs (https://circleci.com/docs/1.0/environment-variables/) I've gotten the "CIRCLE_SHA1" variable which should be the commit hash.

Following that I do an inverted grep which searches for an entry that doesn't match the regex (so source file changes etc), and in that occasion will exit with code 0 (success).

In the event it does find something, it'll return with exit code 1 which I'm hoping is enough for Circle to stop on.

For testing purposes I've added some prints to see what's happening, the regex isn't final as will need to find out what we want to search for and include/exclude!

Creating a quick WIP pull request just so I can get CircleCI to give this a run. (Wasn't letting my build on my fork 🤷‍♂️ )


sunshinejr:
@Sherlouk I've labeled it with `[WIP]` - please remove it once it is ready for review 😉

Sherlouk:
Sorry yea I'm just staring blankly trying to work out whether Circle is just being slow to spin up, or I've done something wrong 🤔 

More likely the latter, but I don't know 😂

marcomorain:
@Sherlouk @sunshinejr you can put the string `[ci skip]` into the commit message to skip CI builds - does that do what you need?

sunshinejr:
@marcomorain That's correct, but new contributors might not know the rule and we are trying to automate the process without additional steps if possible.

Sherlouk:
Pushing a new commit was enough to get Circle running again, couple stupid mistakes later (shebang and file permissions) it's managed to get past that stage.

Next steps is to make sure I can stop a build from continuing but with a successful code!

Sherlouk:
@sunshinejr So noticing two things immediately, due to the way CircleCI separate everything into isolate stages it makes it pretty difficult to exit early with a success flag (obviously you can drop a failure at any point in time).

The other thing is that Circle don't provide an easy way to get a range of commits, only the current commit. This in theory works for things like this pull request where it's running every commit, but in the instance where a user pushed up `x` commits and then creates a pull request where they only update a doc - it would only check the changes on the latest commit.

The second point is easier to fix with a few extra commands in order to work out the range of commits based on the branch, and diffing against master rather than solely on a given commit. Though honestly I'm not sure how to deal with the first one.

Obviously failing is an easy solution, but that would be a royal pain in the ~self censorship~ as you'd always have to find out why it failed. Another possible solution is to define a env variable "trivial_changes" and if that's found you just skip all the build stages (kind of messy, easy to make mistakes with future changes).

Maybe this wasn't the card to learn Circle with 😂 

Willing to take advice at this point?

pedrovereza:
> due to the way CircleCI separate everything into isolate stages it makes it pretty difficult to exit early with a success flag (obviously you can drop a failure at any point in time).

@Sherlouk Have you checked `circleci step halt` ([here](https://discuss.circleci.com/t/stop-build-with-success-failure-at-any-point-in-the-build-life-cycle/2691/2))?

Sherlouk:
@pedrovereza Yeah! I think I saw that exact post in fact, but unfortunately when I did it was saying that the command `circleci` did not exist!

See this build: https://circleci.com/gh/Moya/Moya/1898?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

pedrovereza:
@Sherlouk Okay, maybe it's because we're not using [Circle 2.0](https://circleci.com/docs/2.0/migrating-from-1-2/#steps-to-configure-required-20-keys) yet 🤔

sunshinejr:
@pedrovereza is probably right. If you want more info about CircleCI 2.0 with Moya see #1194. I have already made some progress and will try if skipping the job will be easier to do there.

Sherlouk:
You may be right about the Circle 2.0 stuff! @sunshinejr if you wanna give it a go and just see if it works then great -- The script is fairly simple from that point.

Shall we close this PR and move discussion back to 1286 until we know the situation?

sunshinejr:
Good idea @Sherlouk. Sorry for the trouble :( I'll make a PR with my WIP branch to see how it works!

Issue #1325, opened 2017-09-27
Add Task case to handle request with Encodable type.


sunshinejr:
Long read: #1135. Short one: we need an option to send `Encodable` type that will be converted to data and passed to `httpBody`. It could look somehow similar to:
```swift
enum Task {
    ...
    case requestJsonEncodable(Encodable)
    ...
}
```

afonsograca:
Hi,
I was trying to give this a go, but it seems that this requires more changes than first thought. Since `Encodable` doesn't conform to itself, when trying to encode
```swift
JSONEncoder().encode(encodable)
```
We get the following error `Cannot invoke 'encode' with an argument list of type '((Encodable))'` because it needs a concrete type to conform to the protocol.

This could be solved by defining making the `Task` enum generic, but this would involve a major breaking change, which in my opinion is not worth it. Encoding can be done prior to the request call.

SD10:
@afonsograca Thank's for giving this a shot. There's an open PR #1326 regarding this. It will be a difficult one to work out. As you mentioned, I don't believe making the entire `Task` type generic to support a single case is the right solution.  Hopefully, I can look into this more sometime this week.

afonsograca:
@SD10 damn, only saw your message now. I totally missed the open PR.

I ended type erasing the protocol and from the tests I've done it seems to work. I opened a PR, but given that there's already an open PR regarding this, please disregard mine. 

All in all, this was a good brain teaser :D haven't had to use type erasure in a while.
Keep up the good work!

devxoul:
@sunshinejr Just for making it sure: did you mean `httpBody` instead of `htmlBody`?

sunshinejr:
Right, how the hell did I think of `htmlBody` I'm not sure 😥

sunshinejr:
Not sure why I had to close it manually as the PR had `Resolves #1325` keyword, but yeah, it is already in beta so closing this one!

BasThomas:
GitHub's a little funny there - I know of only "fixes #whatever" that automatically closes the issue.

PR #1326, opened 2017-09-28
Add Task case to handle request with Encodable type. #1325

LeLuckyVint:
Added new enum case for Encodable type


sunshinejr:
Hey @LeLuckyVint, thanks for getting this one up! There are few more things before we can merge, though:
- Handling transformation from `endpoint` to `urlRequest` [here](https://github.com/Moya/Moya/blob/master/Sources/Moya/Endpoint.swift#L66)
- Handling performing the request [here](https://github.com/Moya/Moya/blob/aedd30604d78ac231b9aa295f021681ea62fd4da/Sources/Moya/MoyaProvider%2BInternal.swift#L93)
- Adding a test case around [here](https://github.com/Moya/Moya/blob/aedd30604d78ac231b9aa295f021681ea62fd4da/Tests/EndpointSpec.swift#L127)
- Adding a changelog entry (you can checkout our guidelines [here](https://github.com/Moya/contributors/blob/master/Changelog%20Guidelines.md))
- At the end check if do not have any issues with SwiftLint

The behavior of `requestJsonEncodable` vs `requestData` is very similar, so you can try to base your implementation on that one. If you'll need help in anything, we'll be happy to assist!

LeLuckyVint:
@sunshinejr i have a problem with associated value that confirms protocol `Equatable`.
When I want to create `Data` using `JSONEncoder` with associated value Xcode I have en error
screen shot 2017-09-29 at 1 13 28 pm

This is described in this [thread](https://www.reddit.com/r/swift/comments/6qha9l/why_can_i_not_use_a_protocol_in_the_method/)

sunshinejr:
Oh yeah, this is the tricky one @LeLuckyVint. I believe we would have to think about this - the problem is that we can't use `Encodable` and generic in the case. One solution that comes to my mind is making a type eraser on it. Maybe you have any other idea how to fix that one, @LeLuckyVint?

SD10:
Hey @LeLuckyVint, I just wanted to check if the status has changed on this one at all. It's one of the last things left to do for `v10.0` 😃 Thank you!

LeLuckyVint:
@SD10 can you look into it, please. I have all tests passed on my local machine, but CI is failing and the errors are confusing

sunshinejr:
Hey @LeLuckyVint. Thank you so much for this one! Unfortunately there were still few things to do before this would be merged. And because we want to release new version pretty soon, I finished it while giving credits to you and @afonsograca that also did a similar PR. I would love it if you could review my PR #1349 and let me know what do you think!

PR #1335, opened 2017-10-01
Add support for mapping response to Decodable object

### LINKED ISSUE #1118: Add Moya.Response extension for Decodable types in Swift 4
It seems like handling response JSON is a highly debated topic due to there being so many JSON serialization libraries and no determined best practice. 

However, with the addition of the `Codable` protocol to the Swift 4 Standard Library, I think this will become a widely used approach. I'm wondering if the Moya community would be willing to add some basic extensions to `Moya.Response` to map `Decodable` types. 

Since `Codable` is part of the Standard Library I think it's a safe addition to Moya's core.

A rough API would be:

```Swift
    func mapObject<D: Decodable>(with decoder: JSONDecoder) throws -> D {}
    func mapObject<D: Decodable>(atKeyPath keyPath: String, with decoder: JSONDecoder) throws -> D {}
```

Moya already supports mapping to JSON, Images, and Strings

Regardless, these are not difficult extensions to write on our own or provide in a community extension.

devxoul:
* Resolves #1118
* I recommend you to see commit logs first before reading code.
* Adds `MoyaError.objectMapping(Error, Response)` which contains a `DecodingError` to make it easy to inspect a decoding error.
* Interfaces:

    ```swift
    // Response -> D
    func map<D: Decodable>(_ type: D.Type, atKeyPath keyPath: String? = nil, using decoder: JSONDecoder) throws -> D

    // Obsevable<Response> -> Observable<D>
    func map<D: Decodable>(_ type: D.Type, atKeyPath keyPath: String? = nil, using decoder: JSONDecoder) throws -> Observable<D>

    // Single<Response> -> Single<D>
    func map<D: Decodable>(_ type: D.Type, atKeyPath keyPath: String? = nil, using decoder: JSONDecoder) throws -> Single<D>

    // SignalProducer<Response, MoyaError> -> SignalProducer<D, MoyaError>
    func map<D: Decodable>(_ type: D.Type, atKeyPath keyPath: String? = nil, using decoder: JSONDecoder) throws -> SignalProducer<D, MoyaError>
    ```

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,7 @@
 # Next
 ### Added
 - **Breaking Change** `.parameterEncoding(Swift.Error)` case to `MoyaError`. [#1248](https://github.com/Moya/Moya/pull/1248) by [@SD10](https://github.com/SD10).
+- **Breaking Change** Added `Decodable` object mapping methods to `Moya.Response`. [#1335](https://github.com/Moya/Moya/pull/1335) by [@devxoul](https://github.com/devxoul).
 
 ### Changed
 - **Breaking Change** `Endpoint.init` so it doesn't have any default arguments (removing default argument `.get` for `method` parameter and `nil` for  `httpHeaderFields` parameter). [#1289](https://github.com/Moya/Moya/pull/1289) by [@sunshinejr](https://github.com/sunshinejr).
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,7 +1,7 @@
 # Next
 ### Added
 - **Breaking Change** `.parameterEncoding(Swift.Error)` case to `MoyaError`. [#1248](https://github.com/Moya/Moya/pull/1248) by [@SD10](https://github.com/SD10).
-- **Breaking Change** Added `Decodable` object mapping methods to `Moya.Response`. [#1335](https://github.com/Moya/Moya/pull/1335) by [@devxoul](https://github.com/devxoul).
+- **Breaking Change** `Decodable` object mapping methods to `Moya.Response`. [#1335](https://github.com/Moya/Moya/pull/1335) by [@devxoul](https://github.com/devxoul).
 
 ### Changed
 - **Breaking Change** `Endpoint.init` so it doesn't have any default arguments (removing default argument `.get` for `method` parameter and `nil` for  `httpHeaderFields` parameter). [#1289](https://github.com/Moya/Moya/pull/1289) by [@sunshinejr](https://github.com/sunshinejr).
--- a/Sources/Moya/Response.swift
+++ b/Sources/Moya/Response.swift
@@ -116,4 +116,29 @@ public extension Response {
             return string
         }
     }
+
+    /// Maps data received from the signal into a Decodable object.
+    ///
+    /// - parameter atKeyPath: Optional key path at which to parse object.
+    /// - parameter using: A `JSONDecoder` instance which is used to decode data to an object.
+    func map<D: Decodable>(_ type: D.Type, atKeyPath keyPath: String? = nil, using decoder: JSONDecoder = JSONDecoder()) throws -> D {
+        let jsonData: Data
+        if let keyPath = keyPath {
+            guard let jsonDictionary = (try mapJSON() as? NSDictionary)?.value(forKeyPath: keyPath) as? [String: Any] else {
+                throw MoyaError.jsonMapping(self)
+            }
+            do {
+                jsonData = try JSONSerialization.data(withJSONObject: jsonDictionary, options: [])
+            } catch {
+                throw MoyaError.jsonMapping(self)
+            }
+        } else {
+            jsonData = data
+        }
+        do {
+            return try decoder.decode(D.self, from: jsonData)
+        } catch let error {
+            throw MoyaError.objectMapping(error, self)
+        }
+    }
 }
--- a/Tests/Observable+MoyaSpec.swift
+++ b/Tests/Observable+MoyaSpec.swift
@@ -266,5 +266,92 @@ class ObservableMoyaSpec: QuickSpec {
                 expect(receivedError).to(beOfSameErrorType(expectedError))
             }
         }
+
+        describe("object mapping") {
+            let formatter = DateFormatter()
+            formatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss"
+
+            let decoder = JSONDecoder()
+            decoder.dateDecodingStrategy = .formatted(formatter)
+
+            let json: [String: Any] = [
+                "title": "Hello, Moya!",
+                "createdAt": "1995-01-14T12:34:56"
+            ]
+
+            it("maps data representing a json to a decodable object") {
+                guard let data = try? JSONSerialization.data(withJSONObject: json, options: .prettyPrinted) else {
+                    preconditionFailure("Failed creating Data from JSON dictionary")
+                }
+                let observable = Response(statusCode: 200, data: data).asObservable()
+
+                var receivedObject: Issue?
+                _ = observable.map(Issue.self, using: decoder).subscribe(onNext: { object in
+                    receivedObject = object
+                })
+                expect(receivedObject).notTo(beNil())
+                expect(receivedObject?.title) == "Hello, Moya!"
+                expect(receivedObject?.createdAt) == formatter.date(from: "1995-01-14T12:34:56")!
+            }
+
+            it("maps data representing a json array to an array of decodable objects") {
+                let jsonArray = [json, json, json]
+                guard let data = try? JSONSerialization.data(withJSONObject: jsonArray, options: .prettyPrinted) else {
+                    preconditionFailure("Failed creating Data from JSON dictionary")
+                }
+                let observable = Response(statusCode: 200, data: data).asObservable()
+
+                var receivedObjects: [Issue]?
+                _ = observable.map([Issue].self, using: decoder).subscribe(onNext: { objects in
+                    receivedObjects = objects
+                })
+                expect(receivedObjects).notTo(beNil())
+                expect(receivedObjects?.count) == 3
+                expect(receivedObjects?.map { $0.title }) == ["Hello, Moya!", "Hello, Moya!", "Hello, Moya!"]
+            }
+
+            it("maps data representing a json at a key path to a decodable object") {
+                let json: [String: Any] = ["issue": json] // nested json
+                guard let data = try? JSONSerialization.data(withJSONObject: json, options: .prettyPrinted) else {
+                    preconditionFailure("Failed creating Data from JSON dictionary")
+                }
+                let observable = Response(statusCode: 200, data: data).asObservable()
+
+                var receivedObject: Issue?
+                _ = observable.map(Issue.self, atKeyPath: "issue", using: decoder).subscribe(onNext: { object in
+                    receivedObject = object
+                })
+                expect(receivedObject).notTo(beNil())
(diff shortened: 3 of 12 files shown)


sunshinejr:
Thank you so much @devxoul for this one! 🎉  I didn't go through all of this yet, but few questions at a first glance:
- How do you feel about mapping arrays of `Decodable`s?
- How do you feel about renaming `mapObject(_:atKeyPath:using)` to `map(to:keyPath:encoder)`?
- How do you feel about having default encoder argument?

devxoul:
1. `[Decodable]` is also a decodable so we do not have to write an extra function to deal with an array.
2. I brought interfaces from [RxCodable](https://github.com/devxoul/RxCodable), but I can change it if you think it would be better. (PS: You mean `decoder`, not an `encoder` right? 😃)
3. Yeah, I think we can let the `decoder` parameter optional and use a plain `JSONDecoder` if the parameter is `nil`.

sunshinejr:
@devxoul Definitely, I was talking about decoder but my mind got messed up, sorry!

I didn't know that `[Decodable]` is also `Decodable` - that's awesome! So now the question is for the interface, I don't mind both(or mixture) of them, so maybe let's wait for more inputs :)

SD10:
@devxoul 💯 Thank you for implementing this, you're a life-saver 👍 Thank you for writing tests too!
I can help with updating any docs that revolve around these changes.

I do like @sunshinejr's suggestion of naming the method `map(to:)` over `mapObject`. Object seems redundant and intuitive. I also like shorter names. However, I do like keeping the external parameter names the same as the original method:

```Swift
func map<D: Decodable>(to: D.Type, atKeyPath keyPath: String? = nil, using decoder: JSONDecoder) throws -> D
```
The reason being is because this method reads as fluent English using the `atKeyPath` and `using` parameter names. 🤔

🚲 Shedding

sunshinejr:
I'm up for mixed strategy as well (for interface naming), which is:
```swift
func map<D: Decodable>(to: D.Type, atKeyPath keyPath: String? = nil, using decoder: JSONDecoder = JSONDecoder()) throws -> D
```

So that will be the last thing before documentation (Changelog + documenting the option itself) 👍

devxoul:
I like that interface. It looks much better :+1:

@sunshinejr, Hmm, there are several issues on linting issues:

1. Quick spec is a single class and a single method, so it can easily exceed 200 lines. How about removing this check from the test code?
2. The 55th line of `TestHelpers.swift` is not included in this pull request. I think we should fix this in another pull request to keep this pr tidy. How do you think?

devxoul:
How about removing the parameter name `to`? Because, we're using such like `mapData()` or `mapString()` over `mapToData()` and `mapToString()`. English is not my first language but `map something` is more natural to me than `map to something`.

```swift
- foo.map(to: Issue.self, atKeyPath: "data", using: myDecoder)
+ foo.map(Issue.self, atKeyPath: "data", using: myDecoder)
```

phimage:
The `Decodable` type is not mandatory as parameter of the function
You could specify it by typing the object

with current PR code we must do
```swift
let response = Response(statusCode: 200, data: data)

let receivedObject: Issue = try response.mapObject(atKeyPath: "issue", decoder: decoder)
```

Maybe it's easier for reactivate programming but for standard behaviour I thinks removing it will be the best solution

With reactive programming the code will be something like this
```swift
let single = response.asSingle()
var receivedObject: Issue?
_ = single.mapObject(atKeyPath: "issue", decoder: decoder).subscribe(onSuccess: { (object: Issue) in
      receivedObject = object
})
```
for event , `.subscribe { (event: (SingleEvent<Issue>)) in`

SD10:
Wow, very interesting @phimage. Maybe we can keep the type constraint on the reactive methods and drop it on the normal `Moya.Response` method? I also wonder how obvious it is that you need to specify the type explicitly 🤔

sunshinejr:
👍 for removing `to` parameter name

About skipping type parameter - I feel like this would make more issues about compiler problems as you need to explicitly write that type to let compiler infer it. Making it explicit with a parameter removes this problem as people would have to specify it always as a parameter.

devxoul:
If we remove the `type` parameter, method chaining will not work properly because the compiler cannot infer return types.

Tests/Observable+MoyaSpec.swift
```diff
@@ -266,5 +266,87 @@ class ObservableMoyaSpec: QuickSpec {
                 expect(receivedError).to(beOfSameErrorType(expectedError))
             }
         }
+
+        describe("object mapping") {
```
sunshinejr: Can we add a test that maps an array of objects?

sunshinejr:
Also, I'm sorry again @devxoul but I had to rebase 10.0.0-dev again as we had project changes that were needed on master. You will need to rebase again 😢

devxoul:
@sunshinejr, could you help me with this? Several files are missing from project after rebasing.

screen shot 2017-10-03 at 2 49 31 pm

SD10:
@devxoul It's most likely because of #1292. Looks like the Xcodeproj was corrupted in the rebase

sunshinejr:
Yep, working on it!

sunshinejr:
@devxoul I think it is fixed, please try now. It turned out that we had both on 10.0.0-dev and master changes to `pbxproj` and auto merged kinda broke it 😞

SD10:
I believe I fixed it on a local copy if anyone needs it. It's essentially copying the `project.pbxproj` off of master and deleting the two references to `RxMoya+Availability` & `ReactiveMoya+Availability`.

devxoul:
@sunshinejr @SD10 thanks for quick support! I rebased `10.0.0-dev` branch again and added some test code for json arrays.

Tests/Observable+MoyaSpec.swift
devxoul: Sure! Added.

Sources/Moya/Response.swift
```diff
@@ -116,4 +116,25 @@ public extension Response {
...
+    ///
+    /// - parameter atKeyPath: Optional key path at which to parse object.
+    /// - parameter using: A `JSONDecoder` instance which is used to decode data to an object.
+    func map<D: Decodable>(_ type: D.Type, atKeyPath keyPath: String? = nil, using decoder: JSONDecoder = JSONDecoder()) throws -> D {
+        let jsonData: Data
+        if let keyPath = keyPath {
+            guard let jsonDictionary = (try mapJSON() as? NSDictionary)?.value(forKeyPath: keyPath) as? [String: Any] else {
+                throw MoyaError.jsonMapping(self)
+            }
+            jsonData = try JSONSerialization.data(withJSONObject: jsonDictionary, options: [])
```
sunshinejr: don't we wanna catch the error here as well and make it a `.jsonMapping` as well? 🤔

Tests/Observable+MoyaSpec.swift
```diff
@@ -266,5 +266,103 @@ class ObservableMoyaSpec: QuickSpec {
...
+                        fail("next called for invalid data")
+                    case .error(let error):
+                        receivedError = error
+                    default:
+                        break
+                    }
+                }
+
+                // check whether we properly got a MoyaError
+                expect(receivedError).to(Predicate<Error> { expression in
```
sunshinejr: hmm, isn't `MoyaError` equatable? if not, maybe we can make it so (at least for tests), would make this test cleaner

Tests/Observable+MoyaSpec.swift
```diff
@@ -266,5 +266,103 @@ class ObservableMoyaSpec: QuickSpec {
...
+                    } else {
+                        return PredicateResult(status: .doesNotMatch, message: .expectedActualValueTo("MoyaError.objectMapping"))
+                    }
+                })
+
+                // check whether the MoyaError has a DecodingError as an associated value
+                if case let MoyaError.objectMapping(decodingError, _)? = receivedError {
+                    expect(decodingError).to(beAKindOf(DecodingError.self))
+                } else {
+                    preconditionFailure()
```
sunshinejr: Hmm, is it a `preconditionFailure` though? For me it is a `fail()` as if the test failed, but I'm curious what's your opinion on that one. Also, it is a shame that we cannot avoid this `if` & `preconditionFailure`/`fail` 👿

sunshinejr:
@devxoul last comments from me!

Tests/TestHelpers.swift
```diff
@@ -220,3 +220,14 @@ extension ImageType {
         }
     #endif
 }
+
+// A fixture for testing Decodable mapping
+struct Issue: Codable {
+    let title: String
+    let createdAt: Date
+
+    enum CodingKeys: String, CodingKey {
```
AndrewSB: @devxoul are these keys needed? I thought `Codable` automatically synthesized them if they're omitted.

SD10: Yes, they should be automatically created and specifying the keys is only necessary if you want to change the name.

Sources/Moya/Response.swift
devxoul: Great!

Tests/Observable+MoyaSpec.swift
devxoul: It would be tough because `MoyaError.objectMapping` has an associated value `Error` which is not an Equatable. I ended up with using `if-case-let` instead.

```swift
if case let MoyaError.objectMapping(nestedError, _)? = receivedError {
    expect(nestedError).to(beAKindOf(DecodingError.self))
} else {
    fail("expected <MoyaError.objectMapping>, got <\(String(describing: receivedError))>")
}
```

Tests/TestHelpers.swift
devxoul: Ah, didn't know that. Fixed!

Tests/Observable+MoyaSpec.swift
devxoul: This code will be merged into a single `if-case-let` statement :)

devxoul:
@sunshinejr @AndrewSB Thanks for your reviews! I just updated my commits :)

SD10:
This does require a change to the rarely viewed`ErrorTypes.md` but it can be done when we provide some sort of example documentation for the response decoding.

sunshinejr:
Oh, Jeon, could you also add entry to `Changelog`, please?

devxoul:
Updated `ErrorTypes.md` and `Changelog.md` :)

SD10:
@devxoul 👍 on updating `ErrorTypes.md`. About the CHANGELOG entry, I think that adding `MoyaError.objectMapping(Error, Response)` is a breaking change because the `MoyaError` enum will no longer be considered exhaustive in some cases.

Changelog.md
```diff
@@ -1,6 +1,7 @@
 # Next
 ### Added
 - **Breaking Change** `.parameterEncoding(Swift.Error)` case to `MoyaError`. [#1248](https://github.com/Moya/Moya/pull/1248) by [@SD10](https://github.com/SD10).
+- Added `Decodable` object mappings. #1335 by [@devxoul](https://github.com/devxoul).
```
SD10: Added `Decodable` object mapping methods to `Moya.Response`. 
This could be a little more descriptive. Sorry to be so needy 😅

devxoul: That makes sense!

sunshinejr:
Thank you again, @devxoul and all contributors in this one! 🎉

devxoul:
Thanks for giving me an opportunity to contribute 🎉

PR #1340, opened 2017-10-03
[xcodeproj] add a number of recommended compiler flags

AndrewSB:
XCode 9 recommended these. They're a good idea imo


AndrewSB:
@sunshinejr I just rebased!

PR #1341, opened 2017-10-03
Enable swiftlint for Examples/ + cleanup rules

AndrewSB:


AndrewSB:
2 questions came up while I was doing this

1. Why don't we have a build phase that runs swiftlint for us?
2. Can someone try to remove cyclomatic_complexity from https://github.com/Moya/Moya/blob/master/Sources/Moya/MoyaProvider+Internal.swift#L204, maybe by splitting the function up, or collapsing them into existing `switch`es. Also, can we add documentation for what https://github.com/Moya/Moya/blob/master/Sources/Moya/MoyaProvider+Internal.swift#L224 is doing? Specifically why we don't have to do anything if `progressCompletion` is `nil`, or why we aren't handling the other cases of `progressAlamoRequest`? It no longer looks as if its "making the actual request", but that it's just registering progress callbacks.

BasThomas:
Ooooh, emoji in branch names. 😬

sunshinejr:
@AndrewSB to answer your questions:
1. It did came up before that we could have SwiftLint build phase, but it is not _that_ easy to do as we have to be careful for Carthage users. I remember when I used Carthage with Quick that they had problems because SwiftLint was running on the whole project, instead of their own library. Of course that does not mean we cannot implement it - just be careful with it.
2. About progress: before we assign normal `alamoRequest` to `progressRequest`, in case of given `progressCompletion` we just replace `alamoRequest` with request with progress for given types (namely: `DataRequest`, `UploadRequest`, `DownloadRequest`). We don't do anything for other cases (right now the other case is probably `StreamRequest` because it doesn't have progress in it).

AndrewSB:
1. I added the build phases, is there any way you can think of to test the potential breakage?
2. It sounds complicated, but like it's working. So I just disabled cyclomatic_complexity for the function :) 

Past that, this is ready to go after revew

SD10:
Wow, I wonder if pulling out the nested switch statement in an earlier PR contributed to removing the cyclomatic complexity warning. Thanks for doing this @AndrewSB 👍 I wanted to get rid of that for awhile.

PR #1348, opened 2017-10-07
Feature: Add JSON encodable task

### LINKED ISSUE #1325: Add Task case to handle request with Encodable type.

afonsograca:
Fixes #1325


sunshinejr:
Hey @afonsograca. Thank you so much for this one! Unfortunately there were still few things to do before this would be merged. And because we want to release new version pretty soon, I finished it while giving credits to you and @LeLuckyVint that also did a similar PR. I would love it if you could review my PR #1349 and let me know what do you think!

PR #1355, opened 2017-10-08
Add link to migration guides in README

SD10:
I've added a link to the `MigrationGuides.md` under the Moya version compatability table. I think this will help to make our migration guides more easily accessible. Let me know what you think

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -69,6 +69,8 @@ your Swift version.
 | 2.3   | 7.0.2 - 7.0.4 | 7.0.2 - 7.0.4 | 7.0.2 - 7.0.4 |
 | 2.2   | <= 7.0.1      | <= 7.0.1      | <= 7.0.1      |
 
+**Upgrading to a new major version of Moya? Check out our [migration guides](https://github.com/Moya/Moya/blob/master/docs/MigrationGuides.md).**
+
 ### Swift Package Manager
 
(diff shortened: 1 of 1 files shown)


BasThomas:
Awesome!

I was thinking... would it be nice to create an issue to keep track of changes in the Readme, so that the Chinese Readme can be updated according to that issue? That way we can keep track of the changes to keep both Readmes in sync as best way can.

SD10:
@BasThomas I think that's a great idea. Are we only concerned about the README or are we talking about all documentation in general?

We could have an issue: **Changes in English docs not reflected in Chinese docs**
Then we could reference that issue in any PR that modifies the docs. We could even write a danger rule to remind us 😉

BasThomas:
Exactly what I was thinking. Maybe add other docs once they have an initial translation.

On the Danger rule: awesome! Would never have thought about that. Let's do it!

BasThomas:
@SD10 Done: #1357

SD10:
@BasThomas I had to step away for a bit but can create that issue later unless you’d like to first.

Re: Danger - instead of having it remind us, if it had the Issue # we were using for tracking. Couldn’t the bot just reference it? It’s still probably better for us to organize that information 

BasThomas:
Danger: fancy! I think at this point any of those will do. Let's not put in too much effort for now, we can always expand on it later.

PR #1356, opened 2017-10-08
Add Moya 9.0 -> 10.0 migration guide

### LINKED ISSUE #1354: Create a Moya 9.0 -> 10.0 Migration Guide
Moya `10.0` will have significantly less breaking changes as compared to the `8.0` -> `9.0` migration. However, this is relatively easy to write up and nice to have. I'm going to start working on this later this evening, just creating an issue in case anyone had similar plans.

SD10:
This resolves #1354. It was really easy so I just took care of it quickly. 

I suggest we leave old migration guides `(8.0 -> 9.0)` because I just responded to an issue where someone was migrating from Moya 8.0 as of yesterday. You can never have too much documentation 😃

### DIFF
--- a/docs/MigrationGuides/migration_9_to_10.md
+++ b/docs/MigrationGuides/migration_9_to_10.md
@@ -1,0 +1,21 @@
+# Migration Guide
+
+This project follows [Semantic Versioning](http://semver.org).
+
+Please follow the appropriate guide below when **upgrading to a new major version** of Moya (e.g. 9.0 -> 10.0).
+
+## Upgrade from 9.x to 10.x
+
+### MoyaError Migration
+- Add `.parameterEncoding`, `.objectMapping`, `.encodableMapping`, cases to `MoyaError` or a `default` case to achieve exhaustiveness.
+
+### Task Migration
+- Add `.requestJSONEncodable` case to `Task` or default case to achieve exhaustiveness.
+
+### Endpoint Migration
+- Replace previously default `method` parameter for `Endpoint.init` with `.get`.
+- Replace previously default `httpHeaderFields` parameter for `Endpoint.init` with `nil`.
+- Replace `urlRequest` with `try? urlRequest()` or use `do/catch` syntax to handle thrown errors.
+
+### NetworkActivityPlugin Migration
+- Add `TargetType` as second argument of `NetworkActivityClosure` in `NetworkActivityPlugin` initializer.
--- a/docs/MigrationGuides/migration_9_to_10.md
+++ b/docs/MigrationGuides/migration_9_to_10.md
@@ -1,10 +1,6 @@
-# Migration Guide
+# Migration Guide from 9.x to 10.x
 
 This project follows [Semantic Versioning](http://semver.org).
-
-Please follow the appropriate guide below when **upgrading to a new major version** of Moya (e.g. 9.0 -> 10.0).
-
-## Upgrade from 9.x to 10.x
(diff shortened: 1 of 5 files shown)


BasThomas:
Although I don't mind leaving old migration guides, I'd like to do either one of two things:

- Create a folder "Migration Guides" and have separate documents per migration.
- Remove old migration guides (as we can point to them anyway in another branch - it's git!)

SD10:
@BasThomas I was kind of treating the `MigrationGuides.md` like we do a `CHANGELOG.md`. I wouldn't mind separating these documents and creating a folder for it.

However, I don't think we should remove them or point to them in other branches/earlier commits. I think that migration guides are relevant for several major versions, possibly up to a years time 😕

BasThomas:
Offtopic, but I *really* don't get these Codecov reports. 😐

SD10:
@BasThomas I created a folder to host all the migration guides. Let me know what you think

docs/MigrationGuides/migration_9_to_10.md
```diff
@@ -0,0 +1,21 @@
+# Migration Guide
+
+This project follows [Semantic Versioning](http://semver.org).
+
+Please follow the appropriate guide below when **upgrading to a new major version** of Moya (e.g. 9.0 -> 10.0).
```
sunshinejr: This line is probably not needed now as we have only one migration guide per file.  Maybe also let's merge the titles  `Migration Guide` and `Upgrade from 9.x to 10.x` to something like `Migration guide from 9.x to 10.x`? What do you think?

docs/MigrationGuides/migration_9_to_10.md
```diff
@@ -0,0 +1,21 @@
...
+### MoyaError Migration
+- Add `.parameterEncoding`, `.objectMapping`, `.encodableMapping`, cases to `MoyaError` or a `default` case to achieve exhaustiveness.
+
+### Task Migration
+- Add `.requestJSONEncodable` case to `Task` or default case to achieve exhaustiveness.
+
+### Endpoint Migration
+- Replace previously default `method` parameter for `Endpoint.init` with `.get`.
+- Replace previously default `httpHeaderFields` parameter for `Endpoint.init` with `nil`.
+- Replace `urlRequest` with `try? urlRequest()` or use `do/catch` syntax to handle thrown errors.
```
sunshinejr: We need context for `urlRequest` here (mention somewhere that this is used on `endpoint` or mostly in `endpointClosure`?

SD10: I have it listed under `Endpoint Migration` so I was hoping that would provide enough context 🤔

docs/MigrationGuides/migration_9_to_10.md
SD10: Yeah I agree 👍 I'll make these changes

docs/MigrationGuides/migration_9_to_10.md
sunshinejr: Oh, I didn't get it then, not so obvious - we can make it similar to what we have in `Changelog`, duplicate the name in the bullet points maybe?

SD10: I agree. I will make this more clear

SD10:
@sunshinejr I made the discussed changes. Also, let me know if the `Endpoint` entry is clear now.

PR #1358, opened 2017-10-09
Refactored Decodable parsing

0111b:
Relates to the #1335 and #1118
This PR aims to address some inefficiencies I noticed in the `Decodable` protocol support. In the current implementation if the `keyPath` param is not nill then data is serialized to object and after that back to `Data`. Proposed changes traverse through given `keyPath` and try to find a suitable container to parse the object. In case of failure standard decodable error will be thrown.


SD10:
Hey @0111b, thank you for submitting this PR. I was able to review it and now I understand what you mean. I just wish the solution didn't have to be so heavy-handed 🤔 

It was my hopes that Moya really didn't have to manage anything related to `Decodable` outside of the `Moya.Response` class as this is not the libraries core focus. I'll wait for some more qualified Moya contributors to shed light on this issue

BasThomas:
I am not sure exactly what this PR addresses / how impactful it is, but just as an FYI, this is from the [Vision](https://github.com/Moya/Moya/blob/master/Vision.md#project-goals):

> provide basics of response decoding while exposing customization points for other libraries to help.

devxoul:
It looks interesting. I never knew that there was such a `nestedContainer`. Actually the key idea of this implementation is really simple: "to create an intermediate model". It may improve the performance of large json response.

Sources/Moya/JSONDecoder+KeyPath.swift
```diff
@@ -0,0 +1,85 @@
...
+/// Object which is representing the value
+private final class KeyPathWrapper<T: Decodable>: Decodable {
+
+    enum KeyPathError: Error {
+        case `internal`
+    }
+
+    /// Naive coding key implementation
+    struct Key: CodingKey {
+        init?(intValue: Int) {
```
devxoul: Why do we need a key for integer?

0111b: It is requirements of the `CodingKey` protocol:
> The value to use in an integer-indexed collection (e.g. an int-keyed dictionary).
**Required.**

Sources/Moya/Response.swift
```diff
@@ -122,21 +122,13 @@ public extension Response {
...
-            }
-            do {
-                jsonData = try JSONSerialization.data(withJSONObject: jsonDictionary, options: [])
-            } catch {
-                throw MoyaError.jsonMapping(self)
-            }
-        } else {
-            jsonData = data
-        }
+        let jsonData = data
```
jane-austin: why this assignment?

0111b: stayed from the original code, describes data intention. Agreed that it can be removed

SD10:
@Moya/contributors Can we get a few extra eyes on this pull request? To summarize things, this addition would eliminate a roundtrip from `data -> object -> data` when decoding using a keyPath.

While I don't believe it's Moya's intention to specialize in response decoding, I feel it hard to ignore a potential performance win. We also already have a simple `struct` wrapper around `Encodable`.

sunshinejr:
Hey guys, sorry for the late response. I was actually looking at this PR few times because I didn't really know what to think about it. 

First of all we have really great amount of work by @0111b that optimizes decoding the object whenever there is a `keyPath` involved - that's great! In case of a big list of items to parse/map using `keypath`, it would be beneficial for sure. 

On the other hand we have kinda big refactor around new `Decodable` methods - that surely makes our codebase a little bit more complicated because it isn't clear on first sight as of _why_ do we need such wrapper.

However, I'm up for the change once we see that this is a visible improvement. We can improve the code a little bit to make it more readable to other users. Let's make some performance tests on bigger/smaller objects first, just to see how it relates to each other. This way whenever we have some sort of improvements in this part of the code we can see if the performance is still there.

How do you feel about this guys?

BasThomas:
I felt exactly the same as you did @sunshinejr, so thanks for biting the bullet and posting something. I do feel like adding performance tests would be a great improvement. Also, if we can make it more clear *why* we need this and we can accomplish that with a few code comments, I’d greatly appreciate that as well.

Sources/Moya/JSONDecoder+KeyPath.swift
```diff
@@ -0,0 +1,85 @@
...
+            return try objectContainer(for: Array(keyPath.dropFirst()), in: container, key: key)
+        }
+
+        let rootKey = try getKey(from: keyPath)
+        let rooTContainer = try decoder.container(keyedBy: Key.self)
+        let (keyedContainer, key) = try objectContainer(for: Array(keyPath.dropFirst()), in: rooTContainer, key: rootKey)
+        object = try keyedContainer.decode(T.self, forKey: key)
+    }
+
+    let object: T
```
ashfurrow: Let's put this up under the enum struct and typealias, above the initializers, to be consistent with the rest of the codebase.

Sources/Moya/JSONDecoder+KeyPath.swift
```diff
@@ -0,0 +1,85 @@
...
+                throw KeyPathError.internal // Should never happen
+        }
+
+        /// Creates a `Key` from the first keypath element
+        func getKey(from keyPath: [String]) throws -> Key {
+            guard let first = keyPath.first,
+                let key = Key(stringValue: first)
+                else {
+                    throw KeyPathError.internal // Should never happen
+            }
```
ashfurrow: Can you clean up indentation here?

Sources/Moya/JSONDecoder+KeyPath.swift
```diff
@@ -0,0 +1,85 @@
...
+                   from data: Data,
+                   keyPath: String,
+                   keyPathSeparator separator: String = ".") throws -> T where T: Decodable {
+        userInfo[keyPathUserInfoKey] = keyPath.components(separatedBy: separator)
+        return try decode(KeyPathWrapper<T>.self, from: data).object
+    }
+}
+
+/// The keypath key in the `userInfo`
+private let keyPathUserInfoKey = CodingUserInfoKey(rawValue: "keyPathUserInfoKey")! //swiftlint:disable:this force_unwrapping
```
ashfurrow: We should have an explanation in comments for why the force-unwrap always works.

0111b:
Thank you all for the review. As far of now I see these action items:
1. Add more comments to the source to describe solved problem and implementation details
2. Cover new code with unit tests
3. Implement some performance  tests

Will work on this on the weekend. Is there any guidance for writing tests?

@ashfurrow actually I do not test yet on big responses. It is preliminary update.

SD10:
Thank you to everyone who jumped in and gave input. As someone who is guilty of premature optimization himself, I agree that performance tests are the most logical path forward.

@0111b I would take a look at how long the response decoding on several different sized responses using a measure block in `XCTest`. I would also look at CPU usage with the Time Profiler in instruments. This is also something I can do this weekend.

ashfurrow:
@0111b Awesome, I am looking forward to learning what you find out. Thanks again!

PR #1359, opened 2017-10-09
Improve changelog readability

BasThomas:
Depends on https://github.com/Moya/contributors/pull/18

PR #1365, opened 2017-10-11
update files in  docs_CN and README_CN.md

lovesomeone:
The final version of this update about chinese translation


BasThomas:
Thanks so much for doing this, @lovesomeone! 😀

sunshinejr:
Hey @lovesomeone, thank you for doing this! Really good job 🎉  

Also, for the next PRs, we usually want some accepted reviews before merging. Nothing bad happened, but another pair of eyes is always good. Also, people sometimes comment but don't really have time right now to check the PR (like I assume @BasThomas did), so it is better to wait for the green "accepted PR" thingy like in #1359. Nonetheless, thanks again 🎉

lovesomeone:
OK. I get it  .  Thank you ！

PR #1367, opened 2017-10-11
Add documentation for Response

### LINKED ISSUE #1352: Add example for Moya.Response extensions
After #1335 we should have some examples showing Moya users how to use the `Decodable` protocol with Moya. 

It would be nice to demonstrate some of Moya's `Response` handling extensions in general. I'm not sure if there's a greater demonstration of this but all I could find was [this example](https://github.com/Moya/Moya/blob/master/docs/Examples/Basic.md#see-more-at-httpsgithubcommoyamoyablobmastersourcesmoyaresponseswift).

I think this could be improved by adding a `Response.md` document.

bjarkehs:
Fixes #1352

Is this something along the lines of what was wanted? This should maybe be extended to include the different filters. But this is at least a good first stab at it.


docs/Examples/Response.md
```diff
@@ -0,0 +1,118 @@
+# Response Handling
+
+You've now setup your TargetType correctly, and you've come to the fun part of implementing calls to your API.
+But one thing is not quite clear: How do you actually use the response from the `MoyaProvider`?
```
BasThomas: I think its more readable and easier to understand this document when you are to the point stating the facts; not asking questions / filling in the user's thoughts. I've not seen this in other examples as well, so it feels a bit out of place.

docs/Examples/Response.md
```diff
@@ -0,0 +1,118 @@
...
+        // do something in your app
+    case let .failure(error):
+        // TODO: handle the error == best. comment. ever.
+    }
+}
+```
+
+## Decodable
+
+This is all good, but how do we actually use this to map to our own objects?
```
BasThomas: Same as above.

docs/Examples/Response.md
```diff
@@ -0,0 +1,118 @@
...
+
+Swift 4 introduces `Decodable`, meaning now our objects can easily be decoded from JSON.
+But how do we actually do that?
```
BasThomas: Can be omitted.

docs/Examples/Response.md
```diff
@@ -0,0 +1,118 @@
...
+struct User: Decodable {
+    let id: String
+    let firstName: String
+    let lastName: String
+}
+```
+
+Now, how do we get our response mapped to our `User` type?
+Luckily Moya provides several nice extensions to the `Response` object.
+One of which is `map(_:, atKeyPath:, using:)`.
```
BasThomas: It felt weird reading the example using only `.map(User.self)`. Maybe just omit this line or tell the user that we use a default `keyPath` and `decoder`.

docs/Examples/Response.md
```diff
@@ -0,0 +1,118 @@
...
+
+An basic example usage would then be:
```
BasThomas: A*

docs/Examples/Response.md
```diff
@@ -0,0 +1,118 @@
...
+
+```swift
+provider.request(.user('jp')) { result in
+    switch result {
+    case let .success(moyaResponse):
+        let filteredResponse = moyaResponse.filterSuccessfulStatusCodes() // We want the status code to be within 200...299
```
BasThomas: Unrelated, but now that I am looking at this... I don't like the `filterSuccessfulStatusCodes()` function name. Seems like it would return status codes or something, not responses. Anyone feeling the same way?

docs/Examples/Response.md
```diff
@@ -0,0 +1,118 @@
...
+        let user = filteredResponse.map(User.self) // The user object is automatically decoded using a default decoder
```
BasThomas: Maybe specify the `user` constant is now of type `User`?

docs/Examples/Response.md
```diff
@@ -0,0 +1,118 @@
...
+Now our example has slightly more going on:
+
+```swift
+provider.request(.user) { result in
+    switch result {
+    case let .success(moyaResponse):
+        let filteredResponse = moyaResponse.filterSuccessfulStatusCodes() // We want the status code to be within 200...299
+        let decoder = JSONDecoder()
+        decoder.dateDecodingStrategy = .secondsSince1970
+        let users = filteredResponse.map([User].self, atKeyPath: "list", using: decoder)
```
BasThomas: Shouldn't `keyPath` be `"users"` here?

SD10:
@bjarkehs This is awesome! Thanks for taking the first steps to get this going. It's hard for me to criticize because we have no documentation on this topic but I do have a few suggestions:

- I think we should talk about Moya's use of the `Result` type and how the `.success` case contains a `Response` and the `.failure` case contains a `MoyaError`.
- I think less of the document should be focused on explaining `Codable/Decodable` to the user and more focused on how this specifically relates to Moya.
- I would like to see more on response status code filtering and other extensions
- It may be good to discuss the properties of `Response` in general

Sorry if I'm throwing a lot at you or I'm being unclear. I'll be able to give better feedback later. This is just at first glance. Really appreciate what you have done already 💯

docs/Examples/Response.md
BasThomas: Also, how does the decoder "magically" infer `"updated"` is supposed to be a `Date` here? Does it map the `Int` to `Date` because of `updated`'s type being `Date` in our `User` model? That's awesome!

SD10:
I added a comment but it was showing as being posted twice. I deleted one and now both are gone after a refresh. Thanks GitHub 😐

BasThomas:
Lol. Was trying to thumb it up and the picker just no-opped. Did you lose what you wrote? 😕

SD10:
@bjarkehs This is awesome 💯! I'll be able to do a better review later but these are my thoughts at first glance:

- I think we should talk about Moya's use of the `Result` type and how the `.success` case contains a `Response` and the `.failure` case contains a `MoyaError`.
- I think less of the document should be focused on explaining `Codable/Decodable` to the user and more focused on how this specifically relates to Moya.
- I would like to see more on response status code filtering and other extensions
- It may be good to discuss the properties of `Response` in general

Feel free to make any changes around these points. I totally understand if I'm throwing you a lot. We can always add more to this document over time.

docs/Examples/Response.md
bjarkehs: Yeah it can. And yes it should've been `"users"`. I changed the path sorta last minute.

Here's something to try in playground:

```swift
import UIKit

struct User: Decodable {
    let firstName: String
    let lastName: String
    let updated: Date
}

let json = """
    { "firstName": "James", "lastName": "Potter", "updated": 1507709926 }
    """.data(using: .utf8)!

let decoder = JSONDecoder()
decoder.dateDecodingStrategy = .secondsSince1970

let user = try! decoder.decode(User.self, from: json)
print(user)
```

bjarkehs:
Thanks guys. I'll take a look at all the feedback and make a new and hopefully improved version later today 👍

docs/Examples/Response.md
bjarkehs: I think it's because I read the first line of [Basic.md](https://github.com/Moya/Moya/blob/master/docs/Examples/Basic.md) and then basically assumed that's how it's written. I can definitely be more factual. I'll have a look at how some of the other documentation is worded.

bjarkehs:
I have now reworded it and added more about other extensions such as filtering and mapJSON. I don't know if it makes sense to write about mapString and mapImage. Maybe mention them? I am just worried that the document can very easily become out of sync.

sunshinejr:
@SD10 @BasThomas would you be up for checking this one again after the updates? 😉

bjarkehs:
I’m glad you like it 👍

PR #1370, opened 2017-10-11
Fix SPM instructions for Moya 10.0.0-beta.1

sunshinejr:
Basically, we don't have to write this long Version initializer - `String` literal works as well! Also when you specify that you want `ReactiveMoya` as a dependency, it has `Moya` as a dependency itself so no need to dupe.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -95,12 +95,12 @@ let package = Package(
             targets: ["MyPackage"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/Moya/Moya.git", .exact(Version(10, 0, 0, prereleaseIdentifiers: ["beta", "1"])))
+        .package(url: "https://github.com/Moya/Moya.git", .exact("10.0.0-beta.1")
     ],
     targets: [
         .target(
             name: "MyPackage",
-            dependencies: ["Moya", "ReactiveMoya"])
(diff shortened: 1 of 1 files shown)


Readme.md
```diff
@@ -95,12 +95,12 @@ let package = Package(
             targets: ["MyPackage"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/Moya/Moya.git", .exact(Version(10, 0, 0, prereleaseIdentifiers: ["beta", "1"])))
+        .package(url: "https://github.com/Moya/Moya.git", .exact("10.0.0-beta.1")
```
AndrewSB: does it need to be exact, or can we just specify version 10, and it will choose the latest published beta/stable?

sunshinejr: Unfortunately in this case the resolver would say that the graph can't be resolved because he needs version starting from 10.0.0 :(

BasThomas:
Can this one not be improved? https://github.com/Moya/Moya/pull/1370/files#diff-1e290ac8433d555bce009b162cb869d0R79

sunshinejr:
@BasThomas 🤦🏼‍♂️ you are right, it could, sorry for not noticing 😄 probably Chinese docs want to be updated as well. I'm gonna tackle it in a sec.

PR #1374, opened 2017-10-12
Remove ReactiveCocoa subspec

BasThomas:
This has been deprecated for quite some time. Only thing is that we can't really let the user know that it has been deprecated. If we do not feel like removing it, feel free to close this PR.


sunshinejr:
Good idea @BasThomas. I think this is the right call. Enough time to migrate. I would only ask if we could add this to the migration guide? To remind that if somebody didn't migrate, they should, _now_.

PR #1377, opened 2017-10-16
Test out danger swift on a mac CI build

orta:
Circle CI has always been a dodgy case with Danger ( they don't reliably set CI env vars ) so it's hard to tell whether this will work, but now  that https://github.com/danger/danger-swift/pull/27 is in, this should allow libs to use Danger Swift.

### DIFF
--- a/Dangerfile.swift
+++ b/Dangerfile.swift
@@ -1,0 +1,8 @@
+import Danger
+
+let danger = Danger()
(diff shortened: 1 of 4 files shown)


sunshinejr:
@orta closing this one for now, but we should definitely look into integrating danger swift in Moya in the future 👍

PR #1378, opened 2017-10-17
tighten up our stale rules to 2 weeks #fortnight

AndrewSB:
I think we're doing a good job at dealing with issues and even better than our stalebot. I think leaning on our bot is a good idea, and I can see us changing this to just a week some time in the future.

2 weeks is a good target for now (+1 week for it to actually close)

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -25,7 +25,7 @@ unmarkComment: false
 
 # Comment to post when closing a stale Issue or Pull Request. Set to `false` to disable
 closeComment: >
-  This issue has been auto-closed because there hasn't been any activity for at least 37 days.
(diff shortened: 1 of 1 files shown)


BasThomas:
What are you aiming at here? I feel like the 30 days work fine too. What do you want to encourage / prevent?

AndrewSB:
Resolving our issues faster instead of allowing long issue sit times.

Open to a discussion if that isn't universally accepted as a good idea!

SD10:
I'm not opposed to this but we did just decrease it from 59 days to 30 days... Now to 14 😅

sunshinejr:
I don't mind - I see that our bot is doing fine now, but maybe shortening will help us even better. Whatever we decide to do, I'm fine with it 👍

AndrewSB:
😄

PR #1382, opened 2017-10-18
Fix a bug that decodable mapping won't decode nested json array

### LINKED ISSUE #1381: The response map to decodable array always failed.
See this line of code from branch 10.0.0-dev:
https://github.com/Moya/Moya/blob/a951eeab47ad7f4ef780b63e085704b576bbc3be/Sources/Moya/Response.swift#L127
If the response data at given key path is an array, the type cast to `Dictionary<String, Any>` will always fail, and the result is throw an error.

devxoul:
Fixes #1381

### DIFF
--- a/Sources/Moya/Response.swift
+++ b/Sources/Moya/Response.swift
@@ -124,11 +124,11 @@ public extension Response {
     func map<D: Decodable>(_ type: D.Type, atKeyPath keyPath: String? = nil, using decoder: JSONDecoder = JSONDecoder()) throws -> D {
         let jsonData: Data
         if let keyPath = keyPath {
-            guard let jsonDictionary = (try mapJSON() as? NSDictionary)?.value(forKeyPath: keyPath) as? [String: Any] else {
+            guard let jsonObject = (try mapJSON() as? NSDictionary)?.value(forKeyPath: keyPath) else {
                 throw MoyaError.jsonMapping(self)
             }
             do {
-                jsonData = try JSONSerialization.data(withJSONObject: jsonDictionary, options: [])
+                jsonData = try JSONSerialization.data(withJSONObject: jsonObject)
             } catch {
                 throw MoyaError.jsonMapping(self)
             }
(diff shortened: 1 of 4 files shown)


Sources/Moya/Response.swift
```diff
@@ -124,11 +124,11 @@ public extension Response {
     func map<D: Decodable>(_ type: D.Type, atKeyPath keyPath: String? = nil, using decoder: JSONDecoder = JSONDecoder()) throws -> D {
         let jsonData: Data
         if let keyPath = keyPath {
-            guard let jsonDictionary = (try mapJSON() as? NSDictionary)?.value(forKeyPath: keyPath) as? [String: Any] else {
+            guard let jsonObject = (try mapJSON() as? NSDictionary)?.value(forKeyPath: keyPath) else {
                 throw MoyaError.jsonMapping(self)
             }
             do {
-                jsonData = try JSONSerialization.data(withJSONObject: jsonDictionary, options: [])
+                jsonData = try JSONSerialization.data(withJSONObject: jsonObject, options: [])
```
BasThomas: Nit: you can omit the `options` parameter - it has a default value of `[]`.

Tests/Observable+MoyaSpec.swift
```diff
@@ -326,6 +326,23 @@ class ObservableMoyaSpec: QuickSpec {
                 expect(receivedObject?.createdAt) == formatter.date(from: "1995-01-14T12:34:56")!
             }
 
+            it("maps data representing a json array at a key path to a decodable object") {
```
AndrewSB: nit: it would be nice to mark tests we add from bug reports with the bug report that found it. So adding something like bug from Moya/Moya#1311 would be awesome

Sources/Moya/Response.swift
devxoul: Oh didn't know that. It would be great!

Tests/Observable+MoyaSpec.swift
devxoul: That's a great idea.

sunshinejr:
@devxoul would you rebase it so we can merge this one in? 😉

devxoul:
@sunshinejr Sure!

SD10:
SD10 Bot Has Commented 😂:

`❌  fatal error: URLEncoding is disallowed as bodyEncoding.: file /Users/distiller/Moya/Sources/Moya/Endpoint.swift, line 95`

PR #1385, opened 2017-10-19
Update RxSwift dependency to 4.0.0

devxoul:
Sets RxSwift dependency to `~> 4.0`.


Moya.podspec
```diff
@@ -36,6 +36,6 @@ Pod::Spec.new do |s|
   s.subspec "RxSwift" do |ss|
     ss.source_files = "Sources/RxMoya/"
     ss.dependency "Moya/Core"
-    ss.dependency "RxSwift", "4.0.0-rc.0"
+    ss.dependency "RxSwift", ">= 4.0.0"
```
SD10: Do we want `>=` or just `4.0.0`?

BasThomas: I would say `~> 4.0` for both CocoaPods and Carthage, like we are dong with the other dependencies.

devxoul: Ohh that's right. It is not a beta anymore. I just updated to use `~>` :)

sunshinejr:
@devxoul Thanks! We should also update `Package.swift` I believe.

devxoul:
@sunshinejr sure! Did it :)

Issue #1380, opened 2017-10-18
Weak self in Rx extension causes dispose of the request


MrAsterisco:
I'm using _Moya 10.0.0-beta.1_ together with _RxSwift 4.0.0-rc.0_ and I've found a weird issue when making a request.

I have the following code:

```swift
MoyaProvider<SomeService>().rx.request(.all)
                .subscribe({ (event) in
                    switch event {
                    case let .success(response):
                        print(response)
                        
                    case let .error(response):
                        print(response)
                    }
                })
                .disposed(by: self.disposable)
```

Using the _out-of-the-box_ install with CocoaPods, this request is never fulfilled. The subscribe event is logged (with RxSwift `debug()` method), but no response is received… ever.

I have managed to identify the issue in `MoyaProvider+Rx.swift` in the following function:

```swift
internal func rxRequest(_ token: Target, callbackQueue: DispatchQueue? = nil) -> Single<Response> {
        return Single.create { [weak self] single in
            let cancellableToken = self?.request(token, callbackQueue: callbackQueue, progress: nil) { result in
                switch result {
                case let .success(response):
                    single(.success(response))
                case let .failure(error):
                    single(.error(error))
                }
            }

            return Disposables.create {
                cancellableToken?.cancel()
            }
        }
    }
```

The issue is solved by just removing the `weak self` reference, hence making `self` a strong reference to the object.

```swift
internal func rxRequest(_ token: Target, callbackQueue: DispatchQueue? = nil) -> Single<Response> {
        return Single.create { single in
            let cancellableToken = self.request(token, callbackQueue: callbackQueue, progress: nil) { result in
                switch result {
                case let .success(response):
                    single(.success(response))
                case let .failure(error):
                    single(.error(error))
                }
            }

            return Disposables.create {
                cancellableToken.cancel()
            }
        }
    }
```

I don't know if this is an actual issue with Moya itself or it depends on something from RxSwift, but I couldn't find any way to reproduce the same behaviour with just RxSwift.

Can someone explain me why?

sunshinejr:
Hey @MrAsterisco. This is kind of a duplicate of #1294. TL;DR is that we don't support this way of executing the request - you need to retain your provider.

MrAsterisco:
I'm really sorry for the duplicate: I didn't see that issue. Ok, closing this one.

sunshinejr:
@MrAsterisco don't worry, we probably need to document it better. Would you be interested in improving our existing documentation? :)

[...]

PR #1386, opened 2017-10-19
[WIP] Update RxSwift.md

MrAsterisco:
Add a note to better explain the need to manually retain the MoyaProvider.

This pull request attempts to prevent new issues with the same question as #1380 .

### DIFF
--- a/docs/RxSwift.md
+++ b/docs/RxSwift.md
@@ -12,8 +12,6 @@ Just use your `MoyaProvider` instance.
 let provider = MoyaProvider<GitHub>()
 ```
 
-*You are responsible to retain the provider for as long as you need. If you fail to do so, the object will be automatically disposed and the observable will not work properly.*
-
 After that simple setup, you're off to the races:
 
 ```swift
--- a/docs/RxSwift.md
+++ b/docs/RxSwift.md
@@ -12,6 +12,8 @@ Just use your `MoyaProvider` instance.
 let provider = MoyaProvider<GitHub>()
 ```
 
+*You are responsible to retain the provider for as long as you need. If you fail to do so, the object will be automatically disposed and the observable will not work properly.*
+
 After that simple setup, you're off to the races:
 
 ```swift
--- a/docs/Examples/Basic.md
+++ b/docs/Examples/Basic.md
@@ -114,6 +114,8 @@ provider.request(.updateUser(id: 123, firstName: "Harry", lastName: "Potter")) {
 // POST https://api.myservice.com/users/123?first_name=Harry&last_name=Potter
 ```
 
+Always remember to retain the provider somewhere: if you fail to do so, it will be released automatically, potentially before you receive any response.
+
 The `TargetType` specifies both a base URL for the API and the sample data for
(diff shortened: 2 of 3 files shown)


docs/RxSwift.md
```diff
@@ -12,6 +12,8 @@ Just use your `MoyaProvider` instance.
 let provider = MoyaProvider<GitHub>()
 ```
 
+*You are responsible to retain the provider for as long as you need. If you fail to do so, the object will be automatically disposed and the observable will not work properly.*
```
sunshinejr: Hm, this is more of a general rule for all of the configurations: `Moya`, `Moya/RxSwift`, `Moya/ReactiveSwift`. Can we move it somewhere to e.g. Basic usage? Or maybe we should duplicate it in all the docs to make sure everyone sees it? 🤔

SD10: I agree. It would probably be good in `Providers.md`, `RxSwift.md`, `ReactiveSwift.md`. If you want to add basic too that's fine 😅

sunshinejr: Oh, now that I think about it duplicating it in 4 places might not be the best idea. Maybe just cover it in `Basic.md` and  `Providers.md`?

BasThomas: Yep, I think `Basic.md` would be good. Pretty nice to at least have read when you start using Moya, as the Xcode errors might not lead you to a fix immediately.

SD10: Yeah but this keeps coming up mainly on the reactive providers? Or is it just because most people use the reactive Moya libraries?

sunshinejr: I think the vast majority of people using Moya is using RxSwift extensions. At least it was the case few months back. When you start with Moya, whether your use reactive extensions or not, you should read the basics as well. I'm thinking if providers also should have it, though. 🤔

SD10: Ok sorry to bikeshed the issue 😅 I think `Providers.md` and `Basic.md` will be clear enough

sunshinejr: Okay cool, I agree 👍

sunshinejr: Okay cool, I agree 👍 Sorry for the delay in this one, @MrAsterisco, we got an agreement now 😄

MrAsterisco: Ok, so just to confirm: we need to move the very same sentence to both `Providers.md` and `Basic.md`, right?

sunshinejr: Not exactly the very same sentence, because we have `Observable` and `disposing` here. We would need to adopt it to a more broader use (if you don't retain the provider, it (probably) will be released before the response comes back). But yes, remove it from `RxSwift.md` and move to `Providers.md` and `Basic.md` 👍

MrAsterisco:
Ok, I've tried to adapt the sentence to the other two files and removed it from the previous one. Actually, in `Providers.md` there was already a phrase telling about the same thing, but I've improved it to make it more explicit.

Let me know what you think!

sunshinejr:
Thank you for doing this, @MrAsterisco! :)

PR #1391, opened 2017-10-21
Add Changelog entry for decodable fix from 10.0.0-beta.1

sunshinejr:
In #1382 @devxoul fixed quite big bug so we should let people know about it (I already added this one to the Releases section)


devxoul:
Oh, thanks!

Issue #1283, opened 2017-09-13
Allowing parameters and headers to be reused across providers


SD10:
### Background
This is inspired by issue #1273 and its matching pull request #1279.

Basically, #1273 discusses extending the behavior of Moya's current `AccessTokenPlugin` to support adding an `access_token` as a URL parameter to a request. 

While this parameter *can* be added directly to the parameters of a `TargetType`, the fact that it needs to be applied to every request makes it a good candidate for improving the existing `AccessTokenPlugin`.

### AccessTokenPlugin Current Behavior
Currently, the `AccessTokenPlugin` only supports authentication using an `"Authorization"` header with the values of `"Basic"` or `"Bearer"`. Supporting auth via parameters would be a new concept for this plugin.

### Proposed Solutions

1.) Adding another case to the `AuthorizationType` enum:
```Swift
public enum AuthorizationType: String {
    case basic = "Basic"
    case bearer = "Bearer"
    case urlParamter = "access_token"
    case none
}
```
This raises the question: What if the API has different parameter names? Ex) `api_key` or `api_token`

2.) Using associated values over raw values:
```Swift
public enum AuthorizationType {
    case header(name: String)
    case urlParameter(name: String)
    case none
}
```

3. Specifying the parameter name or header name in the `AccessTokenPlugin` initializer:
```Swift
init(tokenClosure: ..., parameterName: ..., headerName: ...) { ... }

enum AuthorizationType {
    case basicInHeaders
    case bearerInHeaders
    case basicInParameters
    case bearerInParameters
    case none
}
```

4. Splitting option 3 into two enums and requesting optional `AuthorizationType` in the target, where nil will mean no authorization will be appended:
```swift
enum AuthorizationType {
    case basic
    case bearer
}

enum AuthorizationTypePlacement {
    case header
    case parameter
}
```

5. Providing a `HeaderPlugin` and `ParameterPlugin`:
```Swift
protocol HeaderApplicable {
    var applyPluginHeaders: Bool
}
public struct HeaderPlugin: PluginType {
    var headerValue: String
    var headerName: String
}
protocol ParameterApplicable {
    var applyPluginParameters: Bool
}
public struct ParameterPlugin: PluginType {
   var parameterValue: Any
   var parameterName: String
   var encoding: ParameterEncoding
}
```
Approach 5 moves away from the domain of `Authentication` and looks at the problem as being we need to reuse a specific parameter or specific header on requests across multiple providers or targets.

Sorry for the essay 😅

[...]

PR #1393, opened 2017-10-22
Add .custom option to AuthorizationType of AccessTokenPlugin

ffittschen:
#### Description
This PR adds a `.custom` option to the `AuthorizationType` of the `AccessTokenPlugin`, which enables to specify a custom header field and a custom token prefix.

```swift
// Examples:
.custom(headerField: nil, tokenPrefix: "JWT")
// will result in this header 'Authorization: JWT <token>'

.custom(headerField: "X-Access-Token", tokenPrefix: nil)
// will result in this header 'X-Access-Token: <token>'
```

#### Motivation
I recently used Moya in a project of mine and the backend it interacts with requires me to use `x-access-token` as header field and it also requires that only the token value is set as value in that header. I was about to create a `CustomAccessTokenPlugin` in my project, but I thought that there might be other developers who are in a similar situation, so I created this PR instead.

Please let me know what you think and if you require any changes :)

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,4 +1,6 @@
 # Next
+### Added
+- Added a `.custom` case to `AuthorizationType`. [#1393](https://github.com/Moya/Moya/pull/1393) by [ffittschen](https://github.com/ffittschen).
 
 # [10.0.0] - 2017-10-21
 ### Fixed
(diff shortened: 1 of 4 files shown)


SD10:
This is related to #1283

sunshinejr:
Hey @ffittschen, thank you very much for taking your time to do it. However, we have quite a big discussion around this topic at #1283. Before taking any steps further we have to agree on a solution. We would love it if you could help us with your feedback and then you could recreate a PR with implementation.

For now I have to close it, unfortunately, but as I said please help us resolve it issue and we can go back to it really soon. Thanks for understanding!

PR #1399, opened 2017-10-24
Remove redundant reactive abstractions

### LINKED ISSUE #1372: Remove redundant RxSwift & ReactiveSwift request methods.
I saw this https://github.com/Moya/Moya/pull/1320#issuecomment-332476303 and thought it would be a good idea to create an issue so we don't forget to do this

sunshinejr:
Fixes #1372.  I also created new branch, `development`, that we should use for the next features/fixes before we make a release (it doesn't need to be the next major, could be a minor/patch, thus general name `development` instead of `11.0.0-dev`).

### DIFF
--- a/Sources/RxMoya/MoyaProvider+Rx.swift
+++ b/Sources/RxMoya/MoyaProvider+Rx.swift
@@ -15,20 +15,8 @@ public extension Reactive where Base: MoyaProviderType {
     ///   - callbackQueue: Callback queue. If nil - queue from provider initializer will be used.
     /// - Returns: Single response object.
     public func request(_ token: Base.Target, callbackQueue: DispatchQueue? = nil) -> Single<Response> {
-        return base.rxRequest(token, callbackQueue: callbackQueue)
-    }
-
-    /// Designated request-making method with progress.
-    public func requestWithProgress(_ token: Base.Target, callbackQueue: DispatchQueue? = nil) -> Observable<ProgressResponse> {
-        return base.rxRequestWithProgress(token, callbackQueue: callbackQueue)
-    }
-}
-
-internal extension MoyaProviderType {
-
-    internal func rxRequest(_ token: Target, callbackQueue: DispatchQueue? = nil) -> Single<Response> {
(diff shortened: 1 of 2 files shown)


Sources/RxMoya/MoyaProvider+Rx.swift
```diff
@@ -15,20 +15,8 @@ public extension Reactive where Base: MoyaProviderType {
...
-        return base.rxRequestWithProgress(token, callbackQueue: callbackQueue)
-    }
-}
-
-internal extension MoyaProviderType {
-
-    internal func rxRequest(_ token: Target, callbackQueue: DispatchQueue? = nil) -> Single<Response> {
-        return Single.create { [weak self] single in
-            let cancellableToken = self?.request(token, callbackQueue: callbackQueue, progress: nil) { result in
+        return Single.create { [weak base] single in
```
SD10: Duh 😐 I just had to change `self` to `base`. Now I feel stupid 😝

sunshinejr: Don't be! Both in `RxSwift` and `ReactiveSwift` there is this struct called `ReactiveBase` or whatever, that is nothing more than the object that hold `rx` namespace and the object itself. E.g. when you have `provider.rx`, the `base` for it is a `MoyaProvider` object, `provider`.

SD10: Yeah, I see now. `self` was referencing the `ReactiveBase` and `base` references the wrapped value. 🤦‍♂️ Thanks 👍

AndrewSB: I remember seeing this and being super confused. This is just a hack in the language to get nice namespaced Observables

sunshinejr:
Thanks for the review guys! 🐼 

@AndrewSB I created an issue, #1408, would love to hear your feedback around branches and development/release flow there. Apart from that, I'm gonna merge it to development, as we did before, but if anything bigger comes out of #1408 we can easily change it.

PR #1405, opened 2017-10-25
Fix base value types(String, Int, Bool) mapping cause crash.

ufosky:
Currently on the release version 10.0.0, The response data mapping to base value types(String, Int, Bool) at given key path will cause crash. This PR fix these cases, but cannot mapping to other types like URL, Date etc.


BasThomas:
Thanks for looking into this @ufosky! It seems like this would a breaking change, right?

BasThomas:
Also I think this should go into `master`, right @sunshinejr? When / where do we use the `development` branch even? I was surprised to see it was even with `master`. 🤔

ufosky:
@BasThomas It isn't a breaking change, since in this code if jsonObject is not an array or dictionary, it causes a crash, so it's a bug.
```
jsonData = try JSONSerialization.data(withJSONObject: jsonObject)
```

Sources/Moya/Response.swift
```diff
@@ -127,9 +127,18 @@ public extension Response {
             guard let jsonObject = (try mapJSON() as? NSDictionary)?.value(forKeyPath: keyPath) else {
                 throw MoyaError.jsonMapping(self)
             }
-            do {
-                jsonData = try JSONSerialization.data(withJSONObject: jsonObject)
-            } catch {
+            if let value = jsonObject as? D {
+                return value
+            }
+            let dict = jsonObject as? NSDictionary
```
sunshinejr: Instead of checking for `NSDictionary` or `NSArray`, can we use `isValidJSONObject` method? I feel like it will be more scalable this way. Also we can use `guard` to make the code below more user-friendly (with the check for valid JSON object)

Tests/Observable+MoyaSpec.swift
```diff
@@ -343,6 +343,108 @@ class ObservableMoyaSpec: QuickSpec {
                 expect(receivedObjects?.first?.createdAt) == formatter.date(from: "1995-01-14T12:34:56")!
             }
 
+            it("maps data representing an Int at a key path to an Int value") {
```
sunshinejr: should we add tests for mapping these primitives without `keypath`?

Tests/Observable+MoyaSpec.swift
```diff
@@ -343,6 +343,108 @@ class ObservableMoyaSpec: QuickSpec {
...
+
+                var count: Int?
+                _ = observable.map(Int.self, atKeyPath: "count", using: decoder).subscribe(onNext: { value in
+                    count = value
+                })
+                expect(count).notTo(beNil())
+                expect(count) == 1
+            }
+
+            it("maps data representing a Bool at a key path to a Bool value") {
```
sunshinejr: we should also add all of these to `Single+MoyaSpec` + `SignalProducer+MoyaSpec`

Tests/Observable+MoyaSpec.swift
```diff
@@ -343,6 +343,108 @@ class ObservableMoyaSpec: QuickSpec {
...
+
+                var description: String?
+                _ = observable.map(String.self, atKeyPath: "description", using: decoder).subscribe(onNext: { value in
+                    description = value
+                })
+                expect(description).notTo(beNil())
+                expect(description) == "Something interesting"
+            }
+
+            it("maps data representing a String at a key path should not to an Int value") {
```
sunshinejr: I'd rearrange the words for tests that "should not do something", to something similar to:
> it shouldn't map String data to an Int value, at a key path

Ideally we would add a context that says:
```
context("when using key path mapping") {
    it("shouldn't map String data to an Int value")
}
```
And then maybe we could create a `sharedExample` and do not duplicate code for tests with and without a key path.
But it isn't required right now and we could make an issue for improving this in the future :)

sunshinejr:
I also assigned you @devxoul, would love to have your input in it, but no pressure of course :)

Sources/Moya/Response.swift
SD10: Yes, I'm pretty sure when I was digging through Foundation's source, I saw that `isValidJSONObject` just performs all the checks we're attempting here.

BasThomas: Pretty much, yup: https://github.com/apple/swift-corelibs-foundation/blob/3ee0925fceb476e89b26673effb7b9f92edbc3da/Foundation/JSONSerialization.swift#L57-L126

SD10: @BasThomas I seriously love how you're always willing to go grab a reference ❤️

BasThomas: Let’s just say that
1. I was curious
2. I like context

Happy to help 😃

ufosky:
Maybe we can check the value at a key path is string ,int,bool or null,then convert it to data,this way may support URL ,Date or other Decodable types.

ufosky:
I update the code to support all Decodable types, and modify the test cases follow your advices.

Sources/Moya/Response.swift
```diff
@@ -127,10 +127,26 @@ public extension Response {
...
+                }
+            } else {
+                let wrappedJsonObject = ["value": jsonObject]
+                let wrappedJsonData: Data
+                if JSONSerialization.isValidJSONObject(wrappedJsonObject) {
+                    do {
+                        wrappedJsonData = try JSONSerialization.data(withJSONObject: wrappedJsonObject)
+                    } catch {
+                        throw MoyaError.jsonMapping(self)
+                    }
```
SD10: It looks like we're repeating this block again. Can we extract it into a closure?

Sources/Moya/Response.swift
```diff
@@ -127,10 +127,26 @@ public extension Response {
...
+            if JSONSerialization.isValidJSONObject(jsonObject) {
+                do {
+                    jsonData = try JSONSerialization.data(withJSONObject: jsonObject)
+                } catch {
+                    throw MoyaError.jsonMapping(self)
+                }
+            } else {
+                let wrappedJsonObject = ["value": jsonObject]
+                let wrappedJsonData: Data
+                if JSONSerialization.isValidJSONObject(wrappedJsonObject) {
```
SD10: It looks like we're repeating this block again. Can we extract it into a closure?

Sources/Moya/Response.swift
```diff
@@ -142,3 +158,7 @@ public extension Response {
         }
     }
 }
+
+private struct DecodableWrapper<T: Decodable>: Decodable {
```
SD10: Can you explain why we had to add this wrapper?

ufosky: Because the base value types cannot serialize to json data, so I wrap the value to a Dictionary, then serialize the Dictionary to Data. So when decode the data, the result should be unwrapped.

devxoul:
screen shot 2017-10-26 at 2 27 06 pm

Could you please write a test for `SignalProducer`?

SD10:
We need to add a CHANGELOG entry for this under `Fixed`.

For anyone who is interested this is the [intended behavior](https://github.com/apple/swift-corelibs-foundation/blob/3ee0925fceb476e89b26673effb7b9f92edbc3da/Foundation/JSONSerialization.swift#L143-L153) so crashing instead of throwing is not a bug.

Using `isValidJSONObject` will protect us from this 😃

ufosky:
@SD10 Should bump the next release version before add change log entries.

SD10:
@ufosky We add new changelog entries under the `Next Release` section without a specified version. 
You can refer to our [changelog guidelines](https://github.com/Moya/contributors/blob/master/Changelog%20Guidelines.md) if this helps 👍

ufosky:
Do we need a method likes below, it returns nil if the keyPath is not present in the response data, just like the method [decodeIfPresent(_:forKey:)](https://developer.apple.com/documentation/swift/keyeddecodingcontainerprotocol/2893945-decodeifpresent) in protocol `KeyedDecodingContainerProtocol`:
```
func mapIfPresent<D: Decodable>(_ type: D.Type, atKeyPath keyPath: String? = nil, using decoder: JSONDecoder = JSONDecoder()) throws -> D?
```

SD10:
@ufosky The current decoding method should handle a case where the key is invalid, correct? We could consider propagating a more descriptive error regarding which point in the process the decoding failed, but I don't think we need the overhead of an additional method.

sunshinejr:
@ufosky can you also rebase this one into master, please? This is a fix we would like to include in the nearest path release.

ufosky:
@sunshinejr I don't find the correct way to use rebasing. Can I cherry pick the commits into master?

SD10:
@ufosky You can always just resolve this conflict on GitHub. It's just a CHANGELOG conflict. We need to keep the both of them. It will be a little easier to rebase once you don't have to work from a fork 😅

Sources/Moya/Response.swift
```diff
@@ -122,15 +122,38 @@ public extension Response {
     /// - parameter atKeyPath: Optional key path at which to parse object.
     /// - parameter using: A `JSONDecoder` instance which is used to decode data to an object.
     func map<D: Decodable>(_ type: D.Type, atKeyPath keyPath: String? = nil, using decoder: JSONDecoder = JSONDecoder()) throws -> D {
+        let serializeToData: (Any) throws -> Data? = { (jsonObject) in
+            if JSONSerialization.isValidJSONObject(jsonObject) {
```
freak4pc: Minor nit - I would use a guard here and return early

freak4pc:
Regarding rebasing - since this is just CHANGELOG you can just copy it from master and apply your changes over it.

freak4pc:
BTW might worth making a “tryOrThrow” kind of method. Seems like that’s a lot of the complexity of that methods.

freak4pc:
@sunshinejr I would strongly suggest moving all the `let thing = try something(x) { } else { throwError }` to a separate `tryOrThrow` method similar to RxSwift, if you know that one. I think it would really clean up that code. Don't mind making the PR if you'd be interested.

sunshinejr:
@freak4pc yeah, if you have some spare time it would be awesome if you could do a PR. I was talking about the `guard` in this PR only 😄

ufosky:
Is there any `tryOrThrow` implementation in Moya?

freak4pc:
@ufosky sorry I kinda got confused with the specific castOrThrow etc. This will need a bit more though. Anyways, this PR "as it is" should be good to Go IMO. @sunshinejr care to merge?

sunshinejr:
Thank you very much for your hard work, @ufosky! 👍

SD10:
Thanks for jumping in so quickly @freak4pc @sunshinejr and of course @ufosky for the great work 👍

PR #1410, opened 2017-10-25
Added missing content type for requestJSONEncodable task

### LINKED ISSUE #1406: Content type for requestJSONEncodable
When using requestJSONEncodable task there is no "Content-Type": "application/json" header in result request. Is it intended behavior? Now it is necessary to add this header manually via headers method on TargetType.

Version of Moya: 10.0.0

Vict0rS:
Fixes #1406

### DIFF
--- a/Sources/Moya/URLRequest+Encoding.swift
+++ b/Sources/Moya/URLRequest+Encoding.swift
@@ -6,6 +6,11 @@ internal extension URLRequest {
         do {
             let encodable = AnyEncodable(encodable)
             httpBody = try JSONEncoder().encode(encodable)
+
+            if value(forHTTPHeaderField: "Content-Type") == nil {
+                setValue("application/json", forHTTPHeaderField: "Content-Type")
+            }
+
             return self
         } catch {
             throw MoyaError.encodableMapping(error)
--- a/Sources/Moya/URLRequest+Encoding.swift
+++ b/Sources/Moya/URLRequest+Encoding.swift
@@ -7,8 +7,9 @@ internal extension URLRequest {
             let encodable = AnyEncodable(encodable)
             httpBody = try JSONEncoder().encode(encodable)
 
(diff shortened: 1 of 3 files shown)


Sources/Moya/URLRequest+Encoding.swift
```diff
@@ -6,6 +6,11 @@ internal extension URLRequest {
         do {
             let encodable = AnyEncodable(encodable)
             httpBody = try JSONEncoder().encode(encodable)
+
+            if value(forHTTPHeaderField: "Content-Type") == nil {
```
BasThomas: Maybe we can add a local constant `let contentType = "Content-Type"` here and use that instead of typing it twice? To prevent regressions in the future.

Vict0rS: Sounds like a good idea to me. I pushed the change.

Vict0rS:
Changelog and tests are updated.

BasThomas:
Hey @Vict0rS - I think some other tests were impacted by this change. See [here](https://circleci.com/gh/Moya/Moya/2211?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link). Can you check if everything works fine locally?

Vict0rS:
It seems like there are some network connection inconsistencies. When I run tests locally they appear to pass or fail at random. The failing test requires to upload image in 5 seconds so definitely looks like network connection inconsistency.
screen shot 2017-10-26 at 20 11 31

Vict0rS:
I experience the same issue when running tests in version from master, so it doesn't seem like my fault. I guess we either need to increase wait time for this test or rerun tests a couple of times until they pass.

sunshinejr:
You are right @Vict0rS, both tests that were failing are not your fault, I've created an issue for this one as well (#1424).

ashfurrow:

Vict0rS:
Was happy to help :)

PR #1416, opened 2017-10-26
require extension-safe api for RxMoya in release mode as well

spookyvision:
In the RxMoya target, restriction to extension safe APIs is only required for debug, not release builds, meaning [RxMoya cannot be used for app extensions](https://developer.apple.com/library/content/documentation/General/Conceptual/ExtensibilityPG/ExtensionScenarios.html). (this should probably be fixed for ReactiveMoya as well? I'm not a ReactiveMoya user, so I cannot tell if it's safe to change)


BasThomas:
Nice catch! We should enable the flags for the remaining targets as well. Would you be up for that?

spookyvision:
ok, will do!

pedrovereza:
For the record, this seems like a regression: https://github.com/Moya/Moya/pull/1127 had done it already 😅

spookyvision:
it seems I cannot create a new PR with this one open, so closing it

BasThomas:
@pedrovereza - but not for every target in both release+debug. It's not a regression. :)

PR #1425, opened 2017-10-30
update docs_CN with MigrationGuides 

lovesomeone:
sync the docs_CN with  English docs


SD10:
Thanks for doing this @lovesomeone 👍

PR #1443, opened 2017-11-03
Add Task.requestCustomJSONEncodable

### LINKED ISSUE #1442: How to customize JSONEncoder for Task.requestJSONEncodable?
I need a way to pass custom DateFormatter to dateEncodingStrategy, but it seems now it is impossible.

Moya version is 10.0.0.

evgeny-sureev:
Fixes #1442

### DIFF
--- a/Tests/EndpointSpec.swift
+++ b/Tests/EndpointSpec.swift
@@ -175,6 +175,45 @@ final class EndpointSpec: QuickSpec {
                 }
             }
 
+            context("when task is .requestCustomJSONEncodable") {
+                var issue: Issue!
+                var request: URLRequest!
+
+                let formatter = DateFormatter()
+                formatter.locale = Locale(identifier: "en_US_POSIX")
+                formatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSSZ"
+
+                let encoder = JSONEncoder()
+                encoder.dateEncodingStrategy = .formatted(formatter)
+
+                let decoder = JSONDecoder()
+                decoder.dateDecodingStrategy = .formatted(formatter)
+
+                beforeEach {
+                    issue = Issue(title: "Hello, Moya!", createdAt: Date())
+                    endpoint = endpoint.replacing(task: .requestCustomJSONEncodable(issue, encoder: encoder))
+                    request = try! endpoint.urlRequest()
+                }
+
+                it("updates httpBody") {
+                    let expectedIssue = try! decoder.decode(Issue.self, from: request.httpBody!)
+                    expect(formatter.string(from: issue.createdAt)).to(equal(formatter.string(from: expectedIssue.createdAt)))
+                    expect(issue.title).to(equal(expectedIssue.title))
+                }
+
+                it("updates headers to include Content-Type: application/json") {
+                    let contentTypeHeaders = ["Content-Type": "application/json"]
+                    let initialHeaderFields = endpoint.httpHeaderFields ?? [:]
+                    let expectedHTTPHeaderFields = initialHeaderFields.merging(contentTypeHeaders) { initialValue, _ in initialValue }
+                    expect(request.allHTTPHeaderFields).to(equal(expectedHTTPHeaderFields))
+                }
+
+                it("doesn't update any of the other properties") {
+                    expect(request.url?.absoluteString).to(equal(endpoint.url))
+                    expect(request.httpMethod).to(equal(endpoint.method.rawValue))
+                }
+            }
+
             context("when task is .requestCompositeData") {
                 var parameters: [String: Any]!
                 var data: Data!
--- a/Tests/EndpointSpec.swift
+++ b/Tests/EndpointSpec.swift
@@ -151,7 +151,7 @@ final class EndpointSpec: QuickSpec {
                 var request: URLRequest!
 
                 beforeEach {
-                    issue = Issue(title: "Hello, Moya!", createdAt: Date())
+                    issue = Issue(title: "Hello, Moya!", createdAt: Date(), rating: 0)
                     endpoint = endpoint.replacing(task: .requestJSONEncodable(issue))
                     request = try! endpoint.urlRequest()
                 }
@@ -190,7 +190,7 @@ final class EndpointSpec: QuickSpec {
                 decoder.dateDecodingStrategy = .formatted(formatter)
 
                 beforeEach {
-                    issue = Issue(title: "Hello, Moya!", createdAt: Date())
+                    issue = Issue(title: "Hello, Moya!", createdAt: Date(), rating: 0)
                     endpoint = endpoint.replacing(task: .requestCustomJSONEncodable(issue, encoder: encoder))
                     request = try! endpoint.urlRequest()
                 }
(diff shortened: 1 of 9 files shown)


Tests/EndpointSpec.swift
```diff
@@ -175,6 +175,45 @@ final class EndpointSpec: QuickSpec {
                 }
             }
 
+            context("when task is .requestCustomJSONEncodable") {
```
SD10: Maybe we can add test case for when encoding fails?

SD10:
Hey @evgeny-sureev, thanks for doing this 👍 I think this PR is better targeted at development because the addition of a new `Task` case is an API breaking change. However, I see that `development` needs to be updated to reflect `master`

evgeny-sureev:
First SwiftLint error is Cyclomatic Complexity Violation, it is unavoidable in switches for enums with many cases.

AndrewSB:
@evgeny-sureev can associated values in an enum case be given defaults? If so, maybe it would be nice to add an encoder property to the `requestJSONEncodable` case instead of creating a new one. For people who don't want to change the default value, they need not use it

evgeny-sureev:
This would be best variant, but Swift doesn't support defaults for enum associated values yet.

Istered:
You can use static functions to "mimic" cases with default values.
Something like this:
```swift
public enum Task {
   ....
    case requestCustomJSONEncodable(Encodable, encoder: JSONEncoder)

    public static func requestJSONEncodable(_ encodable: Encodable, encoder: JSONEncoder = JSONEncoder()) -> Task {
        return .requestCustomJSONEncodable(encodable, encoder: encoder)
    }
   ....
}
```

Or even skip the second parameter:
```swift
public static func requestJSONEncodable(_ encodable: Encodable) -> Task {
    return .requestCustomJSONEncodable(encodable, encoder: JSONEncoder())
}
```

sunshinejr:
This is really smart @Istered! Doesn't compiler complain about it, though? 🤔

BasThomas:
That should just work - we're actually doing something similar already [here](https://github.com/Moya/Moya/blob/6a37ac5d45dded0ffd008b76a41eb31c5dcf33b1/Sources/Moya/MoyaProvider.swift#L147-L163).

Istered:
@sunshinejr It will just work as long as functions and cases have different names.

sunshinejr:
Oh, so even though the parameter list is different, it still won't work right @Istered? Bummer. Then we probably can't make it pretty without duplicating cases, right?

Istered:
@sunshinejr Yes, it won't work even though parameter list is different, but still there's only one case you need to handle with switch.

So instead of this:
```swift
switch task {
    ....
     /// These are two different cases to handle Encodable in this PR
    case let .requestJSONEncodable(encodable):
     ///

    case let .requestCustomJSONEncodable(encodable, encoder):
     ///
    ....
}
```
You would write only this:
```swift
switch task {
    ....
     /// 
    case let .requestCustomJSONEncodable(encodable, encoder):
     ///
    ....
}
```

evgeny-sureev:
@Istered done.

SD10:
I'm not sure I agree with this approach. We're avoiding adding another enum case (technically replacing an existing one) by adding a static method. 

I don't think there is anything to gain here, we're simply moving the definition from the `case` level to the `method` level and making the API less intuitive.

I also really dislike that the default name of this task is now `requestCustomJSONEncodable`

sunshinejr:
I agree with @SD10. This is why I said I don't see much advantages in that approach with our current situation.

Istered:
@SD10 @sunshinejr 
Maybe it's just me, but I don't see any problems with that since implicit member expressions work with static functions the same way as with enum cases.

What really confuses me are these two lines doing absolutely the same:
```swift
return .requestJSONEncodable(...)
return .requestCustomJSONEncodable(..., JSONEncoder())
```

But I agree with you, guys, it would be better to have only one case. However, the problem is that you can't specify default value for cases as it was mentioned before and also it would be more or less breaking change, right?

SD10:
@Istered I'm not worried about breaking the API, breaking changes are inevitable and this requires a very simple migration.

```Swift
// These may "do" the same thing but the way they are defined have implications
return .requestJSONEncodable(...)
return .requestCustomJSONEncodable(..., JSONEncoder())
```

What bothers me is there is no value in defining `requestJSONEncodable(Encodable, encoder: JSONEncoder)` as a static method.

- It's true that there will be 1 less enum case but the code footprint is no smaller, it actually takes more code to define a method.

-  It makes the API non-intuitive, `Task.requestJSONEncodable` looks like an enum case to the user but it is really a method that returns a `Task.requestCustomJSONEncodable`.

- We no longer get the added benefit of exhaustiveness for having this defined as an enum case.

I think we should solve this problem 1 of two ways:

a) We make one case `requestJSONEncodable(Encodable)` and the user just has to use a boilerplate `JSONEncoder()` or create their own convenience property.

b) We have two cases: `requestJSONEncodable(Encodable, encoder: JSONEncoder)` and `requestCustomJSONEncodable(Encodable, encoder: JSONEncoder)`

SD10:
@evgeny-sureev Can we revert the last commit and move forward on merging this? That would be in line with option (B) I mentioned in my previous comment. Unless any @Moya/contributors disagree with me of course

AndrewSB:
@SD10: tbh I prefer (a), it seems more complete a solution

evgeny-sureev:
I think I was too fast with last commit. Feel free to revert it.

sunshinejr:
We have already a method that requires only to pass `Encodable` and if you don't worry about your `JSONEncoder` (which is probably the case most of the times), it is tedious to pass an empty instance to the method all the time (we have it already e.g. request with/without parameters/encoding etc.). Thus, I'm more keen to just add another case to our `Task` type.

Thoughts @evgeny-sureev @SD10 @AndrewSB?

SD10:
@sunshinejr I agree that adding another case best matches the existing request/parameter encoding API. It doesn’t make sense to treat encodable requests any different. If we’re not going to require users to specify parameter encoding on every request then we shouldn’t require them to pass a JSON decoder (which is more likely to remain constant).

evgeny-sureev:
@sunshinejr I think adding new case to `Task`, as I did earlier, is better.

sunshinejr:
Cool! @evgeny-sureev can we go back to two cases then? :)

evgeny-sureev:
Ok, now we have two cases.

SD10:
@evgeny-sureev Thanks 👍 It looks good to me. If you could just two do more things:
1) Fix the SwiftLint warning
2) Add a CHANGELOG entry under **Added**, this is a breaking change

evgeny-sureev:
Do you have an idea how to fix cyclomatic complexity warning? Other warnings are not a result of this pull request.

SD10:
@evgeny-sureev I don't see a good way to fix that one. When I made my comment Moya had not updated the comment from the results of CI. The build was failing so I restarted it

evgeny-sureev:
Maybe add '// swiftlint:disable:next cyclomatic_complexity'?

SD10:
@evgeny-sureev Sure that would be fine by me. It looks like there are 39 failing tests on CI

Tests/EndpointSpec.swift
```diff
@@ -286,6 +325,26 @@ final class EndpointSpec: QuickSpec {
                 }
             }
 
+            context("when JSONEncoder set with incorrect parameters") {
```
sunshinejr: nice one 👌

ashfurrow:

AndrewSB:
Uggh, I don't particularly like the trade-off we made, but I understand why we did. Thank you for your work and patience @evgeny-sureev 👍

Lang-le:
@evgeny-sureev How can i send url parameters with .requestCustomJSONEncodable? Please help me!

SD10:
@Lang-le Your question is better suited to opening a new GitHub issue

Lang-le:
@SD10 Thank you! I will open it!

PR #1449, opened 2017-11-07
Add Moya's branch guidelines

### LINKED ISSUE #1408: Moya development/release flow.
Hey guys! So lately for versions 9 and 10 we were using a branch called `9.0.0-dev` or `10.0.0-dev` - it was great. I was thinking if we could just remove the version name in the branch and stay with `development`. Additionally, everything that isn't released to the public yet, would stay on this branch. Master would have only production ready code.

Moya's flow could look like:
1. Merge everything to develop that is not a documentation change for current version of Moya (we could also just merge everything to develop and then to master)
2. When there is a need for new patch/minor version before the major one, pick commits for master and then rebase our development.
3. When we want to release a major version, and development contains only changes for this next version, just merge everything into master and rebase/reset development.
4. When we want to release a major version, but development contains also changes for versions after this one, just proceed the same as for patch/minor version.

This is only an idea, I'd love to hear your feedback on that matter. Cheers! 🍻 

cc @Moya/contributors

colinta:
Per #1408 this section of `Contributing.md` outlines our thinking around how and why we maintain `master` and `develop` branches.

- Is it in the right place in the document?
- Clearly titled?
- Clearly worded?

Closes #1408


Contributing.md
```diff
@@ -18,6 +18,18 @@ As an organization contributor, you can merge other people's pull requests, or o
 
 This project is not continuously deployed, there is space for debate after review too. Offering everyone the chance to revert, or make an amending pull request. If it feels right, merge.
 
+#### Bug fixes and Release cycles
+
+We maintain two major branches: `master` and `develop`.
```
ashfurrow: "major" doesn't _quite_ capture what I think you're trying to say here. I would say "persistent" or "permanent" maybe? We might want to add that they are protected as well.

> We maintain two permanent, protected branches: `master` and `develop`.

Contributing.md
```diff
@@ -18,6 +18,18 @@ As an organization contributor, you can merge other people's pull requests, or o
...
+
+`master` is for working on the current release, so any bug fixes or documentation spelling fixes should be merged into this branch.
```
ashfurrow: It's such a shame that HTML dictionary lists never became widely used enough to be in the markdown spec.

Contributing.md
```diff
@@ -18,6 +18,18 @@ As an organization contributor, you can merge other people's pull requests, or o
...
+
+`develop` is where we stage work for the *next* release, i.e. breaking API changes and related documentation updates. Contributors should gently encourage new pull-requests to point to the appropriate branch, and to rebase onto that branch if necessary.
+
+When a new version is ready to be released, please create a pull request to merge `develop` into `master`, named something like "Release 10.0". Then we can have some final discussion before we merge it into `master` and push the release out to the public.
+
+Since `develop` is a *shared* branch, it is important not to ever rebase this branch onto `master`. If a bug fix is applied to `master` it can be merged into `develop` using good old simple `git checkout develop && git merge master`. Yes this will clutter the history a little bit, but it also provides important context to know how/when a patch was applied. Merge commits can be considered necessary historical data, not warts on an idealized history graph.
```
ashfurrow: Not to be the "you should write a blog post about this" guy, but you should think about writing a blog post about this!

colinta:
Yes that occurred to me as I wrote this one - can do.

Contributing.md
```diff
@@ -18,6 +18,18 @@ As an organization contributor, you can merge other people's pull requests, or o
 
 This project is not continuously deployed, there is space for debate after review too. Offering everyone the chance to revert, or make an amending pull request. If it feels right, merge.
 
+#### Bug fixes and Release cycles
```
SD10: Maybe we could make the section header something like **Development Process**?

Contributing.md
SD10: I really like this last paragraph too because it stresses the importance of `merge` vs `rebase` on the development branch and this is something I did not know 👍

colinta:
I agree all around!

A blog post, eh? Boy it’s been a while since I exercised that part of my brain! But talking about merge vs rebase would be fun. I’m very opinionated about all things git, especially the history rewriting. 

I’ll check back in soon, these changes are easy. Thanks all! 🙏

Contributing.md
sunshinejr: Then again, we have [Development.md](https://github.com/Moya/Moya/blob/master/docs/Development.md) already in place, maybe move it there? 🤔

Contributing.md
```diff
@@ -18,6 +18,18 @@ As an organization contributor, you can merge other people's pull requests, or o
...
+
+We maintain two major branches: `master` and `develop`.
+
+`master` is for working on the current release, so any bug fixes or documentation spelling fixes should be merged into this branch.
+
+`develop` is where we stage work for the *next* release, i.e. breaking API changes and related documentation updates. Contributors should gently encourage new pull-requests to point to the appropriate branch, and to rebase onto that branch if necessary.
+
+When a new version is ready to be released, please create a pull request to merge `develop` into `master`, named something like "Release 10.0". Then we can have some final discussion before we merge it into `master` and push the release out to the public.
```
sunshinejr: how about pre-releases? alpha/beta, should it also be merged on master? 🤔

colinta: We're already past the point of alpha/beta, but I would think a pre-release would be synonymous with `develop` (it could be considered the "next" version).

colinta:
Any more thoughts on `Contributing.md` and `Development.md`?  Should those two docs be consolidated into one?  Development just has a small snippet about running tests - that could go in Contributing or the Readme.

Contributing.md
AndrewSB: I don't want to disagree - you're probably right - but whats the upside of putting a bugfix into master and then merging that into develop instead of creating the bugfix commit in develop and then cherry-picking that commit onto master?

colinta: @AndrewSB There's nothing *wrong* with that, per se, but the downside is that you would have two commits that do the same thing, but that have different SHAs.  In the history, after merging, you would have both commits, but one would become effectively empty.

A merge commit would show "on this day, these commits were merged in from master", which is itself useful information.

colinta:
We don't have to keep this Development.md / Contributing.md merger, but thought I would throw it up here for consideration.

Contributing.md
AndrewSB: that makes sense, thanks @colinta 😄

Contributing.md
```diff
@@ -1,7 +1,30 @@
...
+
+Since `develop` is a *shared* branch, it is important not to ever rebase this branch onto `master`. If a bug fix is applied to `master` it can be merged into `develop` using good old simple `git checkout develop && git merge master`. Yes this will clutter the history a little bit, but it also provides important context to know how/when a patch was applied. Merge commits can be considered necessary historical data, not warts on an idealized history graph.
+
+#### Testing
+
+To run tests locally, you will need to download Moya's dependencies.
+To do so, run `carthage update --platform iOS` and take a nap, waiting for it to
```
AndrewSB: let's recommend the [`--cache-builds`](https://github.com/Carthage/Carthage/blob/master/README.md#caching-builds) flag as well!

Contributing.md
```diff
@@ -1,7 +1,30 @@
...
+finish. 😴
+
+If you don't have Carthage installed, check the [installation instructions](https://github.com/Carthage/Carthage#installing-carthage).
+And, of course, don't forget to run `carthage update --platform iOS` after.
```
AndrewSB: cache-builds

colinta:
@AndrewSB I'm not a Carthage user, so I'm not *quite* comfortable putting my name on changes to carthage commands... I'll leave that to another PR 😉 (hint hint!)

Rebasing before merge, I think it'll need another approval before I can merge, we'll see.

AndrewSB:
Haha sounds good @colinta

PR #1454, opened 2017-11-09
Replaced the target's validate property by an enum

### LINKED ISSUE #1447: Custom status code validation
I want to configure my `MoyaProvider` to treat status codes specifically  according to the target I'm requesting.
Ex: a target T1 is expected to return a 404 sometimes, it's catched by the app and treated like a 200 with an empty response.
Another target T2 is not expected to return 404.

So here is my problem: How can I configure my `MoyaProvider` to treat 404 from T1 as valid, and 404 from T2 as invalid?
Short answer: I have not found a way
Long answer: It's possible, but by changing a bit Moya

1. Using the Alamofire's validation
So far, the `Target`'s `validate` property only allows us to use the default Alamofire's validation, i.e validating codes in 200..<300
With Alamofire's `validate(statusCode:)` function, it's possible to customize the validation, but this function is not used by Moya. This issue could be easily addressed by replacing the `Target`'s boolean `validate` property with an enum allowing to use the default 200..<300 range or any custom array and then pass it back to Alamofire.

2. Not using the Alamofire's validation
The Moya's `Response` already contains some `filter()` functions but none allows us to provide an array of status code. In my case, to allow the 404 for T2 would mean to use a 200...404 range, allowing 401 402 and 403 codes which I don't want. A new `Response`'s `filter()` function taking an array as parameter would be needed.

Is all this reasoning correct, or am I missing something that allows me to solve my issue without modifying Moya?

amaurydavid:
Fixes #1447 

This makes the Alamofire's validation fully usable through Moya, and not only the default validation.


SD10:
Hey @amaurydavid thanks for submitting the PR 👍 I think this is an enhancement I can get behind

Sources/Moya/MultiTarget.swift
```diff
@@ -29,8 +29,8 @@ public enum MultiTarget: TargetType {
         return target.task
     }
 
-    public var validate: Bool {
-        return target.validate
+    public var validationType: ValidationType? {
+        return nil
```
SD10: Maybe a `.none` case would be more expressive?

amaurydavid: Initially I had a `.none` case, but then I struggled with the ValidationType's `statusCodes` property: I didn't want it to return an empty array because if used wrongly it would cause Alamofire to error on every status code, so it would have to turn it into an optional. 

Secondly, in `MoyaProvider+Internal` It would meant much more "complex" code than a single line for both 3 methods with a `case let` to match the `.none`, an unwrapping of the newly optional `statusCode` property, and poor factorization capability because of the Alamofire's validation function being duplicated in DataRequest and DownloadRequest ( i.e 2 similar functions except for the generic type parameter).

As I see it, the nil version wins in term of code complexity. As for the usage, I admit being neutral.

amaurydavid:
I've got an issue with the validation test failing because the `TargetType`'s `validationType` default implementation overrides the value given when defining the `StructAPI`'s `validationType`. 
Some help here would be appreciated ^^

SD10:
@amaurydavid Are you sure that is what is happening? Can you try to use a different default like `successCodes` to verify?

amaurydavid:
Yep that's what I did: 
with a default to `nil` it tells me it gets `nil`, 
with a default to `successAndRedirectionCodes` it tells me it gets `Optional(Moya.ValidationType.successAndRedirectionCodes)`
So whatever the value I set in the StructAPI, It only takes the default value.

SD10:
@amaurydavid The problem is probably related to using `String(describing:)`

amaurydavid:
I just checked by implementing Equatable for ValidationType, and ```
it("uses correct validate") {
				expect(target.validationType!) == ValidationType.successCodes
            }
```
results in a `Fatal error: Unexpectedly found nil while unwrapping an Optional value` with `nil`being the default. It looks like the String(describing:) irrelevant here :/

**Comment by SD10** [comment]
@amaurydavid I could be wrong but I can’t replicate the functionality you described. However, when using `String(describing:)` I can replicate what you are experiencing

**Comment by SD10** [comment]
@amaurydavid Any progress here? How did you implement `Equatable` conformance?

File: Tests/MultiTargetSpec.swift [hunk 2]
```diff
@@ -60,7 +60,7 @@ class MultiTargetSpec: QuickSpec {
             }
 
             it("uses correct validate") {
-                expect(target.validate) == true
+				expect(String(describing: target.validationType)).to(endWith("successCodes)")) // Hack to avoid implementing Equatable for ValidationType
```
AndrewSB: I think it's probably just wise to implement Equatable

docs/Targets.md
```diff
@@ -126,6 +126,24 @@ public var sampleData: Data {
 }
 ```
 
+With the `validationType` property, Moya also allows you to use Alamofire's requests validation.
```
AndrewSB: I think it would be helpful to
1. make a recommendation for what the default was (i.e. the upgrade path)
2. Link to the Alamofire docs on validationType

Sources/Moya/MoyaProvider+Internal.swift
```diff
@@ -181,19 +181,19 @@ private extension MoyaProvider {
 
     func sendUploadFile(_ target: Target, request: URLRequest, callbackQueue: DispatchQueue?, file: URL, progress: ProgressBlock? = nil, completion: @escaping Completion) -> CancellableToken {
         let uploadRequest = manager.upload(file, with: request)
-        let alamoRequest = target.validate ? uploadRequest.validate() : uploadRequest
+        let alamoRequest = target.validationType == nil ? uploadRequest : uploadRequest.validate(statusCode: target.validationType!.statusCodes)
```
AndrewSB: nit: [up for discussion, since this actually might be more confusing] can we change these lines to `target.validationType?.statusCodes.map(uploadRequest.validate) ?? uploadRequest` instead?

amaurydavid:
@SD10 Here is my Equatable implementation, I think its pretty straightforward
```
extension ValidationType : Equatable {
	
	public static func ==(lhs: ValidationType, rhs: ValidationType) -> Bool {
		switch (lhs,rhs) {
		case (.successCodes,.successCodes),
			 (.successAndRedirectionCodes, .successAndRedirectionCodes):
			return true
		case (.customCodes(let c1), .customCodes(let c2)):
			return c1 == c2
		default:
			return false
		}
	}
}
```
And my updated test (not used to writing test so there must be a better way):
```
it("uses correct validate") {
	if let type = target.validationType {
		expect(type) == ValidationType.successCodes
	} else {
		expect(false).to(beTrue())
	}
 }
```

Sources/Moya/MoyaProvider+Internal.swift
amaurydavid: Not sure how this is supposed to work as it cause a build error.
If I'm not mistaken, `map` would call the validate func for every `statusCode`, that's not what we want. Also, we have to specifically call the `validate(statusCode:)` func because there are others `validate` func using another first parameter name.

But yeah, It would be great to find a way to get rid of the force unwrap while keeping the code that simple (though as the nil test is performed before using it it won't cause a crash)

docs/Targets.md
amaurydavid: 1. Wouldn't it be better to put this in a migration_10_to_11.md instead?
2. Will do.

SD10:
@amaurydavid Can you check the enum case without using `Quick/Nimble` by using a `switch` statement?
```swift
switch target.validationType {
/// ...
}
```

amaurydavid:
The issue here is that is doesn't even execute the `switch` statement because the validationType is `nil`

sunshinejr:
Hey @amaurydavid, I had to change the base for your PR because of #1483, sorry about that. Could you rebase this PR before further implementation, please?

sunshinejr:
Hey @amaurydavid, as this is a breaking change we would like to introduce it in nearest Moya 11. Do you feel like you would have time to do this next week? If not maybe someone else could help you with that. Please let us know.

amaurydavid:
Hello @sunshinejr , unfortunately I won't be able to work on it this week but I think most of the work is done.
There is still this issue with the default value not being overriden in tests, as I'm not able to fix this, my last though was to simply remove the default value and see what's going on.
After fixing this and editing the migration doc I think it would be done with this PR.

SD10:
Ok @amaurydavid I can take over if you'd like and finish this (if the feature is possible) this weekend.
Are you okay with that?

SD10:
@amaurydavid I'm going to close this because it's difficult for me to sync this PR with the current state of the Moya project. I'm working on a PR for this now. I appreciate you getting the ball rolling 👍

PR #1465, opened 2017-11-15
New example: Composing provider

### LINKED ISSUE #1258: Migrate SubclassingProvider example to use composition instead of subclassing.
This one should be done after #1256 is merged. Look at this [file](https://github.com/Moya/Moya/blob/cd25d627571968e9c79fbddd7e87db51b9876bba/docs/Examples/SubclassingProvider.md). This one was written before Moya 9.0.0 and before we deprecated reactive providers. It was also based on Eidolon's stack, which @ashfurrow already migrated to `9.0.0-alpha.1` version, but we changed few things in `beta.1` so it would need some work anyways. Current Eidolon's almost ready version is [here](https://github.com/artsy/eidolon/blob/de2a3f2987cc83cb213dbc7bd371c14f65660ead/Kiosk/App/Networking/Networking.swift).

pietrocaselani:
This PR adds an example on how to refresh session automatically based on [Artsy's implementation](https://github.com/artsy/eidolon/blob/de2a3f2987cc83cb213dbc7bd371c14f65660ead/Kiosk/App/Networking/Networking.swift).
I updated the implementation based on Moya 10.

Closes #1258

### DIFF
--- a/docs/Examples/ComposingProvider.md
+++ b/docs/Examples/ComposingProvider.md
@@ -1,93 +1,64 @@
-# Composing Provider and refreshing session automatically
+# Composing Provider
 
-Based on [Artsy's implementation](https://github.com/artsy/eidolon/blob/master/Kiosk/App/Networking/Networking.swift).
+You should compose provider when you need to add or customize the behavior from `MoyaProvider`.
 
-Used RxSwift.
+You can use this to do HTTP requests only when your app has internet connection or to handle session refresh automatically and more!
+
+Let's go through some code.
+
+First we will create a provider to do HTTP request when the app is online, We are going to use RxSwift.
 
 ```swift
-final class TokenProvider<Target> where Target: Moya.TargetType {
-	private let provider: MoyaProvider<Target>
+final class OnlineProvider<Target> where Target: Moya.TargetType {
+    private let provider: MoyaProvider<Target>
 
-	// init with all default values to initialize MoyaProvider
-	init(endpointClosure: @escaping MoyaProvider<Target>.EndpointClosure = MoyaProvider.defaultEndpointMapping,
-	     requestClosure: @escaping MoyaProvider<Target>.RequestClosure = MoyaProvider.defaultRequestMapping,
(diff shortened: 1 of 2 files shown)


BasThomas:
Can we use four spaces for indentation instead of a tab? That would keep it in line with our current code samples.

docs/Examples/ComposingProvider.md
```diff
@@ -0,0 +1,93 @@
...
+			.map { (token, refreshToken, expiry) -> String? in
+				return token
+			}
+			.catchError { e -> Single<String?> in
+				guard let error = e as? MoyaError else { throw e }
+				guard case .statusCode(let response) = error else { throw e }
+
+				// If we have 401 error - delete all credentials and handle logout
+				if response.statusCode == 401 {
+					UserInfo.shared.invalidate()
```
BasThomas: Not sure if these things being in a provider would serve as a great example - I think they are not the responsibility of the provider. What do you think?

AndrewSB:
Yeah, maybe as comments throughout the code?
Also, it doesn't look like your `Closes #1258` in the first comment took effect. Not sure what's up there

BasThomas:
> Also, it doesn't look like your `Closes #1258` in the first comment took effect. Not sure what's up there

Seems like that was because they were using the URL rather than the `#`. GitHub 😕

pietrocaselani:
> Can we use four spaces for indentation instead of a tab? That would keep it in line with our current code samples.

@BasThomas Sure! I will change to spaces, sorry about that.

> I'm just wondering if we could maybe add a bit of text explaining to the user what we're doing instead of solely giving them a code snippet.
I also think the document would be better if it didn't discuss refreshing and solely focused on provider composition.

@SD10 I tried to do very likely to the old example, but yes, I agree with you that the example may lost the point that is composition.

About the Closes, I think that will works correctly after the PR gets merged.

pietrocaselani:
Guys, I updated the PR!
Let me know if I need to change or update something!

SD10:
@BasThomas Can we send this through? ⛵️

PR #1473, opened 2017-11-18
Fix linter violations

larryonoff:


larryonoff:
What about disabling some linter rules specifically for spec files?

Tests/EndpointSpec.swift
```diff
@@ -271,8 +271,16 @@ final class EndpointSpec: QuickSpec {
                     // Non-serializable type to cause serialization error
                     class InvalidParameter {}
 
-                    endpoint = endpoint.replacing(task: .requestParameters(parameters: ["":InvalidParameter()], encoding: PropertyListEncoding.default))
-                    let cocoaError = NSError(domain: "NSCocoaErrorDomain", code: 3851, userInfo: ["NSDebugDescription":"Property list invalid for format: 100 (property lists cannot contain objects of type 'CFType')"])
+                    endpoint = endpoint.replacing(
+                        task: .requestParameters(
```
SD10: Just fixing the ones in Package.swift would've been ok... no need for a separate PR 😓
Now this misalignment is killing my OCD

PR #1483, opened 2017-11-28
Replace develop -> development in contributing docs.

sunshinejr:
Right now we have 2 branches that are responsible for development branch: `develop` and `development`. After a quick discussion with @Moya/core-team, we decided to keep `development` as it has already some commits on it. I think I've got all the places that were mentioning `develop`, but might've missed something, so if anyone could double-check it would be awesome.


sunshinejr:
@SD10 oh, what a nice catch! Updated as well :)

Issue #1499, opened 2017-12-11
Step by step sample how to run the examples?


glennposadas:
Hi there, I've never used carthage before, and I believe in your documentation on how to use the examples, just run `pod install`. Carthage is giving me errors.  Thanks.

SD10:
@glennposadas Try running `carthage bootstrap`

glennposadas:
Thanks a lot! I had no choice but to try out Carthage. It's okay now.

glennposadas:
Ooh, now it says Image not found :/

BasThomas:
Can you post the whole console log?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

glennposadas:
It's not a big deal though. :) The error or rather crash happens both on my iPhone 5S with iOS 10+ and on my iPhone 7plus iOS 11+.

When I made my own sample and integrated Moya with Cocoapods, I didn't experience such crash. So I think it's okay and not a big deal. Thank you!

PR #1500, opened 2017-12-11
Change class-only protocol from class to AnyObject

zhongwuzw:
1. Change class-only protocol from `class` to `AnyObject`.
2. Update `README` in `Example` to use `carthage update` instead of `pod install`.


Readme.md
```diff
@@ -43,7 +43,7 @@ You can check out more about the project direction in the [vision document](Visi
 
 ## Sample Project
 
-There's a sample project in the Demo directory. To use it, run `pod install` to download the required libraries. Have fun!
+There's a sample project in the Demo directory. To use it, run `carthage update` to download the required libraries. Have fun!
```
SD10: @BasThomas Looks related to #1499. It's possible we can't run the demo with `pod install` because the Quick/Nimble dependencies are managed with Carthage`?

zhongwuzw: The project don't have `Podfile`, maybe have been switch to `carthage`.

zhongwuzw:
@SD10 , the reason switch from `class` to `AnyObject` is to remove duplication. see [Swift-evolution](https://github.com/apple/swift-evolution/blob/master/proposals/0156-subclass-existentials.md#class-and-anyobject)

Readme.md
sunshinejr: Examples were merged with main project so we don’t use CocoaPods anymore.

ashfurrow:

PR #1502, opened 2017-12-12
[WIP] Refactoring Moya's stubbing API

SD10:
### Summary
This is purely experimental but I was dabbling with the idea of defining the stubbing API in terms of `TestTargetType` as opposed to configuring a `MoyaProvider`.

@Moya/contributors I would appreciate if any of you could review this and provide feedback 🙏 

### How?
We add a protocol that allows Moya users to provide the sample response and stub behavior:
```Swift
protocol TestTargetType {
    var sampleResponse: EndpointSampleResponse { get }
    var stubBehavior: Moya.StubBehavior { get }
}
```

Then internally, we use a `where` clause on `MoyaProvider` to determine if the `Target` we are handling is capable of stubbing requests.

```Swift
extension MoyaProvider where Target: TestTargetType {
/// ... stubbing methods here
}
```

It all comes down to this method:
```Swift
func performRequest(_ target: Target, request: URLRequest, callbackQueue: DispatchQueue?, progress: Moya.ProgressBlock?, completion: @escaping Moya.Completion, endpoint: Endpoint<Target>) -> Cancellable
```

When `Target` != `TestTargetType` this version is called:
```Swift
extension MoyaProvider {
    private func performRequest(_ target: Target, request: URLRequest, callbackQueue: DispatchQueue?, progress: Moya.ProgressBlock?, completion: @escaping Moya.Completion, endpoint: Endpoint<Target>) -> Cancellable {
        return self.performNormalRequest(target, request: request, callbackQueue: callbackQueue, progress: progress, completion: completion, endpoint: endpoint)
    }
}
```

When `Target` == `TestTargetType` this version is called:
```Swift
extension MoyaProvider where Target: TestTargetType {
    private func performRequest(_ target: Target, request: URLRequest, callbackQueue: DispatchQueue?, progress: Moya.ProgressBlock?, completion: @escaping Moya.Completion, endpoint: Endpoint<Target>) -> Cancellable {
        switch target.stubBehavior {
        case .never:
            return self.performNormalRequest(target, request: request, callbackQueue: callbackQueue, progress: progress, completion: completion, endpoint: endpoint)
        default:
            return self.stubRequest(target, request: request, callbackQueue: callbackQueue, completion: completion, endpoint: endpoint, stubBehavior: target.stubBehavior)
        }
    }
}
```

Pros:
- Removes `stubClosure` from MoyaProvider because you can control behavior in `TestTargetType`
- Moves all stubbing behavior to `MoyaProvider+Stubbing.swift`
- Providers not using a `TestTargetType` need not be concerned with anything related to stubbing and perform no checks regarding stubbing at runtime

Cons:
- That's what I want to hear from you all 😅 I'm biased and blind to my own system

### Final thoughts
This breaks a lot of tests, I have not fixed them because it will take some work and I didn't want to start if people think this may not be a good change for the project. However, I think that fixing those tests would help me better understand any drawbacks to this approach.

### DIFF
--- a/Sources/Moya/MoyaProvider+Stubbing.swift
+++ b/Sources/Moya/MoyaProvider+Stubbing.swift
@@ -1,0 +1,102 @@
+import Foundation
+
+/// Mark: Stubbing
+
+/// A TargetType configurable for testing responses
+public protocol TestTargetType {
+
+    var stubBehavior: Moya.StubBehavior { get }
+
+    var sampleResponse: EndpointSampleResponse { get }
+
+}
+
+/// Controls how stub responses are returned.
+public enum StubBehavior {
+
+    /// Do not stub.
+    case never
+
+    /// Return a response immediately.
+    case immediate
+
+    /// Return a response after a delay.
+    case delayed(seconds: TimeInterval)
+}
+
+extension MoyaProvider where Target: TestTargetType {
+
+    // swiftlint:disable function_parameter_count
+    /// When overriding this method, take care to `notifyPluginsOfImpendingStub` and to perform the stub using the `createStubFunction` method.
+    /// Note: this was previously in an extension, however it must be in the original class declaration to allow subclasses to override.
+    @discardableResult
+    open func stubRequest(_ target: Target, request: URLRequest, callbackQueue: DispatchQueue?, completion: @escaping Moya.Completion, endpoint: Endpoint<Target>, stubBehavior: Moya.StubBehavior) -> CancellableToken {
+        let callbackQueue = callbackQueue ?? self.callbackQueue
+        let cancellableToken = CancellableToken { }
+        notifyPluginsOfImpendingStub(for: request, target: target)
+        let plugins = self.plugins
+        let stub: () -> Void = createStubFunction(cancellableToken, forTarget: target, withCompletion: completion, endpoint: endpoint, plugins: plugins, request: request)
+        switch stubBehavior {
+        case .immediate:
+            switch callbackQueue {
+            case .none:
+                stub()
+            case .some(let callbackQueue):
+                callbackQueue.async(execute: stub)
+            }
+        case .delayed(let delay):
+            let killTimeOffset = Int64(CDouble(delay) * CDouble(NSEC_PER_SEC))
+            let killTime = DispatchTime.now() + Double(killTimeOffset) / Double(NSEC_PER_SEC)
+            (callbackQueue ?? DispatchQueue.main).asyncAfter(deadline: killTime) {
+                stub()
+            }
+        case .never:
+            fatalError("Method called to stub request when stubbing is disabled.")
+        }
+
+        return cancellableToken
+    }
+    // swiftlint:enable function_parameter_count
+
+    /// Creates a function which, when called, executes the appropriate stubbing behavior for the given parameters.
+    public final func createStubFunction(_ token: CancellableToken, forTarget target: Target, withCompletion completion: @escaping Moya.Completion, endpoint: Endpoint<Target>, plugins: [PluginType], request: URLRequest) -> (() -> Void) { // swiftlint:disable:this function_parameter_count
+        return {
+            if token.isCancelled {
+                self.cancelCompletion(completion, target: target)
+                return
+            }
+
+            switch target.sampleResponse {
+            case .networkResponse(let statusCode, let data):
+                let response = Moya.Response(statusCode: statusCode, data: data, request: request, response: nil)
+                plugins.forEach { $0.didReceive(.success(response), target: target) }
+                completion(.success(response))
+            case .response(let customResponse, let data):
+                let response = Moya.Response(statusCode: customResponse.statusCode, data: data, request: request, response: customResponse)
+                plugins.forEach { $0.didReceive(.success(response), target: target) }
+                completion(.success(response))
+            case .networkError(let error):
+                let error = MoyaError.underlying(error, nil)
+                plugins.forEach { $0.didReceive(.failure(error), target: target) }
+                completion(.failure(error))
+            }
+        }
+    }
+
+    // swiftlint:disable:next function_parameter_count
+    private func performRequest(_ target: Target, request: URLRequest, callbackQueue: DispatchQueue?, progress: Moya.ProgressBlock?, completion: @escaping Moya.Completion, endpoint: Endpoint<Target>) -> Cancellable {
+        switch target.stubBehavior {
+        case .never:
+            return self.performNormalRequest(target, request: request, callbackQueue: callbackQueue, progress: progress, completion: completion, endpoint: endpoint)
+        default:
+            return self.stubRequest(target, request: request, callbackQueue: callbackQueue, completion: completion, endpoint: endpoint, stubBehavior: target.stubBehavior)
+        }
+    }
+
+    /// Notify all plugins that a stub is about to be performed. You must call this if overriding `stubRequest`.
+    final func notifyPluginsOfImpendingStub(for request: URLRequest, target: Target) {
+        let alamoRequest = manager.request(request as URLRequestConvertible)
+        plugins.forEach { $0.willSend(alamoRequest, target: target) }
+        alamoRequest.cancel()
+    }
+}
--- a/Sources/Moya/MultiTarget.swift
+++ b/Sources/Moya/MultiTarget.swift
@@ -21,9 +21,10 @@ public enum MultiTarget: TargetType {
         return target.method
     }
 
-    public var sampleData: Data {
-        return target.sampleData
-    }
+// TODO: - Not sure how this would affect sample data
+//    public var sampleData: Data {
+//        return target.sampleData
+//    }
 
     public var task: Task {
         return target.task
--- a/Tests/MultiTargetSpec.swift
+++ b/Tests/MultiTargetSpec.swift
@@ -54,10 +54,10 @@ class MultiTargetSpec: QuickSpec {
                 expect(String(describing: target.task)).to(beginWith("requestParameters")) // Hack to avoid implementing Equatable for Task
             }
 
-            it("uses correct sample data") {
-                let expectedData = "sample data".data(using: .utf8)!
-                expect(target.sampleData).to(equal(expectedData))
-            }
+//            it("uses correct sample data") {
+//                let expectedData = "sample data".data(using: .utf8)!
+//                expect(target.sampleData).to(equal(expectedData))
+//            }
 
             it("uses correct validate") {
                 expect(target.validate) == true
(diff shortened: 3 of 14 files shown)


Sources/Moya/MultiTarget.swift
```diff
@@ -21,9 +21,10 @@ public enum MultiTarget: TargetType {
         return target.method
     }
 
-    public var sampleData: Data {
-        return target.sampleData
-    }
+// TODO: - Not sure how this would affect sample data
+//    public var sampleData: Data {
+//        return target.sampleData
+//    }
```
SD10: This is also the only thing I haven't been able to get around with this PR. I don't use the `MultiTarget` so I don't really understand it's purpose.

sunshinejr: Yeah, this is tricky. I will think about this one, but two ideas that come to my mind right now are:
1. Create two MultiTargets, one that stubs and one that does not (like with target types)
2. If target is not a Testing one, use empty data/crash, but this is not the best solution I think.

SD10: This PR is very low priority IMO. It won't make it in Moya 12.0 for sure and I'd like a few Moya members to discuss if they even like this change.

BasThomas:
This is super cool - thanks for sharing your thoughts and code @SD10; much appreciated. I’ll give this a closer inspection soon 👍

Sources/Moya/MoyaProvider+Stubbing.swift
```diff
@@ -0,0 +1,102 @@
...
+    /// When overriding this method, take care to `notifyPluginsOfImpendingStub` and to perform the stub using the `createStubFunction` method.
+    /// Note: this was previously in an extension, however it must be in the original class declaration to allow subclasses to override.
+    @discardableResult
+    open func stubRequest(_ target: Target, request: URLRequest, callbackQueue: DispatchQueue?, completion: @escaping Moya.Completion, endpoint: Endpoint<Target>, stubBehavior: Moya.StubBehavior) -> CancellableToken {
+        let callbackQueue = callbackQueue ?? self.callbackQueue
+        let cancellableToken = CancellableToken { }
+        notifyPluginsOfImpendingStub(for: request, target: target)
+        let plugins = self.plugins
+        let stub: () -> Void = createStubFunction(cancellableToken, forTarget: target, withCompletion: completion, endpoint: endpoint, plugins: plugins, request: request)
+        switch stubBehavior {
```
freak4pc: Mind breaking this up to logical portions with some newlines? (before the switch, before and after the notifyPlugins, etc) 

It's a bit hard to chew down this way

Sources/Moya/MoyaProvider+Stubbing.swift
```diff
@@ -0,0 +1,102 @@
...
+            switch callbackQueue {
+            case .none:
+                stub()
+            case .some(let callbackQueue):
+                callbackQueue.async(execute: stub)
+            }
+        case .delayed(let delay):
+            let killTimeOffset = Int64(CDouble(delay) * CDouble(NSEC_PER_SEC))
+            let killTime = DispatchTime.now() + Double(killTimeOffset) / Double(NSEC_PER_SEC)
+            (callbackQueue ?? DispatchQueue.main).asyncAfter(deadline: killTime) {
```
freak4pc: You should be able to do do `.asyncAfter(deadline: .now() + delay) { ` IIRC

Sources/Moya/MoyaProvider+Stubbing.swift
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        return cancellableToken
+    }
+    // swiftlint:enable function_parameter_count
+
+    /// Creates a function which, when called, executes the appropriate stubbing behavior for the given parameters.
+    public final func createStubFunction(_ token: CancellableToken, forTarget target: Target, withCompletion completion: @escaping Moya.Completion, endpoint: Endpoint<Target>, plugins: [PluginType], request: URLRequest) -> (() -> Void) { // swiftlint:disable:this function_parameter_count
+        return {
+            if token.isCancelled {
```
freak4pc: This would read better as 

```swift
guard !token.isCancelled else { 
    self.cancelCompletion(completion, target: target)
    return
}
```

IMO

Tests/MultiTargetSpec.swift
```diff
@@ -54,10 +54,10 @@ class MultiTargetSpec: QuickSpec {
                 expect(String(describing: target.task)).to(beginWith("requestParameters")) // Hack to avoid implementing Equatable for Task
             }
 
-            it("uses correct sample data") {
-                let expectedData = "sample data".data(using: .utf8)!
-                expect(target.sampleData).to(equal(expectedData))
-            }
+//            it("uses correct sample data") {
```
freak4pc: If it's commented out, just remove it :)

Sources/Moya/MoyaProvider+Stubbing.swift
```diff
@@ -0,0 +1,102 @@
...
+
+    /// Do not stub.
+    case never
+
+    /// Return a response immediately.
+    case immediate
+
+    /// Return a response after a delay.
+    case delayed(seconds: TimeInterval)
+}
```
freak4pc: Really great ideas here ! Super nice work @SD10 .

Sources/Moya/MoyaProvider+Stubbing.swift
SD10: I believe this should work as well. Good clean up 👍

Tests/MultiTargetSpec.swift
SD10: Ah, I commented it out temporarily. I should've added a TODO. I don't think I should be reducing test coverage, but rather fix whatever tests I break 😅

Sources/Moya/MoyaProvider+Stubbing.swift
SD10: Thanks for the review!

Tests/MultiTargetSpec.swift
freak4pc: Adding a TODO is always good :) But if your PR is WIP and you remember to fix this later than its fine :)

SD10:
I'm tagging #1539 because I think it's an issue that will help me better analyze potential cons with this approach.

Sources/Moya/MultiTarget.swift
SD10: Revisited this today... I have a feeling it "just works" now that I understand what `MultiTarget` is doing.
If the multi-target type is a `TestTargetType` the user will have to specify where to get `sampleData` from.

Following the pattern of the existing properties, they would have to extract the associated value of the `target` case. They would have to cast to `TestTargetType` to access the underlying `sampleData`. If casting fails they still have to return something because `sampleData` is non-optional. Really hard to go unnoticed unless you're already writing bad code to begin with.


So as the other properties are doing, they would unwrap the associated value of the enum and return the data from the nested target.

SD10:
Revisited this today with an even better understanding of Moya's internals.

Currently, to stub a request in Moya you have to configure the stubbed request in 3 different places:
1. You need to provide `sampleData` in the `TargetType` protocol
2. You need to provide a `StubBehavior` through the `stubClosure` property of `MoyaProvider
3. You need to provide a new `endpointClosure` to a `MoyaProvider` if you want to manipulate the mapping of a `TargetType` to `EndpointSampleResponse`

I think the drawback of this approach is it requires so much configuration in many different places throughtout the library. This also results in intertwining testing logic with the business logic of your codebase.

Furthermore, `stubClosure` and `endpointClosure` both take in a `TargetType` as their only parameters. 

I think this alludes to the behavior that people really want: **to customize behavior per `TargetType`**.

So, this PR aims to address this issue by making all these things configurable in the `TargetType` directly. Looking for any feedback from @Moya/contributors again

Sources/Moya/MultiTarget.swift
sunshinejr: @SD10 I'm not sure I follow... Are you saying that we won't provide any support for `MultiTarget` with `TestTargetType` so the users would have to do it themselves?

SD10: @sunshinejr I wouldn't say "we won't provide any support" but I was going to suggest that they extend `MultiTarget` to conform to `TestTargetType` inside their test target. We could add a `MultiTestTarget` if you believe this is expecting too much.

sunshinejr: @SD10 Oh, sorry if it sounded like that, I was just trying to understand if that's what you would do. I don't think there is a 100% good solution for this, `MultiTestTarget` is okay, but still doesn't cover all cases (like mixed targets & test targets). I wonder if making few examples instead would be enough... What do you think?

SD10: `MultiTestTarget` would cover all cases because the enum associated value would have to be `TestTargetType` as well. 

An edge case would be a `MultiTarget` with an inner `TestTargetType`. Stubbing would be determined by the type of the root `MultiTarget` so it would never stub.

We could document this. Even provide an assertion/error if the inner target is a `TestTargetType` but `MultiTarget` is not? The only way to fully prevent this is to check every target at runtime by trying to cast it to `TestTargetType`.

SD10: Honestly, the edge case allows you to opt in or out of testing by wrapping your `TargetType` or `TestTargetType`. It's valuable behavior in a way 😂 

Let's sit on this a bit more for me to explore casting. To be honest, I really liked the code separation that the `where` clause provides, but if I can protect people from their own design decisions then that should be explored.

SD10: Realized last night in a dream that nothing can be done with casting. The `MultiTarget` abstracts away any sense of an underlying `TargetType`.

sunshinejr:
@SD10 I know that we are still trying to pass this one, so please rebase it when you find time so we can   review it again.

gligorkot:
Hey @SD10, just wanted to chime in as an outsider as I've been using Moya since v7.0 and love it. I think this would really be a worthwhile change. I've been reading through the issues and I believe this would address #1126 so I thought I'd link it here as well, but correct me if I'm wrong?

One thing I wanted to add, I am big on unit testing my frameworks and have been writing tests for the network layer for a while, so having the stubs really makes this easy; however if it was possible to only include the stubs and `TestTargetType` in the `UnitTest` package this would be awesome! Wasn't sure if this was your intention so I thought I'd add it.

Happy to provide help on this matter and jump more into the open source waters.

SD10:
@gligorkot Thanks for the feedback 💯 This does in fact address #1126 and I agree it's a worthwhile change.

 It allows greater flexibility, a better separation of concerns, and would allow us to remove `MoyaProvider`'s closure based customization in favor of plugins (should we choose to do so) in the future.

I do plan on fixing this up, however, I just started a new job so my time for open source has significantly decreased. Over the next month, I will be moving our apps networking layer to Moya, so after that, I will look at this again.

If the Moya community accepts this proposal, I don't see this change being made until Moya 13. In the meantime, if you want to move your tests to the test target you can use this work around _could_ work?

```Swift
// In your app target
extension TargetType {
   var data: Data { return Data() }
}

// In your test target
extension YourObjectThatConformsToTargetType {
    var data: Data { 
          // return your real data here
    }
}
```

gligorkot:
@SD10 thanks for getting back to me and providing the workaround, however I just tried it and the extension implementation didn't override the `TargetType` implementation in this case, so I'm back to including my `sampleData` in the main target instead of only the test target. Hopefully this makes it in the goals for one of the next versions of Moya.

SD10:
@gligorkot This will definitely be in Moya 13.0 if the community agrees it's a good approach. 

Regarding your work around, I tried the exact same thing yesterday and figured out you can't change the dynamic dispatch for the `Target` argument of the `defaultEndpointMapping` closure in the test target. I got around this by providing my own `endpointClosure` to the `MoyaProvider` in the test target. The `Endpoint` closure may also need it's `target` argument typed as your specialized `TargetType`.

```swift
// In your test target
let endpointClosure: (MyTargetType) -> Endpoint = { target in
   return Endpoint(...
}
```

amaurydavid:
Hello :) Is there any update on this matter? The subject is opened since more than 1 year now :|

Issue #1492, opened 2017-12-06
AccessTokenPlugin hidden coupling


biboran:
I am using `AccessTokenPlugin` to implement authorization in our app, and I just updated Moya from **8.0.2** to **10.0.1** and was surprised to see authorization stop working. 

After looking at the changelogs and Moya source files I found that decision to append an authorization header is now decided by [optionally type casting target](https://github.com/Moya/Moya/blob/c29fb183ed917c5fb53e71260abe02ee32668a66/Sources/Moya/Plugins/AccessTokenPlugin.swift) (line 57) to `AccessTokenAuthorizable`. This is surprising because subtype casting is not something Moya uses actively as a pattern and is not something I think Moya should adopt. It introduces unnecessary hidden coupling, breaks encapsulation, and makes it impossible to understand the behavior of `AccessTokenPlugin` without looking at the documentation (which I didn't find) or the source code.

I think at the very least Moya should use an `assert()` to warn its users in non-production mode that they are using `AccessTokenPlugin` and their `Target` doesn't conform to `AccessTokenAuthorizable`. However, I believe a better solution would be to pass a `AccessTokenAuthorizationSource` in the `AccessTokenPlugin` ctor, so that `AccessTokenPlugin` will be able to ask source wether it's supposed to append an authorization header for this particular target. This way the decision will be delegated to the source which can be proxy, target, or anything else. If users of the plugin choose to break encapsulation and cast their target to some marker interface, they can do it in the source.

SD10:
Hey @biboran, thanks for submitting this issue 😃

A few of my thoughts:
- The casting behavior was present in [Moya 8.0.2](https://github.com/Moya/Moya/blob/f398a5a468e78bb7cceb74685b687520159e06b1/Sources/Moya/Plugins/AccessTokenPlugin.swift#L54).
- An `assert()` could work, but plugins are applied through the Moya provider, which can accept multi targets and not all targets are `AccessTokenAuthorizable`.

> It introduces unnecessary hidden coupling, breaks encapsulation, and makes it impossible to understand the behavior of AccessTokenPlugin without looking at the documentation (which I didn't find) or the source code.

I do agree it would be quite difficult to infer this behavior without looking at the documentation. You can find the documentation for the `AccessTokenPlugin` [here](https://github.com/Moya/Moya/blob/master/docs/Authentication.md).

Could you help me better understand what kind of problem you ran into before we discuss adding an `AccessTokenAuthorizationSource`? 

[...]

PR #1521, opened 2018-01-02
Add custom header / query parameter auth to AccessTokenPlugin

ffittschen:
This is a result of the discussions in https://github.com/Moya/Moya/issues/1283, but does not add the ability to reuse parameters/headers. It is only an improvement to the `AccessTokenPlugin`.

Based on the discussion, I changed the `AccessTokenPlugin` to support custom header / query parameter authentication in addition to the existing Basic and Bearer authentication schemes. Doing so, I also cleaned up the naming:

| Old | New | Reason |
|-----|-----|--------|
|`AccessTokenAuthorizable`|`AccessControllable`| The plugin does not add the capability of targets to be authorized. It merely adds the ability to comply to access control rules of endpoints |
|`AuthorizationType`|`AccessControlType`| Same as above, these are types of access control and not authorization |

The behavior mention in https://github.com/Moya/Moya/issues/1492#issuecomment-350091896 remains unchanged: 
> `AccessTokenPlugin` does nothing by default. You must conform to `AccessControllable` for the plugin to take effect.

### Breaking Changes

* Renamed `AccessTokenAuthorizable` to `AccessControllable`
* Renamed `AuthorizationType` to `AccessControlType`
  * And therefore renamed `authorizationType` to `accessControlType`
* Added another "enum layer" to `accessControlType`, therefore removed the `.basic` and `.bearer` cases. Their new equivalent is `.http(scheme: .basic)` and `.http(scheme: .bearer)`, and added a new case: `.apiKey(name: String, placement: TokenPlacement)` to support custom header / query parameter access control
* Removed `@autoclosure` from `tokenClosure`

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -7,6 +7,7 @@
 - **Breaking Change** Updated minimum version of `ReactiveSwift` to 3.0.
 [#1470](https://github.com/Moya/Moya/pull/1470) by [@larryonoff](https://github.com/larryonoff).
 - **Breaking Change** Changed the `validate` property of `TargetType` to use new `ValidationType` enum representing valid status codes. [#1505](https://github.com/Moya/Moya/pull/1505) by [@SD10](https://github.com/sd10), [@amaurydavid](https://github.com/amaurydavid). 
+- **Breaking Change** Changed `AccessTokenPlugin` to support custom header / query parameter authentication in addition to the current `Basic` and `Bearer` auth [#1521](https://github.com/Moya/Moya/pull/1521) by [@ffittschen](https://github.com/ffittschen).
 
 # [10.0.1] - 2017-11-23
 ### Fixed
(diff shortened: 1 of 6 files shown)


SD10:
Hey @ffittschen, first and foremost thanks for taking the time to put together this PR 💪.

Are there any @Moya/contributors that want to give this a look through? No pressure because we won't have another major release for some time. I can only make minor contributions at the moment

sunshinejr:
Hey @ffittschen. First of all thanks so much for the hard work not only when creating the PR, but also by participating in finding a good, flexible solution. This is a hard one though, as the migration is a big one and seems like the placement of the token is rather a custom usage.  We also didn't get enough feedback on this big change.

Because of that we decided to go with a simpler solution proposed in #1611, at least for now. Although, I've already sent you an invite to the organization as a thank you for your hard work. I hope that this won't stop you from contributing more in the future!

PR #1525, opened 2018-01-04
Development

NguyenDucThanhTam:


sunshinejr:
Hey @NguyenDucThanhTam. I think the purpose of this PR is to use the newest changes of Moya that we have in development - in that case you just need to wait just a little bit more, but the beta release its on its way. After #1527 we should be ready to go. Sorry for the trouble!

Issue #1357, opened 2017-10-08
Changes in English docs not reflected in Chinese docs


BasThomas:
**PLEASE DO NOT CLOSE THIS ISSUE**

This issue has been created to keep track of changes in the English documentation not translated into Chinese.

This issue will continuously be updated to reflect documents that are out of sync.
Documents apart from the README will be added once an initial Chinese counterpart has been written.

If you would like to update the Chinese documentation, please open a pull request and mention which part you've updated.

# To do list

## README
- [ ] #1355: Add link to migration guides in README

## Other
- [ ] The intro in this issue itself!

lovesomeone:
oh  I will update the doc  . Since we have a National Day（from 10.1- 10.8 ），it 's a long holiday 。I didn't focus on the project

weakish:
#1409 has been merged.

SD10:
@Moya/contributors If any of you speak Chinese and have some spare time, we could really use some help to sync up our English docs and Chinese docs. You can see what needs to be done above.
If I had to prioritize things:
- [Add Moya 10 to 11 migration guide](https://github.com/Moya/Moya/blob/master/docs/MigrationGuides/migration_10_to_11.md)
- Add ValidationType to TargetType #1505

SD10:
@bestwnh No need to be sorry! 不要紧。谢谢你!

bestwnh:
@SD10 I would try to completed more Chinese documents if I got some time. But maybe not so soon. If you thought some of them are urgent. Just tag me and I would complete them ASAP.

emmanuelkehinde:
#1997  needs to be updated in the Chinese doc.

Issue #434, opened 2016-03-17
Endpoint doesn't need to be generic


ashfurrow:
If the `Endpoint` type isn't generic, users can build a non-generic provider that may be better-suited for their needs. 

Background:

Today, @jasl and I had a Skype call about #376 to discuss their fork of Moya that uses a far less generic-based approach: https://github.com/jasl/MoyaX There are some really cool ideas there, and I realized that by forcing `Endpoint` to be a generic restricts those ideas. There's no need for it to be generic, that I can think of.

Other ideas were a little more radical, and might not fit within Moya, but would make excellent library extensions. For example, the non-generic providers used in MoyaX help define more complex APIs involved in social network apps (other people have cited social network apps as a use case for using structs instead of enums, too).

Once we remove the generic `Endpoint` requirement, there are some other ideas we should consider borrowing from MoyaX:

- Make the backend (Alamofire) abstracted away so users can replace their own. This might be an interesting way to stub network calls instead, like having a "StubBackend" to stub requests. 
- I'm sure I missed some, @jasl please add the ones I missed.

This is a longer-term project, and Moya as it exists today is relatively stable. We need to be mindful about how we introduce changes like this to our existing users, and balance the trade-offs appropriately. 

Thanks again @jasl for your great ideas :cake:

jasl:
Seems `Backend` should be `Strategy` more clearly.

The Moya current data flow is
`TargetType - endpointClosure -> Endpoint<T> - requestClosure -> NSURLRequest - plugins -> Alamofire's Request -> Cancellable`
Because of generic Endpoint, we can do some things like adding authentication headers for specified Target in `requestClosure`, I think that's why `Endpoint` needs be generic. @ashfurrow is this right?

If `Endpoint` becoming non-generic, how do we do the same thing? maybe moving this job to `TargetType` itself is a good choice, so MoyaX's flow is
`TargetType - TargetType#endpoint-> Endpoint -> Provider#willTransformToRequest -> NSMutableURLRequest - plugins - Backend -> Cancellable`

Besides, `Endpoint` is a Moya internal presentation of `TargetType`, I think there is no side effect and should keeping in a single thread, so changing `Endpoint` no need returns a new instance and It could be a `struct`.

After these changes, I think generic Provider is just for convenient.
I admit non-generic provider is a little radical, but If we have non-generic Provider first, we can easily wrapping it as generic provider but harder the other way.
[...]

PR #1529, opened 2018-01-06
remove endpoint generics

### LINKED ISSUE #1524: Remove generics of Endpoint
I want to make a `PR` to remove generics of `Endpoint`. The reason based on:

`Endpoint` is a concrete class, and store some basic info about request. I cannot found any reason that `Endpoint` need uses generics.

Maybe my thoughts is one-sided, please let me know if it is.:beers::beers:

zhongwuzw:
Fixes #1524


freak4pc:
@zhongwuzw What problem does this PR solve, except for the "pure" aspect of that Concrete class not necessarily needing Target? There might be other developers who count on expecting Endpoints of certain Targets as inputs of classes, etc. Seems like a bit of a breaking change for no real benefit, unless I'm missing out. 

I believe the original author of that class would be @ashfurrow - if you have a few seconds to say what you think about this, and what was the original reasoning behind using a Generic here, that might be helpful :)

ashfurrow:
Hmm, yeah I'm undecided here. I think there _was_ once a reason to have the `Endpoint` type be generic but it's beyond me now. I've looked through [Eidolon](https://github.com/artsy/eidolon) and can't find any use of the generics, and there's not any member in the `Endpoint` type that even references the generic. 

So on philosophical grounds, I think this PR is good. However, it's hard to say what the impact could be for the larger community. Additionally, there's code churn to be concerned with; this is a big change and I don't want to give the community a bunch of work to do without having a concrete benefit in mind. I mean, it's not a lot of work to migrate, but it _is_ work, and it _is_ a breaking change, so we'd need to roll this into Moya 11.0.

So I'm kind of split on this. I like the idea of removing the generic for code hygiene reasons, but I worry about unintended impact and code churn. Hopefully others who are more active users of Moya than I am can chime in about how they feel about it. Hope that context helps.

SD10:
I think the PR is good too. However, I can’t see any positive gains from doing so other than clean code. The generic is resolved at compile time so there shouldn’t be any performance boosts.

SD10:
@ashfurrow How do you recommend Moya handles code churn going forward? I realize it's important given the maturity of the project, but Moya was also built over 4 years of an unstable Swift.

For example, we still have class methods on MoyaProvider because a generic type couldn't have a nested type. 

Though important now, maybe this becomes even more important when Swift itself reaches ABI stability?

sunshinejr:
I see the value of this PR in terms of cleaner `endpoint`/`endpointClosure`, but as others stated I'm not sure it outweighs the cons from migration/relying on the generic part of an Endpoint. We would have to investigate it a bit more to see what really would happen after the flattening.

zhongwuzw:
@SD10 , Aside `Endpoint`, `Generics` would increase build time. In this situation of `Endpoint`, we need two steps, one is to find wether we need generics in `Endpoint` in some situations(by now, I can not found, maybe it has but rarely), another step is to investigate that how we can handle this in these situations.

@sunshinejr , yep, I love clean code😊, if we just need in rarely situations, I think we should remove it, because it leads to many useless code, build time, and increase executable size because specialization if we not need generics at all.

SD10:
@zhongwuzw I don't expect this to significantly impact build times, even on a large project using Moya, but it is relevant.

Honestly, the migration for this is trivial. That does not worry me in the slightest. I just want to figure out if it is possible for someone to be relying on the generic before removing it. We need to figure out if it has any meaning.

SD10:
Ironically, I found an issue #434 Endpoint doesn't need to be generic by @ashfurrow 🤣

ashfurrow:
That's pretty funny – looks like #434 was closed and merged into #556, but _that_ discussion itself didn't lead to changes and we never reconsidered #434. After a year and a half, I'm glad we're finally wrapping that up!

SD10:
@ashfurrow I've been keeping track of all the changes to the English docs that are not reflected in the Chinese docs in #1357. Last week I released Moya 11.0.0-beta.1, should this change be merged before the official release of 11.0?

@zhongwuzw How about I take care of the documentation Ash has requested since my native language is English and you can translate them into Chinese for me?

ashfurrow:
If we already have a major release in the works, I would recommend prioritizing this change and shipping it in the next beta.

zhongwuzw:
@SD10 , 💯 I'll updated Chinese docs after you updated the English docs.

SD10:
I got it @BasThomas. So sorry, I will really do this today.

- [ ] CHANGELOG
- [ ] Docs
- [ ] Migration Guide
- [ ] Inline Endpoint Docs

BasThomas:
No pressure @SD10!

SD10:
Alright, really really sorry for the delays. I completed the checklist I made above. 

I updated all the occurrences of the `Endpoint` generic in the `docs_CN` directory as well @zhongwuzw. So there doesn't seem to be anything to translate to Chinese. We could use the `docs/migration_10_to_11.md` translated but that seems like it should be another issue.

SD10:
We had a lot of input on this one. Let me know if everything looks good to merge and we can get this merged into 11.0.0 ⛵️

Jeroenbb94:
We did use the generic on our project but just as @ashfurrow mentioned, we don't know the reason anymore. The migration from 10 to 11 was easily done and our unit test still work without changing them.

sunshinejr:
@Jeroenbb94 Thank you very much for this feedback! We didn't hear any in beta stage so each report is really valuable to us.

Issue #1530, opened 2018-01-06
Update to RxSwift 4.1


lucas34:
Hi,

RxSwift just recently released a new version aka 4.1. Current dependency is pointing to 4.0.
I think we can considerer upgrading the dependency to 4.1 for the next release (11.0). If it's not too late.

I upgraded my project and base on the release note we should not have any breaking changes.

I can make a PR by updating both podfile and cartfile. Should I target the PR to branch 'master' or 'development' ?
  

Need to update:
- [x] Podfile
- [x] Cartfile and Cartfile.resolved
- [x] Package.Swift and Package.resolved

SD10:
Hey @lucas34, I think `development` is the best place to submit this change.

PR #1532, opened 2018-01-06
Bump RxSwift to 4.1

lucas34:
Link to issue #1530
Updating RxSwift to 4.1

Pod lib lint -> Pass
swift build -> Pass


lucas34:
I forgot to update the Changelog

sunshinejr:
Hey @lucas34. I'm curious: why are we doing this change? '~> 4.0' should be enough to use with both 4.0 and 4.1. Unless something big happened between these 2 releases that we should support in our code?

lucas34:
You are right. I was wondering if I should update only the .resolved files. There is no issue as keeping ~> 4.0 in both Podspec and cartfile. 

This PR is mostly to give the latest released version of the dependency at this date when using 
`carthage bootstrap` or `swift package resolve`

sunshinejr:
Oh, now I understand! Thanks @lucas34, nice catch!

PR #1550, opened 2018-01-19
Fix wrong reference to target instead of endpoint #1546

### LINKED ISSUE #1546: Modifying Task.uploadMultipart through endpointClosure has no effect
I'm trying to add authentication data (username/password) to an uploadMultipart task. First thing I tried was modifying `URLRequest.httpBody` in my authenticator plugin in `PluginType.prepare(_:target:)`, but this doesn't work for uploadMultipart tasks. Next, I tried modifying target.task in an endpointClosure like this:
```swift
        let endpointClosure: ((T) -> Endpoint<T>) = { target in
            var outtask = target.task
            if case .uploadMultipart(let multipart) = outtask {
              outtask = .uploadMultipart(multipart + [ /* Add authentication multipartData here */ ]
            }
            return Endpoint(url: target.path.absoluteString,
                            sampleResponseClosure: { .networkResponse(200, target.sampleData) },
                            method: target.method,
                            task: outtask,
                            httpHeaderFields: target.headers)
        }
```
This doesn't work either, since, if I understand MoyaProvider+Internal correctly, the original target gets passed to `performRequest(_:request:callbackQueue:progress:completion:endpoint:stubBehavior:)` alongside `endpoint`. Is there a way to modify multipart data outside of `TargetType` in Moya 10.0.1? Thanks in advance!

SD10:
This resolves #1546 by switching on `endpoint.task` instead of the original `target.task`. 

This way when a request is performed it will have any changes made to the: `.uploadMultipart`, `.uploadCompositeMultipart`, `.uploadFile`, `.downloadDestination`, and `.downloadParameters` tasks through an `endpointClosure`.

Thanks to @sunshinejr for setting up the test

### DIFF
--- a/Tests/EndpointClosureSpec.swift
+++ b/Tests/EndpointClosureSpec.swift
@@ -29,7 +29,7 @@ final class EndpointClosureSpec: QuickSpec {
             provider = MoyaProvider<HTTPBin>(endpointClosure: endpointClosure, manager: sessionManager)
         }
 
-        it("returns a new endpoint for adding(newHTTPHeaderFields:)") {
+        it("appends additional multipart body in endpointClosure") {
             let multipartData1 = Moya.MultipartFormData(provider: .data("test1".data(using: .utf8)!), name: "test1")
             let multipartData2 = Moya.MultipartFormData(provider: .data("test2".data(using: .utf8)!), name: "test2")
 
--- a/Tests/EndpointClosureSpec.swift
+++ b/Tests/EndpointClosureSpec.swift
@@ -38,7 +38,10 @@ final class EndpointClosureSpec: QuickSpec {
 
             _ = provider.request(.uploadMultipart(providedMultipartData, nil)) { _ in }
             let stringData1 = String(data: try! sessionManager.uploadMultipartFormData!.encode(), encoding: .utf8)
-            let stringData2 = String(data: try! sessionManager.convertMoyaMultipartFormDataToAlamofire(sentMultipartData).encode(), encoding: .utf8)
+
+            let requestMultipartFormData = RequestMultipartFormData()
+            requestMultipartFormData.applyMoyaMultipartFormData(sentMultipartData)
+            let stringData2 = String(data: try! requestMultipartFormData.encode(), encoding: .utf8)
 
             let formData1 = "data; name=\"test1\"\r\n\r\ntest1\r\n"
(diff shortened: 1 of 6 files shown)


Tests/EndpointClosureSpec.swift
```diff
@@ -0,0 +1,83 @@
...
+
+    var uploadMultipartFormData: Alamofire.MultipartFormData?
+
+    override func upload(multipartFormData: @escaping (Alamofire.MultipartFormData) -> Void, usingThreshold encodingMemoryThreshold: UInt64, with urlRequest: URLRequestConvertible, encodingCompletion: ((SessionManager.MultipartFormDataEncodingResult) -> Void)?) {
+        let uploadMultipartFormData = Alamofire.MultipartFormData()
+        multipartFormData(uploadMultipartFormData)
+        self.uploadMultipartFormData = uploadMultipartFormData
+    }
+
+    func convertMoyaMultipartFormDataToAlamofire(_ multipartBody: [Moya.MultipartFormData]) -> Alamofire.MultipartFormData {
```
sunshinejr: I've added this method here, but maybe we can add it as an internal extension to `Moya.MultipartFormData`?

SD10: I think it is better to keep this in the test target if we're not using it in the core library?

sunshinejr: [Actually, we use it in the core as well!](https://github.com/Moya/Moya/blob/f071d9703fbbd3f4a309fbbe32eae3ac6ca2f32d/Sources/Moya/MoyaProvider%2BInternal.swift#L154). If you have time to refactor this a little bit, go for it, if not, let's just make this an extension to `MultipartFormData` and add an issue for a little refactor?

SD10: Ok, thanks for pointing that out. I'll take care of it tonight 👍

SD10:
@sunshinejr Sorry for the delay. I played around with the refactor in 91bb548. Let me know if that is what you were expecting. If not, we can kill the commit and make it an issue.

sunshinejr:
@SD10 almost there! Everything is correct except we don't need `convertMoyaMultipartFormDataToAlamofire()` now 😉

SD10:
@sunshinejr Fixed. Thanks for the eyes 👀

PR #1553, opened 2018-01-25
Add Danger rule to help track English and Chinese docs

### LINKED ISSUE #1548: Add Danger rule that reminds us to track changes in English & Chinese docs
Moya currently maintains two sets of docs: English & Chinese.

It would be awesome if when a file is updated in the [docs](https://github.com/Moya/Moya/tree/master/docs) folder, Danger can warn us to update the file in the [docs_CN](https://github.com/Moya/Moya/tree/master/docs_CN) folder. 

Bonus points if you get Danger to mention issue #1357 😃

yeskunall:
This change makes it so that Danger warns the user to try and update the corresponding Chinese docs, if possible. Otherwise, it asks the user to have the modified file(s) to be added to the list of Chinese docs to be translated being tracked in #1357.

I'm not an expert Ruby developer, so if I can trim some lines of code, LMK, I'll update this PR. 😅

💖

<hr />
Closes #1548

### DIFF
--- a/Dangerfile
+++ b/Dangerfile
@@ -6,9 +6,11 @@ has_app_changes = !git.modified_files.grep(/Sources/).empty?
 # Make it more obvious that a PR is a work in progress and shouldn't be merged yet
 warn("PR is classed as Work in Progress") if github.pr_title.include? "WIP"
 
-# Warn, asking to update Chinese translation of docs if file(s) in `docs` is/are modified
-if !git.modified_files.grep(/docs/).empty?
-  warn("Consider **also** updating the Chinese version of the docs, if possible. Otherwise, request the modified file(s) to be added to the list [here](https://github.com/Moya/Moya/issues/1357) for someone else to translate.")
+# Warn, asking to update Chinese docs if only English docs are updated and vice-versa
+en_docs_modified = git.modified_files.grep(%r{docs/}).empty? # Necessary to exclude `docs_CN` from the grep.
+cn_docs_modified = git.modified_files.grep(%r{docs_CN}).empty?
+if en_docs_modified ^ cn_docs_modified
+  warn("Consider **also** updating the #{ en_docs_modified ? "English" : "Chinese" } docs. For Chinese translations, request the modified file(s) to be added to the list [here](https://github.com/Moya/Moya/issues/1357) for someone else to translate, if you can't do so yourself.")
 end
 
 # Warn when there is a big PR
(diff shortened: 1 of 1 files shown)


yeskunall:
Sorry, I knew I would make the mistake of requesting this to be merged into `master`. Closing this one and opening a new one. 😓

yeskunall:
Oops, looks like it's okay to have this merged into `master`. LMK if I'm wrong, and I'll open one to have it merged into the `development` branch instead!

SD10:
@yeskunall This is fine to be targeted at `master`. You can read more about our development process in CONTRIBUTING.md. You also don’t need to tag other Moya contributors in your PRs unless you feel that a specific contributor should review it. Whenever other Moya contributors have time, they’ll stop by to review the PR 🏎

Dangerfile
```diff
@@ -6,6 +6,11 @@ has_app_changes = !git.modified_files.grep(/Sources/).empty?
 # Make it more obvious that a PR is a work in progress and shouldn't be merged yet
 warn("PR is classed as Work in Progress") if github.pr_title.include? "WIP"
 
+# Warn, asking to update Chinese translation of docs if file(s) in `docs` is/are modified
+if !git.modified_files.grep(/docs/).empty?
```
ashfurrow: This is a good start, but we should be more clever. For example, if someone _does_ update both docs, this warning is still triggered. What do you think about this instead?

```rb
en_docs_modified = !git.modified_files.grep(/docs\//).empty? # Necessary to exclude `docs_CN` from the grep.
cn_docs_modified = !git.modified_files.grep(/docs_CN/).empty?
if en_docs_modified ^ cn_docs_modified
  warn(...)
end
```

This uses the exclusive-or (xor) operator, which is true if only one or the other is true (but not both). This way, if someone: 

- updates both, the xor will be false
- if they update neither, the xor will be false
- updates either one but not the other, the xor will be true

I think that makes sense. What do you think?

yeskunall: No, no, completely agreed. I kinda had the feeling this would be brought up right after I made the PR 🙊

I will update the PR now. I love the use of the XOR operator too! Couldn't be more cleaner!

BasThomas:
This is awesome! Do we want to exclude the `Readme`, `Code of Conduct` and `Vision` though?

ashfurrow:
Looks good! Right now, those files are excluded since they're not in the docs directories. But they do also have CN versions. I think the new Danger rule works as-is, since we modify those files a lot less often than the docs, but the rule could be expanded for checking for updates to `Readme.md` but not `Readme_CN.md`, for example. Not a blocker on merging, imo.

yeskunall:
All right, ping me if I need to change anything 😁

PR #1577, opened 2018-02-09
Add discussion label to exemptLabels for StaleBot

sunshinejr:
I've added "discussion" label to the `exemptLabels`, which just prevents StaleBot from closing/making it stale. Discussions tend to be long and most of the time we reopen them by ourselves whenever StaleBot is taking action. Thus, I figured we could automate it 😄


SD10:
Awesome 🙌 Good idea

Issue #1559, opened 2018-01-30
Add xcode_summary to Danger to show build errors


SD10:
One thing I hate about CI failing is I have to go and dig through the build log. It would be great if we could use `xcode_summary` to propagate test failures and build errors. 

Here is an example of it being used at the [EZSwiftExtensions repo](https://github.com/goktugyil/EZSwiftExtensions/pull/379#issuecomment-279291138)

SD10:
@LucianoPAlmeida I'm tagging you because I see you setting this up for https://github.com/SwifterSwift/SwifterSwift/pull/498. If you get it working would you be so kind as to see if it can be applied to Moya too? ❤️

LucianoPAlmeida:
For sure @SD10 :))
I just start running some tests to see how it works, let's see if I can put this to run there 😆

LucianoPAlmeida:
I think I can try to make this one on this weekend :)) 
Let's see if we can make this run 😊

PR #1579, opened 2018-02-10
Updated Dangerfile

### LINKED ISSUE #1555: Add Danger rule to warn about changes to README and not Chinese README
Related to #1548 #1553. 

The current Danger rule uses the `docs` & `docs_CN` directories to determine if documentation has been modified. However, some docs like README.md have a Chinese counterpart and are not located in these directories. It would be good to add a Danger rule to warn users about keeping both files in sync. See [comment](https://github.com/Moya/Moya/pull/1553#issuecomment-360507030).

kimdv:
Resolves #1555 

This is first time I'm working with Danger. 

This should warn if you change the English readme and not the Chinese and vice-versa. 

I've got inspiration from the other rules and tried to implement it. 🚀

### DIFF
--- a/Dangerfile
+++ b/Dangerfile
@@ -13,6 +13,13 @@ if en_docs_modified ^ cn_docs_modified
   warn("Consider **also** updating the #{ en_docs_modified ? "English" : "Chinese" } docs. For Chinese translations, request the modified file(s) to be added to the list [here](https://github.com/Moya/Moya/issues/1357) for someone else to translate, if you can't do so yourself.")
 end
 
+# Warn, asking to update Chinese docs if only English README are updated and vice-versa
+en_readme_modified = git.modified_files.grep(%r{Readme.md}).empty?
+cn_readme_modified = git.modified_files.grep(%r{Readme_CN.md}).empty?
+if en_readme_modified ^ cn_readme_modified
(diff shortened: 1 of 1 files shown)


kimdv:
Maybe I should add #1559 in this PR too?

Dangerfile
```diff
@@ -13,6 +13,13 @@ if en_docs_modified ^ cn_docs_modified
   warn("Consider **also** updating the #{ en_docs_modified ? "English" : "Chinese" } docs. For Chinese translations, request the modified file(s) to be added to the list [here](https://github.com/Moya/Moya/issues/1357) for someone else to translate, if you can't do so yourself.")
 end
 
+# Warn, asking to update Chinese docs if only English README are updated and vice-versa
+en_readme_modified = git.modified_files.grep(%r{Readme.md}).empty?
```
pedrovereza: Since you're using an `xor` this still works either way, but you need to negate the expression to know if Readme was modified (it should be `!empty`) 😉

Dangerfile
```diff
@@ -13,6 +13,13 @@ if en_docs_modified ^ cn_docs_modified
...
+cn_readme_modified = git.modified_files.grep(%r{Readme_CN.md}).empty?
+if en_readme_modified ^ cn_readme_modified
+  warn("Consider **also** updating the #{ en_readme_modified ? "English" : "Chinese" } README. For Chinese translations, request the modified file(s) to be added to the list [here](https://github.com/Moya/Moya/issues/1357) for someone else to translate, if you can't do so yourself.")
```
pedrovereza: Same thing here on `en_readme_modified ? "English" : "Chinese" `: if only English Readme was modified, you should ask for the Chinese change as well (so basically make it `"Chinese" : "English"`).

PR #1581, opened 2018-02-10
mapOptional<D: Decodable>( ... ) for Observable and Signal

jdisho:
This pull request covers the case when `nil` is returned by `Observable/Signal` instead of error.

### DIFF
--- a/Sources/RxMoya/Observable+Response.swift
+++ b/Sources/RxMoya/Observable+Response.swift
@@ -10,53 +10,66 @@ extension ObservableType where E == Response {
     /// Filters out responses that don't fall within the given range, generating errors when others are encountered.
     public func filter(statusCodes: ClosedRange<Int>) -> Observable<E> {
         return flatMap { response -> Observable<E> in
-            return Observable.just(try response.filter(statusCodes: statusCodes))
+            return .just(try response.filter(statusCodes: statusCodes))
         }
     }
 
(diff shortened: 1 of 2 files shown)


jdisho:
Hey @SD10, thanks for your reply 😄
I can use try? but I like more `do{ ... } catch { ... }`for being more expressive and doing more while catching the error. 

Regarding to the question, _why I need this method?_
Suppose that I want to fetch something from an API. I want to return `nil` if  the object from response is not mapped, otherwise return the object and with `mapOptional` I can do it.

I didn't submit the ReactiveSwift version of mapOptional because of my lack of experience in there, but if it is approved for RxSwift, definitely I will implement the ReactiveSwift version.

SD10:
> I can use try? but I like more do{ ... } catch { ... }for being more expressive and doing more while catching the error.

Forgive me, but it doesn't do anything "more" to handle the error, it just returns `nil`. 

I really think handling the error is the best approach here. The `MoyaError` could occur for a variety of reasons: problems with endpoint mapping, invalid response, failure to decode an object.

jdisho:
Valid argument. 
I just wanted to generalise the errors while returning `nil`. 
In my opinion it is useful if you don't care about the `errors` and just return `nil`.

If this approach is against Moya core principles, you can close it. 😊

SD10:
No problem @jdisho, we really appreciate you taking the time to open this PR 👍

PR #1591, opened 2018-02-22
Fix Alamofire validation not being performed on uploadMultipart

### LINKED ISSUE #1590: ValidationType not working with uploadMultipart
Moya 11.0 using CocoaPods

- The new ValidationType has something with the uploadMultipart, why there is no validation on the multipart upload?
- The below is the debugger output when printing down the **result** from the *requestNormal* function in the **MoyaProvider+Internal.swift**

```lldb
(lldb) po result
▿ .success(Status Code: 400, Data Length: 104)
  ▿ success : Status Code: 400, Data Length: 104
```

SD10:
Resolves #1590 

Alamofire validation was not being performed on `.uploadMultipart` requests. I've added a backing unit test for this.

### DIFF
--- a/Sources/Moya/MoyaProvider+Internal.swift
+++ b/Sources/Moya/MoyaProvider+Internal.swift
@@ -162,7 +162,9 @@ private extension MoyaProvider {
                     self.cancelCompletion(completion, target: target)
                     return
                 }
-                cancellable.innerCancellable = self.sendAlamofireRequest(alamoRequest, target: target, callbackQueue: callbackQueue, progress: progress, completion: completion)
+                let validationCodes = target.validationType.statusCodes
+                let validated = validationCodes.isEmpty ? alamoRequest : alamoRequest.validate(statusCode: validationCodes)
+                cancellable.innerCancellable = self.sendAlamofireRequest(validated, target: target, callbackQueue: callbackQueue, progress: progress, completion: completion)
             case .failure(let error):
                 completion(.failure(MoyaError.underlying(error as NSError, nil)))
             }
--- a/Sources/Moya/MoyaProvider+Internal.swift
+++ b/Sources/Moya/MoyaProvider+Internal.swift
@@ -163,8 +163,8 @@ private extension MoyaProvider {
                     return
                 }
                 let validationCodes = target.validationType.statusCodes
-                let validated = validationCodes.isEmpty ? alamoRequest : alamoRequest.validate(statusCode: validationCodes)
-                cancellable.innerCancellable = self.sendAlamofireRequest(validated, target: target, callbackQueue: callbackQueue, progress: progress, completion: completion)
+                let validatedRequest = validationCodes.isEmpty ? alamoRequest : alamoRequest.validate(statusCode: validationCodes)
+                cancellable.innerCancellable = self.sendAlamofireRequest(validatedRequest, target: target, callbackQueue: callbackQueue, progress: progress, completion: completion)
             case .failure(let error):
                 completion(.failure(MoyaError.underlying(error as NSError, nil)))
             }
--- a/Tests/MoyaProviderIntegrationTests.swift
+++ b/Tests/MoyaProviderIntegrationTests.swift
@@ -306,21 +306,21 @@ final class MoyaProviderIntegrationTests: QuickSpec {
                 let formData = HTTPBin.createMultipartFormData()
                 let target = HTTPBin.validatedUploadMultipart(formData, nil)
                 let provider = MoyaProvider<HTTPBin>()
-                provider.request(target) { result in
-                    switch result {
-                    case .success(let response):
-                        let validCodes = target.validationType.statusCodes
-                        expect(validCodes).to(contain(response.statusCode))
-                    case .failure(let error):
-                        switch error {
-                        case .underlying(_, let response):
-                            let validCodes = target.validationType.statusCodes
-                            expect(validCodes).toNot(contain(response!.statusCode))
-                        default:
-                            fatalError("Received unexpected error")
+                var receievedResponse: Response?
+                var receivedError: Error?
+                waitUntil(timeout: 5.0) { done in
+                    provider.request(target) { result in
+                        switch result {
+                        case .success(let response):
+                            receievedResponse = response
+                        case .failure(let error):
+                            receivedError = error
                         }
+                        done()
                     }
                 }
+                expect(receievedResponse).to(beNil())
+                expect(receivedError).toNot(beNil())
             }
         }
     }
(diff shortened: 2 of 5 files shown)


Tests/TestHelpers.swift
```diff
@@ -171,6 +181,20 @@ extension GitHubUserContent: TargetType {
     }
 }
 
+// MARK: - Upload Multipart Helpers
+
+extension HTTPBin {
+    static func createMultipartFormData() -> [MultipartFormData] {
+        let url = Bundle(for: MoyaProviderSpec.self).url(forResource: "testImage", withExtension: "png")!
```
BasThomas: I know it's a helper, but maybe `fatalError` here too?

Tests/MoyaProviderIntegrationTests.swift
```diff
@@ -299,6 +299,30 @@ final class MoyaProviderIntegrationTests: QuickSpec {
...
+                provider.request(target) { result in
+                    switch result {
+                    case .success(let response):
+                        let validCodes = target.validationType.statusCodes
+                        expect(validCodes).to(contain(response.statusCode))
+                    case .failure(let error):
+                        switch error {
+                        case .underlying(_, let response):
+                            let validCodes = target.validationType.statusCodes
+                            expect(validCodes).toNot(contain(response!.statusCode))
```
BasThomas: Can we assume to always have a `response` here?

Tests/TestHelpers.swift
SD10: Sounds like a good idea to me 👍

Tests/MoyaProviderIntegrationTests.swift
SD10: I think our hands are tied on this one. Response validation is only done on Alamofire integration tests. So I can't stub and ensure a response 😕 This could be an improvement

Sources/Moya/MoyaProvider+Internal.swift
```diff
@@ -162,7 +162,9 @@ private extension MoyaProvider {
                     self.cancelCompletion(completion, target: target)
                     return
                 }
-                cancellable.innerCancellable = self.sendAlamofireRequest(alamoRequest, target: target, callbackQueue: callbackQueue, progress: progress, completion: completion)
+                let validationCodes = target.validationType.statusCodes
+                let validated = validationCodes.isEmpty ? alamoRequest : alamoRequest.validate(statusCode: validationCodes)
```
sunshinejr: `validatedRequest`?

Tests/MoyaProviderIntegrationTests.swift
```diff
@@ -299,6 +299,30 @@ final class MoyaProviderIntegrationTests: QuickSpec {
...
             }
         }
+
+        // Resolves ValidationType not working with multipart upload #1590
+        describe("a provider performing a multipart upload with Alamofire validation") {
+            it("only allows status code 287") {
+                let formData = HTTPBin.createMultipartFormData()
+                let target = HTTPBin.validatedUploadMultipart(formData, nil)
+                let provider = MoyaProvider<HTTPBin>()
+                provider.request(target) { result in
```
sunshinejr: I would restructure this testing code a little bit. Instead of using `expect`s inside the closures, we could use `waitUntil` structure. This makes sure we don't skip this test, where with using `expect`s inside the closure we are risking the situation where this test is skipped, e.g. when closure is not executed. See tests above for examples.

Sources/Moya/MoyaProvider+Internal.swift
SD10: Ugh, that is what I had originally. I knew I should've left it 😆

Tests/MoyaProviderIntegrationTests.swift
SD10: Great catch. I totally forgot about that

Tests/TestHelpers.swift
```diff
@@ -113,14 +114,23 @@ enum HTTPBin: TargetType {
...
         }
     }
 
     var headers: [String: String]? {
         return nil
     }
+
+    var validationType: ValidationType {
+        switch self {
+        case .validatedUploadMultipart:
```
sunshinejr: hmm, how about adding `customCodes` as an associated value, so we can provide it in tests? we could also comment in tests _why_ do we use this constant

Tests/MoyaProviderIntegrationTests.swift
```diff
@@ -299,6 +299,30 @@ final class MoyaProviderIntegrationTests: QuickSpec {
                 }
             }
         }
+
+        // Resolves ValidationType not working with multipart upload #1590
+        describe("a provider performing a multipart upload with Alamofire validation") {
+            it("only allows status code 287") {
```
sunshinejr: how about `it returns error for status code different than 287` and then we could add another case `it returns valid response for status code 287`

Tests/TestHelpers.swift
```diff
@@ -171,6 +181,24 @@ extension GitHubUserContent: TargetType {
     }
 }
 
+// MARK: - Upload Multipart Helpers
+
+extension HTTPBin {
+    static func createMultipartFormData() -> [MultipartFormData] {
```
sunshinejr: nitpick: we could change the name to include test in it, e.g. `createTestMultipartFormData()`

Tests/TestHelpers.swift
SD10: I can do that 👌. To be honest the reason why this is such a hack of a test is I can't validate a stubbed request. Or else I wouldn't need another enum case 😕

Tests/MoyaProviderIntegrationTests.swift
SD10: I wasn't able to test for `287` specifically because I have no control over the actual response since this is not a stub. I just tested for `ValidationType.successCodes` instead

Tests/MoyaProviderIntegrationTests.swift
```diff
@@ -299,6 +299,49 @@ final class MoyaProviderIntegrationTests: QuickSpec {
                 }
             }
         }
+
+        // Resolves ValidationType not working with multipart upload #1590
+        describe("a provider performing a multipart upload with Alamofire validation") {
+            let provider = MoyaProvider<HTTPBin>()
+            let formData = HTTPBin.createTestMultipartFormData()
+            it("returns an error for status code different than 287") {
```
sunshinejr: last one, can we add few empty lines to the code? looks really stacked 😅 

e.g. before `it("")`, in `it("")` we could use a 3 blocks scenario similar to Act-Act-Assert which would add new line before `waitUntil` and one before expects.

SD10:
@sunshinejr No problem, I added some spacing between the tests. LMK if this is 👌

PR #1593, opened 2018-02-22
Perform validation on stubbed responses

### LINKED ISSUE #1592: Perform validation on stubbed response
Currently, Moya is not doing any validation on stubbed requests. Only integration tests done through Alamofire are validated. This incongruency can make targets hard to test. I'm going to label this as an enhancement but it could also be arguably labeled as a bug.

SD10:
Resolves #1592 

This adds the ability to validate a stubbed response.

#### Something to consider:
The `EndpointSampleResponse` has a case `.response(HTTPURLResponse, Data)`. 

We have a unit test for this where we use the `URLResponse` initializer to create the `HTTPURLResponse`.

You can see the two initializers below:
```Swift
// HTTPURLResponse initializer
public init?(url: URL, statusCode: Int, httpVersion HTTPVersion: String?, headerFields: [String: String]?)

// URLResponse initializer
public init(url: URL, mimeType MIMEType: String?, expectedContentLength length: Int, textEncodingName name: String?)
```
 
This change would require that you use the `HTTPURLResponse` initializer because that is the only way to set the `statusCode` property (it is `get-only`). The drawback, however, is arguments like `mimeType` are also `get-only`. 

However, both `HTTPURLResponse` and `URLResponse` are open, so you could get around this by subclassing to make a mock response. 

Looking for thoughts from @Moya/contributors

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -3,7 +3,7 @@
 ### Fixed
 - Fixed Alamofire validation not being performed on `.uploadMultipart` requests.
 [#1591](https://github.com/Moya/Moya/pull/1591) by [@SD10](https://github.com/SD10).
-- Fixed `Response` validation not being performed on stubbed requests.
+- Fixed Alamofire validation not being performed on stubbed requests.
 [#1593](https://github.com/Moya/Moya/pull/1593) by [@SD10](https://github.com/sd10).
 
 # [11.0.0] - 2018-02-07
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,9 @@
 # Next
 
+### Fixed
+- Fixes `Response` validation not being performed on stubbed requests.
+[#1593](https://github.com/Moya/Moya/pull/1593) by [@SD10](https://github.com/sd10).
+
 # [11.0.0] - 2018-02-07
 
 # [11.0.0-beta.2] - 2018-01-27
--- a/Tests/MoyaProviderSpec.swift
+++ b/Tests/MoyaProviderSpec.swift
@@ -409,8 +409,7 @@ final class MoyaProviderSpec: QuickSpec {
             }
 
             it("returns identical sample response") {
-                // TODO: Fix this failing test
-                let response = HTTPURLResponse(url: URL(string: "http://example.com")!, mimeType: nil, expectedContentLength: 0, textEncodingName: nil)
+                let response = HTTPURLResponse(url: URL(string: "http://example.com")!, statusCode: 200, httpVersion: nil, headerFields: nil)!
                 let endpointResolution: MoyaProvider<GitHub>.EndpointClosure = { target in
                     return Endpoint(url: URL(target: target).absoluteString, sampleResponseClosure: { .response(response, target.sampleData) }, method: target.method, task: target.task, httpHeaderFields: target.headers)
                 }
--- a/Tests/MoyaProviderSpec.swift
+++ b/Tests/MoyaProviderSpec.swift
@@ -973,6 +973,7 @@ final class MoyaProviderSpec: QuickSpec {
             }
         }
 
+        // Resolves #1592 where validation is not performed on a stubbed request
         describe("a provider for stubbed requests with validation") {
             var stubbedProvider: MoyaProvider<GitHub>!
 
--- a/Tests/MoyaProviderSpec.swift
+++ b/Tests/MoyaProviderSpec.swift
@@ -409,10 +409,10 @@ final class MoyaProviderSpec: QuickSpec {
             }
 
             it("returns identical sample response") {
+                // TODO: Fix this failing test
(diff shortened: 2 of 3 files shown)


Tests/MoyaProviderSpec.swift
```diff
@@ -409,8 +409,7 @@ final class MoyaProviderSpec: QuickSpec {
             }
 
             it("returns identical sample response") {
-                // TODO: Fix this failing test
-                let response = HTTPURLResponse(url: URL(string: "http://example.com")!, mimeType: nil, expectedContentLength: 0, textEncodingName: nil)
+                let response = HTTPURLResponse(url: URL(string: "http://example.com")!, statusCode: 200, httpVersion: nil, headerFields: nil)!
```
SD10: This is the change I discuss in the body of the issue above

SD10:
Doing a little more research on this - there should be no issue here. Worst case scenario a user would have to implement a convenience initializer on `HTTPURLResponse`.

Tests/MoyaProviderSpec.swift
```diff
@@ -973,5 +972,63 @@ final class MoyaProviderSpec: QuickSpec {
...
+                        return Endpoint(
+                            url: URL(target: target).absoluteString,
+                            sampleResponseClosure: { .networkResponse(400, target.sampleData) },
+                            method: target.method,
+                            task: target.task,
+                            httpHeaderFields: target.headers
+                        )
+                    }
+                    stubbedProvider = MoyaProvider<GitHub>(endpointClosure: endpointClosure, stubClosure: MoyaProvider.immediatelyStub)
+                    stubbedProvider.request(.zen) { result in
```
sunshinejr: can we change it to use `waitUntil` as well here? 😉

Tests/MoyaProviderSpec.swift
```diff
@@ -973,5 +972,63 @@ final class MoyaProviderSpec: QuickSpec {
                 }
             }
         }
+
+        describe("a provider for stubbed requests with validation") {
+            var stubbedProvider: MoyaProvider<GitHub>!
+
+            context("response contains invalid status code") {
+                it("returns a failure case") {
```
sunshinejr: how about `returns an error`?

Tests/MoyaProviderSpec.swift
```diff
@@ -973,5 +972,63 @@ final class MoyaProviderSpec: QuickSpec {
                 }
             }
         }
+
+        describe("a provider for stubbed requests with validation") {
+            var stubbedProvider: MoyaProvider<GitHub>!
+
+            context("response contains invalid status code") {
```
sunshinejr: few empty lines in these would be awesome as well 😄

Tests/MoyaProviderSpec.swift
```diff
@@ -973,5 +972,63 @@ final class MoyaProviderSpec: QuickSpec {
...
+                    }
+                    stubbedProvider = MoyaProvider<GitHub>(endpointClosure: endpointClosure, stubClosure: MoyaProvider.immediatelyStub)
+                    let validStatusCodes = GitHub.zen.validationType.statusCodes
+                    stubbedProvider.request(.zen) { result in
+                        switch result {
+                        case .success(let response):
+                            expect(validStatusCodes).toNot(beEmpty())
+                            expect(validStatusCodes).to(contain(response.statusCode))
+                        case .failure(let error):
+                            guard case let MoyaError.underlying(_, response) = error else {
```
sunshinejr: here a `failure` case is probably a failed test, as we expect only `success` case, right? and similar in the above test with the `success case`?

Tests/MoyaProviderSpec.swift
```diff
@@ -409,10 +409,9 @@ final class MoyaProviderSpec: QuickSpec {
             }
 
             it("returns identical sample response") {
-                let response = HTTPURLResponse(url: URL(string: "http://example.com")!, mimeType: nil, expectedContentLength: 0, textEncodingName: nil)
+                let response = HTTPURLResponse(url: URL(string: "http://example.com")!, statusCode: 200, httpVersion: nil, headerFields: nil)!
```
sunshinejr: makes sense 👍

SD10:
Alright @sunshinejr, I cleaned up the unit tests and added a CHANGELOG entry

Changelog.md
```diff
@@ -1,8 +1,10 @@
 # Next
 
-## Fixed
-- Fixes Alamofire validation not being performed on `.uploadMultipart` requests.
+### Fixed
+- Fixed Alamofire validation not being performed on `.uploadMultipart` requests.
 [#1591](https://github.com/Moya/Moya/pull/1591) by [@SD10](https://github.com/SD10).
+- Fixed `Response` validation not being performed on stubbed requests.
```
sunshinejr: can we phrase it the same as the entry above? it is basically the same thing, just fixed for different parts of the library, but might seem like something different

SD10: I didn't mention Alamofire because technically it doesn't have anything to do with Alamofire 😂 They don't need to know that though

PR #1605, opened 2018-03-12
Adding convenience Range filter.

LucianoPAlmeida:
Hey guys :))
Notice the filter API in response has a range filter that takes a ClosedRange type, this PR is a proposal to have also a function that takes a Range as an option.


SD10:
Hey @LucianoPAlmeida, thanks for taking the time to submit this PR 👍 

I'm not all that opposed to adding this because there is a low maintenance cost but I'll get some others opinions. 

Here are my thoughts:
- We can already achieve the same functionality through `ClosedRange`
- If we do decide to add this then we need to maintain the parity of the RxSwift and ReactiveSwift APIs:
   - Update RxMoya's Observable response type to have this method as well
   - Provide this method to ReactiveSwift as well.

Sources/RxMoya/Single+Response.swift
```diff
@@ -7,13 +7,20 @@ import RxSwift
...
+    /// Filters out responses that don't fall within the given closed range, generating errors when others are encountered.
     public func filter(statusCodes: ClosedRange<Int>) -> Single<ElementType> {
         return flatMap { response -> Single<ElementType> in
             return Single.just(try response.filter(statusCodes: statusCodes))
         }
     }
 
+    /// Filters out responses that don't fall within the given range, generating errors when others are encountered.
+    public func filter(statusCodes: Range<Int>) -> Single<ElementType> {
+        return flatMap { response -> Single<ElementType> in
```
BasThomas: Can we not forward this to the `ClosedRange` `filter` to prevent duplicating the implementation?

Tests/Single+MoyaSpec.swift
```diff
@@ -23,6 +23,23 @@ final class SingleMoyaSpec: QuickSpec {
                 expect(errored).to(beTruthy())
             }
 
+            it("filters out unrequested status codes with range") {
+                let data = Data()
+                let single = Response(statusCode: 10, data: data).asSingle()
+
+                var errored = false
+                _ = single.filter(statusCodes: 0..<10).subscribe { event in
```
BasThomas: Nice!

Sources/Moya/Response.swift
```diff
@@ -55,6 +55,20 @@ public extension Response {
...
 
+    /**
+     Returns the `Response` if the `statusCode` falls within the specified range.
+     
+     - parameters:
+     - statusCodes: The range of acceptable status codes.
+     - throws: `MoyaError.statusCode` when others are encountered.
+     */
+    public func filter(statusCodes: Range<Int>) throws -> Response {
+        guard statusCodes.contains(statusCode) else {
```
BasThomas: This could also be forwarded.

Sources/RxMoya/Single+Response.swift
LucianoPAlmeida: Sure :))

LucianoPAlmeida:
@SD10 
Totally agree, this was just more of a convenience API, to provide options for those who prefer to use Range instead. Me personally I prefer use ranges, for example, 200..<300 instead of 200...299 😂
And since it is very simple to add, I did it. I send it more as a proposal, thought will be cool :)) 
@BasThomas Forwarded :))

Sources/ReactiveMoya/SignalProducer+Response.swift
```diff
@@ -14,18 +14,26 @@ extension SignalProducerProtocol where Value == Response, Error == MoyaError {
         }
     }
 
+    /// Filters out responses that don't fall within the given range, generating errors when others are encountered.
+    public func filter(statusCodes: Range<Int>) -> SignalProducer<Value, MoyaError> {
+        return filter(statusCodes: statusCodes.lowerBound...statusCodes.upperBound-1)
```
sunshinejr: this shouldn't have to use logic again, it is defined on `Response` and this way we can have inconsistent condense whenever we forgot to change the same line in 3 places after a fix/change... let's change it to:
```swift
     public func filter(statusCodes: Range<Int>) -> SignalProducer<Value, MoyaError> {
         return producer.flatMap(.latest) { response -> SignalProducer<Value, Error> in
             return unwrapThrowable { try response.filter(statusCodes: statusCodes) }
          }
      }
```

Sources/RxMoya/Single+Response.swift
```diff
@@ -7,25 +7,33 @@ import RxSwift
...
-    /// Filters out responses that don't fall within the given range, generating errors when others are encountered.
+    /// Filters out responses that don't fall within the given closed range, generating errors when others are encountered.
     public func filter(statusCodes: ClosedRange<Int>) -> Single<ElementType> {
         return flatMap { response -> Single<ElementType> in
             return Single.just(try response.filter(statusCodes: statusCodes))
         }
     }
 
+    /// Filters out responses that don't fall within the given range, generating errors when others are encountered.
+    public func filter(statusCodes: Range<Int>) -> Single<ElementType> {
```
sunshinejr: the same situation here, as per ReactiveSwift version

Tests/Observable+MoyaSpec.swift
```diff
@@ -25,6 +25,25 @@ final class ObservableMoyaSpec: QuickSpec {
                 expect(errored).to(beTruthy())
             }
 
+            it("filters out unrequested status codes") {
```
sunshinejr: can we also have a test for lower bound, please? (e.g. with range `200..<300`)

SD10:
@LucianoPAlmeida Can you also add a CHANGELOG entry for this? It’s non api breaking under added

Sources/ReactiveMoya/SignalProducer+Response.swift
LucianoPAlmeida: @sunshinejr  Sure :))

LucianoPAlmeida:
Should CI be running now? 😂

sunshinejr:
Yeah, @LucianoPAlmeida, sorry for the problem but our org is currently bugged on CircleCI. Long-story-short we were using a beta of CircleCI 2.0 on macOS and we had to be given a trial to make it work. I already messaged support and we should have it resolved soon. Then we can proceed with this PR. 😅

ashfurrow:

Issue #1542, opened 2018-01-10
Moya 10.0.1 installation via Carthage is not working


eMdOS:
Cartfile:

```
github "Moya/Moya" ~> 10.0
```

Cartfile.resolved:
```
github "Moya/Moya" "10.0.1"
```

Error:
```
*** Building scheme "Moya" in Moya.xcodeproj
Build Failed
	Task failed with exit code 65:
	/usr/bin/xcrun xcodebuild -project /Users/emdos/Development/eMdOS/Playground/Carthage/Checkouts/Moya/Moya.xcodeproj -scheme Moya -configuration Release -derivedDataPath /Users/emdos/Library/Caches/org.carthage.CarthageKit/DerivedData/9.2_9C40b/Moya/10.0.1 -sdk iphoneos ONLY_ACTIVE_ARCH=NO BITCODE_GENERATION_MODE=bitcode CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY= CARTHAGE=YES archive -archivePath ./ SKIP_INSTALL=YES GCC_INSTRUMENT_PROGRAM_FLOW_ARCS=NO CLANG_ENABLE_CODE_COVERAGE=NO (launched in /Users/emdos/Development/eMdOS/Playground/Carthage/Checkouts/Moya)

This usually indicates that project itself failed to compile. Please check the xcodebuild log for more details: /var/folders/m7/7xddbtzx2rz52pdm6n_ybgv40000gn/T/carthage-xcodebuild.h549xk.log
```

SD10:
@lucas34 Any input on this one? I don’t use Carthage but it looks like maybe we need to use 10.0.1 in the Cartfile?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

lucas34:
Not sure about this issue,
Could you please provide more information such as:
- The command that you ran
- Carthage Version
- Xcode Version

And also, can you please run your command with `--verbose` ?

I just made a quick test with only `github "Moya/Moya" ~> 10.0` in a cartfile. Running `carthage bootstrap` works just fine. My Cartfile is resolved with those dependencies:

Cartfile.resolved
```
github "Alamofire/Alamofire" "4.6.0"
github "Moya/Moya" "10.0.1"
github "ReactiveCocoa/ReactiveSwift" "2.0.1"
github "ReactiveX/RxSwift" "4.1.1"
github "antitypical/Result" "3.2.4"
```

[...]

PR #1615, opened 2018-03-29
Remove SwiftLint build phases

### LINKED ISSUE #1458: Error while updating Moya with carthge
Hello,
I'm using carthage version 0.26.2, and when trying to update with `carthage update --platform iOS` I get, only for Moya, this error:
```
*** Skipped installing Moya.framework binary due to the error:
	"Incompatible Swift version - framework was built with 4.0 (swiftlang-900.0.65.2 clang-900.0.37) and the local version is 4.0.2 (swiftlang-900.0.69.2 clang-900.0.38)."
*** Checking out Moya at "10.0.0" 
```

Here is my Cartfile:
```github "Moya/Moya" "10.0.0"```

Would you have any idea of why or how I could make this work ?

sunshinejr:
Fixes #1458 and #1542.

I removed the build phases because it does more harm than it helps right now. Not only there is a problem for people that do not have SwiftLint installed, but also there is a problem with new versions of Swift and new versions of SwiftLint, whenever there is a linter error the build just doesn't pass.

Right now we have a linter on every PR, so that's a good start. We could try a fix that Quick did for the same problem. I'm gonna create an issue for the improvement with link to their fix.

Also, I'm targeting master so we can release the next patch version after the merge, for Carthage users.

Oh, and while I was at it, fixed some of the SwiftLint errors.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,6 @@
 # Next
-
+### Fixed
+- Fixed Carthage compatibility by removing a SwiftLint build phase for now. [#1615](https://github.com/Moya/Moya/pull/1615) by [@sunshinejr](https://github.com/sunshinejr).
 # [11.0.1] - 2018-02-26
 
 ### Fixed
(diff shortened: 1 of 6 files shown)


Dschee:
I'm personally against removing SwiftLint entirely. There's a reason to have a linter and that's to have a common code style. If any rule is problematic in this project, we can still configure SwiftLint to match our purposes. Regarding Carthage builds, there's a simple solution by only running the script in debug mode – Carthage does release builds so that fixes the issues. I've created a PR #1619 with my suggestion.

Dschee:
I have just also included the identifier_name fixes of @sunshinejr and continued his work (he didn't fix all violations of that rule).

sunshinejr:
@Dschee I'm not against the build phase as well, just that we need to have this working for all, especially Carthage users. That's why I created #1616. I'm happy to close this one if #1619 will work correctly.

SD10:
We're closing this one in favor of #1619. Thanks for getting the ball rolling @sunshinejr 💪

PR #1617, opened 2018-03-29
Xcode 9.3 test run & dependencies update.

sunshinejr:
I've run a test for Swift 4.1 with Moya and it seems to be working just fine without any warnings. Only change I think was that there is this new file, `IDEWorkspaceChecks.plist`, and it seems that [it's better committing this file instead of ignoring it](https://twitter.com/NeoNacho/status/978339756374351872), so I've did it as well.

Apart from that I've updated the dependencies for SPM & Carthage.

### DIFF
--- a/Cartfile.resolved
+++ b/Cartfile.resolved
@@ -1,4 +1,4 @@
-github "Alamofire/Alamofire" "4.6.0"
+github "Alamofire/Alamofire" "4.7.0"
 github "AliSoftware/OHHTTPStubs" "f90c2bb0fb882e43761ab963ca8869d349d2c6e3"
 github "Quick/Nimble" "v7.0.3"
 github "Quick/Quick" "v1.2.0"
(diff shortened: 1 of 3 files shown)


Dschee:
When specifying this branch and running `carthage update --platform ios` the build of RxMoya failed. I just created a new branch named `xcode_9.3` which includes this branch and the branch `swiftlint-fix` which is part of #1619. Now it works.

sunshinejr:
@Dschee yeah because this one doesn't fix the Carthage problem, that's why you have these errors. After fixing Carthage Swift 4.1 will work without any additional works. This one just adds a file that Xcode 9.3 added and updates dependencies, that's why it's targeted to the `development` branch as well.

SD10:
@sunshinejr Can we merge this one?

sunshinejr:
@SD10 Yeah, it's good to go.

PR #1619, opened 2018-03-30
Run SwiftLint only on debug builds

### LINKED ISSUE #1458: Error while updating Moya with carthge

Dschee:
Attempts to fix problems with the SwiftLint script when building via Carthage.

Fixes #1458 and #1542. Intended to supersede #1615.
Related to #1616.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,6 +29,9 @@ jobs:
           name: Update Homebrew
           command: brew update
       - run:
+          name: Test Carthage Build before installing SwiftLint
+          command: carthage build --no-skip-current --cache-builds
+      - run:
           name: Install Swiftlint
           command: brew install swiftlint
       - run:
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,4 +1,6 @@
 # Next
+### Fixed
+- Fixed Carthage compatibility by disabling the SwiftLint build phase in release builds. [#1619](https://github.com/Moya/Moya/pull/1619) by [@Dschee](https://github.com/Dschee).
 
 # [11.0.1] - 2018-02-26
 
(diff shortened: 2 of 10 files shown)


sunshinejr:
I appreciate that you jumped in on this, @Dschee. However, we don't want to make the same mistake twice and just believe that this is working. Could you please create a step in Circle CI config that runs carthage script without SwiftLint installed?

Dschee:
I just added something to the Circle CI config. Let's see what the result's gonna be ...

Dschee:
The new step running this before SwiftLint install:

```
carthage build --no-skip-current --cache-builds
```

Resulted in a successful build of all Moya targets:

```
*** Building scheme "Moya" in Moya.xcodeproj
*** Building scheme "RxMoya" in Moya.xcodeproj
*** Building scheme "ReactiveMoya" in Moya.xcodeproj
```

Does this meet your expectations?

SD10:
@Dschee Thanks for finding an alternative solution to removing part of our tooling

Dschee:
Just as a side note: I had similar issues with SwiftLint and Carthage like I saw in several issues here in my own open source projects. This approach successfully fixed them all. That‘s why when I saw you were removing SwiftLint I felt like I should share my experience there.

PR #1624, opened 2018-04-04
Making range filter generic to any range type.

LucianoPAlmeida:
Hey guys :)) 
Proposal 
I did a PR adding a convenience filter for Range #1605 
But now I think is a better idea make instead of adding a convenience function for any Range type, just make the existing function generic and maintain only on function implementation that can handle any range type for a filter.

### DIFF
--- a/Sources/Moya/Response.swift
+++ b/Sources/Moya/Response.swift
@@ -42,7 +42,7 @@ public final class Response: CustomDebugStringConvertible, Equatable {
 public extension Response {
 
     /**
-     Returns the `Response` if the `statusCode` falls within the specified closed range.
+     Returns the `Response` if the `statusCode` falls within the specified range.
 
      - parameters:
         - statusCodes: The range of acceptable status codes.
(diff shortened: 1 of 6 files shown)


sunshinejr:
@LucianoPAlmeida just wanted to say that this is the first time I'm seeing `RangeExpression` type and I'm amazed. Will probably take a closer look on this and PR later on, but great job on finding it 💯

LucianoPAlmeida:
Hey @sunshinejr I felt the same way when I find out about it, I always amazed at these things on the  Swift :))

Sources/Moya/Response.swift
```diff
@@ -48,24 +48,13 @@ public extension Response {
         - statusCodes: The range of acceptable status codes.
```
SD10: May want to remove the reference to a closed range in the comment above

SD10:
We'd also need a CHANGELOG entry, this is API breaking cuz we've changed signatures

PR #1628, opened 2018-04-05
No Brown M&M's

SD10:
Adds another section to Contributing.md to let us know who has read the document before opening a PR. I really like this idea.

Twitter thread for reference:
https://twitter.com/alloy/status/981959592337903618

Let me know what you all think 🍦

### DIFF
--- a/Contributing.md
+++ b/Contributing.md
@@ -72,3 +72,7 @@ Both [Ash Furrow](https://github.com/ashfurrow) and [Orta Therox](https://github
 #### Where can I get more info about this document?
 
 The original source of this document can be found at [https://github.com/moya/contributors](https://github.com/moya/contributors).
+
+## [No Brown M&M's](http://en.wikipedia.org/wiki/Van_Halen#Contract_riders)
+
+If you made it all the way to the end, bravo dear user, we love you. You can include this emoji in the top of your ticket to signal to us that you did in fact read this file and are trying to conform to it as best as possible:🍦 


Contributing.md
```diff
@@ -72,3 +72,7 @@ Both [Ash Furrow](https://github.com/ashfurrow) and [Orta Therox](https://github
 #### Where can I get more info about this document?
 
 The original source of this document can be found at [https://github.com/moya/contributors](https://github.com/moya/contributors).
+
+## [No Brown M&M's](http://en.wikipedia.org/wiki/Van_Halen#Contract_riders)
+
+If you made it all the way to the end, bravo dear user, we love you. You can include this emoji in the top of your ticket to signal to us that you did in fact read this file and are trying to conform to it as best as possible:🍦
```
ashfurrow: I like this idea a lot! But I'd suggest we should go with a more generic, non-food celebratory emoji. I took and CocoaPods really cornered the market with the rainbow, how about we use ✨? 

I would also make sure to include both the emoji in the `Contributing.md` as well as the shortcode, which you'll need to enclose in backticks. So something like:


```md
If you made it all the way to the end, bravo dear user, we love you. You can include the ✨ emoji in the top of your ticket to signal to us that you did in fact read this file and are trying to conform to it as best as possible: `:sparkles:`
```

SD10:
@ashfurrow Thanks. I agree a more generic emoji is a better fit. I guess I'm just hungry right now 😋 I just wanted to use something different than CocoaPods so the same emoji didn't become common use.

SD10:
@BasThomas Good idea 👍

PR #1630, opened 2018-04-07
Added Testing Documentation

### LINKED ISSUE #632: Improve documentation surrounding testing
See https://github.com/Moya/Moya/issues/541 for a conversation about how to test, specifically around stubbing. We should document the outcomes of that discussions somewhere in the repo.

fbeeper:
✨
As discussed on #632, there was the need to add some documentation for Testing... so, here is my take on it. Look forward to getting your feedback!

Fixes #632

### DIFF
--- a/docs/Testing.md
+++ b/docs/Testing.md
@@ -4,7 +4,7 @@ Moya has been created with testing at its heart. In this document, you will find
 
 ## `sampleData`
 
-When creating your `TargetType` you are required to provide `sampleData` for your targets. All you need to do there is to provide `Data` that represents a sample response from every particular target.
+When creating your `TargetType` you are required to provide `sampleData` for your targets. All you need to do there is to provide `Data` that represents a sample response from every particular target. This can be used later for tests or just for providing offline support while developing. 
 
 For example:
 
@@ -16,8 +16,6 @@ public var sampleData: Data {
     }
 }
 ```
-
-This can be used later for tests or just for providing offline support while developing.
 
 ## `stubClosure`
 
@@ -37,21 +35,21 @@ Before you continue, it is worth mentioning that all the sample responses will b
 
 With the previous `sampleData` and `stubClosure`, we could only specify the data returned when stubbing. But you have more options.
 
-Moya offers you the opportunity to configure an [`endpointClosure`](https://github.com/Moya/Moya/blob/master/docs/Endpoints.md#from-target-to-endpoint) on your provider. In that closure, you convert from `Target` to `Endpoint`. And this `Endpoint` is where you are going to be able to specify more details for your testing. More concretely, on its `sampleResponseClosure`.
+Moya offers you the opportunity to configure an `endpointClosure` on your provider. In this closure, your `Target` needs to be mapped to an `Endpoint`. `Endpoint` is a semi-internal data structure used by Moya to reason about the request. And in this `Endpoint` is where you are going to be able to specify more details for your testing. More concretely, on its `sampleResponseClosure`.
 
-As we discussed above, the default stubbing behavior is to respond to requests with your sample data with a `200` HTTP status code. This is, because the default `endpointClosure` defines a default `sampleResponseClosure` as follows:
+As we discussed above, the default stubbing behavior is to respond to requests with your sample data with a `200` HTTP status code. This is because the default `endpointClosure` defines its `sampleResponseClosure` as follows:
 
 ```swift
 { .networkResponse(200, target.sampleData) }
 ```
 
-A `sampleResponseClosure` should return a `EndpointSampleResponse`, an it can be:
+If you need to setup your own `sampleResponseClosure`, your implementation should return a case of the `EndpointSampleResponse` enum:
 
 - A `.networkResponse(Int, Data)` where `Int` is a status code and `Data` is the returned data.
  Useful to customize either response codes or data to your own specification.
 - A `.response(HTTPURLResponse, Data)` where `HTTPURLResponse` is the response and `Data` is the returned data.
  Useful to *fully* stub your responses.
-- A `.networkError(NSError)` where `NSError` is the error occurred when sending the request or retrieving a response.
+- A `.networkError(NSError)` where `NSError` is the error that occurred when sending the request or retrieving a response.
  Useful to test for any network errors: Timeouts, reachability issues, etc.
 
 For example, the following code creates a provider to stub with *immediate* `401` responses:
--- a/docs/Testing.md
+++ b/docs/Testing.md
@@ -19,17 +19,19 @@ public var sampleData: Data {
 
 ## `stubClosure`
 
-Sample data is not meant to be the default response for your calls. This substitution will not happen without configuring your provider for stubbing:
+The `stubClosure` property of a `MoyaProvider` is responsible for configuring whether a provider returns an actual or a stubbed response. This configuration is made in this closure by mapping a `TargetType` to a case of the `StubBehavior` enum:
+
+- `.never` to not stub responses.
+- `.immediate` to stub responses immediately.
+- `.delayed(seconds: TimeInterval)` to stub responses after a given delay (to simulate the delays the real network calls may have).
+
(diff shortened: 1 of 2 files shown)


docs/Testing.md
```diff
@@ -0,0 +1,69 @@
+# Testing
+
+Moya has been created with testing at its heart. In this document, you will find all the customization points that will allow you to tailor your testing.
+
+## `sampleData`
+
+When creating your `TargetType` you have been required to provide `sampleData` for your targets. As you already know, all you want to do there is to provide `Data` that represents a sample response from every particular target.
```
pedrovereza: I think we should avoid making assumptions about user knowledge here. Using `As you already know` can cause some bad feelings to users who are not aware of what `sampleData` does/means.

More in [this post](http://alistapart.com/blog/post/the-most-dangerous-word-in-software-development)

docs/Testing.md
```diff
@@ -0,0 +1,69 @@
...
+
+You can also use the `MoyaProvider.delayedStub(:)`. This will allow you to delay the stubbed responses to simulate the delays the real network calls may have. Further, you can also implement your own `StubClosure`. It will allow you to define the stubbing behavior target by target (and/or any condition related to the targets that you may have).
+
+Before you continue, it is worth mentioning that all the sample responses will be delivered with an HTTP status `200` by default.
+
+## `sampleResponseClosure`
+
+With the previous `sampleData` and `stubClosure`, we could only specify the data returned when stubbing. But you have more options.
+
+As you may already know, Moya offers you the opportunity to configure an [`endpointClosure`](https://github.com/Moya/Moya/blob/master/docs/Endpoints.md#from-target-to-endpoint) on your provider. In that closure, you convert from `Target` to `Endpoint`. And this `Endpoint` is where you are going to be able to specify more details for your testing. More concretely, on its `sampleResponseClosure`.
```
pedrovereza: Same point, I think we should avoid using `As you may already know` here 😉

docs/Testing.md
```diff
@@ -0,0 +1,69 @@
...
+A `sampleResponseClosure` should return a `EndpointSampleResponse`, an it can be:
+
+- A `.networkResponse(Int, Data)` where `Int` is a status code and `Data` is the returned data.
+ Useful to customize either response codes or data to your own specification.
+- A `.response(HTTPURLResponse, Data)` where `HTTPURLResponse` is the response and `Data` is the returned data.
+ Useful to *fully* stub your responses.
+- A `.networkError(NSError)` where `NSError` is the error occurred when sending the request or retrieving a response.
+ Useful to test for any network errors: Timeouts, reachability issues, etc.
+
+For example, the following code create a provider to stub with *inmediate* `401` responses:
```
pedrovereza: Typo on `inmediate` 😉 

Also, should it be `creates` instead?

docs/Testing.md
```diff
@@ -0,0 +1,69 @@
...
+
+You can also use the `MoyaProvider.delayedStub(:)`. This will allow you to delay the stubbed responses to simulate the delays the real network calls may have. Further, you can also implement your own `StubClosure`. It will allow you to define the stubbing behavior target by target (and/or any condition related to the targets that you may have).
+
+Before you continue, it is worth mentioning that all the sample responses will be delivered with an HTTP status `200` by default.
+
+## `sampleResponseClosure`
+
+With the previous `sampleData` and `stubClosure`, we could only specify the data returned when stubbing. But you have more options.
+
+Moya offers you the opportunity to configure an [`endpointClosure`](https://github.com/Moya/Moya/blob/master/docs/Endpoints.md#from-target-to-endpoint) on your provider. In that closure, you convert from `Target` to `Endpoint`. And this `Endpoint` is where you are going to be able to specify more details for your testing. More concretely, on its `sampleResponseClosure`.
```
SD10: I'm not sure if we do this in our other docs but I would remove the link. These are hard to track down if our doc structure changes.

docs/Testing.md
```diff
@@ -0,0 +1,69 @@
...
+
+As we discussed above, the default stubbing behavior is to respond to requests with your sample data with a `200` HTTP status code. This is, because the default `endpointClosure` defines a default `sampleResponseClosure` as follows:
```
SD10: Nit: `This is, because` -> `This is because`

docs/Testing.md
```diff
@@ -0,0 +1,69 @@
...
+
+```swift
+{ .networkResponse(200, target.sampleData) }
+```
+
+A `sampleResponseClosure` should return a `EndpointSampleResponse`, an it can be:
```
SD10: `an it can be` -> `and it can be`

SD10: Can we change this line to:
A `sampleResponseClosure` should return a case of the `EndpointSampleResponse` enum:

I just want to mention that this is an enum.

docs/Testing.md
```diff
@@ -0,0 +1,69 @@
...
+
+- A `.networkResponse(Int, Data)` where `Int` is a status code and `Data` is the returned data.
+ Useful to customize either response codes or data to your own specification.
+- A `.response(HTTPURLResponse, Data)` where `HTTPURLResponse` is the response and `Data` is the returned data.
+ Useful to *fully* stub your responses.
+- A `.networkError(NSError)` where `NSError` is the error occurred when sending the request or retrieving a response.
```
SD10: Nit: `error occurred` -> `error that occurred`

docs/Testing.md
```diff
@@ -0,0 +1,69 @@
...
+    case .userRepositories(let name):
+        return "[{\"name\": \"Repo Name\"}]".data(using: String.Encoding.utf8)!
+    }
+}
+```
+
+This can be used later for tests or just for providing offline support while developing.
+
+## `stubClosure`
+
```
SD10: I feel like this section is a little unclear but we could expand on this later in another PR?  I would like it to feel a bit more explanatory.


> The `stubClosure` property of a `MoyaProvider` is responsible for configuring whether a provider returns an actual or stubbed response. The `stubClosure` is responsible for mapping a `TargetType` to a case of the `StubBehavior` enum:

Then I would list all cases of `StubBehavior` and explain what each of them does.

I just think it's clearer this way and also matches the setup of the next section. 😓

SD10:
@fbeeper Don't worry if some of my feedback is asking too much. Just the nitty ones are blocking! 😆 Thanks for doing this 👍

docs/Testing.md
```diff
@@ -0,0 +1,69 @@
...
+```swift
+public var sampleData: Data {
+    switch self {
+    case .userRepositories(let name):
+        return "[{\"name\": \"Repo Name\"}]".data(using: String.Encoding.utf8)!
+    }
+}
+```
+
+This can be used later for tests or just for providing offline support while developing.
```
sunshinejr: I would move this line to the paragraph before "For example", as the last sentence.

docs/Testing.md
sunshinejr: Yeah I like it better as well.

docs/Testing.md
sunshinejr: Yeah, and also I'd add a quick explanation what this closure does as well.

docs/Testing.md
sunshinejr: Right now the sentence doesn't feel connected to the paragraph above, at least for me. Maybe we could change it to something like:
> If you want to setup your own `sampleResponseClosure`, you have few options to return in it:

Or any other sentence that goes: _"this is a default one, but you could also specify one of the following"._

sunshinejr:
First of all thanks so much for this piece @fbeeper! This is a huge amount of work and many years in the making so you did a great job 👏 

I've left few comments myself as well - I know that with all of us commenting this is a lot of work, but I hope we can get this finished.

fbeeper:
Thanks for your kind comments, patience, and feedback. I really appreciate your attention to detail!

I tried my best at incorporating your notes, including a rewrite of the `stubClosure` section because I really agreed with @SD10's comment about it. This way is more detailed and formal. 

Hope I didn't introduce more typos 😊

ashfurrow:

BasThomas:
I 😍 this, thanks so much @fbeeper!

PR #1642, opened 2018-04-17
Workaround to fix long Carthage build times

### LINKED ISSUE #345: Ability to checkout only specific type of Moya via Carthage
Similar to way it's done via CocoaPods where podspec contains different types(Core/Rx/Reactive).

Right now when we pull Moya via Carthage — all 3 versions are downloaded and compiled (as well as specific dependencies for each one). 
Obviously, that is much more time-consuming than pulling only 1 version.
Example: `carthage update --platform tvos  856.54s user 145.58s system 445% cpu 3:44.89 total`
And yup, i think this is not highest-priority thing because this is an issue mostly for first Carthage run with Moya at Cartfile.

I've checked Carthage and found zero answers about such possibility. 

From all stuff i wrote above i have 3 different assumptions: 

* I'm wrong (feel free to tell me) 
* This is an issue and Carthage actually has way to solve this 
* This is an issue and Carthage has no ways to solve this (probably i'll create an issue on it's repo than). 

I had a thought about dealing with that via 
* Tags 
* Separate branches (seems to be painful to support) 

But i just hope to get any advice before starting to dig into that without even knowing if it's reasonable to spend time on this. 

Thanks for attention :)

Dschee:
Wow, I can't believe I waited so long for https://github.com/Carthage/Carthage/issues/1227 to be tackled. But I've given up on that long ago after even trying to [fix it myself](https://github.com/Carthage/Carthage/pull/1990) (with no success). Now, I just started [developing a new framework](https://github.com/JamitLabs/LingoHubSync) supporting multiple platforms and plan to use plain Moya (without the reactive extensions). And here's what I saw:

```
*** Building scheme "Alamofire iOS" in Alamofire.xcworkspace
*** Building scheme "Alamofire macOS" in Alamofire.xcworkspace
*** Building scheme "Alamofire watchOS" in Alamofire.xcworkspace
*** Building scheme "Alamofire tvOS" in Alamofire.xcworkspace
*** Building scheme "Result-Mac" in Result.xcodeproj
*** Building scheme "Result-iOS" in Result.xcodeproj
*** Building scheme "Result-watchOS" in Result.xcodeproj
*** Building scheme "Result-tvOS" in Result.xcodeproj
*** Building scheme "ReactiveSwift-macOS" in ReactiveSwift.xcworkspace
*** Building scheme "ReactiveSwift-watchOS" in ReactiveSwift.xcworkspace
*** Building scheme "ReactiveSwift-tvOS" in ReactiveSwift.xcworkspace
*** Building scheme "ReactiveSwift-iOS" in ReactiveSwift.xcworkspace
*** Building scheme "RxBlocking-tvOS" in Rx.xcworkspace
*** Building scheme "RxBlocking-macOS" in Rx.xcworkspace
*** Building scheme "RxBlocking-iOS" in Rx.xcworkspace
*** Building scheme "RxBlocking-watchOS" in Rx.xcworkspace
*** Building scheme "RxCocoa-macOS" in Rx.xcworkspace
*** Building scheme "RxCocoa-watchOS" in Rx.xcworkspace
*** Building scheme "RxCocoa-tvOS" in Rx.xcworkspace
*** Building scheme "RxCocoa-iOS" in Rx.xcworkspace
*** Building scheme "RxSwift-macOS" in Rx.xcworkspace
*** Building scheme "RxSwift-tvOS" in Rx.xcworkspace
*** Building scheme "RxSwift-watchOS" in Rx.xcworkspace
*** Building scheme "RxSwift-iOS" in Rx.xcworkspace
*** Building scheme "RxTests-macOS" in Rx.xcworkspace
*** Building scheme "RxTests-tvOS" in Rx.xcworkspace
*** Building scheme "RxTests-iOS" in Rx.xcworkspace
*** Building scheme "Moya" in Moya.xcodeproj
*** Building scheme "ReactiveMoya" in Moya.xcodeproj
Build Failed
	Task failed with exit code 65:
	/usr/bin/xcrun xcodebuild -project /Users/JamitLabs/Code/Apple/LingoHubSync/Carthage/Checkouts/Moya/Moya.xcodeproj -scheme ReactiveMoya -configuration Release -derivedDataPath /Users/JamitLabs/Library/Caches/org.carthage.CarthageKit/DerivedData/9.3_9E145/Moya/11.0.2 -sdk watchos ONLY_ACTIVE_ARCH=NO BITCODE_GENERATION_MODE=bitcode CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY= CARTHAGE=YES archive -archivePath /var/folders/2l/3tqwt0g91x18hql0jwsz0y280000gn/T/Moya SKIP_INSTALL=YES GCC_INSTRUMENT_PROGRAM_FLOW_ARCS=NO CLANG_ENABLE_CODE_COVERAGE=NO STRIP_INSTALLED_PRODUCT=NO (launched in /Users/JamitLabs/Code/Apple/LingoHubSync/Carthage/Checkouts/Moya)

This usually indicates that project itself failed to compile. Please check the xcodebuild log for more details: /var/folders/2l/3tqwt0g91x18hql0jwsz0y280000gn/T/carthage-xcodebuild.TakmdP.log
Command '/bin/bash -c "carthage update --platform iOS,tvOS,macOS,watchOS --cache-builds"' returned with error code 1.
⚠️  Command '/bin/bash -c /Users/JamitLabs/Documents/beak/builds/_Users_JamitLabs_Code_Apple_LingoHubSync/.build/debug/BeakFile' returned with error code 1.
```

Not only does Moya compile dozens of frameworks I don't need, but it also fails building one of them. This is plain frustrating so even though #345 was closed long time ago, I think we should finally tackle this for those Moya users who don't ever want to run into this issue again.

**And here's my simple yet working solution:**

Let's **introduce support branches** for each Moya variant (plain Moya, RxMoya, ReactiveMoya) and document that people could opt for using those branches instead of the actual git versioning. So long as we clearly state the advantage and disadvantage it's the developers choice to go for one or the other.

I have actually already created these branches based on the latest release (11.0.2), see `carthage/moya/11.x`, `carthage/rxmoya/11.x` and `carthage/reactivemoya/11.x`. I've just also tested how a build with `carthage/moya/11.x` looks like:

```
*** Building scheme "Alamofire iOS" in Alamofire.xcworkspace
*** Building scheme "Alamofire macOS" in Alamofire.xcworkspace
*** Building scheme "Alamofire watchOS" in Alamofire.xcworkspace
*** Building scheme "Alamofire tvOS" in Alamofire.xcworkspace
*** Building scheme "Result-Mac" in Result.xcodeproj
*** Building scheme "Result-iOS" in Result.xcodeproj
*** Building scheme "Result-watchOS" in Result.xcodeproj
*** Building scheme "Result-tvOS" in Result.xcodeproj
*** Building scheme "Moya" in Moya.xcodeproj
```

Wow, it's much faster and it doesn't fail! Yay 🎉 

Please note the following:
* As you can see in the docs, I've marked this workaround as Beta as we don't have much experience with it yet. We can remove that once we have.
* In order for this to work, we would need to rebase all three branches every time we do a release. I'm not sure if we can automate this properly though, so someone might need to do that manually (and I would actually volunteer myself at the beginning).
* The rationale behind the branch naming had three goals:
  1. It should be clear that it's only for Carthage support.
  2. The name of the Moya variant should be clear.
  3. The major version should be clear.

I hope you like the idea. What do you think? Especially @Moya/core-team?

Fixes #345, finally.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -139,7 +139,33 @@ Then run `carthage update`.
 
 If this is your first time using Carthage in the project, you'll need to go through some additional steps as explained [over at Carthage](https://github.com/Carthage/Carthage#adding-frameworks-to-an-application).
 
-> NOTE: At this time, Carthage does not provide a way to build only specific repository submodules. All submodules and their dependencies will be built with the above command. However, you don't need to copy frameworks you aren't using into your project. For instance, if you aren't using `ReactiveSwift`, feel free to delete that framework along with `ReactiveMoya` from the Carthage Build directory after `carthage update` completes. Or if you are using `ReactiveSwift` but not `RxSwift`, then `RxMoya`, `RxTest`, `RxCocoa`, etc. can safely be deleted.
+> NOTE: At this time, Carthage does not provide a way to build only specific repository submodules. All submodules and their dependencies will be built with the above command. However, you don't need to copy frameworks you aren't using into your project. For instance, if you aren't using `ReactiveSwift`, feel free to delete that framework along with `ReactiveMoya` from the Carthage Build directory after `carthage update` completes. Or if you are using `ReactiveSwift` but not `RxSwift`, then `RxMoya`, `RxTest`, `RxCocoa`, etc. can safely be deleted. Read the next section for a workaround to this.
+
+#### Workaround to enhance build time (Beta)
+
+If you want Carthage to only build a specific variant of Moya with only the dependencies needed for that, we have a workaround for you that's less flexible regarding version specification but still provides you with minor and patch updates to Moya without editing the `Cartfile`.
+
+For plain Moya, use this entry instead:
+
+```
+github "Moya/Moya" "carthage/moya/11.x"
+```
+
+This will checkout the branch `carthage/moya/11.x` which only includes the plain Moya version and its dependencies leaving to a reduced build time. When a new minor or patch version of Moya is released, we will rebase that branch onto the latest release. New major releases will get their own branch.
+
+For RxMoya you would use this:
+
+```
+github "Moya/Moya" "carthage/rxmoya/11.x"
+```
+
+For ReactiveMoya, use this:
+
+```
+github "Moya/Moya" "carthage/reactivemoya/11.x"
+```
+
+NOTE: Please be aware that we don't have long-time experience with this workaround yet and might decide against it and remove it entirely. Once the solution proves to work fine with updates, we will remove the Beta status. We might also remove it once [this issue](https://github.com/Carthage/Carthage/issues/1227) on Carthage is solved.
 
 ### Manually
 


pedrovereza:
I agree with @ashfurrow on this one.

I can see how the branch approach works and it's in fact a nice workaround, but I'd like to see if the tag approach suggested by Ash would work. If it does, I have a feeling it would end up being a cleaner solution to the problem 👍

SD10:
@Dschee Kudos on trying to enhance the experience for Carthage users. As others have mentioned, my only concern with this is regarding the cost of maintaining 3 branches with every release 😢

Dschee:
Thank you guys for the quick feedback.

I can absolutely understand your concerns. And using a tag-based approach was my initial idea, too. But we're currently right before a beta release and that's why being fast (both in building and finding a working solution) mattered so much. And to find a tag-based solution I would need to check Carthage first to make sure it can actually find those tags and update versions automatically. But I would also need to make sure those tags don't interfere with CocoaPods or SPM.

I agree with you though and thinking long term this will probably be the better approach, so I will investigate this once our release is done (currently planned for tomorrow). I'll update this PR once I have a working solution. Thanks again for your feedback!

Dschee:
I just had a quick look into how Carthage version specification behaves and found that there's currently ongoing work to change it to fully comply to SemVer 2.0 [here](https://github.com/Carthage/Carthage/pull/2412). I think we could use the metadata option and write something like `11.0.2+moya`, `11.0.2+rxmoya` and `11.0.2+reactivemoya` and note in the Readme, that people who want fast buildtimes would need to check for updates manually. The big downside is, that when trying to use this approach in a framework specifying a specific version is very bad for dependency graph resolution ...

Therefore it would be great to get support for something like `~> 11.0.2+rxmoya` first into Carthage alongside the suggested solution. I just asked for that [here](https://github.com/Carthage/Carthage/pull/2412#issuecomment-382266450).

ashfurrow:
Hmm, the plot thickens! I think that another, more short-term approach _could_ be to tell users to fork Moya and remove the schemes they don't need. It would solve the immediate problem and share that solution with others while we work towards a more maintainable one. Any thoughts?

ashfurrow:
Oh, interesting idea! Hmm, that _would_ be kind of nice, yeah. I'd like to hear what other Moya members – particularly Carthage users! – have to say about it before moving forward, but it sounds promising!

sunshinejr:
Sorry, but I'm not really following how these branches would work without any actions from us. Could you please a little bit more? 

Apart from that, I still think that it should be fixed on Carthage level - this is just adding overhead to our organization that we need to maintain somehow. Plus these additional repos might be confusing - we would need to add guides or automation etc. And every other repo that uses similar approach would need to do the same, this way we apply even less pressure to Carthage to change it.

But maybe you could create the forks on your account and we could add a note to the README that "as a current workaround for Carthage building all schemes" they could use these branches. This way we could say that we officially support it, people have a workaround, we don't confuse contributors and you'll be commited to the project as an owner. I think that is a decent compromise.

SD10:
I feel like maintaining three new repositories would actually lead to more overhead than the original solution of using branches. I'm not convinced that Moya needs to bear the burden of what seems to be a shortcoming in Carthage. As @sunshinejr mentioned, I think this needs to be fixed on a Carthage level if it is a significant issue for its users.

I also think what @ashfurrow is the most reasonable approach here:
> I think that another, more short-term approach could be to tell users to fork Moya and remove the schemes they don't need.

Dschee:
I completely agree with everybody who blames Carthage for not fixing this themselves. But they haven't done anything to solve this since more than two years now (see https://github.com/Carthage/Carthage/issues/1227) so I'm not sure if they feel like there is any pressure on them at all. And most dependencies I've used don't have the problem that they have reactive extensions which takes ages to build (at least that's what it feels like), thus I'm sure some people over there at Carthage would argue that it **actually is** a Moya issue.

Having this said, here's the overview of possbilities we discussed until now (as I see it):

1. We create forks in the main Moya repo and rebase them on version updates via CI.
  **Pro:** The basic problem is fixed. **Con:** It's work to configure once. Will `bootstrap` even work after a rebase?
2. We require every Moya user interested in fast build times to create the forks themselves.
  **Pro:** We don't have to lift a finger and simply blame Carthage. **Con:** User madness.
3. We automate the CI and use repos on the Moya organization to make this happen (as suggested).
  **Pro:** Users have all SemVer updating possibilities. **Con:** It's work to configure this once.
4. We don't automate the CI and use forks maintained by me to make this happen.
  **Pro:** Users have all SemVer updating possibilities. **Con:** Updates will be delayed.

I actually just created a "konsensing" where you can submit more suggestions until Sunday and rate the options until Monday. The results will be available once the rating deadline is over. See here:
https://www.konsensieren.eu/en/konsensierung/kd8mvrdnd/anzeigen

Note: A "konsensing" is an innovative democratic decision making idea with the goal to find the best consensus which makes everybody happy instead of finding just the idea with the most thumbs up. Unfortunately I couldn't find any English translation to the great German expalantion site [here](http://www.sk-prinzip.eu/). But maybe the [Google Translated Version](https://translate.google.com/translate?sl=de&tl=en&js=y&prev=_t&hl=de&ie=UTF-8&u=http%3A%2F%2Fwww.sk-prinzip.eu%2F&edit-text=&act=url) is good enough?

ashfurrow:
Thanks for the summary, very clear! I’ll check out those links. 

One thing I’d like to clarify is that I don’t “blame” Carthage or its developers for not adopting this feature, just as I don’t think they blame Moya for creating need for skipped schemes. As open source communities, we are both entitled to set our own priorities and our own ways of running things; it just so happens that we don’t line up in this case. As you have correctly pointed out, whatever decision we make will involve trade offs, and while I don’t agree with the trade offs that Carthage has made, I 100% respect their decision.

ashfurrow:
@Moya/contributors As [summarized above](https://github.com/Moya/Moya/pull/1642#issuecomment-382993368), we are discussing different approaches for improving build times for Moya's Carthage users. Any feedback you have, especially if you integrate Moya using Carthage, is valuable and we'd like to hear from you!

freak4pc:
It seems to me unfortunately this is not in Carthage's roadmap at all. I kinda get a vibes that PRs aren't easily (or friendly) supported by that community as well. I hope I'm wrong with the impression I got.

Dschee:
Hey guys, just a small update: I don't know why right now, but somebody finally helped me finishing my PR over at Carthage which adds a feature to skip specific schemes when building: https://github.com/Carthage/Carthage/pull/1990. I think, in case that the PR gets merged, we have another alternative:

Documenting that people can add a `Cartfile.ignore` in order to skip building the extensions they don't need like so:

```
# Ignore RxSwift variant of Moya
scheme "RxBlocking-tvOS"
scheme "RxBlocking-macOS"
scheme "RxBlocking-iOS"
scheme "RxBlocking-watchOS"
scheme "RxCocoa-macOS"
scheme "RxCocoa-watchOS"
scheme "RxCocoa-tvOS"
scheme "RxCocoa-iOS"
scheme "RxSwift-macOS"
scheme "RxSwift-tvOS"
scheme "RxSwift-watchOS"
scheme "RxSwift-iOS"
scheme "RxTests-macOS"
scheme "RxTests-tvOS"
scheme "RxTests-iOS"
scheme "RxMoya"

# Ignore ReactiveSwift variant of Moya
scheme "ReactiveSwift-macOS"
scheme "ReactiveSwift-watchOS"
scheme "ReactiveSwift-tvOS"
scheme "ReactiveSwift-iOS"
scheme "ReactiveMoya"
```

Far away from perfect, but it's working and has higher chances to get merged fast than a "real" dependency-graph based solution.

ashfurrow:
Great work 👏 Please keep us informed about your progress.

Dschee:
I'm closing this since I now have a better solution for this: A new SwiftPM based dependency manager!

See #1834 for more details.

Issue #1635, opened 2018-04-10
Allow customization of formatting in NetworkLoggerPlugin


alanyjw:
✨ 

NetworkLoggerPlugin currently provides the ability to inject a different `output` closure and it will be even better if the customization extends to formatting.

There's a couple of reasons for this:
- Use `NSLog` instead of `print`. We would like to ship these logs to Crashlytics and it supports `NSLog` for logging.
- Omit `date`. `NSLog` already provides timestamp.

I believe there may be other metadata which would be useful for debugging purposes and having the ability to customize it would be helpful 😄

SD10:
@alanyjw To clarify, are you suggesting both support for `NSLog` and `Print` via the output closure? If so, think this is a reasonable enhancement.

alanyjw:
@SD10 Not exactly.

What I meant was to change the initializer to include a `outputFormat` parameter:

```swift
public init(verbose: Bool = false,
            cURL: Bool = false,
            output: ((_ separator: String, _ terminator: String, _ items: Any...) -> Void)? = nil,
            outputFormat: ((_ loggerId: String, date: String, identifier: String, message: String) -> String)? = nil, // add this parameter
            requestDataFormatter: ((Data) -> (String))? = nil,
            responseDataFormatter: ((Data) -> (Data))? = nil) {
        self.cURL = cURL
        self.isVerbose = verbose
        self.output = output ?? NetworkLoggerPlugin.reversedPrint
        self.outputFormat = outputFormat ?? NetworkLoggerPlugin.printFormatter
        self.requestDataFormatter = requestDataFormatter
        self.responseDataFormatter = responseDataFormatter
    }
```

Where the fallback implementation will be what it is in the codebase:

```swift
fileprivate extension NetworkLoggerPlugin {
    static func printFormatter(_ loggerId: String, date: String, identifier: String, message: String) -> String {
        return "\(loggerId): [\(date)] \(identifier): \(message)"
    }
}
```

Let me know if this sounds reasonable and I'll submit a PR 😄

SD10:
@alanyjw Sounds good to me 👍 A PR would be welcome

PR #1644, opened 2018-04-19
Allow customization of formatting in NetworkLoggerPlugin

alanyjw:
Enhancement #1635

### DIFF
--- a/Sources/Moya/Plugins/NetworkLoggerOutput.swift
+++ b/Sources/Moya/Plugins/NetworkLoggerOutput.swift
@@ -1,0 +1,25 @@
+//
+//  NetworkLoggerOutput.swift
+//  Moya
+//
+//  Created by Alan Yeo on 24/4/18.
+//
+
+import Foundation
+
+public protocol NetworkLoggerOutput {
+    func print(_ separator: String, terminator: String, items: Any...)
+    func format(_ loggerId: String, date: String, identifier: String, message: String) -> String
+}
+
+class DefaultNetworkLoggerOutput: NetworkLoggerOutput {
+    func print(_ separator: String, terminator: String, items: Any...) {
+        for item in items {
+            Swift.print(item, separator: separator, terminator: terminator)
+        }
+    }
+
+    func format(_ loggerId: String, date: String, identifier: String, message: String) -> String {
+        return "\(loggerId): [\(date)] \(identifier): \(message)"
+    }
+}
--- a/Sources/Moya/Plugins/NetworkLoggerPlugin.swift
+++ b/Sources/Moya/Plugins/NetworkLoggerPlugin.swift
@@ -10,6 +10,7 @@ public final class NetworkLoggerPlugin: PluginType {
     fileprivate let terminator = "\n"
     fileprivate let cURLTerminator = "\\\n"
     fileprivate let output: (_ separator: String, _ terminator: String, _ items: Any...) -> Void
+    fileprivate let outputFormat: (_ loggerId: String, _ date: String, _ identifier: String, _ message: String) -> String
     fileprivate let requestDataFormatter: ((Data) -> (String))?
     fileprivate let responseDataFormatter: ((Data) -> (Data))?
 
@@ -21,11 +22,13 @@ public final class NetworkLoggerPlugin: PluginType {
     public init(verbose: Bool = false,
                 cURL: Bool = false,
                 output: ((_ separator: String, _ terminator: String, _ items: Any...) -> Void)? = nil,
+                outputFormat: ((_ loggerId: String, _ date: String, _ identifier: String, _ message: String) -> String)? = nil,
                 requestDataFormatter: ((Data) -> (String))? = nil,
                 responseDataFormatter: ((Data) -> (Data))? = nil) {
         self.cURL = cURL
         self.isVerbose = verbose
         self.output = output ?? NetworkLoggerPlugin.reversedPrint
+        self.outputFormat = outputFormat ?? NetworkLoggerPlugin.printFormatter
         self.requestDataFormatter = requestDataFormatter
         self.responseDataFormatter = responseDataFormatter
     }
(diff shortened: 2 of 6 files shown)


Sources/Moya/Plugins/NetworkLoggerPlugin.swift
```diff
@@ -18,10 +19,16 @@ public final class NetworkLoggerPlugin: PluginType {
     public let cURL: Bool
 
     /// Initializes a NetworkLoggerPlugin.
-    public init(verbose: Bool = false, cURL: Bool = false, output: ((_ separator: String, _ terminator: String, _ items: Any...) -> Void)? = nil, requestDataFormatter: ((Data) -> (String))? = nil, responseDataFormatter: ((Data) -> (Data))? = nil) {
+    public init(verbose: Bool = false,
+                cURL: Bool = false,
+                output: ((_ separator: String, _ terminator: String, _ items: Any...) -> Void)? = nil,
+                outputFormat: ((_ loggerId: String, _ date: String, _ identifier: String, _ message: String) -> String)? = nil,
```
SD10: This looks good 👍 I'm just curious if another Moya contributor thinks that we should extract this to a `Formatter` object at this point.

SD10:
@alanyjw Looks like there is a problem with the build. I also changed the target branch to `development` because this is an API breaking change.

Sources/Moya/Plugins/NetworkLoggerPlugin.swift
sunshinejr: We could. We could also extract `output` as it also is not that clear. Maybe we could create two `typealiases` for these?

alanyjw: @SD10 Are you thinking of a `Formatter` object that is responsible for both `output` and `outputFormat`?

@sunshinejr If both of you are keen on the separate `Formatter` object, then I guess there wouldn't be a need for the refactor you are proposing?

sunshinejr: Oh, if you are talking about both `output` & `ouputFormat` then I guess yeah 👍 Maybe the `Formatter` object name a bit more verbose, like `Output` as it is an output + output format.

alanyjw:
@SD10 @sunshinejr I've made the changes based on the suggestions. The build seems to be failing due to an email verification issue with GitHub. It doesn't seem to be from my account because I've verified mine 🤔

SD10:
@alanyjw I think it's possible that this is due to you working off of the `master` branch and not `development`. Just a wild guess

Sources/Moya/Plugins/NetworkLoggerOutput.swift
```diff
@@ -0,0 +1,25 @@
+//
+//  NetworkLoggerOutput.swift
+//  Moya
+//
+//  Created by Alan Yeo on 24/4/18.
+//
+
+import Foundation
+
+public protocol NetworkLoggerOutput {
```
SD10: Nit: Think we need some work on naming here. 
Maybe `NetworkLoggerOutputFormatter`? Need advice from another contributor

Sources/Moya/Plugins/NetworkLoggerOutput.swift
```diff
@@ -0,0 +1,25 @@
...
+    func print(_ separator: String, terminator: String, items: Any...)
+    func format(_ loggerId: String, date: String, identifier: String, message: String) -> String
+}
+
+class DefaultNetworkLoggerOutput: NetworkLoggerOutput {
```
SD10: Can't we just make this an `open` class and remove the protocol?
`DefaultNetworkLoggerOutput` -> `NetworkLoggerOutputFormatter`

Sources/Moya/Plugins/NetworkLoggerPlugin.swift
```diff
@@ -18,17 +18,21 @@ public final class NetworkLoggerPlugin: PluginType {
     public let cURL: Bool
 
     /// Initializes a NetworkLoggerPlugin.
-    public init(verbose: Bool = false, cURL: Bool = false, output: ((_ separator: String, _ terminator: String, _ items: Any...) -> Void)? = nil, requestDataFormatter: ((Data) -> (String))? = nil, responseDataFormatter: ((Data) -> (Data))? = nil) {
+    public init(verbose: Bool = false,
+                cURL: Bool = false,
+                output: NetworkLoggerOutput = DefaultNetworkLoggerOutput(),
+                requestDataFormatter: ((Data) -> (String))? = nil,
+                responseDataFormatter: ((Data) -> (Data))? = nil) {
```
SD10: If we made it a `NetworkLoggerFormatter` we could throw these 2 properties in there as well to remain consistent

SD10: Possibly?
`public init(verbose: Bool = false, cURL: Bool = false, formatter: NetworkLoggerFormatter = .defaultFormatter)`

Sources/Moya/Plugins/NetworkLoggerOutput.swift
sunshinejr: Yeah, I agree.

Sources/Moya/Plugins/NetworkLoggerPlugin.swift
sunshinejr: I like the idea for the generic `Formatter` class, but imo all of the properties should have a default value, as I don't usually want to override everything. Also, how do you feel about namespacing the `Formatter` object in the `NetworkLoggerPlugin`?
```swift
public final class NetworkLoggerPlugin: PluginType {
    class Formatter {}
}
```

sunshinejr:
@SD10 sorry, but it's kinda not clear to me: how would the branch impact the build fail? Didn't investigate it _yet_, but didn't see that error before 😅

Sources/Moya/Plugins/NetworkLoggerPlugin.swift
SD10: I'm fine with namespacing it, but then it can't be `open`. All the properties would have default properties, declared as vars so you can assign a different closure to them.

SD10:
Yeah ignore what I said about CI, this is an odd one. The error isn't even the same as the last build.

SD10:
I'm not sure about these changes making it in Moya 12.0. Anyone else have thoughts at this point?

sunshinejr:
I know it's a huge delay, but would you like to finish the works on this one @alanyjw? I think it would be valuable addition to Moya 13.

alanyjw:
@sunshinejr When is the release cutoff date? I'll see if I could work towards that date.

sunshinejr:
@alanyjw there is no definitive date for the moment, but we will try to align our release with new Alamofire 5 I think. No pressure tho, let us know if you’d need any help :)

sunshinejr:
This was added in #1906. Thanks a lot for the time you put into this! 🧡

PR #1656, opened 2018-04-26
Update project Swift version && CocoaPods integration

sunshinejr:
So `.swift_version` is deprecated now and we shall use `swift_version` in `*.podspec` instead. The thing is that this option is available only from `CocoaPods` 1.4, thus the requirement as well.

Additionally, I changed the projects Swift version to 4.0 (it was still pointing to 3.0). Oh, and also updated `flatMap` to `compactMap` in Tests, fortunately we don't have these in the Moya core so we can still support Swift 4.0 and 4.1 without any shimming at all.


sunshinejr:
@SD10 Oh, right. This is tricky because `SwiftEnv` still uses it but we don't really use SwiftEnv yet (was thinking that I implemented the Swift 3 & Swift 4 tasks in CircleCI, but I didn't 😅). Updated!

SD10:
Looks like this is failing CI though 🤔

sunshinejr:
Hmm, really not sure what the error is. Was suspecting that Carthage is the problem since it had to rebuild all the dependencies and it always crashed in that case. Might need a deeper investigation :shipit:

simonrice:
Chipping in a little here with the CI build issue, the issue at hand is Carthage attempts to build your project with dependencies without actually checking them out - as a result, Alamofire, Result, etc. is not found given you have no cache for your nice new Xcode 9.3 based CI, so your build fails.  This isn't a problem on the Xcode 9.2 based CI as there's already a valid cache, so the dependencies don't need building.

Slightly further down the build process, a `carthage bootstrap` phase is run, which checks out the dependencies and once again builds them (therefore, I believe, it's equivalent to `carthage checkout && carthage build`).  I wonder if this should be moved up the build process and replace the `carthage build` step.

SD10:
Thanks for adding some input @simonrice 👍 I'm just starting to get familiar with Carthage now but if anyone wants to try and debug this that would be totally awesome. I probably won't have time for any open source work until mid next month.

pedrovereza:
@sunshinejr You may wanna check out branch `circle-workflow` 😉

sunshinejr:
Okay so the build passes now! I've made a separate job for carthage integration, we should do few more of these like @pedrovereza in his branch, but for this PR I think it's enough.

The only problem for now is that Danger doesn't pass, although this is a problem for quite some time now. @SD10 might it be something connected to the token capabilities? 🤔  We should resolve this as we are missing out on PR checks.

pedrovereza:
@sunshinejr I pushed a change to rollback parallelism to 1. CircleCI currently does not support parallelism for macOS/iOS, and the build simply doesn't start (your last commit was queued for 5h until I killed it 😄). I faced the same problem when I was playing on the `circle-workflow` branch 😉 

image

sunshinejr:
Oh, didn't know that, thanks @pedrovereza! Is it somewhere documented and maybe is there a road map that would have this option available for macOS as well?

Also, I've sent an email about this seed plan as well, we'll update once I got the info.

pedrovereza:
@sunshinejr They have [a note, but on version `1.0`](https://circleci.com/docs/1.0/ios-builds-on-os-x/#constraints-on-macos-based-builds), but I couldn't find anything on `2.0`. 

I wasted almost an hour waiting for my build to start after I tried to change parallelism config. Only worked after I reverted the change 😅 (cc. @orta @ashfurrow)

SD10:
@sunshinejr In regards to access tokens, I deleted the one for Aeryn when we switched to peril. Is it possible that they were using the same token? I'm just throwing this comment out there blindly

sunshinejr:
@SD10 Hm, not sure if I understand. We don't have MoyaBot right now and we use Peril for both issues/PRs, right? So in theory they should use the same token?

ashfurrow:
Yeah so we currently have two Danger's running, which is totally okay:

- Peril, the server, runs on webhook events.
- Danger, [the Dangerfile](https://github.com/Moya/Moya/blob/master/Dangerfile), runs on CI.

It's the second one that's causing problems. I looked and the email associated with the account Danger uses, @moyabot, doesn't have a verified email address. I'm taking care of that now and will retrigger CI when I think it's ready to go.

sunshinejr:
Oh, okay, that explains everything, thanks @ashfurrow! If we are at it, could we add all tokens/credentials of our bots to our 1password keychain?

ashfurrow:
Good idea! I'll do that. I have retriggered the build.

ashfurrow:
@sunshinejr That's been added.

ashfurrow:
Haha, careful what you wish for I guess 😄

SD10:
@sunshinejr @pedrovereza @ashfurrow Thanks so much for getting everything working again 🎉 This was really stressing me out 😅 You guys rock!

sunshinejr:
Thanks for the help guys! A little bit overtime, but seems like we’ve got it covered! I think we can fix the linter problems in another PR, so if you could check one last time - it would be awesome 😎

PR #1660, opened 2018-05-03
Issue templates

sunshinejr:
So GitHub is testing an early preview of multiple Issue templates. Basically, we can set up multiple templates for Bugs/Questions/Feature requests and so on - I think these are great. Then the user that want to create an issue would see something similar to the UI below:

image

What I proposed in this PR are just templates directly from GitHub - we might want to evaluate what we could use from these and what we want to add from our own `Issue_template.md`.

### DIFF
--- a/.github/ISSUE_TEMPLATE/Bug_report.md
+++ b/.github/ISSUE_TEMPLATE/Bug_report.md
@@ -1,0 +1,35 @@
+---
+name: Bug report
+about: Create a report to help us improve Moya
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behaviour:
+1. Go to '...'
+2. Click on '....'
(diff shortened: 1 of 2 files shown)


BasThomas:
I am not a big fan. I discussed MessageKit's issue template a while back with @SD10, which they feel is not working like they'd want to.

In the [original issue template](https://github.com/Moya/Moya/pull/704) I intentionally made the barrier for creating an issue as low as possible, and these "please fill out this form" will not make it easier, I am afraid.

That being said, do you feel like there is a specific area of improvement? Maybe we can take a look at how we can tackle that?

sunshinejr:
I think that the templates are still just templates, you don't necessarily have to stick fully to the template - it should help in forming your opinion or bug report. Anyway, I would still pursue the idea of multiple issue templates, but I don't really have time to finish it, so I'm gonna close it 😢

PR #1668, opened 2018-05-19
Bump dependencies for Result and ReactiveSwift

sunshinejr:
Fixes #1657.

**Edit:** This was a live contribution on the Wro Open Source conference, now I see that I forgot about updating the lockfiles, will do that later on :D

PR #1670, opened 2018-05-20
Adds a danger rule to remind us to include resolved files when updating

pedrovereza:
Saw #1668 and thought it was weird that @MoyaBot didn't drop by 😄

### DIFF
--- a/Dangerfile
+++ b/Dangerfile
@@ -52,6 +52,8 @@ end
 podspec_updated = !git.modified_files.grep(/Moya.podspec/).empty?
 cartfile_updated = !git.modified_files.grep(/Cartfile/).empty?
 cartfile_resolved_updated = !git.modified_files.grep(/Cartfile.resolved/).empty?
+package_updated = !git.modified_files.grep(/Package.swift/).empty?
+package_resolved_updated = !git.modified_files.grep(/Package.resolved/).empty?
 
 if podspec_updated && (!cartfile_updated || !cartfile_resolved_updated)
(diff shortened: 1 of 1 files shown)


Dangerfile
```diff
@@ -52,6 +52,8 @@ end
 podspec_updated = !git.modified_files.grep(/Moya.podspec/).empty?
 cartfile_updated = !git.modified_files.grep(/Cartfile/).empty?
 cartfile_resolved_updated = !git.modified_files.grep(/Cartfile.resolved/).empty?
+package_updated = !git.modified_files.grep(/Package.swift/).empty?
+package_resolved_updated = !git.modified_files.grep(/Package.resolved/).empty?
 
 if podspec_updated && (!cartfile_updated || !cartfile_resolved_updated)
```
sunshinejr: Would you be up for adding SPM to the checks here as well? Seems like we check if Carthage was updated and CP not or CP was updated and Carthage not.

sunshinejr:
Thanks @pedrovereza this is great! Let me know how do you feel about my comment, otherwise it's good to go 😉

Dangerfile
SD10: Yeah, I think this could be a beneficial addition. If we don't add it here, let's open an issue for it

pedrovereza:
@sunshinejr @SD10 Agree, the addition makes sense! If any of you wanna go ahead and do it, feel free to push to this branch! I probably won't have time to do it until Thursday 😕

PR #1680, opened 2018-06-10
[WIP] Adding Danger xcode_sumary 

LucianoPAlmeida:
Hey guys :)) 
Related to #1559 
For now is just a test to see if it works :))


SD10:
@LucianoPAlmeida Thanks for taking this on 💪 Can you squash all these testing commits into one commit when you get things working?

Cartfile.resolved
```diff
@@ -1,7 +1,7 @@
-github "Alamofire/Alamofire" "4.7.1"
```
BasThomas: Is this supposed to be in this pull request?

Dangerfile
```diff
@@ -77,3 +77,11 @@ end
 
 # Run SwiftLint
 swiftlint.lint_files
+
+# Xcode summary
+xcode_summary.ignored_results { |result|
+  result.message.start_with?('ld') # Ignore ld_warnings
```
BasThomas: Why?

Rakefile
```diff
@@ -73,13 +73,15 @@ def open_simulator_and_sleep(platform)
...
+def xcodebuild(tasks, platform, xcprety_args: '', xcode_summary: false)
   sdk = sdks[platform]
   scheme = schemes[platform]
   destination = devices[platform]
 
   open_simulator_and_sleep(platform)
-  safe_sh "set -o pipefail && xcodebuild -project '#{moya_project}' -scheme '#{scheme}' -configuration '#{configuration}' -sdk #{sdk} -destination '#{destination}' #{tasks} | bundle exec xcpretty -c #{xcprety_args}"
+  xcpretty_json_output_name = xcode_summary == true ? " XCPRETTY_JSON_FILE_OUTPUT=\"xcodebuild-#{platform}.json\"" : ""
+  xcpretty_formatter = xcode_summary == true ? " -f `bundle exec xcpretty-json-formatter`" : ""
+  safe_sh "set -o pipefail && xcodebuild -project '#{moya_project}' -scheme '#{scheme}' -configuration '#{configuration}' -sdk #{sdk} -destination '#{destination}' #{tasks} |#{xcpretty_json_output_name} bundle exec xcpretty -c #{xcprety_args}#{xcpretty_formatter}"
```
BasThomas: Does this need a space after the `|`? There isn't any at this moment.

Cartfile.resolved
LucianoPAlmeida: Nope 🤣  
Now that this is working I will adjust those things :))

Dangerfile
LucianoPAlmeida: I'm not sure if it's important to show ld linker warnings, and since there are some of them in the builds I just skip them :))
Maybe I can try to fix them in this and remove this 😊

Rakefile
LucianoPAlmeida: I'm not sure but even if it's not needed is good to have separation :)

LucianoPAlmeida:
@SD10 For sure man, is working now, just a few more adjustments and I'll remove the WIP and it's gonna be good to go 👍

LucianoPAlmeida:
Just requesting review from everyone involved on #1559 :))

Dangerfile
LucianoPAlmeida: @BasThomas The ld: warnings of the project seems to have something to do with the libraries that are imported via Carthage. Maybe I'm missing something but as far as I could understand the libraries have to set the target’s “Require Only App-Extension-Safe API” build setting to Yes. If you don’t, Xcode reminds you to do so by displaying the warning “linking against dylib not safe for use in application extensions”. You can see more [here](https://developer.apple.com/library/archive/documentation/General/Conceptual/ExtensibilityPG/ExtensionScenarios.html)
Because of that I just Ignore them :))

BasThomas: Hm, I am not so sure about this one. Moya and its dependencies should definitely be extension-safe APIs. What does the warning output look like if we remove this?

sunshinejr:
Thanks for doing this @LucianoPAlmeida! Seems like it's working now, but can we have the results labeled with the platform? Seems like we are testing 3 platforms and we've got 3 results, but it might be confusing as of which one is for which platform currently:

zrzut ekranu 2018-06-11 o 14 33 16

Dangerfile
```diff
@@ -77,3 +77,11 @@ end
...
+}
+xcode_summary.report 'xcodebuild-ios.json'
```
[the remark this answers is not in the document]
LucianoPAlmeida: @BasThomas Nice idea, I can change it :))

Dangerfile
LucianoPAlmeida: I can remove this so you can see the logs :)) I'll ping you then

SD10:
@LucianoPAlmeida it doesn't seem to be ignoring ld warnings 🤔

LucianoPAlmeida:
@SD10 I removed the ignore ld warnings lines because @BasThomas asked me what kind warnings were being shown then I disable so he can take a look :))
I probably going back working on that this weekend

LucianoPAlmeida:
Moving back it to WIP for some tests :))

LucianoPAlmeida:
I think it is good for review now guys @SD10 @sunshinejr @BasThomas :))

SD10:
@LucianoPAlmeida why are all the lines strikethrough above?

LucianoPAlmeida:
@SD10 I'm not sure, thought was a danger thing since the texts are Xcode summaries from previous passing pushs. I can take a look on that :))

LucianoPAlmeida:
@SD10 Remove the stick_summary and deleting the comment seems to be nice now 👍

PR #1730, opened 2018-09-28
updating to xcode 10 and swift 4.2 warnings

nicktrienensfuzz:
the change from `open` to `public` on let variables keeps xcode happy

PR #1740, opened 2018-10-09
Fix warnings generated by Xcode 10

lexorus:
### Background
I wanted to debug some of the Moya functionalities within the library. So I opened it using latest Xcode, which is 10.0, and I found that there are plenty of warnings. So I decided to fix them first.
I noticed that there is one opened PR (#1730) with update to Xcode 10/Swift 4.2, but it seems to be inactive.
@Moya/contributors This is a pretty easy PR to review, so it will be nice if somebody will find some time :sparkles:
Any related changes that I made has a separate commit, so if you would like to separate this PR into multiple, just let me know.

### What has been done
1. Update Swift version to 4.2
2. Update project to recommended settings. This includes enabling `CLANG_WARN_DEPRECATED_OBJC_IMPLEMENTATIONS` and `CLANG_WARN_OBJC_IMPLICIT_RETAIN_SELF` build settings
3. Disable `APPLICATION_EXTENSION_API_ONLY` of MoyaTests target. This fixes the linker errors `linking against a dylib which is not safe for use in application extensions`
4. Update circleci config to use Xcode 10.0.0
5. Updated RxSwift dependency to version 4.3.1

### How to test
Just open the project using the Xcode 10, build the `Moya`/`ReactiveMoya`/`RXMoya` targets, run their tests and run the example applications.


lexorus:
Should I also update `.circleci/config.yml`'s xcode version?
```yml
macos:
  xcode: '10.0.0'
```

SD10:
@lexorus Thanks for taking care of this 👍 Sorry for the delay!

> Should I also update .circleci/config.yml's xcode version?

I think that would be great 👍 

We should be able to proceed with a Moya 12.0.0-beta.1 release after this.

SD10:
@lexorus If you also want to add a CHANGELOG entry regarding support for Swift 4.2 that would be 🔥

lexorus:
It seems like the circleci is using cached dependencies that are compiled with older version of swift. Hopefully disabling the cache builds and forcing carthage to rebuild dependencies with new swift version will solve the problem.
I'll bring the option back once the issue will be solved.

SD10:
@lexorus Looks like CI is still failing. That aside, I definitely think we should keep the cached builds. There used to be a rebuild without cache option on Circle CI but I'm not sure where it went now that we switched to Circle CI 2.0 🤔 

Btw, are you using Moya at Crunchyroll? I only ask because... I ❤️ Crunchyroll 😆

freak4pc:
Didn’t review this yet - but why are your removing SwiftLint from the test target ?

lexorus:
@SD10 
> @lexorus Looks like CI is still failing. That aside, I definitely think we should keep the cached builds. There used to be a rebuild without cache option on Circle CI but I'm not sure where it went now that we switched to Circle CI 2.0 🤔

I disabled the cache only to force the build machine to rebuild the dependencies. After that I reverted the changes, but the cache is now updated and rebuilt with the new specified Xcode.

lexorus:
@freak4pc There was some warnings related to the file length in test files, if I remember correctly. This is not really a good check for the tests. Adding `swiftlint disable` comments in testing files also isn't a really good solution. And, from my practice, we are usually disabling the Swiftlint for the test target.

But, I can propose another solution for this if you really want to persist the checks. I guess we can try to add different `.swiftlint.yml` config file for test target, and specify it in the build phases.

I'm also open for any other solutions.

freak4pc:
> from my practice, we are usually disabling the Swiftlint for the test target.

I think it's not correct to bring in personal intents / habits into an open-source projects. When so many people touch this code base, the only way to make a unified coding-style is to have a proper lint tool. 

The questions that needs asking are:
1. Is that line of code actually too long? 
2. If not, is it happening a ton of times? 
3. If it does, maybe a `.swiftlint.yml` specific to the test target would work.

Removing it altogether seems wrong to me.

lexorus:
@freak4pc This wasn't a personal intent. More like a fast possible fix. I'll recheck that and come with an update. Thanks for the feedback. Can you please also review the other changes?

lexorus:
@SD10 @freak4pc Reverted the change in `.swiftlint.yml` as you requested. I'm ok if we'll take care of this in a separate PR.

lexorus:
The CI failed:
```

PR #1748, opened 2018-10-23
Enhancement/enable danger prose plugin

### LINKED ISSUE #1560: Enable Danger's prose plugin
I think it would be awesome to enable Danger's prose stuff on at least our docs, if nothing else. Maybe the commit message as well.
To enable to prose plugin, go check out http://danger.systems/ruby/, and scroll down to the prose plugin. It will probably just be an edit in our Dangerfile

Also, forcing commit messages to be <72 characters seems like a good idea to me. I dont know how the rest of the contributors feel about this, so we can debate this change if need be 😛

maxxx777:
Fixes #1560


maxxx777:
I added some setting for check spelling like "en-US" language and ignoring acronyms, numbers and some words. Also, lint and check spelling will run only if doc files modified.
If those are not necessary changes please let me know, I will revert them.

SD10:
@maxxx777 This is super cool 💯 I briefly took a look last evening and was happy to see you decide to add some ignored words. Based on the output above, it looks like we may want to add some more: iOS, Codable, Deallocated, Initializer? Whatever is not a typo really.  

On one hand, it's almost as if we'll never remove all warnings entirely because proselint is so opinionated. However, if we keep up with it by adding ignored words on doc changes/enhancements I think this will scale.

I almost wish it collapsed the warning though and we could expand as needed 🤔

maxxx777:
@SD10 I just wanted to clarify what would you suggest to change in MR? 
Not sure if it's possible to collapse/expand warnings on github page with danger-prose currently.

SD10:
@maxxx777 Sorry for the slow follow up. I don't have any specific changes. I like the idea behind this -- I just think this will add a lot of noise to PRs. Maybe some @Moya/contributors have ideas/feedback?

Thanks 👍

SD10:
> I'd recommend only having it run on new or modified files instead of every text file when there's a single change - easier to get buy in from whoever is working on that file

💯 I think this is a great idea

PR #1758, opened 2018-11-07
Bump RxMoya's watchOS deployment target to 3.0

simonrice:
This PR fixes an issue I encountered when building RxMoya for watchOS via Carthage from both development & master on Xcode 10.1.

RxSwift's watchOS deployment target [is now at watchOS 3.0 in v4.4.0](https://github.com/ReactiveX/RxSwift/commit/e8aa1d892a0d8a153a28b74cbad25be534926f49#diff-373dd22559ced25f9a1725127b241a12).  As a result, RxMoya fails to build for watchOS via Carthage since its deployment target is at watchOS 2.0.  This PR bumps RxMoya's watchOS deployment target to match RxSwift's, allowing a successful build via Carthage.

All other OS deployment targets appear to be fine.

**Question**: should the sub-podspec for RxMoya also be bumped to watchOS 3.0?


SD10:
Hey @simonrice, yes we should bump the version in the podspec as well 👍 Could you please add a CHANGELOG entry under *Changed* to convey this to users?

gityubj:
Please Merge Pull Request Right now!!!

BasThomas:
You could always use a fork in the meantime, @gityubj :)

PR #1769, opened 2018-11-19
Fix pod install error for RxMoya and ReactiveMoya subspecs due to watchOS version

### LINKED ISSUE #1768: pod install Error
[!] The platform of the target `Mine` (iOS 10.1) is not compatible with `Moya/RxSwift (12.0.0)`, which does not support `ios`.

SD10:
### Summary

This fixes #1768 

The problem was introduced by #1766 #1758 

### Problem

We bumped the watchOS deployment target for the RxMoya subspec from 2.0 -> 3.0 because RxSwift 4.0 now requires a minimum target of 3.0. We tried defining different watchOS deployment targets for different subspecs to account for this. The `pod lib lint` command passed but I'm not sure if defining different watchOS deployment targets for different subspecs is legal in CocoaPods.

After publishing version 12.0.0 to CocoaPods, users installing the `Moya/RxSwift` or `Moya/ReactiveMoya` subspecs experienced the errors in #1768 

### Solution

I think the best thing to do here is to just bump the minimum deployment target for all of Moya to 3.0. I don't like making executive decisions like this, but I'd like to get the installs working again ASAP, and this seems like a reasonable minimum version to support.

### Notes

I've tested the install locally for all subspecs: `Moya/Core`, `Moya/RxSwift`, `Moya/ReactiveSwift` and they're working.

This could've been avoided by releasing another beta for the changes to the watchOS version. I thought it was trivial -- lesson learned 😓


cwalo:
LGTM 👍🏻

PR #1777, opened 2018-11-28
Remove internal access control.

pinddFull:
internal access level is default.


BasThomas:
Hey @pinddFull, thanks! I think a large chunk of this was done by @SD10, and the reason we did so was to be more explicit about access control, and to make sure we question it.

Not sure if we want to change that. :)

BasThomas:
Thanks for your effort though! 😊

PR #1783, opened 2018-12-20
Conform MoyaError to CustomNSError in addition to LocalizedError

dpoggi:
Since the existing version of `MoyaError` conforms to `LocalizedError`, its cases with associated (underlying) `Swift.Error` values have no means of returning that information from `userInfo[NSUnderlyingErrorKey]`, where it can be seen by crash reporting tools that speak NSError.

### Motivation

My frustration, looking at a run-of-the-mill "handled" error in Bugsnag, when I realized I couldn't determine the key path at which Codable's object mapping had failed.

In this PR branch, `MoyaError` conforms to `CustomNSError` in addition to `LocalizedError` so it can supply values for both `NSLocalizedDescriptionKey` and `NSUnderlyingErrorKey`.

### Implementation

* The error domain (`String(reflecting: MoyaError.self)`), codes (Swift's internal ordinal value for each enum case), and localized descriptions stay the same.
* Even though it's returned as part of the user info dictionary now, the localized description is still available as `someMoyaError.localizedDescription`.
* Tests have been added for `MoyaError`'s bridged user info dictionary. They ensure that `NSLocalizedDescriptionKey` and the Swift `localizedDescription` property return the same string, and that the appropriate enum cases return an underlying error for `NSUnderlyingErrorKey`.

### P.S.

I recognize that an unsolicited PR is less than ideal, and I do apologize. This is something I thought I ought to do while I had the spare minutes and the inspiration.

If this gets closed, so be it — "the people have spoken." Just thought I'd offer, since it's not a whole ton of code, and I don't think I'm the only person whose time it could save.

### DIFF
--- a/Sources/Moya/MoyaError.swift
+++ b/Sources/Moya/MoyaError.swift
@@ -39,11 +39,26 @@ public extension MoyaError {
         case .jsonMapping(let response): return response
         case .stringMapping(let response): return response
         case .objectMapping(_, let response): return response
+        case .encodableMapping: return nil
         case .statusCode(let response): return response
         case .underlying(_, let response): return response
-        case .encodableMapping: return nil
         case .requestMapping: return nil
         case .parameterEncoding: return nil
+        }
+    }
+
+    /// Depending on error type, returns an underlying `Error`.
+    internal var underlyingError: Swift.Error? {
+        switch self {
+        case .imageMapping: return nil
+        case .jsonMapping: return nil
+        case .stringMapping: return nil
+        case .objectMapping(let error, _): return error
+        case .encodableMapping(let error): return error
+        case .statusCode: return nil
+        case .underlying(let error, _): return error
+        case .requestMapping: return nil
+        case .parameterEncoding(let error): return error
         }
     }
 }
@@ -65,12 +80,23 @@ extension MoyaError: LocalizedError {
             return "Failed to encode Encodable object into data."
         case .statusCode:
             return "Status code didn't fall within the given range."
+        case .underlying(let error, _):
+            return error.localizedDescription
         case .requestMapping:
             return "Failed to map Endpoint to a URLRequest."
         case .parameterEncoding(let error):
             return "Failed to encode parameters for URLRequest. \(error.localizedDescription)"
-        case .underlying(let error, _):
-            return error.localizedDescription
         }
+    }
+}
+
+// MARK: - Error User Info
+
+extension MoyaError: CustomNSError {
+    public var errorUserInfo: [String: Any] {
+        var userInfo: [String: Any] = [:]
+        userInfo[NSLocalizedDescriptionKey] = errorDescription
+        userInfo[NSUnderlyingErrorKey] = underlyingError
+        return userInfo
     }
 }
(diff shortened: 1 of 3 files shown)


SD10:
Hey @dpoggi 

> I recognize that an unsolicited PR is less than ideal, and I do apologize. This is something I thought I ought to do while I had the spare minutes and the inspiration.

Always recommend opening an issue but nonetheless thanks for the PR 👍 

> My frustration, looking at a run-of-the-mill "handled" error in Bugsnag, when I realized I couldn't determine the key path at which Codable's object mapping had failed.

I'm definitely a little skeptical about modifying Moya for better integration with other libraries that use `NSError`, one of our principles is to [provide a Swift-first API/abstraction for making network requests](https://github.com/Moya/Moya/blob/master/Vision.md#project-goals).

If I'm reading this right (I'm not familiar with `CustomNSError`), it looks like the `errorDescription` property would be deprecated with this change, so it's API breaking.

@Moya/contributors do you have any input here?

Sources/Moya/MoyaError.swift
```diff
@@ -52,25 +52,36 @@ public extension MoyaError {
 
 extension MoyaError: LocalizedError {
     public var errorDescription: String? {
+        return localizedDescription
```
AndrewSB: what does localizedDescription return here? Could you add some tests with expected values?

dpoggi: With Swift's current implementation of error bridging, this is the same as writing `return errorUserInfo[NSLocalizedDescriptionKey]`. Since `errorDescription` is declared as an optional `String?`, this line could be changed to that to clarify its intent and eliminate possible future issues from Swift implementation changes?

AndrewSB: adding a comment that it bridges to the NSLocalizedDescriptionKey, and a link to the apple documentation of that would be awesome

sunshinejr:
Hey @dpoggi! Will you be able to finish the work on this one? I think it would be a valuable contribution to Moya 13.

dpoggi:
Happy to pick this back up. I'll be in touch this evening.

sunshinejr:
@dpoggi awesome, thanks!

dpoggi:
Updated with a solution that accomplishes what the PR originally set out to do with a smaller diff and less convoluted code.

If there's anything now that should have comments with those documentation links, let me know, I've got 'em on deck and ready to go.

Sources/Moya/MoyaError.swift
```diff
@@ -39,13 +39,28 @@ public extension MoyaError {
...
         case .statusCode(let response): return response
         case .underlying(_, let response): return response
-        case .encodableMapping: return nil
         case .requestMapping: return nil
         case .parameterEncoding: return nil
         }
     }
+
+    /// Depending on error type, returns an underlying `Error`.
+    var underlyingError: Swift.Error? {
```
sunshinejr: I think that we should just leave this property internal. More than half of the errors do not have the underlying error inside of it so it might confuse more people than it could help.

Also, we already have a case named `underlying` in `Moya.error` and so it might be confusing as well: is the `underlyingError` an error of the underlying case or just an error associated with the case? Do you have an idea of a different name for this purpose?

dpoggi: I can totally get behind leaving it `internal`. Its purpose is a lot less about providing convenient error "unwrapping" and a lot more about being a right(-ish?) way to express the mapping to `Error?` for each enum case.

Hypothetical (agreeing with the point): if one or a number of Moya's consumers needed to unwrap many different underlying `Error`s in many different ways, and were seeking out or looking to build conveniences for doing so...

1. They'd need conveniences that operate on a lot more than just `MoyaError`
2. `moyaError.underlyingError` would be a moot point compared to the far more generic and not-that-much-longer `(error as NSError).userInfo[NSUnderlyingErrorKey] as? Error`

dpoggi:
@sunshinejr should be all set. Thanks for the feedback! 😄

sunshinejr:
@dpoggi thanks so much for taking care of that!

PR #1790, opened 2019-01-16
Added discardable result attribute to status code filters

alexandergmacleod:
I often use the function `Response.filterSuccessfulStatusCodes` to catch status code errors. However, rarely do I need the result of these filter functions and thus resort to using the following code in order to hide the complier warning.

```
do {
    _ = try response.filterSuccessfulStatusCodes()
   ...
} catch {
   ...
}
```

I have added `@discardableResult` attribute to the status code filter functions. I believe that this improves the quality of my code.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,8 @@
 # Next
 
+### Added
+- Added `@discardableResult` attribute to status code filtering methods in `Response`. [#1790](https://github.com/Moya/Moya/pull/1790) by [@alexandergmacleod](https://github.com/alexandergmacleod).
+
 # [12.0.1] - 2018-11-19
 
 # [12.0.0] - 2018-11-18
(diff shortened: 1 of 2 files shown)


BasThomas:
This sounds reasonable. Wondering what others think.

Can you add an entry to the changelog for this in the meantime, @alexandergmacleod?

SD10:
Hey @alexandergmacleod, thanks for taking the time to submit this PR 👍 

I'm actually not a huge fan of this functionality. It seems odd to allow the user to discard the result of a filter. If you don't need the result then it seems like you're not trying to "filter" but are rather trying to "validate" the response. 

How about using Moya's [validation property on TargetType](https://github.com/Moya/Moya/blob/master/docs/Examples/AlamofireValidation.md#alamofire-automatic-validation)?

PR #1799, opened 2019-02-11
Observable.mapImage() returns non-optional image

sunshinejr:
This is a continuation of #1789. @bjarkehs forgot to update `Observable` extension and because it was a long time ago I created a complimentary PR.


sunshinejr:
The build error seems to be an issue with CocoaPods linting, gonna open a new PR for that one. Thanks for the review @LucianoPAlmeida!

PR #1800, opened 2019-02-11
Make a restriction for CocoaPods version ~> 1.6

sunshinejr:
Related: https://github.com/CocoaPods/CocoaPods/issues/8089

Our CI is struggling to make a green light on builds that succeed. The problem seems to be our `pod lint` step and I had the same problem locally. For me, the fix was to restrict CocoaPods version to ~> 16.0.

Let's see if it fixes the issue on Circle as well 🐼

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -90,7 +90,7 @@ jobs:
           command: bundle exec danger
       - run:
           name: CocoaPods Spec linting
-          command: pod lib lint
+          command: bundle exec pod lib lint
       - run:
           name: Send Code Coverage to Codecov.io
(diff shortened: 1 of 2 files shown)


LucianoPAlmeida:
Hey @sunshinejr :)
I remember adding that step on #1767, and I think CI passes at the time (not sure). It is only failing on lint? Maybe we can do a test and running `pod lib lint --skip-tests --verbose` to have a better output of the error?

sunshinejr:
Hey @LucianoPAlmeida! No worries, it seems to be working just fine after the fix. The culprit was the CocoaPods version (see referenced issue in the PR body). The build passes fine now, though, so it's all good! 🎉

PR #1803, opened 2019-02-12
Replace #if os() with #if canImport

sunshinejr:


jdisho:
Code LGTM, but `Danger` failed.

sunshinejr:
@jdisho yeah, danger got stuck for some reason but retriggering the build helped! Thanks for the review!

PR #1805, opened 2019-02-13
Use new Alamofire append(_ fileURL, …) function with optional parameters

rivera-ernesto:
This PR relies in https://github.com/Alamofire/Alamofire/pull/2718 getting merged first. Until then PR validation checks will also fail.

Before we had to specify both a `fileName` and a `mimeType`. If one of them was missing the other would be discarded resulting in potential bugs.

 In our case we were removing invalid characters to create a valid server fileName, which in turn was getting discarded by this method resulting on failing uploads.

### DIFF
--- a/Sources/Moya/MultipartFormData.swift
+++ b/Sources/Moya/MultipartFormData.swift
@@ -47,11 +47,7 @@ internal extension RequestMultipartFormData {
     }
 
     func append(fileURL url: URL, bodyPart: MultipartFormData) {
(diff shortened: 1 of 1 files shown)


sunshinejr:
Hey @rivera-ernesto, as far as I can see the work you did was merged into Alamofire beta 4 - are you interested in finishing this PR maybe?

sunshinejr:
Updated the PR and it's ready to go 👍 thanks for the contribution @rivera-ernesto

PR #1810, opened 2019-02-18
Alamofire 5!

sunshinejr:


BasThomas:
Ooh awesome, thanks for all the work recently, @sunshinejr!

sunshinejr:
@BasThomas thanks! It's still a long way to go, though!

sunshinejr:
Update: it compiles, but I did a little workaround for `willSend` plugin method. I'm currently discussing if we could get a way of having `willSend` method provided by Alamofire, so gonna hold with this PR until that conversation is resolved.

Dschee:
@sunshinejr I just added two commits with an update to Alamofire 5 beta 3 and with also updating to Alamofire 5 when installing Moya via SwiftPM. I hope that's fine with you, I pushed since they were just small changes.

sunshinejr:
@Dschee yeah no worries, thanks!

sunshinejr:
@Moya/core-team Would anyone be so kind to check this one out? I'd love it if anyone could do a quick check on an existing app to see if everything works correctly.

sunshinejr:
@larryonoff, @LucianoPAlmeida, @pedrovereza, @Moya/contributors, anyone familiar with Alamofire/Moya core please take a look and let me know if you see any potential issues. Thanks!

Sources/Moya/Moya+Alamofire.swift
```diff
@@ -16,18 +16,19 @@ public typealias Method = Alamofire.HTTPMethod
...
 
 /// Multipart form data encoding result.
-public typealias MultipartFormDataEncodingResult = Manager.MultipartFormDataEncodingResult
-public typealias DownloadDestination = Alamofire.DownloadRequest.DownloadFileDestination
+public typealias DownloadDestination = Alamofire.DownloadRequest.Destination
 
 /// Make the Alamofire Request type conform to our type, to prevent leaking Alamofire to plugins.
 extension Request: RequestType { }
 
+/// Represents Request interceptor type that can can modify/act on Request
```
LucianoPAlmeida: Nit: typo `can can` :))

Sources/Moya/Moya+Alamofire.swift
```diff
@@ -75,16 +76,45 @@ internal protocol Requestable {
...
+            }
         }
     }
 }
 
 extension DownloadRequest: Requestable {
     internal func response(callbackQueue: DispatchQueue?, completionHandler: @escaping RequestableCompletion) -> Self {
-        return response(queue: callbackQueue) { handler  in
-            completionHandler(handler.response, handler.request, nil, handler.error)
+        if let callbackQueue = callbackQueue {
```
LucianoPAlmeida: Not related, but we are doing this `if let callBack` twice (here and on DataRequest), should we move this logic to `response(queue:` so we could just pass the callback and it could be handled there?

Tests/PropertyListEncoding.swift
```diff
@@ -0,0 +1,67 @@
+import Alamofire
+
+internal struct PropertyListEncoding: ParameterEncoding {
```
LucianoPAlmeida: Since this is internal, why all the properties and methods are public?

Sources/Moya/MoyaProvider.swift
```diff
@@ -75,8 +75,8 @@ open class MoyaProvider<Target: TargetType>: MoyaProviderType {
     /// of a request for a given `TargetType`.
     public let stubClosure: StubClosure
 
-    /// The manager for the session.
-    public let manager: Manager
+    /// The session for the session.
+    public let session: Session
```
larryonoff: > The session for the session

I think this should be changed a bit, e.g. The session for the network session / Moya provider

Sources/Moya/Plugin.swift
```diff
@@ -39,8 +39,8 @@ public protocol RequestType {
     var request: URLRequest? { get }
 
     /// Authenticates the request with a username and password.
-    func authenticate(user: String, password: String, persistence: URLCredential.Persistence) -> Self
+    func authenticate(username: String, password: String, persistence: URLCredential.Persistence) -> Self
 
     /// Authenticates the request with an `NSURLCredential` instance.
-    func authenticate(usingCredential credential: URLCredential) -> Self
+    func authenticate(with credential: URLCredential) -> Self
```
larryonoff: just personal suggestion. what about replacing `with` with `using`?

Sources/Moya/Moya+Alamofire.swift
sunshinejr: Ah yes it would be but the `response(queue:` is an Alamofire func and these two funcs are just helpers that enable queue as an optional queue instead of this if/else 😄

Tests/PropertyListEncoding.swift
sunshinejr: ah good catch, I just copied the encoding for testing as Alamofire removed it from version 5 😮

Sources/Moya/MoyaProvider.swift
sunshinejr: oh damn didn't notice that🤦‍♂ nice catch!

sunshinejr: Actually I think this doesn't really need a comment, seems self-explanatory so I will remove it for now.

Sources/Moya/Plugin.swift
sunshinejr: I like it!

sunshinejr:
@larryonoff @LucianoPAlmeida thanks for the fast CR! I've updated the PR & added some comments as well.

sunshinejr:
Alright, bumped Alamofire to 5.0.0-beta.7 as it had minimal changes - merging it!

LucianoPAlmeida:
Awesome @sunshinejr, now that this was merged maybe #1878 could get in too? 
Also do you have plans for a pre-release(alpha/beta) of this one?

sunshinejr:
@LucianoPAlmeida Yeah, I'm in the process of merging #1878 and then releasing new alpha. Wanted #1880 as well but it needs a little bit more time so we will add it to the next pre-release.

PR #1812, opened 2019-03-04
Fix proselint on Dangerfile

f-meloni:
Hi 👋 
I was having a look to convert Danger to Danger-Swift, that is not possible because there is no proselint and mdspell support on Danger-Swift (yet :P) when I saw something that looked wrong to me, it runs prose only if the modified file list is empty.


sunshinejr:
Also finally fixed the "waiting for Circle status" thingy that wouldn't make a green status even though the build was correct. Turns out that we had a relict from CircleCI 1.0 (more [here](https://circleci.com/docs/2.0/workflows-waiting-status/) if someone is interested).

Issue #1813, opened 2019-03-05
Refreshing auth token confusion


user6724161:
I replied in #744, but I continue to experience difficulties trying to sort this out, so I've made a new thread.

I need to be able to pass a bearer token (for some of my calls) and refresh it when necessary. Currently, I have the following class to handle all of my API calls:

```swift
import Foundation
import Moya

enum ApiService {
    case signIn(email: String, password: String)
    case like(id: Int, type: String)
}

extension ApiService: TargetType, AccessTokenAuthorizable {
    var authorizationType: AuthorizationType {
        switch self {
        case .signIn(_, _):
            return .basic
        case .like(_, _):
            return .bearer
        }
    }

    var baseURL: URL {
        return URL(string: Constants.apiUrl)!
    }

    var path: String {
        switch self {
            case .signIn(_, _):
                return "user/signin"
            case .like(_, _):
                return "message/like"
        }
    }

    var method: Moya.Method {
        switch self {
            case .signIn, .like:
                return .post
        }
    }

    var task: Task {
        switch self {
            case let .signIn(email, password):
                return .requestParameters(parameters: ["email": email, "password": password], encoding: JSONEncoding.default)
            case let .like(id, type):
                return .requestParameters(parameters: ["messageId": id, "type": type], encoding: JSONEncoding.default)
        }
    }

    var sampleData: Data {
        return Data()
    }

    var headers: [String: String]? {
        return ["Content-type": "application/json"]
    }
}

private extension String {
    var urlEscaped: String {
        return addingPercentEncoding(withAllowedCharacters: .urlHostAllowed)!
    }

    var utf8Encoded: Data {
        return data(using: .utf8)!
    }
}
```

As @pietrocaselani suggested, [here](https://github.com/Moya/Moya/issues/744#issuecomment-468691723), I have tried to make sense of how to implement this in accordance with my above code, but I've been stuck now for many days.

Here is what that class looks like right now:

```swift
import Foundation
import Moya

public final class ApiProvider {
    private var token: String = ""
    
    private func isTokenValid() -> Bool {
        // check expiration date
        return false
    }
    
    func createProvider() -> MoyaProvider<ApiService> {
        let requestClosure = createRequestClosure()
        
        return MoyaProvider<ApiService>(requestClosure: requestClosure)
    }
    
    func createRequestClosure() -> MoyaProvider<ApiService>.RequestClosure {
[...]

PR #1815, opened 2019-03-09
Fix progress object with invalid Content-Length headers.

### LINKED ISSUE #1771: RxSwift: requestWithProgress().filterCompleted() no onNext events
Moya version: 11.0.2, but also tested with Moya 12

I am using Moya with RxSwift, using the `requestWithProgress` function to get the response and progress on uploading a file.

The progress part is working perfectly, however, I found that when I used the response, my other Rx operations on the provided Observable weren't working correctly. When I debugged what was going on, I noticed that the Observable I received from doing:

```swift
requestWithProgress(...).filterCompleted()
```

was only firing an `onCompleted` and then dispose. There was no `onNext` that contained the actual response.

To get this working instead of using the provided `filterCompleted` I just had to do it myself with:

```swift
requestWithProgress(...)
  .map { $0.response }
  .filter { $0 != nil }
  .map { $0! }
```

In short, this is pulling out the response field, filtering for when it isn't nil and then unwrapping the optional.

When I use this instead, everything works as expected. I think the `filterCompleted` implementation is firing no `onNext` events. The type `Observable<Response>` and the comment on the function suggest that it should be providing an Observable that gives you the response and then an onCompleted event, so I think there is a bug there.

For now my code is working with what I put above, so its not blocking me from using this, but just raising this issue, as I think there is a bug that may be affecting others.

sunshinejr:
Fixes #1771.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,7 +5,7 @@
 - **Breaking Change** `.mapImage()` extension on `Single` and `Observable` now returns non-optional image. [#1789](https://github.com/Moya/Moya/pull/1789), [#1799](https://github.com/Moya/Moya/pull/1799) by [@bjarkehs](https://github.com/bjarkehs) and [@sunshinejr](https://github.com/sunshinejr).
 
 ### Fixed
-- Fixed `Progress` object on responses that did not specify correct `Content-Length` header. Now, whenever there is no valid header, the progress will be 0.0 until the completion of the request. Also, the `completed` property is now `true` only when the response was serialized, we do not rely on progress being 1.0 anymore. [#1813](https://github.com/Moya/Moya/pull/1783) by [@sunshinejr](https://github.com/sunshinejr).
+- Fixed `Progress` object on responses that did not specify correct `Content-Length` header. Now, whenever there is no valid header, the progress will be 0.0 until the completion of the request. Also, the `completed` property is now `true` only when the response was serialized, we do not rely on progress being 1.0 anymore. [#1815](https://github.com/Moya/Moya/pull/1815) by [@sunshinejr](https://github.com/sunshinejr).
 
 # [12.0.1] - 2018-11-19
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,11 @@
 # Next
+
+### Changed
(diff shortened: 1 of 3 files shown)


Tests/MoyaProviderSpec.swift
```diff
@@ -794,7 +796,85 @@ final class MoyaProviderSpec: QuickSpec {
...
+                            error = err
+                        }
+                        done()
+                    }
+
+                    provider.request(target, callbackQueue: nil, progress: progressClosure, completion: progressCompletionClosure)
+                }
+
+                expect(error).to(beNil())
+                expect(progressValues) == [0, 0, 0, 0, 1.0]
```
LucianoPAlmeida: This is just curiosity, there is a reason for all moya equal tests use `expect(value) ==` instead of `.to(equal(value))`?

sunshinejr: @LucianoPAlmeida I guess I just use it because it's a lot more convenient to just type `==` instead of `.to(equal(value))`.

Changelog.md
```diff
@@ -1,7 +1,12 @@
 # Next
+
+### Changed
 - `MoyaError` now conforms to `CustomNSError` protocol, makes underlying errors available in its user-info dictionary. [#1783](https://github.com/Moya/Moya/pull/1783) by [@dpoggi](https://github.com/dpoggi).
 - **Breaking Change** `.mapImage()` extension on `Single` and `Observable` now returns non-optional image. [#1789](https://github.com/Moya/Moya/pull/1789), [#1799](https://github.com/Moya/Moya/pull/1799) by [@bjarkehs](https://github.com/bjarkehs) and [@sunshinejr](https://github.com/sunshinejr).
 
+### Fixed
+- Fixed `Progress` object on responses that did not specify correct `Content-Length` header. Now, whenever there is no valid header, the progress will be 0.0 until the completion of the request. Also, the `completed` property is now `true` only when the response was serialized, we do not rely on progress being 1.0 anymore. [#1813](https://github.com/Moya/Moya/pull/1783) by [@sunshinejr](https://github.com/sunshinejr).
+
```
pedrovereza: I think the link should be `https://github.com/Moya/Moya/pull/1813`, it's pointing to pr `1783` right now 😄

pedrovereza: Wait, it should actually point to this pr:

```
[#1815](https://github.com/Moya/Moya/pull/1815)
```

sunshinejr: 🤦‍♂️ thanks @pedrovereza!

PR #1817, opened 2019-03-18
Bump ReactiveSwift to 5.0 and Result to 4.1

larryonoff:
Bump ReactiveSwift to 5.0 and Result to 4.1


sunshinejr:
Hey @larryonoff, thanks for the PR! Was there something specific that happened in `Result` 4.1 that we don't want to support version 4.0?

larryonoff:
@sunshinejr ReactiveSwift 5.0 depends on Result 4.1. 

ReactiveSwift 5.0 and Result 4.1 both add Swift 5.0 and Xcode 10.2.

sunshinejr:
@larryonoff makes sense, thanks! Would you please update SPM lock (probably `swift package update` if I'm not mistaken)? 

+ I see that the project doesn't want to build tests due to Alamofire update... Interestingly enough it seems like they've made a breaking change without a major update 🤔

larryonoff:
@sunshinejr done

larryonoff:
@sunshinejr 
> I see that the project doesn't want to build tests due to Alamofire update... Interestingly enough it seems like they've made a breaking change without a major update 🤔

I'm not sure, but it was without update.

sunshinejr:
@larryonoff when you look at the `Cartfile.resolved` (we use Carthage dependencies for building/testing), you can see that the Alamofire has bumped version to 4.8.1. Then, when we look at the build error in tests:
```swift
❌  /Users/distiller/Moya/Moya/Tests/EndpointClosureSpec.swift:65:19: method does not override any method from its superclass

    override func upload(multipartFormData: @escaping (Alamofire.MultipartFormData) -> Void, usingThreshold encodingMemoryThreshold: UInt64, with urlRequest: URLRequestConvertible, encodingCompletion: ((SessionManager.MultipartFormDataEncodingResult) -> Void)?) {
```

it seems like Alamofire updated this method in their minor/patch release. If you have time to fix this one it would be great! If not, I could try to look around and help you out so no worries 👍

larryonoff:
@sunshinejr I have fixed it in the current branch. Please review

PR #1834, opened 2019-04-05
[README] Document support & installation via Accio

### LINKED ISSUE #345: Ability to checkout only specific type of Moya via Carthage

Dschee:
This documents support for the new SwiftPM based dependency manager [Accio](https://github.com/JamitLabs/Accio).
Please note that this project is part of Accio's official [integration tests](https://github.com/JamitLabs/Accio/tree/work/1000-frameworks/Demo/Shared/AppDependencies) within the [Demo project](https://github.com/JamitLabs/Accio/tree/work/1000-frameworks/Demo).

This supersedes my previous approach #1642 and is therefore related to #345.

Checkout [this thread](https://github.com/Carthage/Carthage/pull/1990) and my last [comment](https://github.com/Carthage/Carthage/pull/1990#issuecomment-479442640) if you want to learn more about the history of how I decided to write my own dependency manager to fix #345 properly. Please note that Moya was one of the main motivations for me to write Accio. So I really hope you like it! Any feedback is much appreciated.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -107,6 +107,10 @@ let package = Package(
 
 Note that as of Moya 10, SPM only works with Swift 4 toolchain and greater.
 
+### Accio
+
+Accio is based on Swift Package Manager (with the addition of iOS/macOS/tvOS/watchOS support) and therefore integration of Moya is exactly the same as described above. Once your `Package.swift` file is configured, run `accio update`.
+
 ### CocoaPods
 
 For Moya, use the following entry in your Podfile:
@@ -144,28 +148,6 @@ Then run `carthage update`.
 If this is your first time using Carthage in the project, you'll need to go through some additional steps as explained [over at Carthage](https://github.com/Carthage/Carthage#adding-frameworks-to-an-application).
 
 > NOTE: At this time, Carthage does not provide a way to build only specific repository submodules. All submodules and their dependencies will be built with the above command. However, you don't need to copy frameworks you aren't using into your project. For instance, if you aren't using `ReactiveSwift`, feel free to delete that framework along with `ReactiveMoya` from the Carthage Build directory after `carthage update` completes. Or if you are using `ReactiveSwift` but not `RxSwift`, then `RxMoya`, `RxTest`, `RxCocoa`, etc. can safely be deleted.
-
-### Accio
-
-Add the following to your Package.swift:
-
-```swift
-.package(url: "https://github.com/Moya/Moya.git", .upToNextMajor(from: "13.0.0-beta.1")),
-```
-
-Next, choose either `Moya`, `RxMoya` or `ReactiveMoya` and add it to your App targets dependencies, for example:
-
-```swift
-.target(
-    name: "App",
-    dependencies: [
-        "Moya",
-    ]
-),
-```
-
-Then run `accio update`.
-
 
 ### Manually
 
--- a/Readme.md
+++ b/Readme.md
@@ -109,7 +109,7 @@ Note that as of Moya 10, SPM only works with Swift 4 toolchain and greater.
 
 ### Accio
 
-Accio is based on Swift Package Manager (with the addition of iOS/macOS/tvOS/watchOS support) and therefore integration of Moya is exactly the same as described above. Once your `Package.swift` file is configured, run `accio update`.
+[Accio](https://github.com/JamitLabs/Accio) is a dependency manager based on SwiftPM which can build frameworks for iOS/macOS/tvOS/watchOS. Therefore the integration steps of Moya are exactly the same as described above. Once your `Package.swift` file is configured, run `accio update` instead of `swift package update`.
 
 ### CocoaPods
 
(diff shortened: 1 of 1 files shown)


Readme.md
```diff
@@ -144,6 +145,28 @@ If this is your first time using Carthage in the project, you'll need to go thro
 
 > NOTE: At this time, Carthage does not provide a way to build only specific repository submodules. All submodules and their dependencies will be built with the above command. However, you don't need to copy frameworks you aren't using into your project. For instance, if you aren't using `ReactiveSwift`, feel free to delete that framework along with `ReactiveMoya` from the Carthage Build directory after `carthage update` completes. Or if you are using `ReactiveSwift` but not `RxSwift`, then `RxMoya`, `RxTest`, `RxCocoa`, etc. can safely be deleted.
 
+### Accio
+
+Add the following to your Package.swift:
```
freak4pc: I feel like this entire thing should just be "Do the same as SPM but use `accio update` instead". Any reason the repetition is needed?

Dschee: You're right about that. We could also combine both into one section names "SwiftPM & Accio" or would you rather keep it separate?

pedrovereza: I vote for separation, mainly because `SwiftPM & Accio` _may_ be confusing for users looking just for the SwiftPM instructions. 

So basically a 👍 to @freak4pc suggestion

Dschee: I just changed the installation instructions for Accio to refer to the SwiftPM section, like suggested.

Readme.md
```diff
@@ -106,6 +107,10 @@ let package = Package(
 
 Note that as of Moya 10, SPM only works with Swift 4 toolchain and greater.
 
+### Accio
+
+Accio is based on Swift Package Manager (with the addition of iOS/macOS/tvOS/watchOS support) and therefore integration of Moya is exactly the same as described above. Once your `Package.swift` file is configured, run `accio update`.
```
freak4pc: Just a sidenote, I tried to google "Accio Dependency GitHub" and it found not only a same-named project (which is fine usually), but it's also a dependency manager. Might make it a bit hard to find :) 

I'd add a link on the word "Accio" in the beginning of the sentence and change it to a

Accio is a dependency manager based on Swift Package Manager.

I'd get rid of the addition of support thing because SwiftPM does support these platforms in v5 AFAIK: 

image

orta:
Looks cool!

Readme.md
Dschee: Interesting, seems like someone had the same name idea, I didn‘t optimize for SEO at all but I hope the fact that the other one is for C++ will make them still distinguishable.

Regarding platforms support, I disagree. SwiftPM does NOT support all Apple platforms, it only now has a way of specifying what the minimum required version of a platform is - that‘s a feature of the manifest file though, not of the tool itself. I think to most it is clear that SwiftPM doesn‘t support iOS, otherwise the community would have opted to use it a long time ago.

Regarding linking to Accio, no problem, I can add a link.

Readme.md
```diff
@@ -106,6 +107,10 @@ let package = Package(
 
 Note that as of Moya 10, SPM only works with Swift 4 toolchain and greater.
 
+### Accio
+
+[Accio](https://github.com/JamitLabs/Accio) is a dependency manager based on SwiftPM which can build frameworks for iOS/macOS/tvOS/watchOS. Therefore the integration steps of Moya are exactly the same as described above. Once your `Package.swift` file is configured, run `accio update` instead of `swift package update`.
```
lucas34: This sounds like a presentation of the Accio library. I suggest then to only keep the last sentence "Once your `Package.swift` file is configured, run `accio update`"

freak4pc: I don't think it's a big deal to present it in a single sentence given it's not known. Carthage was also presented this way before it was well known.

Dschee: Thanks @freak4pc, I thought it was necessary to explain the connection between Accio and SwiftPM so it's easier to understand why the instructions are the same. 

Also someone told me that there seems to be another dependency manager named exactly the same written at about the same time like I have written this but for C++. So, for now it might be easier for the users to have a link, that's why I also added that.

PR #1839, opened 2019-04-12
[WIP] Drop Result dependency

larryonoff:
This PR drops Result dependency

* Bump ReactiveSwift dependency to 6.0 that also drops Result

I wasn't able to update Package.resolved since of Alamofire, so it may be correct to rebase this PR on #1810.


metisdev:
Tried and faced no problems with a big project 👍

metisdev:
Will this be merged?

larryonoff:
@metisdev I really want to have it merged. But not sure about Alamofire dependency. We may need to wait new Alamofore version to be merged. In other case SPM cannot be used.

I also want some feedback from core team

sunshinejr:
@larryonoff yeah we need to finish Alamofire transition first. Then we would focus on this one. I’m really excited about droping Result in the next version, though.

larryonoff:
@sunshinejr awesome! I really looking forward to the update. I also would like to bump ReactiveSwift and RxSwift dependencies.

PR #1841, opened 2019-04-26
Remove alamoRequest.cancel() from stubbing closure

### LINKED ISSUE #1833: How to avoid error: Error Domain=NSURLErrorDomain Code=-999 "cancelled"
in provider test,

```swift
import Quick
import Nimble
import Alamofire
import RxSwift
import RxBlocking
import RxTest
import OHHTTPStubs

@testable import Moya
@testable import DSZHClientFramework

class DSZHApiServiceSpec: QuickSpec {
    override func spec() {
        describe("provider with Single") {
            var provider: MoyaProvider<DSZHApi>!
            
            beforeEach {
                provider = MoyaProvider<DSZHApi>(endpointClosure: endpointClosure, stubClosure: MoyaProvider.immediatelyStub)
            }
            
            it("emits a Response object") {
                var response: Response?
                
                let target: DSZHApi = .getForums
                response = try! provider.rx.request(target).toBlocking().single()
                expect(response).toNot(beNil())
            }
            
            it("returns stubbed data for get forums request") {
                var json: String?
                
                let target: DSZHApi = .getForums
                provider.request(target) { result in
                    if case let .success(response) = result {
                        json = String(data: response.data, encoding: .utf8)
                    }
                }
                
                let sampleData = target.sampleData
                expect(json).to(equal(String(data: sampleData, encoding: .utf8)))
            }
        }
    }
}
```

I always get error message in xcode output:

2019-04-04 16:18:37.906181+0800 xctest[89156:12465616] Task <CD73025B-02D6-4262-BA85-BCDD84108C79>.<1> load failed with error Error Domain=NSURLErrorDomain Code=-999 "cancelled"

but test case is passed

How to fix it?

sunshinejr:
Fixes #1833.

The request was never really resumed and so it most likely doesn't need to be cancelled. This was only making random noise about cancelled request in console.

Targeted master as I'd like to make a 13.0.1 release with this fix.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,9 +23,6 @@ jobs:
           name: Set Ruby Version
           command: echo "ruby-2.4" > ~/.ruby-version
       - run:
-          name: Update Homebrew
-          command: brew update    
-      - run:
           name: Carthage checkout
           command: carthage checkout
       - run:
@@ -47,9 +44,6 @@ jobs:
       - run:
           name: Set Ruby Version
           command: echo "ruby-2.4" > ~/.ruby-version
-      - run:
-          name: Update Homebrew
-          command: brew update
       - run:
           name: Install Swiftlint
           command: git clone git@github.com:realm/SwiftLint.git && cd SwiftLint && git submodule update --init --recursive && make install && cd ../
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,7 +52,7 @@ jobs:
           command: brew update
       - run:
           name: Install Swiftlint
-          command: brew install swiftlint
+          command: git clone git@github.com:realm/SwiftLint.git && cd SwiftLint && git submodule update --init --recursive && make install && cd ../
       - run:
           name: Add Python directory to PATH
           command: echo 'export PATH=~/Library/Python/2.7/bin:$PATH' >> $BASH_ENV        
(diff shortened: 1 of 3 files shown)


sunshinejr:
[SwiftLint is having problems on newest Xcode/macOS](https://github.com/realm/SwiftLint/issues/2695). Wonder if we should try to build it from source or just disable it until it's released as a new version on brew. 🤔

.circleci/config.yml
```diff
@@ -47,12 +44,9 @@ jobs:
       - run:
           name: Set Ruby Version
           command: echo "ruby-2.4" > ~/.ruby-version
-      - run:
-          name: Update Homebrew
-          command: brew update
       - run:
           name: Install Swiftlint
-          command: brew install swiftlint
+          command: git clone git@github.com:realm/SwiftLint.git && cd SwiftLint && git submodule update --init --recursive && make install && cd ../
```
LucianoPAlmeida: Hey :)) Have you considered use [Mint](https://github.com/realm/SwiftLint#using-mint) here?

sunshinejr: That's a good point @LucianoPAlmeida. I just didn't want to add another dependency to CI in this PR. We may discuss potential updates to the CI script, as it needs refreshing anyways.

LucianoPAlmeida: Sure, that makes sense, I just raised the point because it appears to be a simpler script :) Also we could avoid having to clone all source and compile it, iff mint supports just download the binary(not sure if it does).

PR #1846, opened 2019-05-01
RXSwift 5

LucianoPAlmeida:
Updating RxSwift Dependency to RxSwift 5.0.


LucianoPAlmeida:
Just notice that the development branch is behind master, should we update it?

sunshinejr:
Oh, seems you forgot Changelog entry. I'll make a quick commit so no worries.

PR #1856, opened 2019-05-09
Updated Chinese docs per MoyaBot recommendation.

raid5:


sunshinejr:
good bot @MoyaBot 🤖

sunshinejr:
@raid5 also, btw, you can add [skip ci] in your commit message/description next time you update the Readme, so we won't be building the library if we don't have to 👍

raid5:
> @raid5 also, btw, you can add `[skip ci]` in your commit message/description next time you update the Readme, so we won't be building the library if we don't have to 👍

Ah that's good to know, thanks!

PR #1862, opened 2019-05-22
Update Releasing.md with few additional steps to "After release" paragraph.

sunshinejr:
This includes few additional steps that I usually do after the release. Also, I totally forgot about our [jazzy-generated docs](https://moya.github.io) and I think everyone did 😄 Already updated the docs and gonna try to add a danger check as well in a few minutes.

### DIFF
--- a/docs/Releasing.md
+++ b/docs/Releasing.md
@@ -42,7 +42,9 @@ To make a release:
 
 What you might need to do manually afterwards:
 
-1. Update the Swift Package Manager instructions in the Readme to use the release you just made public.
+1. Update the instructions for our supported package managers in the Readme to use the release you just made public (e.g. update version numbers if it was a major release).
+1. Update Swift/Moya version table if needed.
+1. Update the release description in [GitHub releases tab](https://github.com/Moya/Moya/releases/tag).
 1. Check if we need to regenerate docs for [moya.github.io](https://github.com/Moya/moya.github.io).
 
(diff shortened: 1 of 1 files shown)


sunshinejr:
I've cancelled the builds as this is just a doc change and I forgot about adding `[skip ci]` in the commit message 🤦‍♂

sunshinejr:
Now that I think about it, Danger would need to add this message on every PR as we not only need to regenerate docs on docs changes, but also comments/code updates depending on the change. Wonder if we could automate a commit to master to regenerate the docs on [moya.github.io](https://github.com/Moya/moya.github.io) 🤔

f-meloni:
@sunshinejr Don't know if can help but on Danger we automatically re-generate the docs on every release.
https://github.com/danger/swift/blob/master/Package.swift#L57

sunshinejr:
@f-meloni oh nice! do you think it would be possible to regenerate docs on separate repo in the same org?

f-meloni:
@sunshinejr `Rocket` just runs steps that can be written in bash or swift (+ has some pre defined steps).
Then I think that potentially can.
For example if you look in Danger after the release it does

```
"after": [
    "Scripts/create_homebrew_tap.sh",
]
```

That pulls and updates the Danger homebrew repo

sunshinejr:
@f-meloni alright, I've created an issue for that one so we do not forget about it 😉 #1864

PR #1866, opened 2019-06-03
Add target type in output of network logger plugin

### LINKED ISSUE #1853: Send TargetType as a parameter in output in NetworkLoggerPlugin
NetworkLoggerPlugin currently has an output closure to print logs. I think the TargetType should be added in the output.
 
Reason for this:
- Able to print logs according to TargetType. I would like to print logs for only particular API targets. 

My thoughts and changes below. 
```swift   
    fileprivate let output: (_ taget: TargetType, _ separator: String, _ terminator: String, _ items: Any...) -> Void
.
.
.
    /// Initializes a NetworkLoggerPlugin.
    public init(..., output: ((_taget: TargetType, _ separator: String, _ terminator: String, _ items: Any...) -> Void)? = nil, ...) {
...
        self.output = output ?? NetworkLoggerPlugin.reversedPrint
...
    }

    public func willSend(_ request: RequestType, target: TargetType) {
        if let request = request as? CustomDebugStringConvertible, cURL {
            output(target, separator, terminator, request.debugDescription)
            return
        }
        outputItems(logNetworkRequest(request.request as URLRequest?), target: target)
    }

    public func didReceive(_ result: Result<Moya.Response, MoyaError>, target: TargetType) {
        if case .success(let response) = result {
            outputItems(logNetworkResponse(response.response, data: response.data, target: target), target: target)
        } else {
            outputItems(logNetworkResponse(nil, data: nil, target: target), target: target)
        }
    }

    fileprivate func outputItems(_ items: [String], target: TargetType ) {
        if isVerbose {
            items.forEach { output(target, separator, terminator, $0) }
        } else {
            output(target, separator, terminator, items)
        }
    }
}

hasankose:
It provides print logs according to TargetType.
Fixes #1853


hasankose:
Let me a quick update, I've checked the TODOs below. 
- [ ] CHANGELOG.md is updated.
- [ ] NetworkActivityPlugin and CredentialsPlugin have the TargetType in the closure. 
- [ ] TargetType is not needed for AccessTokenPlugin. 
- [ ] The Output which I've changed is not used in the Plugins doc and it is not needed to update(I can add an example if needed).

pedrovereza:
@sunshinejr @hasankose Hey, how about we make `target` an optional parameter? This will keep the plugin working as it used to, and also enable users to filter by `TargetType` when they need.

sunshinejr:
@pedrovereza is it possible to do so without making a breaking change? Even if there is an optional parameter, you’d still need to handle it in the closure, right?

pedrovereza:
@sunshinejr Ah, you're right! Ignore me, I miss understood how the changes were being added here 😅

sunshinejr:
@pedrovereza no worries, this was something I didn't consider before so good question!

sunshinejr:
@hasankose alright I've pushed a small polish to the Changelog entry but we're good to go! Thanks again for looking into this, good timing with the new major release as well 👍

Issue #1864, opened 2019-05-30
Regenerate moya.github.io docs on Moya release.


sunshinejr:
We have a repository that gathers our docs and represents it beautifully on our website https://moya.github.io.

The problem is that these docs currently have to be generated manually and that's easy to forget. @f-meloni suggested that we could use setup similar to the one they use at [danger-swift](https://github.com/danger/swift/blob/master/Package.swift#L57) which generates the docs automatically.

This task is for creating an automation to regenerate the docs on each release. For releasing new Moya version we use our [Rake script](https://github.com/Moya/Moya/blob/master/Rakefile#L158) and this would be a good starting point.

Also see #1862 discussion as this is where the idea came from.

Let me know if you would be interested in helping out or if you have any questions regarding this task!

MetalDent:
Hi, anyone working on this?

sunshinejr:
hey @MetalDent, I don't think anyone started working on this one yet - would you like to try?

MetalDent:
@sunshinejr yeah I can try! 

I saw the danger script to auto-generate docs. Do you want similar implementation? And where should this change go?

sunshinejr:
@MetalDent I don’t have a preference for implementation - you can do whatever you are good at or think will be enough.

The docs should auto-generate on our second repository (https://github.com/Moya/moya.github.io)

MetalDent:
@sunshinejr I'll also use `rocket` . So which files from this repository do you want to auto-generate in the other repository?

sunshinejr:
@MetalDent so once you ran the docs generating using jazzy:
```bash
jazzy -o URL_TO_THE_DOCS
```

we should move all the docs to the `moya.github.io` repo

sunshinejr:
@MetalDent now that I think about it... can it be done from the GitHub Actions?

MetalDent:
@sunshinejr sorry I can't work on this my iOS VM is not working right now :(

sunshinejr:
@MetalDent don't worry, we can wait, it's not something urgent right now - just nice to have 👍

hamdshah:
Why don't we use the same repo for the documentation instead of the separate one ?
I mean we can configure this one for Github pages.

PR #1873, opened 2019-06-23
Danger swift

f-meloni:
Hi 👋 ,
The goal of this PR is convert the current ruby Dangerfile to a swift Dangerfile


f-meloni:
Ok, this is now ready!
@sunshinejr can you please have a look when you have time? :)
The cache is giving to Danger acceptable performances, even if probably the configuration with `Rocket` (https://danger.systems/swift/guides/about_the_dangerfile.html#swift-package-manager-more-performant) would give a lot better performances, would probably help with https://github.com/Moya/Moya/issues/1864 and would maybe also allow use the `Package.swift` also for swiftlint.
But this can be discussed later, and can be, in case, implemented in another PR.

Dangerfile.swift
```diff
@@ -0,0 +1,144 @@
...
+if #available(OSX 10.12, *),
+    !addedAndModifiedCnDocsMarkdown.isEmpty {
+    Proselint.performSpellCheck(files: addedAndModifiedCnDocsMarkdown, excludedRules: ["misc.scare_quotes", "typography.symbols"])
+}
+
+// Run danger-prose to lint and check spelling English docs
+func isEnDocsMarkdown(_ file: String) -> Bool {
+    return file.fileType == .markdown && file.contains("docs/")
+}
+let addedAndModifiedEnDocsMarkdown = danger.git.createdFiles.filter(isEnDocsMarkdown) + danger.git.modifiedFiles.filter(isEnDocsMarkdown)
```
sunshinejr:
```suggestion
let addedOrModifiedEnDocsMarkdown = allSourceFiles.filter { $0.fileType == .markdown && $0.contains("docs/") }
```

Dangerfile.swift
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// Warn when library files has been updated but not tests.
+let testsUpdated = danger.git.modifiedFiles.first { $0.hasPrefix("Tests") } != nil
+if sourceChanges && !testsUpdated {
+    warn("The library files were changed, but the tests remained unmodified. Consider updating or adding to the tests to match the library changes.")
+}
+
+// Run Swiftlint
+SwiftLint.lint(inline: true)
```
sunshinejr: I remember we had a problem with SwiftLint plugin to Danger Swift and we had to use it two times, once to lint `Sources` and once to lint `Tests` (see https://github.com/ashfurrow/danger-swiftlint/issues/4#issuecomment-364797573) - do you know if this is fixed already?

sunshinejr: Also, can we add `Dangerfile.swift` to SwiftLint? 😄

Dangerfile.swift
```diff
@@ -0,0 +1,144 @@
...
+if sourceChanges && missingDocChanges && docChangeRaccomanded && !isTrivial {
+    warn("Consider adding supporting documentation to this change. Documentation can be found in the `docs` directory.")
+}
+
+// Run danger-prose to lint Chinese docs
+func isCnDocsMarkdown(_ file: String) -> Bool {
+    return file.fileType == .markdown && file.contains("docs_CN")
+}
+let addedAndModifiedCnDocsMarkdown = danger.git.createdFiles.filter(isCnDocsMarkdown) + danger.git.modifiedFiles.filter(isCnDocsMarkdown)
+if #available(OSX 10.12, *),
```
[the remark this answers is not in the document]
f-meloni: Because `Proselint` is using https://developer.apple.com/documentation/foundation/filemanager/1642807-homedirectoryforcurrentuser
(Proselint wants the config in the user home :( https://github.com/amperser/proselint#checks)

Dangerfile.swift
f-meloni: We can and we should :P

f-meloni: Danger Swiftlint supports the config file then I would try to create one that includes `Sources` and `Tests` and see what happens, we can still go with the double lint if needed

f-meloni:
@sunshinejr Thank you very much for the review!
Will have a look today/tomorrow to it and will address the points :)

Dangerfile.swift
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// Don't let testing shortcuts get into master by accident
+if danger.utils.exec("grep -r \"fit Demo/Tests/\"").count > 1 {
+    fail("fit left in tests")
+}
+
+// Added (or removed) library files need to be added (or removed) from the
+// Carthage Xcode project to avoid breaking things for our Carthage users.
+let addedSwiftLibraryFiles = danger.git.createdFiles.first { $0.fileType == .swift && $0.hasPrefix("Sources") } != nil
```
LucianoPAlmeida: Those `.first` calls the after test for `!= nil`  could be replaced by `contains(where: )` ? 
Maybe its a bit more clear and we avoid the check since contains is already boolean. What you guys think?

f-meloni: Yes, I think it can be replaced with a `contains(where: )`, thank you :)

Dangerfile.swift
f-meloni: Swiftlint is not using`.swiftlint.yml` config file that contains
```yaml
included:
  - Sources
  - Tests
  - Examples
  - Dangerfile.swift
```
Let's see how it goes, if there is any problem I will change it :)

PR #1878, opened 2019-07-05
Add missing session headers to Request/NetworkLoggerPlugin.

### LINKED ISSUE #1057: Missing Headers in Response->Request->allHTTPHeaderFields
I am using Moya version 8.0.3 and I want to log failing requests, their responses and the respective body data to crashlytics

When I log to console via the networkLoggingPlugin I get the following request readers:

```swift
$ curl -i \
	-X POST \
	-H "Content-Type: application/json" \
	-H "Accept-Encoding: gzip;q=1.0, compress;q=0.5" \
	-H "Accept-Language: en-US;q=1.0" \
	-H "User-Agent: Example/0.0.2 (com.abc.example; build:1; iOS 9.3.0) Alamofire/4.4.0" \
	-d "{\"data\":{\"phone_number\":\"+49123\"}}" \
	"https://www.example.com/api/v1/user"
```

In contrast when I want to extract the request headers from the response after a network request I only get those headers:

```swift
HTTP Error: POST https://www.example.com/api/v1/user {
   ["Content-Type": "application/json"]
 }
  {"data":{"phone_number":"+49123"}})
```

Am I something missing? Is there another way doing this?

sunshinejr:
Fixes #1057. Totally forgot about that one and when updating Alamofire I couldn't find the session headers and then I remembered!

This can be reviewed but will be merged only after #1810.

---

Before:
> Moya_Logger: [05/07/2019 12:32:14] Request Headers: [:]

After:

> Moya_Logger: [05/07/2019 12:33:50] Request Headers: ["User-Agent": "Basic/1.0 (Moya.Basic; build:1; iOS 12.2.0) Alamofire/5.0.0", "Accept-Encoding": "br;q=1.0, gzip;q=0.9, deflate;q=0.8", "Accept-Language": "en;q=1.0, pl-PL;q=0.9"]

### DIFF
--- a/Sources/Moya/Plugins/NetworkLoggerPlugin.swift
+++ b/Sources/Moya/Plugins/NetworkLoggerPlugin.swift
@@ -30,7 +30,7 @@ public final class NetworkLoggerPlugin: PluginType {
             output(separator, terminator, request.debugDescription)
             return
         }
-        outputItems(logNetworkRequest(request.request as URLRequest?))
+        outputItems(logNetworkRequest(request))
     }
 
     public func didReceive(_ result: Result<Moya.Response, MoyaError>, target: TargetType) {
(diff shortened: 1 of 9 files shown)


Sources/Moya/Plugins/NetworkLoggerPlugin.swift
```diff
@@ -62,25 +62,33 @@ private extension NetworkLoggerPlugin {
...
+                return requestHeader
+            }
+        }
+
+        if !headers.isEmpty {
             output += [format(loggerId, date: date, identifier: "Request Headers", message: headers.description)]
         }
 
-        if let bodyStream = request?.httpBodyStream {
+        if let bodyStream = request.request?.httpBodyStream {
```
LucianoPAlmeida: Nit: `request.request` can be a little confuse for people reading this later. What you think?

Tests/NimbleHelpers.swift
```diff
@@ -0,0 +1,18 @@
+import Nimble
+
+/// A Nimble matcher that succeeds when at least one of the substrings
+public func containOne(of substrings: String...) -> Predicate<String> {
+    return containOne(of: substrings)
+}
+
+/// A Nimble matcher that succeeds when at least one of the substrings
+public func containOne(of substrings: [String]) -> Predicate<String> {
```
LucianoPAlmeida: This could be upstreamed to nimble :))
what you think?

sunshinejr: I was thinking about it, but then I noticed a PR that allows concatenating expectations: https://github.com/Quick/Nimble/pull/493. This way we could introduce `or` and make it more generic. Already bumped the PR to see if we can get it updated & reviewed.

Sources/Moya/Plugins/NetworkLoggerPlugin.swift
sunshinejr: Well yeah, this is from the `RequestType` - we could rename the `request` to `urlRequest` maybe. What do you think?

LucianoPAlmeida: Seems nice 👍

PR #1880, opened 2019-07-08
Fixed NetworkLoggerPlugin behavior with errors.

amaurydavid:
Hi!

So the last few days I've been dealing with an API that errors a lot, and I was frustrated to se that when using the NetworkLoggerPlugin, errors would not be correctly displayed.

If using the Alamofire http codes validation and the API responds with a HTTP 500, `NetworkLoggerPlugin` would just log `Received empty network response for ...`, without any details put in the response's body by the API. To get them I had to copy the cURL and re-run manually the request in <Put here your preferred HTTP request testing service> to get the full response detail.

So I dived in NetworkLoggerPlugin and made a few changes.
First, I now fetch the error's response in the `didReceive(:target)` if there is any.
Second, I added a "Body: " prefix before the body content so that we can retrieve it more easily from the logs.

Examples using the same request:
Before:
`["Moya_Logger: [08/07/2019 16:28:54] Response: Received empty network response for <Here is a description of the target>"]`
After:
`["Moya_Logger: [08/07/2019 15:56:08] Response: <NSHTTPURLResponse: 0x1c002eaa0> { URL: <Some URL Here> } { Status Code: 500, Headers { <Some headers here>} }", "Body: {\"error\":{\"message\":\"Here is an error message from the API\",\"code\":178912}}"]`

edit: I will update tests so that they don't brake.


amaurydavid:
I've reworked a bit the logging to improve once again the behavior, and also updated tests.
The PR is now ready to be reviewed :)

Sources/Moya/Plugins/NetworkLoggerPlugin.swift
```diff
@@ -87,21 +90,32 @@ private extension NetworkLoggerPlugin {
...
 
         var output = [String]()
 
-        output += [format(loggerId, date: date, identifier: "Response", message: response.description)]
+        output += [format(loggerId, date: date, identifier: "Response", message: httpResponse.description)]
 
-        if let data = data, let stringData = String(data: responseDataFormatter?(data) ?? data, encoding: String.Encoding.utf8), isVerbose {
-            output += [stringData]
+        if let stringData = String(data: responseDataFormatter?(response.data) ?? response.data, encoding: String.Encoding.utf8),
+            isFromError || isVerbose {
```
sunshinejr: hmm, why does it matter if the response is from error or not?

sunshinejr:
Hey @amaurydavid, sorry for the delay, didn't see that you finish the work earlier! Really nice improvement, got one question but other than that it looks great. I think many people will appreciate the updates to the `NetworkLoggerPlugin`, myself included 😉

Sources/Moya/Plugins/NetworkLoggerPlugin.swift
amaurydavid: I guess it's based on the way I work which, as you ask the question, may not be the same as most of other users :p
I don't want the responses's body to be logged every time because it may contains a lot of informations that I don't care about if the request suceeded, so most of the time the verbose mode is off.
But what I do care about is to know why a certain request failed, and useful informations are often sent by APIs in the body, which is often a simple object with some properties: not such risk at clogging the console here.

sunshinejr: Ah, now it makes sense! Thanks for the explanation.

amaurydavid: I agree and will remove the forced log in error. About the log level, yes i've thought about that too, but it may deserve its own PR. I'll work on it.

amaurydavid:
> Hey @amaurydavid, would you be able to also add a changelog entry and check if we need to update our existing documentation for this plugin? Thanks!

I've added a changelog entry, the actual doc seems fine to me.

sunshinejr:
Awesome @amaurydavid! Do you think you can also resolve the conflicts? Sorry, we released new alpha version in the meantime...

PR #1886, opened 2019-07-29
Fix a typo

g-enius:


LucianoPAlmeida:
@f-meloni Just pinging you, since you implemented danger-swift here :)) do you have any idea about this error?

f-meloni:
The reason of that error is that there is a mismatch between the danger you are getting and what the plugins expects, this is due to the due to the 2.0 release, that has breaking changes on the models.
My best guess is that the problem is the cache that just needs to be deleted.

g-enius:
I only changed a single character of a typo in a Doc. Hadn't touch any source code. :)

LucianoPAlmeida:
@g-enius This is just  a CI issue, not related to the changes :))

LucianoPAlmeida:
> My best guess is that the problem is the cache that just needs to be deleted.
Thank's @f-meloni 
Handling to @sunshinejr or someone @Moya/core-team that have access to clear the CI cache and re-run it to see if works :))

sunshinejr:
Yeah we don't need to block this PR due to the CI failing, we will see if the next PR with the cache clearing will work. Thanks for the fix @g-enius!

PR #1891, opened 2019-08-02
Use CACHE_VERSION environment variable for CircleCI caches

sunshinejr:
Should fix our problems with caching in CircleCI. So in CircleCI 1.0 there was a button called "rerun without cache" that would usually fix our problems with inconsistent cache states for given builds. Unfortunately, with CircleCI 2.0 and workflows, there is no such button. 

[There is a thread on CircleCI forums](https://discuss.circleci.com/t/workflows-rerun-without-cache/22901) and maybe there will be such a button in the future, but for now the advice is [to use environment variables](https://support.circleci.com/hc/en-us/articles/115015426888-Clear-project-cache) and increment it when the cache changes. And this is what the PR does.

I added an environment variable in CircleCI dashboard `CACHE_VERSION` which currently has value of `2`.

Now, let's see if it builds 😄


sunshinejr:
Ah, almost worked! Need to ignore warnings, for some reason there is a warning  in `pod lint`:
> WARN  | url: The URL (http://twitter.com/ashfurrow) is not reachable.

PR #1894, opened 2019-08-05
Adding more customization to NetworkLoggerPlugin.

amaurydavid:
Following the discussion with @sunshinejr in #1880, here is a draft of a potential rework of the NetworkLoggerPlugin.

The issue we had was that response's body was only logged if we were in verbose mode. This means that if a request returns with an error, the error's body wasn't logged if not in verbose. It's ok if you don't want it, but it can be necessary according to your workflow. So we had to provide an option so that responses' bodies are logged in errors.
I could have used a simple enum `LogLevel { case error(showBodies: Bool), info, verbose }` but I wanted the user to fully customize the logs.

So I came with the following idea: list all the components that can be logged in requests & responses, and allow the user to opt-in for each one.
That's the purpose of the new `RequestLoggingOptions` and `ResponseLoggingOptions`.

And to handle the issue we were facing with bodies in errors, there are 2 sets of options for responses: the `successResponseLoggingOptions` will be used when the response returns as a success, and `errorResponseLoggingOptions` when it's an error. This way, we can configure the logger to log the error's body, but not success's.

Additionnaly, I've moved all of the plugin's config into a.... `Configuration` struct, making the Plugin more readable imo.

So it's just a draft of the direction I want to go, It's not even tested and things may be broken atm.
But I want to have some feedback before spending some time on it: do you feel it may be a suitable rework for NetworkLoggerPlugin?


Sources/Moya/Plugins/NetworkLoggerPlugin.swift
```diff
@@ -1,135 +1,197 @@
...
-public final class NetworkLoggerPlugin: PluginType {
-    fileprivate let loggerId = "Moya_Logger"
-    fileprivate let dateFormatString = "dd/MM/yyyy HH:mm:ss"
-    fileprivate let dateFormatter = DateFormatter()
-    fileprivate let separator = ", "
-    fileprivate let terminator = "\n"
-    fileprivate let cURLTerminator = "\\\n"
-    fileprivate let output: (_ separator: String, _ terminator: String, _ target: TargetType, _ items: Any...) -> Void
-    fileprivate let requestDataFormatter: ((Data) -> (String))?
-    fileprivate let responseDataFormatter: ((Data) -> (Data))?
```
amaurydavid: Not sure here why both dataFormatters did not have the same return type. As for now, i've made both of them return a String - which I believe is the logical choice for a logger -, but if it causes some trouble let me now.

Sources/Moya/Plugins/NetworkLoggerPlugin.swift
```diff
@@ -1,135 +1,197 @@
...
 
-fileprivate extension NetworkLoggerPlugin {
-    static func reversedPrint(_ separator: String, terminator: String, target: TargetType, items: Any...) {
-        for item in items {
-            print(item, separator: separator, terminator: terminator)
+//MARK: - Configuration
+public extension NetworkLoggerPlugin {
+    struct Configuration {
+
+        public typealias OutputType = (_ target: TargetType, _ items: [Any]) -> Void
```
amaurydavid: `[Any]` was previously `Any...`
Some questions here:
- the previous logger actually sent strings wrapped in an array. I do not understand why. Feel free to enlight me.
- why `Any` and not simply `String`? Entries sent by the logger are always `String`.

Tests/MoyaProviderIntegrationTests.swift
```diff
@@ -250,8 +251,7 @@ final class MoyaProviderIntegrationTests: QuickSpec {
                             provider.request(.zen) { _ in done() }
                         }
 
-                        expect(log).to(contain("Request:"))
-                        expect(log).to(contain("{ URL: https://api.github.com/zen }"))
+                        expect(log).to(contain("Request: https://api.github.com/zen"))
```
amaurydavid: The request's description wasn't correct here. I don't think it's due to my work but rather an update of AF 5 that wasn't fixed.

Tests/NetworkLoggerPluginSpec.swift
```diff
@@ -74,37 +73,34 @@ final class NetworkLoggerPluginSpec: QuickSpec {
...
+        it("outputs the formatted response data") {
             let response = Response(statusCode: 200, data: "cool body".data(using: .utf8)!, response: HTTPURLResponse(url: URL(string: url(GitHub.zen))!, mimeType: nil, expectedContentLength: 0, textEncodingName: nil))
             let result: Result<Moya.Response, MoyaError> = .success(response)
 
-            pluginWithRequestDataFormatter.didReceive(result, target: GitHub.zen)
+            pluginWithResponseDataFormatter.didReceive(result, target: GitHub.zen)
 
             expect(log).to(contain("Response:"))
             expect(log).to(contain("{ URL: https://api.github.com/zen }"))
-            expect(log).to(contain("formatted request body"))
```
amaurydavid: There seemed to be a mismatch between the test's name and its content: it was supposed to check the requestDataFormatter, but acted like it was testing the response. So I fixed it by trully checking the request.

amaurydavid:
I didn't have any feedback on the draft so I continued along the path.
Tests are now ok.
There are some changes I've made to simplify the previous behavior, but if it's a mistake let me now.
I've added some comments in the review on shadowy points.

Tests/NetworkLoggerPluginSpec.swift
```diff
@@ -7,37 +7,36 @@ final class NetworkLoggerPluginSpec: QuickSpec {
...
 
-        let pluginWithResponseDataFormatter = NetworkLoggerPlugin(verbose: true, output: { (_, _, _, printing: Any...) in
-            //mapping the Any... from items to a string that can be compared
-            let stringArray: [String] = printing.map { $0 as? String }.compactMap { $0 }
-            let string: String = stringArray.reduce("") { $0 + $1 + " " }
-            log += string
-        }, responseDataFormatter: { _ in
-                return "formatted body".data(using: .utf8)!
-        })
+
```
MoyaBot: - :warning: Limit vertical whitespace to a single empty line. Currently 2. (`vertical_whitespace`)

Sources/Moya/Plugins/NetworkLoggerPlugin.swift
```diff
@@ -1,135 +1,197 @@
...
-        if isVerbose {
-            items.forEach { output(separator, terminator, target, $0) }
-        } else {
-            output(separator, terminator, target, items)
+            configuration.output(target, logNetworkError(error, target: target))
         }
     }
 }
 
+//MARK: - Logging
```
MoyaBot: - :warning: MARK comment should be in valid format. e.g. '// MARK: ...' or '// MARK: - ...' (`mark`)

Sources/Moya/Plugins/NetworkLoggerPlugin.swift
```diff
@@ -1,135 +1,195 @@
...
 
-fileprivate extension NetworkLoggerPlugin {
-    static func reversedPrint(_ separator: String, terminator: String, target: TargetType, items: Any...) {
-        for item in items {
-            print(item, separator: separator, terminator: terminator)
+// MARK: - Configuration
+public extension NetworkLoggerPlugin {
+    struct Configuration {
+
+        public typealias OutputType = (_ target: TargetType, _ items: [Any]) -> Void
```
MoyaBot: - :warning: Types should be nested at most 1 level deep (`nesting`)

Sources/Moya/Plugins/NetworkLoggerPlugin.swift
```diff
@@ -1,135 +1,195 @@
...
+        public typealias DataFormatterType = (Data) -> (String)
```
MoyaBot: - :warning: Types should be nested at most 1 level deep (`nesting`)

amaurydavid:
I agree, the necessity of having to use .verbose on all 3 parameters struck me when updating the tests.
However i think the cURL option should stay with other options because the cURL format rules which component is logged and which is not.
Another solution could also be to reunite RequestLoggingOptions and ResponseLoggingOptions as ResponseLoggingOptions only has 1 useful option.
We would only have to rename  `RequestLoggingOptions` into `LoggingOptions`, and add a `successResponseBody` and `errorResponseBody` option.
Bonus: 2 less parameters in Configuration's init.

Tests/NetworkLoggerPluginSpec.swift
```diff
@@ -7,37 +7,28 @@ final class NetworkLoggerPluginSpec: QuickSpec {
...
```
MoyaBot: - :warning: Limit vertical whitespace to a single empty line. Currently 2. (`vertical_whitespace`)

Sources/Moya/Plugins/NetworkLoggerPlugin.swift
sunshinejr: if all entries are strings, feel free to replace the `Any` to `String` and see if it builds! if so, we can probably ship it as well

sunshinejr:
@amaurydavid thanks, I think it looks better! gonna checkout the branch and play with the logger :) 

cc @LucianoPAlmeida @pedrovereza are u guys up for a CR maybe?

Sources/Moya/Plugins/NetworkLoggerPlugin.swift
amaurydavid: i've replaced Any with String and it works fine :)

Tests/MoyaProviderIntegrationTests.swift
```diff
@@ -231,16 +231,11 @@ final class MoyaProviderIntegrationTests: QuickSpec {
 
                 describe("a provider with network logger plugin") {
                     var log = ""
-                    var plugin: NetworkLoggerPlugin!
+                    let plugin = NetworkLoggerPlugin(configuration: .init(output: { log += $1.joined() },
```
sunshinejr: oh nice this looks really clean now!

Sources/Moya/Plugins/NetworkLoggerPlugin.swift
```diff
@@ -1,135 +1,188 @@
 import Foundation
 
 /// Logs network activity (outgoing requests and incoming responses).
-public final class NetworkLoggerPlugin: PluginType {
-    fileprivate let loggerId = "Moya_Logger"
-    fileprivate let dateFormatString = "dd/MM/yyyy HH:mm:ss"
-    fileprivate let dateFormatter = DateFormatter()
-    fileprivate let separator = ", "
-    fileprivate let terminator = "\n"
-    fileprivate let cURLTerminator = "\\\n"
```
LucianoPAlmeida: Did we remove those constants and are using that hard coded now or they are not needed anymore?

amaurydavid: Previously the `dateFormatter` was set using a `dateFormat` provided in the `NetworkLoogerPlugin's` init. Now we have to directly provide the `DateFormatter`, which allow use to set it up using an hard coded `dateFormat`, or Apple's `dateStyle` and `timeStyle`. The default value uses a short date and time style.

`separator` and `terminator` are now hardCoded in the default output method.
`cURLTerminator` is no longer used.

LucianoPAlmeida: Nice @amaurydavid :))

Sources/Moya/Plugins/NetworkLoggerPlugin.swift
LucianoPAlmeida: Should we set the swiftlint flag to ignore this?

amaurydavid: Definitively, though I'm not sure whether to disable it for the whole project or just those two typealiases.

LucianoPAlmeida: I think we could just let for those two. Maybe this warning is good to have and decide if we fix or disable the warning on a case by case basis. cc @sunshinejr thoughts?

docs/MigrationGuides/migration_13_to_14.md
```diff
@@ -0,0 +1,9 @@
+# Migration Guide from 13.x to 14.x
```
sunshinejr: 💯

PR #1896, opened 2019-08-13
[WIP] SPM testing! (+ some other minor updates)

sunshinejr:
Thanks to @jeffctown, OHHTTPStubs is getting a support for SPM! This was something that was keeping us from adopting SPM testing. It's still in PR so not merged and so this PR is in [WIP] state.


f-meloni:
```
Skipping Danger due to this run not executing on a PR.
```
😮 what happened to Danger js?

sunshinejr:
@f-meloni how did you catch that? 😄 Not sure what happened, but it seems like for some reason [there is no source](https://github.com/danger/danger-js/blob/b5c09199da7e3ec309cb12a7071e1f5d16e0aefe/source/commands/ci/runner.ts#L41-L43) (doubt the `isPr` is set to `false`, but maybe)?

f-meloni:
Was having a look to see if a cache reset was still needed (and I think it was, but it didn't arrive to that point :D)
It looks everything ok in that build, the only thing I can think is that `CIRCLE_CI_API_TOKEN ` is missing for any reason https://github.com/danger/danger-js/blob/master/source/ci_source/providers/Circle.ts#L62? 🤔

sunshinejr:
Interesting! I'm pretty sure we didn't change anything in the CircleCI settings since https://github.com/Moya/Moya/pull/1891 🤔

f-meloni:
Didn't see that PR, it looks that solved the cache problem for Danger-swift as well (pretty cool :P)
The only thing I can think is that, given the other values are almost always there, it recognises is in CI and Circle recognises is running on PR, then should have all the requested PR values like  `CIRCLE_PR_NUMBER`,  the only thing I could think of is some problem with the token :/

sunshinejr:
Ahh, @f-meloni, I think [this might be the case](https://github.com/danger/danger/blob/56286aab4d3b9644eb6da5e81d0ab5b32575e9d4/lib/danger/ci_source/circle.rb#L13-L16):
>  A common scenario is when CircleCI begins building a commit before the commit becomes associated with a PR (e.g. a developer pushes their branch to the remote repo for the first time. CircleCI spins up and begins building. Moments later the developer creates a PR on GitHub. Since the build process started before the PR existed, Danger won't be able to use the Circle-provided environment variables to retrieve PR metadata.)

sunshinejr:
Although this is still in WIP process, I'm gonna merge it as this has some value in it (SPM fixes, version bump for ReactiveSwift 6.1). If the PR on OHHTTPStubs is updated, I will make another PR with a follow-up as well.

PR #1904, opened 2019-09-02
Combine support!

### LINKED ISSUE #1870: Combine-based version of Moya?
I'm personally not a reactive programming advocate (actually far away from it) but now that Apple has their own take on it with the [Combine](https://developer.apple.com/documentation/combine) library in Xcode 11, I will definitely give it a second chance and therefore also wonder if it would make sense to add `MoyaCombine` to the list of variants we already provide for Moya. That one, I might actually use some day. :)

### LINKED ISSUE #1871: Result should be 5.0.0?
Result should be 5.0.0?

sunshinejr:
Fixes #1871


Sources/Moya/Combine/MoyaPublisher.swift
```diff
@@ -0,0 +1,43 @@
...
+        }
+    }
+
+    private let callback: (AnySubscriber<Output, MoyaError>) -> Cancellable?
+
+    init(callback: @escaping (AnySubscriber<Output, MoyaError>) -> Cancellable?) {
+        self.callback = callback
+    }
+
+    public func receive<S>(subscriber: S) where S : Subscriber, Failure == S.Failure, Output == S.Input {
```
MoyaBot: - :warning: Colons should be next to the identifier when specifying a type and next to the key in dictionary literals. (`colon`)

Tests/MoyaTests/MoyaProvider+CombineSpec.swift
```diff
@@ -0,0 +1,322 @@
...
+
+                _ = provider.combine.request(.zen)
+                    .sink(receiveCompletion: { completion in
+                        switch completion {
+                        case let .failure(error):
+                            receivedError = error
+                        case .finished:
+                            ()
+                        }
+                    }, receiveValue: { response in
```
MoyaBot: - :warning: Unused parameter "response" in a closure should be replaced with _. (`unused_closure_parameter`)

sunshinejr:
@Moya/contributors anyone able to make a review for this one? 🙏

Sources/Moya/Combine/AnyPublisher+Response.swift
```diff
@@ -0,0 +1,120 @@
...
+
+#if canImport(UIKit)
+    import UIKit.UIImage
+#elseif canImport(AppKit)
+    import AppKit.NSImage
+#endif
+
+/// Extension for processing raw NSData generated by network access.
+@available(OSX 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
+extension AnyPublisher where Output == Response, Failure == MoyaError {
```
LucianoPAlmeida: You could make it a public extension and avoid have make every method public, like the one in `Sources/Moya/Combine/MoyaProvider+Combine.swift` :))

Sources/Moya/Combine/MoyaPublisher.swift
```diff
@@ -0,0 +1,42 @@
+#if canImport(Combine)
+
+import Combine
+
+// This should be already provided in Combine, but it's not.
+// Ideally we would like to remove it, in favor of a framework-provided solution, ASAP.
+
+@available(OSX 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
+internal class MoyaPublisher<Output>: Publisher {
```
sunshinejr: @freak4pc I know you're really busy, but would you mind taking a look at this publisher code? would really appreciate that

Sources/Moya/Combine/AnyPublisher+Response.swift
sunshinejr: ah good one! I copied the base for it from Rx/Reactive extensions and they have the same issue, gonna fix them all!

freak4pc: There's a Swiftlint rule for this if you'd like :]

Sources/Moya/Combine/MoyaPublisher.swift
freak4pc: Hey - I'll only be available in a week and a half (flying to try! Swift tomorrow), not sure if it would help you by then :(

sunshinejr:
Seems like tests on macOS are crashing because the CircleCI environment has macOS Mojave, not Catalina? There is this `@available(OSX 10.15...)` so it shouldn't run but let me check this one.

Sources/Moya/Combine/AnyPublisher+Response.swift
sunshinejr: ah nice good idea, added the rule 👍

Sources/Moya/Combine/MoyaPublisher.swift
sunshinejr: Ah no problem, seems to be working fine so I just wanted another pair of eyes on it but it can go to beta and see how it works 👍

sunshinejr:
Okay it built, but Danger seems to have troubles (currently xcode-summary plugin). cc @f-meloni

f-meloni:
:O today I will have a look

sunshinejr:
@f-meloni no rush!

f-meloni:
Could be Xcode 11.0 the problem 🤔

f-meloni:
Found it!
```
<unknown>:0: error: invalid value '5.1' in '-swift-version 5.1'
<unknown>:0: note: valid arguments to '-swift-version' are '4', '4.2', '5'
```
Marathon sets 5.1 as version, but apparently that is not recognised as valid argument from `swiftc` yet.
Now this is the problem... Will have a thinking on how solve it, before proposing SPM + Rocket, that is a possibility and would remove the need of using Marathon

sunshinejr:
@f-meloni I think 5.1 is just specified as 5 but depending on Xcode it will use either 5.0 or 5.1 so they probably won't add 5.1 as a valid argument. 5.1 seems to be valid argument only for SPM tools version for the moment.

I'm gonna merge this PR for now so we have one valid cache for the PRs. If anyone finds something in this PR don't hesitate to comment/open a PR/open an issue!

f-meloni:
@sunshinejr It accepts  `'4', '4.2'`, then I would expect it to accept `'5', '5.1'` too `:thinking:`

sunshinejr:
@f-meloni but 4.1 was still 4 in terms of swift version, I think the 4.2 is because there was a lot of breaking changes and it should’ve been named 5.0 but they couldnt’t because the API wasn’t stable. So in a sense these are all just major releases.

f-meloni:
@sunshinejr Yeah you probably are right!

f-meloni:
Added a fix to Marathon that should solve the problem

https://github.com/JohnSundell/Marathon/pull/209

vikmeup:
@sunshinejr is there any extra steps requires to see it when you do pod install of `14.0.0-beta.3` version?  those classes not exposed that introduced in the app.

PR #1906, opened 2019-09-03
NetworkLoggerPlugin formatter

amaurydavid:
Continuation of PR #1644 using the new NetworkLoggerPlugin.

I've kept the idea of having a Formatter object acting as a namespace for all formatting closures used by the plugin. It contains a closure to format entries (formerly an internal func) and request and response bodies formatter that moved 1 level deeper in objects hierarchy.
Having the entries customizable made the `Configuration.dataFormatter` useless and it's now specific to the default implementation. The `Configuration.loggerId` was removed for the same reason.

I've also made most plugin's properties `public var` instead of `fileprivate let` because now that there is a new `Formatter` object, the plugin's init would grow a bit dirtier in my opinion. Having `var` properties now allow us to update them on the fly after the init:
```
//This works
let pluginWithRequestDataFormatter = NetworkLoggerPlugin(configuration: .init(formatter: .init(requestData: { _ in return "formatted request body" }),
                                                                                          output: customLoggerOutput,
                                                                                          logOptions: .verbose))

//This also works
 let pluginWithRequestDataFormatter: NetworkLoggerPlugin = {
                let plgn = NetworkLoggerPlugin()
                plgn.configuration.output = customLoggerOutput
                plgn.configuration.logOptions = .verbose
                plgn.configuration.formatter.requestData = { _ in return "formatted request body" }
                return plgn
            }()
```

I didn't change the `Configuration.ouput` closure because because it's a matter of where to write the log entries whereas the `Formatter` is a matter of how format entries.


Sources/Moya/Plugins/NetworkLoggerPlugin.swift
```diff
@@ -109,67 +104,46 @@ private extension NetworkLoggerPlugin {
...
             self.logOptions = logOptions
         }
 
-        public static var defaultDateFormatter: DateFormatter {
-            let formatter = DateFormatter()
-            formatter.timeStyle = .short
-            formatter.dateStyle = .short
-            return formatter
-        }
+        //MARK: - Defaults
```
MoyaBot: - :warning: MARK comment should be in valid format. e.g. '// MARK: ...' or '// MARK: - ...' (`mark`)

Examples/_shared/GitHubAPI.swift
```diff
@@ -15,8 +15,14 @@ private func JSONResponseDataFormatter(_ data: Data) -> String {
     }
 }
 
-let gitHubProvider = MoyaProvider<GitHub>(plugins: [NetworkLoggerPlugin(configuration: .init(responseDataFormatter: JSONResponseDataFormatter,
-                                                                                             logOptions: .verbose))])
+let loggerPlugin: NetworkLoggerPlugin = {
```
sunshinejr: can we leave the inline initialization in the tests/examples? (plugins: [NetworkLoggerPlugin(configuration...]) - this will show us better what impact it will have on usage

sunshinejr:
@amaurydavid thanks for this! can you take care of my comment & SwiftLint comments? this is a breaking change so ideally I would love to get it into beta which should happen in the next few days as Alamofire 5 RC is out.

Examples/_shared/GitHubAPI.swift
amaurydavid: The inline declaration is back :)

amaurydavid:
Tests are failling but I suppose it's due to some configuration changes for combine?

Tests/MoyaTests/NetworkLoggerPluginSpec.swift
```diff
@@ -11,19 +11,12 @@ final class NetworkLoggerPluginSpec: QuickSpec {
...
-                                                                      logOptions: [.formatRequestAscURL]))
-
-        let pluginWithRequestDataFormatter = NetworkLoggerPlugin(configuration: .init(output: customLoggerOutput,
-                                                                                      requestDataFormatter: { _ in return "formatted request body" },
-                                                                                      logOptions: .verbose))
-
-        let pluginWithResponseDataFormatter = NetworkLoggerPlugin(configuration: .init(output: customLoggerOutput,
-                                                                                       responseDataFormatter: { _ in return "formatted response body" },
-                                                                                       logOptions: .verbose))
+        let plugin: NetworkLoggerPlugin = {
```
sunshinejr: can you rename the `plgn` to something else? `plugin` might work with shadowing variables but if not, maybe `networkPlugin`?

sunshinejr:
@amaurydavid ah yes we have different cache for the PRs now so it failed... I just merged the Combine support, once you push the minor changes for the `plgn` naming it should build correctly (but still fail on Danger as we are in progress of improving it for Swift 5.1...)

PR #1912, opened 2019-09-09
Alamofire 5.0.0-rc.2

sunshinejr:
Alamofire 5.0.0-rc.2 was just released with a quickfix - so we will release beta 2 as well!


sunshinejr:
Tests passed (only Danger failed but its a known issue) so merging!

PR #1915, opened 2019-09-11
Link Combine weakly

### LINKED ISSUE #1914: Issue with Combine on iOS 12
In an app of mine, I included Moya Version `14.0.0-beta2` via Accio (As Accio is built on Carthage, I effectively installed it via Carthage).

Running my app on iOS 13 (Simulator) works fine, but when I run on my iPhone with iOS 12, the app crashes on launch and Xcode logs the following error message:

```
dyld: warning: could not load inserted library '__PLATFORMS__/iPhoneOS.platform/Developer/usr/lib/libXCTestBundleInject.dylib' into hardened process because image not found
dyld: Library not loaded: /System/Library/Frameworks/Combine.framework/Combine
  Referenced from: /private/var/containers/Bundle/Application/10FE3E5A-2604-473C-ACB1-B52AD8B6E469/Debug.app/Frameworks/Moya.framework/Moya
  Reason: image not found
```

Seemingly, Combine is expected, but, of course, it's not available when running iOS 12.

As Moya is already pre-compiled and doesn't know which iOS version it will be deployed to at compile-time, the Combine framework must somehow be made "optional" and not required in order to keep iOS 12 compatibility...

fredpi:
Fixes #1914.

I linked Combine weakly as proposed here: https://stackoverflow.com/questions/57168931/optional-linking-for-swift-combine-framework-in-xcode-11

After installing Moya via my branch, the issue went away so I could run my app on iOS 12 devices again.

As proposed in the SO answer, maybe the weak linking configuration should be done via a XCConfig file to maintain compatibility with Xcode 10:

> or if you still want to support building with older Xcode versions:
> 
> `OTHER_LDFLAGS[sdk=iphoneos13.0] = -weak_framework Combine`

As there are no XCConfig files currently and I needed a quick fix, I didn't opt for this approach for now.

### DIFF
--- a/Config/Moya.xcconfig
+++ b/Config/Moya.xcconfig
@@ -1,0 +1,2 @@
+
+OTHER_LDFLAGS[sdk=iphoneos13.0] = -weak_framework Combine
--- a/Config/Moya.xcconfig
+++ b/Config/Moya.xcconfig
@@ -1,2 +1,12 @@
 
 OTHER_LDFLAGS[sdk=iphoneos13.0] = -weak_framework Combine
+OTHER_LDFLAGS[sdk=iphonesimulator13.0] = -weak_framework Combine
+
+OTHER_LDFLAGS[sdk=watchos6.0] = -weak_framework Combine
+OTHER_LDFLAGS[sdk=watchsimulator6.0] = -weak_framework Combine
+
+OTHER_LDFLAGS[sdk=appletvos13.0] = -weak_framework Combine
+OTHER_LDFLAGS[sdk=appletvsimulator13.0] = -weak_framework Combine
+
+OTHER_LDFLAGS[sdk=macosx10.15] = -weak_framework Combine
+OTHER_LDFLAGS[sdk=macosx10.15] = -weak_framework Combine
--- a/Config/Moya.xcconfig
+++ b/Config/Moya.xcconfig
@@ -9,4 +9,3 @@ OTHER_LDFLAGS[sdk=appletvos13.0] = -weak_framework Combine
 OTHER_LDFLAGS[sdk=appletvsimulator13.0] = -weak_framework Combine
 
 OTHER_LDFLAGS[sdk=macosx10.15] = -weak_framework Combine
-OTHER_LDFLAGS[sdk=macosx10.15] = -weak_framework Combine
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,8 @@
 # Next
 
+### Fixed
+- Fixes an issue that caused apps to crash on startup when using the latest, Combine-enabled Moya version on non-iOS 13 devices. [#1915](https://github.com/Moya/Moya/pull/1915) by [@fredpi](https://github.com/fredpi).
+
 # [14.0.0-beta.2] - 2019-09-09
 
 ### Changed
--- a/Moya.xcodeproj/project.pbxproj
+++ b/Moya.xcodeproj/project.pbxproj
@@ -1010,6 +1010,10 @@
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/../Frameworks @loader_path/Frameworks";
 				MACOSX_DEPLOYMENT_TARGET = 10.12;
 				MTL_ENABLE_DEBUG_INFO = NO;
+				OTHER_LDFLAGS = (
+					"-weak_framework",
+					Combine,
+				);
 				PRODUCT_BUNDLE_IDENTIFIER = com.cocoapods.Moya;
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				SDKROOT = macosx;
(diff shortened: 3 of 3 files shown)


fredpi:
@sunshinejr In contrast to what I stated earlier, I have now used a config file to realize the weak linking, thereby maintaining Xcode 10 compatibility - turned out not to be that complicated ;)

Can I do anything to fix the tests?

sunshinejr:
@fredpi ah no we currently have an issue with Danger... It seems like all of the tests are passing so it's all fine. Can you add a changelog entry at the top? Similar to what I've done [here](https://github.com/Moya/Moya/pull/1916/files#diff-9b8fab691c00b9e5380b19ce882f3271). 

Also, do you think it's Carthage-specific issue or rather infected all package managers?

fredpi:
@sunshinejr It definitely affects all package managers that precompile the framework. I have no idea how CocoaPods works exactly, but I guess the issue was also there with CocoaPods.

fredpi:
@sunshinejr Turns out, while the first approach worked for iOS 12 devices and the iOS 12 simulator, the second one (using the config file used to maintain Xcode 10 compatibility) does work for iOS 12 devices but not for the iOS 12 simulator. Trying to fix...

fredpi:
@sunshinejr Should work properly now ;)

sunshinejr:
Thanks @fredpi! It’s interesting though, I was thinking that we would need to link weakly for iOS 12 instead of 13... Do you know why ia that?

Config/Moya.xcconfig
```diff
@@ -0,0 +1,11 @@
...
+OTHER_LDFLAGS[sdk=iphoneos13.0] = -weak_framework Combine
+OTHER_LDFLAGS[sdk=iphonesimulator13.0] = -weak_framework Combine
+
+OTHER_LDFLAGS[sdk=watchos6.0] = -weak_framework Combine
+OTHER_LDFLAGS[sdk=watchsimulator6.0] = -weak_framework Combine
+
+OTHER_LDFLAGS[sdk=appletvos13.0] = -weak_framework Combine
+OTHER_LDFLAGS[sdk=appletvsimulator13.0] = -weak_framework Combine
+
+OTHER_LDFLAGS[sdk=macosx10.15] = -weak_framework Combine
```
amaurydavid: Does this means we'll have to update this for every sdk update?

fredpi: I'm not entirely sure, but I think it means iOS 13+, so we won't have to update it

sunshinejr: what if we drop the config per sdk and just use `OTHER_LDFLAGS = -weak_framework Combine`? is there any drawback in linking Combine everywhere? will Xcode 10 be able to just ignore this statement?

sunshinejr: nvm just tried that and and it didn't ignore it but stopped the build... hmm...

fredpi: I tried, and for me, Xcode 10 failed to build it without the sdk specification

sunshinejr: unfortunately I tried adding a random weak framework to `sdk=iphonesimulator12.4` and it didn't build (as it should) on 12.4 but it built correctly on 13.0 so @amaurydavid raises a valid point - we would need to update that on each sdk release... is there any other way we could do the linking on 13+?

sunshinejr:
Also I'm still trying to get it to reproduce 🤔 I have macOS Catalina and iOS 12.4 device and it seems to be working correctly on SPM/CocoaPods. Carthage is still left to be tested but if it's also working in there, will this be a Mojave problem maybe? @fredpi what macOS version do you have?

Config/Moya.xcconfig
sunshinejr: hey @fredpi, was thinking, we can actually just make sure everyone uses Xcode 11 in the development process anyways - the only question is if it will work for people that use Moya in Xcode 10? did you try that maybe?

fredpi:
@sunshinejr I also tried with SwiftPM and everything worked fine without the adjustments from this PR, even though I'm running Mojave. Accio, in contrast, still fails. I've uploaded a test project I created to check whether Accio (built on Carthage) is working: https://github.com/fredpi/TestAccioMoya

Would you please try this project and check whether it also fails for you?

Config/Moya.xcconfig
fredpi: In the SO answer I referenced, there's also another suggestion:

```
OTHER_LDFLAGS_XCODE_SPECIFIC_1100 = -weak_framework Combine

OTHER_LDFLAGS = $(inherited) ${OTHER_LDFLAGS_XCODE_SPECIFIC_$(XCODE_VERSION_ACTUAL)}
```

But it seems like that needs updating, too.

I also think, dropping Xcode 10 support altogether isn't feasible at this point of time...

sunshinejr: Yeah that doesn't seem as a better alternative. And yeah we don't want to remove support for Xcode 10 for our users for sure.

I'm just wondering if we remove the `sdk` constraint will it:
- break development in Xcode 10 (meaning Moya developers will have to use Xcode 11)
- break both development and client-side for Xcode 10

sunshinejr:
@fredpi of course, will try that. Can you check if the framework built by Carthage, not Accio, will also work for you?

sunshinejr:
**Update:** I tested both with Accio and Carthage and it **doesn't work**. I also tested with CocoaPods and SPM and it **does work**.

Now, I added weak-linking for the Combine framework in [this branch](https://github.com/Moya/Moya/tree/fix/weak_link) but it didn't help 😮 I checked with `otool` and the framework is indeed weakly referenced (`LC_LOAD_WEAK_DYLIB`) but it still crashes on my phone with iOS 12.4. Really interesting.

sunshinejr:
Alright I think I found a way: a custom build script that will update build settings based on the Xcode major version... Really, really hacky but it seems to work. Gonna test it a little bit more though.

sunshinejr:
see #1920

PR #1916, opened 2019-09-11
NetworkLoggerPlugin curl fixes.

sunshinejr:
Alamofire changed how they print `curlDescription` so we have to adapt as well!


Sources/Moya/RequestTypeWrapper.swift
```diff
@@ -28,4 +28,9 @@ struct RequestTypeWrapper: RequestType {
         let newRequest = _request.authenticate(with: credential)
         return RequestTypeWrapper(request: newRequest, urlRequest: _urlRequest)
     }
+
+    func cURLDescription(calling handler: @escaping (String) -> Void) -> RequestTypeWrapper {
+        _request.cURLDescription(calling: handler)
```
amaurydavid: Is there a specific reason why you used the async version of `cURLDescription(calling:)` while there is a sync one: `cURLDescription()` ?

PR #1920, opened 2019-09-17
Weak-link Combine.

### LINKED ISSUE #1914: Issue with Combine on iOS 12

sunshinejr:
Fixes #1914. Supersedes #1915. 

(You can read the latter to see the discussion around possible fixes) 

I have no more ideas to get this one done without introducing more problems. And we need to have something working really soon. Two not-so-perfect but viable options are either to use weak linking with SDK/Xcode constraints or without. 

Using weak linking without constraints:
- will work fine on Xcode 11+
- won't break a lot of apps in a year or so
- will introduce another step for Xcode 10 users that use Carthage and Moya 14+

Using weak linking with constraints:
- will work fine on any Xcode
- we would need to remember to update build settings every SDK/Xcode release
- even if we do the above, every app that uses Moya 14 (without updated xcodeproj for newest SDKs) and Carthage will crash on runtime on iOS 12 and this is really scary

And so the one without constraints seems like a reasonable way to go at least for now. Would be awesome to figure out a way without a script for Carthage + Xcode 10 users but this should do for now.

cc @fredpi


fredpi:
@sunshinejr 

> Edit: My idea was to create a build-phase script... It worked but it was too late in the compilation process and so it won't work for Carthage unfortunately. Still looking for a viable solution but running out of ideas...

Do you mean this PR by your comment in the other thread, or is this yet another, actually working approach?

sunshinejr:
Unfortunately the build phase odea wasn’t working for carthage as we replaced the build settings too late in the build process. So this is the best solution I found for now @fredpi.

sunshinejr:
@fredpi gonna merge & release new beta with this one but lemme know if you figure out a better way of handling this.

PR #1921, opened 2019-09-17
Fix stubbed requests that didn't persist plugin changes.

### LINKED ISSUE #1918: AccessTokenPlugin not working in Unit Test
Moya Version: 14.0.0-beta.2

```swift
import Quick
import Nimble
import Moya

class Testing: QuickSpec {

    struct TestTarget: TargetType, AccessTokenAuthorizable {
        let baseURL = URL(string: "http://www.api.com/")!
        let path = ""
        let method = Method.get
        let task = Task.requestPlain
        let sampleData = Data()
        let headers: [String: String]? = nil
        let authorizationType: AuthorizationType
    }

    override func spec() {

        var sut: MoyaProvider<TestTarget>!

        beforeEach {
            let token = UUID().uuidString
            let authPlugin = AccessTokenPlugin { token }
            sut = MoyaProvider<TestTarget>(stubClosure: MoyaProvider.immediatelyStub,
                                           plugins: [authPlugin])
        }

        it("test") {
            var header: String?
            _ = sut.rx.request(TestTarget(authorizationType: .bearer))
                      .subscribe { event in
                          switch event {
                              case .success(let response):
                                  header = response.request?.value(forHTTPHeaderField: "Authorization")
                              case .error:
                                  fail()
                          }
                      }

            expect(header).toNot(beNil())
        }
    }
}
```
Testing Result `expected to not be nil, got <nil>`.

I got confused that the plugin seems not working in Stub mode, but when I connect to real network the plugin work as usual.

My question: Is it normal that plugin is not working in test ? Or am I testing in wrong way ?
Thanks.

sunshinejr:
Fixes #1918. 

Also replaced current test image with a lot lighter version (1x1 pixel image) so our integration tests  are quicker on slow connection.


sunshinejr:
cc @f-meloni, seems like 2.0.3 had problems with fetching the plugin xcode-summary

f-meloni:
It probably doesn't compile with xcode 11, let me try it

f-meloni:
I've reproduced it locally, I think the problem is related to Marathon pointing to a revision and not a version, and that was conflicting with the other dependencies.
I've released `2.0.4`, that locally is working.
Schermata 2019-09-17 alle 20 36 53

f-meloni:
🎉

Issue #1639, opened 2018-04-12
[Suggestion] Simpler version of Moya?


jdisho:
Hey guys 👋

I am using `Moya` in some projects and I love how it abstracts `Alamofire` and how easy it is to use. ❤️
But, personally, I see a problem with `Moya` when it comes to projects that does not necessarily need that much networking. 

For instance, there are projects that just need a small networking provider to request `JSON` data from an endpoint and map it to a `Decodable` object. I would rather write my own provider rather that installing `Moya`(together with `Alamofire` and `Result`) which increases the amount of dependencies for something that supposed to be easy.

My suggestion would be to introduce a lightweight Moya version (i.e `Moya-Lite` as a separate library) with no other dependencies to support just some basic, common networking tasks. 

-  A tiny wrapper around `NSURLSession`
- Supporting CRUD methods (`GET`, `POST`, `PUT`, `DELETE`).
- Working only with `Decodable` objects.

Similar to this: https://github.com/jdisho/TinyNetworking 
(My intention is **not to advertise** my networking library, but more as an example to better understand my suggestion.)

Hopefully I was clear otherwise I am open to answer some questions. 

Let me know what do you think. 😊

SD10:
@jdisho Thanks for writing this up 👍 

To be honest, I’m not sure what a lighter weight version of Moya would look like. The Moya codebase is already extremely small and well focused. I’m not sure if a thinner layer [could even be considered Moya anymore](https://github.com/Moya/Moya/blob/master/Vision.md). 

It sounds like in this situation someone should just be using Alamofire directly. Making the request you described above with Alamofire is less than 15 lines of code with decoding included.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

jdisho:
You misunderstood my suggestion @SD10, probably I am not explaining it well 😄

What I am suggesting is to have a `Moya-Lite` version with no dependencies on `Alamofire` and `Result`, with the same [project goals](https://github.com/Moya/Moya/blob/master/Vision.md) and covering basic networking cases like exchanging data with some endpoints using CRUD methods and working only with `Codable` objects.

I don't have anything against Moya's usability, it is easy, understandable, concise and as you mentioned, well focused. I am concerned about the dependencies it introduces when installing it, `Alamofire` and `Result` in some cases.

[...]

PR #1924, opened 2019-09-20
Built-in retry! 🎉

sunshinejr:
_(Note: we should wait with this PR until all the current fixes are merged & released in the next beta, but this is something I want to support soon)_

TODO:
- [ ] tests
- [ ] documentation
- [ ] examples (maybe GitHub provider should use a retry closure)

### DIFF
--- a/Moya.xcodeproj/project.pbxproj
+++ b/Moya.xcodeproj/project.pbxproj
@@ -68,7 +68,7 @@
 		85C98006231D41E400AAAFB2 /* MoyaPublisher.swift in Sources */ = {isa = PBXBuildFile; fileRef = 85C98005231D41E400AAAFB2 /* MoyaPublisher.swift */; };
 		85C98008231D4BE900AAAFB2 /* AnyPublisher+Response.swift in Sources */ = {isa = PBXBuildFile; fileRef = 85C98007231D4BE900AAAFB2 /* AnyPublisher+Response.swift */; };
 		85D5277A2302BEF30093E9C1 /* TestImage.swift in Sources */ = {isa = PBXBuildFile; fileRef = 85D527792302BEF30093E9C1 /* TestImage.swift */; };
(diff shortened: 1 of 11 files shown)


amaurydavid:
I've recently struggled using a combination of Alamofire's `requestRetrier` and `requestAdapter` with Moya plugins, and finally ended up handling the retrying by myself with embedded `moyaprovider.request()` calls.
The main issue was that when retrying with the `requestRetrier`, `requestAdapter` was used upon retry but Moya plugins were not called. The reason is that for Moya perspective, Alamofire was only performing 1 request even if underhood Alamofire was performing multiple times the request through `requestRetrier`.
Note: tested on AF 4, I don't know if it still works the same way with AF 5, but I think `requestRetrier` and `requestAdapter` had their share of changes.

Having a built-in retry feature would allow plugins to be called upon every retry, which is nice.

There are some points I would like to raise, though it may be best to discuss them in a dedicated thread:
- If we provide a retry feature, we will want to make sure it's not possible to use Moya's and Alamofire's at the same time. At the moment, it's possible to access the Alamofire's sessionManager using `moyaProvider.manager`, maybe we should make the property internal? The drawback is that we would have to provide a proxy for any property not directly handled by Moya.

- Do we have a list of Alamofire's features overriden by Moya? It may be a good idea to keep one up to date. I believe the status code validation can also be overriden according to user's preference.
If we end up overriding too many of Alamofire's features, may it would be time to think about not relying on Alamofire? It may be related to issue #1639 at some point.

sunshinejr:
Hey @amaurydavid - great points. Generally you shouldn't need to use the session at all, we try to provide as much flexibility as possible but if something's not there you can just customize the session (or sessionManager in Moya < 14).  

As for the list there is nothing like that AFAIK. 

The resolution of #1639 was that we don't really want to remove Alamofire as a dependency due to the fact that we would need to provide our own networking layer (which Moya shouldn't really care about as we are just a layer on top of that).

But this discussion was year ago and maybe we could try again? Maye instead of removing Alamofire (and using our own networking layer in Moya) we could to write a small library under Moya organization that would handle networking? Not sure if we have the capacity for that.

adrianodiasx93:
+1

iWECon:
Why no merger?

Issue #1928, opened 2019-10-01
Combine support mega-thread.


sunshinejr:
We've introduced Combine support in 14.0.0-beta.1 and unfortunately we introduced some linking issues with that. Unfortunately, it doesn't seem to be an easy fix as we have to do a lot of workarounds to not crash your app:

## Carthage
When adding `-weak_framework Combine` linker flags it's actually working great on both debug and release builds! Oh, unfortunately only on Xcode 11. On Xcode 10 we created a compatibility script that you would need to use instead of normal `carthage update/bootstrap`:
```bash
carthage update Moya --platform iOS --no-build
./Carthage/Checkouts/Moya/scripts/carthage-xcode10-compat.sh
carthage build Moya --platform iOS
```

## CocoaPods
It works good, but only in debug builds. It's correctly weak-linking the Combine library but when you archive your project, it suddenly links strongly. I'm no expert to know why would that happen, but I guess it might be something with automatic linking? Not sure.

We also have a workaround for that: adding `s.pod_target_xcconfig = { 'OTHER_LDFLAGS' => '-weak_framework Combine' }` to the .podspec works (probably `s.weak_framework` would work as well) but it breaks Xcode 10 compatibility as well. I haven't found a way of adding a workaround for Xcode 10 yet.

## Swift Package Manager
It also works good in debug builds, but doesn't on release builds. Unfortunately the only workaround for release builds I found is using linker flags in `swift build` phase:
```bash
swift build -c release -Xlinker "-weak_framework" -Xlinker "Combine"
```

It's really bad in terms of user experience and also not sure how to pass that to the Xcode-SPM integration. SPM has ability to pass flags but weak linking isn't something we can pass right now so it will take some time to introduce it (FWIW there is a way to strongly link a framework so there is a hope).

# Next steps
I'm still investigating possible options but we might just remove Combine support from version 14 and introduce it in later on.

Note: If anyone has more experience in this area and could help I would really appreciate that. I'll keep you all posted on my findings as well.

sunshinejr:
An update: I was talking with Ankit from SwiftPM and basically it seems like a bug in Combine framework (see linked FB issue) and there is nothing we can do more for SwiftPM integration in terms of workarounds, unfortunately.

I will be most likely removing Combine extensions from Moya for now and maybe introduce it later on, but not sure yet.

sunshinejr:
Just release 14.0.0-beta.4 which removes Combine support & thus removed runtime crashes on iOS lower than 13.

[...]

PR #1930, opened 2019-10-03
Add Combine folder to source_files in Podspec

vikmeup:
Currently if you use Moya via cocoa-pods install, `Combine` extensions not being visitble to the xcode project, adding `Combine` folder to `source_files` should fix it.

I was able to use combine extension on my branch via: `provider.requestPublisher()`


sunshinejr:
@vikmeup @leoneparise thanks for the fix guys! Unfortunately due to #1928 we will remove Combine support from version 14, but it’s good that we have a proper version to get back to after we figure it out

PR #1933, opened 2019-10-05
Remove Combine.

sunshinejr:
Due to #1928, we are removing Combine extensions for now. Probably gonna be back once we stop supporting Xcode 11 (should be after Moya 14).


sunshinejr:
Alright tests passed, merging this one & releasing new beta...

Issue #1932, opened 2019-10-04
MoyaProvider.RequestClosure is there a way to get the enum that is being called?


DMCApps:
Hello

I'm attempting to add a middleware step to my requests that attempts to refresh my tokens but I only want it to be done when the request requires a token. I don't see anywhere that I can get the request being made and check if it's a request that requires a token.

e.g.
```swift
// Closure run on each request in order to check the validity of the token
// The token will be refreshed whenever needed.
let requestClosure: MoyaProvider.RequestClosure = { (endpoint: Endpoint, done: @escaping MoyaProvider.RequestResultClosure) in
    
    // Only proceed if the task is AccessTokenAuthorizable and it's authorization type is bearer.
    
    // TODO: How do I get the enum being called to check if it's type is AccessTokenAuthorizable and get the value of that key for this request? 
    // Something like below except .task is not the proper value as I can't call these methods on it.
    let authorizationType: AuthorizationType = (endpoint.task as? AccessTokenAuthorizable)?.authorizationType ?? .none
    guard authorizationType  == .bearer else {
            
        done(.success(try! endpoint.urlRequest()))
        return
    }
    
    // There is no state to be updated. Proceed with the request
    guard authManager.state != nil else {
        done(.success(try! endpoint.urlRequest()))
        return
    }
    
    /// State is available and the request uses the auth token. Attempt to refresh before performing request
    authManager.performAction(freshTokens: { (accessToken, idToken, error) in
        guard error == nil else {
            log.error("Failed to refresh token with Error: \(error!.localizedDescription)")
            done(.failure(MoyaError.underlying(error!, nil)))
            return
        }
        
        done(.success(try! endpoint.urlRequest()))
    })
}
```

sunshinejr:
Hey @DMCApps. I'm afraid it's not possible currently. You might want to see https://github.com/Moya/Moya/issues/1671, which is I believe a similar use-case that you specified.

DMCApps:
@sunshinejr aw that's unfortunate. Is there any reason against passing in the target to that closure? e.g. 

This:
```swift
public typealias RequestClosure = (Endpoint, @escaping RequestResultClosure) -> Void
```
becomes:
```swift
public typealias RequestClosure = (Target, Endpoint, @escaping RequestResultClosure) -> Void
```

[...]

Issue #1671, opened 2018-05-21
Sign specific endpoints with Authorization header in `requestClosure`


oluckyman:
I use Firebase and `getIDToken` method is asynchronous so I cannot use `AccessTokenPlugin` with `tokenClosure`.
Instead I implemented a request closure: 

```swift
let requestClosure = { (endpoint: Endpoint, done: @escaping MoyaProvider.RequestResultClosure) in
    do {
        var request = try endpoint.urlRequest()
        // TODO: sign only specific endpoints and leave other unsigned (how?)
        if shouldSignEndpoint {
            Auth.auth().currentUser?.getIDToken { token, error in
                if let token = token {
                    request.setValue("Token \(token)", forHTTPHeaderField: "Authorization")
                    done(.success(request))
                }
            }
        } else {
            done(.success(request))
        }
    } catch {
        done(.failure(MoyaError.requestMapping(error.localizedDescription)))
    }
}
```

The problem is that some endpoints are public, so I do not need to add 'Authorization' header. But in the `requestClosure` I see no way to decide which endpoints are public and which are not.

What is the best way to specify which endpoints are needed to sign and which are not?

SD10:
@oluckyman You probably want to create a custom `PluginType` for this. You can’t really do this using the `requestClosure` because the underlying `TargetType` has been abstracted away at this point.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

pietrocaselani:
Hi @oluckyman 

Maybe this will help you.

Instead of setting the Authorization header inside the `requestClosure`, you could just save the token on a property for example, and then use the `AccessTokenPlugin` to set the header.

Take a look at this file, where I request and save the token, using the `requestClosure` https://github.com/pietrocaselani/Trakt-Swift/blob/master/Trakt/TraktTokenInterceptor.swift

And here I use the `AceessTokenPlugin` to set the header  https://github.com/pietrocaselani/Trakt-Swift/blob/master/Trakt/Trakt.swift

Let me know if this solve your problem or if you need any further help.

oluckyman:
Hi @pietrocaselani I'm pretty new in iOS development, so your code is too complex for me. As I understood, you're getting token from response and save it in a property.
In my case token is requested from Firebase Auth module asynchronously . And to be sure that the token is valid I have to retrieve it right before request.

SD10:
[...]

Issue #1706, opened 2018-08-01
Has anyone thought about an AsyncPluginType for Moya?


SD10:
```Swift
protocol AsyncPluginType {
    /// Called to modify a request before sending.
    func prepare(_ 
        target: TargetType, 
        callback: @escaping (URLRequest) -> Void
    )

    /// Called to modify a result before completion.
    func process(_ 
        result: Result<Moya.Response, MoyaError>, 
        target: TargetType, 
        callback: @escaping (Result<Moya.Response, MoyaError>) -> Void
    )
}
```

Just throwing an issue out there, I'm not even sure this is compatible with the Moya internals. I'd need to take a deep dive again.

pietrocaselani:
@SD10 I think this would be a great addition to Moya!
I already can see `RefreshTokenPlugin` working with this!

And looking at Moya internals, more specific, the method `MoyaProvider.requestNormal`, I think would be possible to add this without too much effort.

SD10:
Thanks for the input @pietrocaselani, I think there could be huge benefits to something like this as well. 

In my app I'm faced with a situation where I need to modify a request's parameters asynchronously but not through a request per say. Rather, I just need to access an optional value on some other object that could or could not be `nil`. So chaining callbacks, promises, or Rx are all useless to me

Right now I'm writing a wrapper around my `MoyaProvider` to wait for this value to not be `nil`.
TL;DR this doesn't scale well and creates some tight coupling between my provider and this resource.

I do agree this would work awesome for a `RefreshTokenPlugin` as well.

pedrovereza:
The idea seems interesting, but I think it'd be helpful to have a more detailed usage example.

> In my app I'm faced with a situation where I need to modify a request's parameters asynchronously but not through a request per say. Rather, I just need to access an optional value on some other object that could or could not be nil. So chaining callbacks, promises, or Rx are all useless to me

@SD10 I'm not sure I understood your scenario, can you clarify a bit more? Specially why callbacks and promises are useless and how the `AsyncPluginType` would help you 🤔 

@pietrocaselani @SD10 since both of you mentioned the `RefreshTokenPlugin`, any chance you could share how you think the plugin would look like using `AsyncPluginType`?

SD10:
Hey @pedrovereza, as I was explaining it I was realizing it was a slippery slope and difficult for people to understand my situation.

[...]

Issue #744, opened 2016-10-27
Creating a PluginType to handle request reauthentication 


raphaelcruzeiro:
I'm using Moya to talk to an API that may return HTTP 401 if a requests needs to be authenticated. Unfortunately, there's no way to know at the client if the session has expired or not, and the only way to know that is to make a request to an endpoint which requires authentication and checking the status code.

What I want to do is to make all requests to the authenticated endpoints automatically check the status code of the response, reauthenticate if needed and retry the request with the refreshed session. This way, my client doesn't need to know the specifics of the API and can just make simple requests without worrying about authentication. From what I understand, creating a `PluginType` would be the way to go. However, I'm not sure on how to replace the whole request on `didReceiveResponse`.

Is there a way to achieve this with the `PluginType` or am I completely on the wrong track here?

ashfurrow:
Plugins probably aren't the best way to approach this, I'd suggest subclassing either the RxMoyaProvider or ReactiveMoyaProvider and use RxSwift or ReactiveCocoa to `retry` on failures. There is no straightforward approach, I'm afraid. Even if you don't use RxSwift or ReactiveCocoa, I would still recommend subclassing and putting the necessary logic in the `request` function.

raphaelcruzeiro:
I'm trying a naive implementation for this:

```
public typealias AuthenticationBlock = (_ done: () -> Void) -> Void

enum Error: Swift.Error {
    case missingAuthenticationBlock
    case invalidCredentials
}

class RxAuthenticatedMoyaProvider<Target>: RxMoyaProvider<Target> where Target: Moya.TargetType {

    private let disposeBag = DisposeBag()

    public var authenticationBlock: AuthenticationBlock?

    override func request(_ token: Target) -> Observable<Response> {
        return _request(token)
    }

    private func _request(_ token: Target, isSecondTryAfterAuth: Bool = false) -> Observable<Response> {
        return super.request(token)
            .flatMap { [unowned self] response -> Observable<Response> in
                if response.statusCode == 401 { // We need to authenticate

                    if isSecondTryAfterAuth { // Server is still asking for authentication. Give up
                        return Observable<Response>.error(Error.invalidCredentials)
                    }

                    guard let authenticationBlock = self.authenticationBlock else {
                        throw Error.missingAuthenticationBlock
                    }

                    return Observable.create { observer in
                        authenticationBlock {
                            self._request(token, isSecondTryAfterAuth: true)
                                .subscribe { event in
[...]

PR #1935, opened 2019-10-08
RequestClosure include Target

DMCApps:
:sparkles: Enhancement driven by #1671 and #1932 

## Driving force

This pull request is to add the `Target` to the `RequestClosure`. There have been 2 tickets opened (#1671 and #1932) pertaining to this ability. The suggested ways to fix are complex or unnecessary when the information is easily available through the `requestNormal` method.

One use case for adding the closure to the `Target` is when an authorization token needs to be renewed. It is nice to have the ability to see the type of request being made and only refreshing the token on those specific requests.

## Sample Usage

```swift
// Closure run on each request in order to check the validity of the token
// The token will be refreshed whenever needed.
let requestClosure: MoyaProvider.RequestClosure = { (target: Target, endpoint: Endpoint, done: @escaping MoyaProvider.RequestResultClosure) in
    
    // Only proceed if the task is AccessTokenAuthorizable and it's authorization type is bearer.
    let authorizationType: AuthorizationType = (target as? AccessTokenAuthorizable)?.authorizationType ?? .none
    guard authorizationType == .bearer else {
        done(.success(try! endpoint.urlRequest()))
        return
    }
    
    /// State is available and the request uses the auth token. Attempt to refresh before performing request
    authManager.performAction(freshTokens: { (accessToken, idToken, error) in
        guard error == nil else {
            log.error("Failed to refresh token with Error: \(error!.localizedDescription)")
            done(.failure(MoyaError.underlying(error!, nil)))
            return
        }
        
        done(.success(try! endpoint.urlRequest()))
    })
}
```

## Migration Notes:

The RequestClosure has changed from:
```swift
public typealias RequestClosure = (Endpoint, @escaping RequestResultClosure) -> Void
```
to:
```swift
public typealias RequestClosure = (Target, Endpoint, @escaping RequestResultClosure) -> Void
```

## Unit tests

Added a unit test to `MoyaProviderSpec`. Not sure if it's redundant or necessary.


Tests/MoyaProviderSpec.swift
```diff
@@ -388,6 +390,13 @@ final class MoyaProviderSpec: QuickSpec {
 
                 expect(executed).to(beTruthy())
             }
+            
+            it("has the proper internal target") {
+                let target: GitHub = .zen
+                provider.request(target) { _ in  }
+
```
MoyaBot: - :warning: Lines should not have trailing whitespace. (`trailing_whitespace`)

sunshinejr:
@DMCApps thanks for the PR! Would you be able to update the PR to use the `underlyingTarget` we discussed in #1932? (and also resolve the conflicts?)

DMCApps:
I should have some time over the weekend to get to this.

DMCApps:
@sunshinejr I was looking at implementing this tonight but I feel I may have ran into a snag. I know that `Endpoint` had it's generic requirement removed in previous tickets, but I feel like this implementation may require the Generic typing to be required again. 

I'm attempting to write a unit test to ensure the target type is there and set properly however the tests do not like the comparison between the protocol type and a concrete type.

```
it("has the proper underlyingTarget") {
    let endpoint = self.simpleGitHubEndpoint
    // Error here: Binary operator '==' cannot be applied to operands of type 'Expectation<TargetType>' and 'GitHub'
    expect(endpoint.underlyingTarget) == GitHub.zen
}
```

If I add in `Endpoint<Target: TargetType>` and insert the proper `Target` generic throughout the library it seems to work.

Thoughts?

sunshinejr:
Hey @DMCApps, just got to check the PR and I agree it's a tough one... I don't feel like we should go back to the generic `Endpoint ` as the property is not too big to justify it at least in my opinion. 

I'm currently not sure what we should do next, maybe someone from @Moya/core-team would like to chime in?

andre991:
@sunshinejr and @DMCApps, do you have any plans to make this feature available?

SD10:
This has gotten a bit stale. I'd be curious how many people in the Moya community are looking for this feature as I haven't seen a large number of reported issues around it. 

I do think this fixes some issues people run into with the `endpointClosure` but I have two main concerns:

1. Making `Endpoint` generic again after removing it in Moya 11.0.0

2. Forcing all `Endpoint` types to have an `underlyingTarget: Target`

DMCApps:
I'm still interested in the feature being added. I stopped working on it as I was waiting on guidance of how it would be best implemented after my attempts were discussed and thought to not be the best.

pietrocaselani:
Hello people!

I was looking at the reasons for this and I found this

From #1671

> "I use Firebase and getIDToken method is asynchronous so I cannot use AccessTokenPlugin with tokenClosure. Instead I implemented a request closure:"

From #1932

> "I'm attempting to add a middleware step to my requests that attempts to refresh my tokens..."

And I already found myself facing the same issue for refreshing tokens and intercepting requests on #744 and more.

The way I see things is that we already have a good API that can act as middleware, the `PluginType`. The problem is `PluginType` is a sync API and for the examples above, refreshing tokens require some async work, so we keep relying on the `requestClosure` API because it's async.

So I will ask again @SD10, [Has anyone thought about an AsyncPluginType for Moya?](https://github.com/Moya/Moya/issues/1706)😝 

I share the same concerns about making `Endpoint` generic again.

Maybe what we really need is `AsyncPluginType`.

@DMCApps What do you think? `AsyncPluginType` would help your situation?

(Side note: I had the same problem last weeks at my work trying to use `Firebase.getIDToken` in our internal network client 😞)

raphaelcruzeiro:
@pietrocaselani Not OP, but an `AsyncPluginType` would save me sooo much pain.

PR #1931, opened 2019-10-04
Update development to Swift 5.1

BasThomas:
Hi!

Wanted to take another look at Moya and thought I'd start with something of a simple cleanup.

Not sure if it is expected to upgrade to Swift 5.1 for Moya 14? As far as I know that is source compatible with 5.0 and should be OK to handle in Xcode 11 — but we should verify that, I guess!

What this pull request does:

- Updates the podspec to Swift 5.1 (to check: updating Carthage / SwiftPM)
- Omits `return`s in single-expression functions (to check: causes invalid linter warnings)
- Use non-nullable `Data` to `String` conversion

PR #1936, opened 2019-10-19
Refactoring Task

### LINKED ISSUE #1655: Add uploadData(Data) case to Task enum
The inspiration for this issue is this comment https://github.com/Moya/Moya/issues/1565#issuecomment-383371379 by @ffittschen 

### Problem
He's using `.uploadFile(URL)` to upload `Data` by first writing that `Data` to a `URL`. 

### Solution
It seems reasonable to add a `.uploadData(Data)` case to the `Task` enum. 

My main concern is the scalability of the enum as we continue to add more cases. I think this is one of the drawbacks of flattening it and not using the previous strategy of composing associated values.

amaurydavid:
Hello :)

The Task enum has become a monster with so much variations of the same tasks, especially requests, that it is no longer trivial to know which one to use. I wanted to shrink it drastically to have as few cases as possible. And add the long overdue (one year and a half) upload case with `Data` (Fixes #1655) 🎉 

Edit: This is now the second version of this PR.

Taking advantage of Swift 5.1 and the ability to have default associated values in enums (thanks @sunshinejr  for the heads up), there are now only 3 case instead of 12.
Additionally, with Alamofire 5 `ParameterEncoding` is now stated to be soon depreciated in favor of the new `ParameterEncoder` so I replaced all usages accordingly.

Some particular points I'ld like to discuss:
- Atm it's possible to perform any upload or download task by providing body or query params. I don't really know how upload or download are performed and when and where it's possible to have parameters. I suspect i'll have to tweak it a bit to work as intended. 
- ~~Having based Tasks on `BodyParams` and `QueryParams`, I've rejected the possibility for users to use an `URLEncodedFormParameterEncoder` with a `.methodDependant` destination. To make sure encoders then behave like there are supposed to, I've added runtime checks on provided encoders. Do you think I should allow users to use this destination?~~ This point have been resolved by not exposing the ParameterEncoder subclasses in the Task but only their nested encoders.

Feel free to enlight me whatever the subject.


Tests/MoyaTests/TaskSpec.swift
```diff
@@ -0,0 +1,120 @@
...
+            let task = context()["task"] as! Task
+            let expectedDestination = context()["destination"] as! URLEncodedFormParameterEncoder.Destination
+            let encodable = context()["encodable"] as! [String: String]
+
+            it("uses the correct encoder and encodable") {
+                let entry = task.params!.first { $0.0 is URLEncodedFormParameterEncoder }
+                expect(entry).toNot(beNil())
+                let encoder = entry!.0 as! URLEncodedFormParameterEncoder
+                expect(encoder.destination).to(equal(expectedDestination))
+                expect(entry!.1 as! [String: String]).to(equal(encodable))
```
MoyaBot: - :warning: treating a forced downcast to '[String : String]' as optional will never produce 'nil'
- :warning: treating a forced downcast to '[String : String]' as optional will never produce 'nil'
- :warning: treating a forced downcast to '[String : String]' as optional will never produce 'nil'

amaurydavid:
Some questions regarding tests:
- Looking at the Endpoint file, i'm wondering about the purpose of the `adding(newHTTPHeaderFields:)`, `replacing(task:)` and `add(httpHeaderFields:)` functions. They only seems to be used by the tests, and. Maybe we could get rid of them? If it's really necessary to be able to change the task or headers of a Endpoint we may just make the proprties var instead of let ?

- Until now, parameters encoding was performed by an homemade extension of URLRequest. The correct encoding and headers update was then checked in EndpointSpec. With the new Alamofire, this extension is no longer necessary because AF's encoders now perform these tasks. Is it still pertinent to check for both params presence & headers update in our EndpointSpec ?

Sources/Moya/Task.swift
```diff
@@ -3,39 +3,91 @@ import Foundation
...
 public enum Task {
 
-    /// A request with no additional data.
-    case requestPlain
+    public typealias TaskParameters = [(ParameterEncoder, Encodable)]
 
-    /// A requests body set with data.
-    case requestData(Data)
+    /// A task to request some data
+    case request(rawBody: Data?, params: TaskParameters?)
```
sunshinejr: hmm not sure if I get this - you can specify both `rawBody` (which will be a `httpBody`) and also add params which can also append to body? 

also keep in mind that `URLEncoding` or the new encoder can also append to body

sunshinejr:
Hey @amaurydavid, thanks a lot for the hard work on this one! I didn't have time to fully get into it, just got a quick look at the changes but it's not an easy refactoring for sure 😅

Also wanted to let you know that AFAIK in Swift 5.1 there is an ability to add default values for cases and we were thinking of improving the `Task` to also benefit from that (we could make a requirement for Swift 5.1 in Moya 15 as we are for sure gonna enforce Xcode 11+ anyways). E.g.:
```diff
enum Task {
-    /// A request body set with `Encodable` type
-   case requestJSONEncodable(Encodable)
-
-   /// A request body set with `Encodable` type and custom encoder
-    case requestCustomJSONEncodable(Encodable, encoder: JSONEncoder)
+    case request(encodable: Encodable, encoder: JSONEncoder = JSONEncoder())
}
```

---

> Looking at the Endpoint file, i'm wondering about the purpose of the adding(newHTTPHeaderFields:), replacing(task:) and add(httpHeaderFields:) functions. They only seems to be used by the tests. Maybe we could get rid of them? If it's really necessary for the user to be able to change the task or headers of a Endpoint we may just make the properties var instead of let ?

yes we use it in tests to make sure these work - we don't use it anywhere in the core as it's a pure utility for Moya users 

> Until now, parameters encoding was performed by an homemade extension of URLRequest. The correct encoding and headers update was then checked in EndpointSpec. With the new Alamofire, this extension is no longer necessary because AF's encoders now perform these tasks. Is it still pertinent to check for both params presence & headers update in our EndpointSpec ?

if it's now in Alamofire then great - the more code we can remove from the core and leave in Alamofire the better 😄 in that case we won't really need these tests to be there, but I'd love to leave them for at least one version to make sure they didn't break our assumptions about the params/headers

also I think important note that could impact your refactor is that this PR seems like a really big change and we would love to keep the API changes to bare minimum - we want our users to keep up with version bumps and if they have too much things to update they will just stay at the previous ones as long as possible. With that in mind use your best judgement for the changes (e.g. the `adding(newHTTPHeaderFields:)` method updates).

amaurydavid:
> Also wanted to let you know that AFAIK in Swift 5.1 there is an ability to add default values for cases and we were thinking of improving the `Task` to also benefit from that (we could make a requirement for Swift 5.1 in Moya 15 as we are for sure gonna enforce Xcode 11+ anyways)

Damn, I didn't catch this one! I've indeed read that you wanted to wait for defaults in enums but I've missed it would be possible with Siwft 5.1. Its strange, I just checked in playground with Xcode 11.1 and it works. Still, I don't see any mention of this from Swift 5.1 changes. Is it linked to https://github.com/apple/swift-evolution/blob/master/proposals/0242-default-values-memberwise.md ?
Now that you pointed it I'll ... refacto my refacto to use defaults, which should make things a bit simpler.

About the potential updates of mutating functions in Endpoint I understand your point of view about not changing too much API at once, but isn't it the point of updating to Moya 15? Shall we make properties var instead of let and mark those functions as deprecated? I think it would be a good compromise.

sunshinejr:
@amaurydavid yeah I’m not saying we don’t want breaking changes at all, just to try and make the migration as easy as possible. Also if we want to make big changes then I’m not opposed to that as well, but we would have to decide if the output of the change  is worth having people spend hours on migrating to newest version in bigger apps. It’s hard to do that though so we have to use our best judgement.

Sources/Moya/Task.swift
amaurydavid: Yes that was a flaw of this design. The idea here was that if some other parameter was to also be added as a body, it would override the rawData.
But in the 2nd version this behavior is now prevented by using only 1 kind of body param.

Sources/ReactiveMoya/MoyaProvider+Reactive.swift
```diff
@@ -30,7 +30,7 @@ public extension Reactive where Base: MoyaProviderType {
     /// Designated request-making method with progress.
     func requestWithProgress(_ token: Base.Target, callbackQueue: DispatchQueue? = nil) -> SignalProducer<ProgressResponse, MoyaError> {
         let progressBlock: (Signal<ProgressResponse, MoyaError>.Observer) -> (ProgressResponse) -> Void = { observer in
-            return { progress in
+            { progress in
```
MoyaBot: - :warning: Opening braces should be preceded by a single space and on the same line as the declaration. (`opening_brace`)

Tests/MoyaTests/TaskSpec.swift
```diff
@@ -0,0 +1,115 @@
...
+                let encoder = PropertyListEncoder()
+                let task = Task.request(bodyParams: .custom(encodable, encoder))
+
+                let allParameters = try? task.allParameters()
+                expect(allParameters).toNot(beNil())
+                let entry = allParameters!.first { $0.1 is PropertyListEncoder }
+                expect(entry).toNot(beNil())
+                expect(entry!.1).to(beAKindOf(PropertyListEncoder.self))
+                expect(entry!.1 as! PropertyListEncoder).to(equal(encoder))
+                expect(entry!.0 as! [String: String]).to(equal(encodable))
```
MoyaBot: - :warning: treating a forced downcast to '[String : String]' as optional will never produce 'nil'
- :warning: treating a forced downcast to '[String : String]' as optional will never produce 'nil'
- :warning: treating a forced downcast to '[String : String]' as optional will never produce 'nil'

Sources/Moya/Task+Internal.swift
```diff
@@ -0,0 +1,74 @@
+import Foundation
+import Moya
```
MoyaBot: - :warning: file 'Task+Internal.swift' is part of module 'Moya'; ignoring import
- :warning: file 'Task+Internal.swift' is part of module 'Moya'; ignoring import
- :warning: file 'Task+Internal.swift' is part of module 'Moya'; ignoring import

Sources/Moya/MoyaProvider.swift
```diff
@@ -171,24 +171,40 @@ public enum StubBehavior {
...
     /// Return a response immediately.
     final class func immediatelyStub(_: Target) -> Moya.StubBehavior {
-        return .immediate
+        .immediate
     }
 
     /// Return a response after a delay.
     final class func delayedStub(_ seconds: TimeInterval) -> (Target) -> Moya.StubBehavior {
-        return { _ in return .delayed(seconds: seconds) }
+        { _ in .delayed(seconds: seconds) }
```
MoyaBot: - :warning: Opening braces should be preceded by a single space and on the same line as the declaration. (`opening_brace`)

Sources/Moya/RawDataParameterEncoder.swift
```diff
@@ -0,0 +1,21 @@
+import Foundation
+import Moya
```
MoyaBot: - :warning: file 'RawDataParameterEncoder.swift' is part of module 'Moya'; ignoring import
- :warning: file 'RawDataParameterEncoder.swift' is part of module 'Moya'; ignoring import
- :warning: file 'RawDataParameterEncoder.swift' is part of module 'Moya'; ignoring import

amaurydavid:
I think I'm done here. I'm satisfied with the actual `Task`, and spent (a lot of time) on tests though I'm sure they might be improved.

- There are warnings in `Task+internal` and `RawDataParameterEncoder` because they are part of the Moya target and the test target as I need them in tests, but don't want to make them public. Is there a better way?

- Similarly I wanted in `Moya+Alamofire` to make the following typealiases internal
```
public typealias URLEncodedFormParameterEncoder = Alamofire.URLEncodedFormParameterEncoder
public typealias JSONParameterEncoder = Alamofire.JSONParameterEncoder
```
but they are used in tests so it was either having them public or using `Alamofire.URLEncodedFormParameterEncoder` and `Alamofire.JSONParameterEncoder`.

- I've let reflexions on https://github.com/Moya/Moya/pull/1936#issuecomment-546611729 aside as it's not the focus of this PR.

- I'll update doc and guides once the PR has been reviewed and validated.

Sources/Moya/RawDataParameterEncoder.swift
```diff
@@ -0,0 +1,21 @@
...
+
+/// An "encoder" that expects the given parameters to be of type `Data` and just sets the request's httpBody with it, without any additional encoding.
+struct RawDataParameterEncoder: ParameterEncoder {
+    func encode<Parameters>(_ parameters: Parameters?, into request: URLRequest) throws -> URLRequest where Parameters: Encodable {
+        guard let parameters = parameters else { return request }
+
+        var request = request
```
LucianoPAlmeida: This is more of a question,  if we are assign the request to a mutating, modifying and then returning... could be a valid approach reveice this request here as `inout` and modify it directly?

amaurydavid: It would indeed be nice. But if we add the inout keyword in the function definition it is no longer the one from `ParameterEncoder`, raising the error `Type 'RawDataParameterEncoder' does not conform to protocol 'ParameterEncoder'`

Sources/Moya/Task.swift
```diff
@@ -3,39 +3,48 @@ import Foundation
...
+
+        /// The given encodable will be encoded in the request's body using the provided encoder.
+        ///
+        /// The provided encoder must not be a `URLEncodedFormParameterEncoder` (use `BodyParams.urlEncoded`instead)
+        /// or `JSONParameterEncoder`(use `BodyParams.json` instead). If this is the case, a `MoyaError.encodableMapping` will be raised.
+        case custom(Encodable, ParameterEncoder)
+    }
+
+    /// All different ways to set parameters in an HTTP request's query.
+    public enum QueryParams {
```
sunshinejr:
```suggestion
    public enum URLParams {
```

Sources/Moya/Task.swift
```diff
@@ -3,39 +3,48 @@ import Foundation
...
+        case json(Encodable, JSONEncoder = JSONEncoder())
+
+        /// The given encodable will be url encoded in the request's body.
+        case urlEncoded(Encodable, URLEncodedFormEncoder = URLEncodedFormEncoder())
+
+        /// The given encodable will be encoded in the request's body using the provided encoder.
+        ///
+        /// The provided encoder must not be a `URLEncodedFormParameterEncoder` (use `BodyParams.urlEncoded`instead)
+        /// or `JSONParameterEncoder`(use `BodyParams.json` instead). If this is the case, a `MoyaError.encodableMapping` will be raised.
+        case custom(Encodable, ParameterEncoder)
```
sunshinejr: interesting - so the `ParameterEncoder` is only accepting `Encodable` type as an input, no more parameters as `[String: Any]`? in that case imo this is too big of a change, and the `ParameterEncoding` is not deprecated in 5.0.0, so I would keep using `ParameterEncoding` for now as personally I wouldn't have the energy to create a new `Encodable` type for each endpoint as this might be get tedious pretty fast

Sources/Moya/Task.swift
```diff
@@ -3,39 +3,48 @@ import Foundation
...
+    }
+
+    /// All different ways to set parameters in an HTTP request's query.
+    public enum QueryParams {
+        /// The given encodable will be url encoded in the request's query.
+        case query(Encodable, URLEncodedFormEncoder = URLEncodedFormEncoder())
```
sunshinejr:
```suggestion
        case urlEncoded(Encodable, URLEncodedFormEncoder = URLEncodedFormEncoder())
        case custom(Encodable, ParameterEncoder)
```

Sources/Moya/Task.swift
```diff
@@ -3,39 +3,48 @@ import Foundation
...
+
+        /// Upload the file at the given url
+        case file(URL)
+
+        /// Upload some multipart content.
+        case multipart([MultipartFormData])
+    }
+
+    /// A task to request some data
+    case request(bodyParams: BodyParams? = nil, queryParams: QueryParams? = nil)
```
sunshinejr: 👌 really like it how much it simplifies both the source code and call site

Sources/Moya/Task.swift
```diff
@@ -3,39 +3,48 @@ import Foundation
...
+
+    /// A task to upload some data
+    case upload(source: UploadSource, bodyParams: BodyParams? = nil, queryParams: QueryParams? = nil)
```
sunshinejr: should body parameters be possible to add for an upload? if we want to add parameters to the upload we are doing, we would use multipart/form-data (in that case `.multipart` as a source)

sunshinejr:
I encourage all @Moya/contributors to at least get a quick look at the changes in this PR - this is a pretty big one and I would love some opinions on that 🙏

Sources/Moya/Task.swift
amaurydavid: That is a good question ^^ I don't have the answer as I never had to upload something.

Sources/Moya/Task.swift
amaurydavid: Dictionaries are automatically encodable if their keys and values are. So ["Hello": "moya"] is Encodable and can be used as parameter, you don't have to create a custom Encodable object.
This is actually was I use in tests.
JSONEncoding needs a valid JSON object, which in fact are all encodables.
URLEncoding uses the String description of values that are not NSNumber, Array of Dictionary.
So I think in a large majority of use cases no changes will be needed, and if so they should be small.

Sources/Moya/Task.swift
amaurydavid: Do you have a use case for the .custom value? I struggle to find one way to add parameters to an url without it being with the query. Or is it just for the purpose of providing a fully customizable way, "just in case of" ?

Sources/Moya/Task.swift
sunshinejr: ah you're probably correct - in that case nevermind 👍

Sources/Moya/Task.swift
sunshinejr: yeah I just try to leave a room for things we didn't think about at the moment but users will probably find a use case to use their own parameter encoder

amaurydavid: I've just commited some changes regarding custom params.
The issue with your proposition is that it would result in a `custom` value in both `BodyParams` and `Queryparams`, while they would work exactly the same.
So I've created a `CustomParams` struct we can now provide as associated value of the `Task` value and removed the `BodyParams.custom` value.
At the same time, `QueryParams` is now a struct, which is less weird than an enum with only one case.

Sources/Moya/Task.swift
sunshinejr: I was thinking more about removing the body parameters as `MultipartFormData` already have an option to provide properties alongside data upload.

LucianoPAlmeida:
There are few unrelated commits(combine related) on this PR I think is because this branch was created from another other than the target(I think master). So maybe we can merge master/develop into 5.1 ?
cc @sunshinejr

sunshinejr:
@LucianoPAlmeida good catch! this is because this PR targets a different branch - `development_swift_5.1` which is not synced with `development` yet (see #1931)

LucianoPAlmeida:
Hummm I didn't notice that PR, there are a few conflicts so we can't sync it right now...
Perhaps we can target this to develop for now, so we can have a sync diff which is cleaner for us @Moya/contributors to take a look. What do you think @sunshinejr ?

Tests/MoyaTests/MultiTargetSpec.swift
```diff
@@ -32,28 +32,43 @@ final class MultiTargetSpec: QuickSpec {
...
                 expect(target.method) == Method.get
             }
 
             it("uses correct task") {
-                expect(String(describing: target.task)).to(beginWith("requestParameters")) // Hack to avoid implementing Equatable for Task
+                guard case Task.request = target.task else {
+                    fail("expected task type `.request`, was \(String(describing: target.task))")
+                    return
+                }
+                expect(true) == true
```
LucianoPAlmeida: Is this necessary?

Tests/MoyaTests/TestHelpers.swift
```diff
@@ -100,16 +100,22 @@ enum HTTPBin: TargetType, AccessTokenAuthorizable {
 
     var task: Task {
         switch self {
-        case .basicAuth, .post, .bearer:
-            return .requestParameters(parameters: [:], encoding: URLEncoding.default)
+        case .basicAuth, .bearer:
+            return .request(queryParams: .init(["": ""]))
```
LucianoPAlmeida: If we make QueryParam conform to `ExpressibleByDictionaryLiteral` we can simplify this to `.request(queryParams: ["": ""])`

Sources/Moya/MoyaError.swift
```diff
@@ -26,39 +26,42 @@ public enum MoyaError: Swift.Error {
...
-        case .requestMapping: return nil
-        case .parameterEncoding: return nil
+        case let .imageMapping(response),
+             let .jsonMapping(response),
+             let .stringMapping(response),
+             let .objectMapping(_, response),
+             let .statusCode(response):
+            return response
+
+        case let .underlying (_, response):
```
LucianoPAlmeida: This could be together with the cases above?

Tests/MoyaTests/MultiTargetSpec.swift
amaurydavid: I'm not familiar with tests yet so I wanted to make sure the test passed. Will it pass by default if no `expect` statement is used?

Sources/Moya/MoyaError.swift
amaurydavid: underlying's Error is a `Swift.Error` while users are `MoyaError` so the compiler raise a type mismatch.

Tests/MoyaTests/TestHelpers.swift
amaurydavid: That is a great idea! I'll update it.

amaurydavid:
> Hummm I didn't notice that PR, there are a few conflicts so we can't sync it right now...
> Perhaps we can target this to develop for now, so we can have a sync diff which is cleaner for us @Moya/contributors to take a look. What do you think @sunshinejr ?

Initially the PR was to be merged on `develop` but when I started using Swift 5.1 features I also included commits for the `development-swift-5.1 branch`, resulting in this PR to contains both last commits from `develop` and commits from `development-swift-5.1 branch` (with conflicts resolved). Updating #1931 to sync it with dev should eliminate those files related to Combine from the diff.

Tests/MoyaTests/TestHelpers.swift
amaurydavid: I'm struggling with the fact that to be encodable, the dict should have its keys `Hashable & Encodable`. I can't express it with `public typealias Key = Encodable & Hashable` because it raises the error `Protocol 'Hashable' can only be used as a generic constraint because it has Self or associated type requirements` when implementing the init.
The only way I see is to have the new init being like `public init<Key: Encodable & Hashable, Value: Encodable>(dictionaryLiteral elements: (Key, Value)...)` but it's not the same definition from `ExpressibleByDictionaryLiteral`.

Sources/Moya/Task.swift
amaurydavid: Ok I didn't understand at first ;)
Task.upload no longer has a BodyParams.

Sources/Moya/Task.swift
amaurydavid: I've renamed it ;)

Tests/MoyaTests/TestHelpers.swift
LucianoPAlmeida: Ohh, I thought it was [String: String] ... Sad :(
But being a constrained generic type I'm not sure you can express that in swift with type alias. 
The only other way you could express that would be if the class itself was generic ... 
```swift
class QueryParam<Key, Value>: ExpressibleByDictionaryLiteral where Key: Codable & Hashable, Value: Codable {
  var d: [Key: Value] = [:]
  required convenience init(dictionaryLiteral elements: (Key, Value)...) {
    self.init()
    self.d = Dictionary(uniqueKeysWithValues: elements)
  }
}
``` 
But that will be harder to the user I think. 

One question tho is if it makes sense for this query param to be generic?
because I'm really not sure a use case where someone would use this 
with another thing other than `[String: String]`...

Tests/MoyaTests/MultiTargetSpec.swift
LucianoPAlmeida: Yeah ... and since the `fail` is the actual `assertion` that is important on this test(we are failing if the task is what we don't expect) ... we could remove that I think

Tests/MoyaTests/TestHelpers.swift
amaurydavid: We coud restraint the key to be a `String` and let the Value be an `Encodable` as it does seems to be an issue with `ExpressibleByDictionaryLiteral`. But I fear it closes some possibilities just for the sake of being `ExpressibleByDictionaryLiteral`.
About habing de Generic `QueryParam`, it would allow `ExpressibleByDictionaryLiteral` but then if we provide pairs which type can not be inferred it will quickly be cumbersome to had the generic type to the QueryParam. 
Imo either we find a way to implement `ExpressibleByDictionaryLiteral` without generics or we don't.

amaurydavid:
I have rebased the PR to remove all commits that were included from the `develop` branch.

Sources/Moya/Endpoint.swift
```diff
@@ -84,37 +83,15 @@ public extension Endpoint {
...
-            return try request.encoded(parameters: urlParameters, parameterEncoding: parameterEncoding)
-        case let .requestCompositeParameters(bodyParameters: bodyParameters, bodyEncoding: bodyParameterEncoding, urlParameters: urlParameters):
-            if let bodyParameterEncoding = bodyParameterEncoding as? URLEncoding, bodyParameterEncoding.destination != .httpBody {
-                fatalError("Only URLEncoding that `bodyEncoding` accepts is URLEncoding.httpBody. Others like `default`, `queryString` or `methodDependent` are prohibited - if you want to use them, add your parameters to `urlParameters` instead.")
-            }
-            let bodyfulRequest = try request.encoded(parameters: bodyParameters, parameterEncoding: bodyParameterEncoding)
-            let urlEncoding = URLEncoding(destination: .queryString)
-            return try bodyfulRequest.encoded(parameters: urlParameters, parameterEncoding: urlEncoding)
+        //Encode params
+        try task.allParameters().forEach { tuple in
```
jdisho: This seems like a good case to use the `reduce` function. Will this work?

```swift 
return task.allParameters().reduce(request) { initialRequest, parameters -> URLRequest in
    let encodable = parameters.0
    let encoder = parameters.1
    return try encoder.encode(AnyEncodable(encodable), into: initialRequest)
}
```

Sources/Moya/MoyaError.swift
```diff
@@ -76,16 +79,14 @@ extension MoyaError: LocalizedError {
             return "Failed to map data to a String."
         case .objectMapping:
             return "Failed to map data to a Decodable object."
-        case .encodableMapping:
-            return "Failed to encode Encodable object into data."
+        case let .encodableMapping(message):
```
jdisho: Just for the sake of consistency
``` swift 
case .encodableMapping(let message):
```

Sources/Moya/MoyaProvider+Internal.swift
```diff
@@ -63,15 +63,15 @@ public extension MoyaProvider {
             }
 
             let networkCompletion: Moya.Completion = { result in
-              if self.trackInflights {
-                self.inflightRequests[endpoint]?.forEach { $0(result) }
+                if self.trackInflights {
```
jdisho: Why are these spaces here? Am I missing something?

Sources/Moya/MoyaProvider+Internal.swift
```diff
@@ -83,24 +83,67 @@ public extension MoyaProvider {
...
-        switch stubBehavior {
-        case .never:
-            switch endpoint.task {
-            case .requestPlain, .requestData, .requestJSONEncodable, .requestCustomJSONEncodable, .requestParameters, .requestCompositeData, .requestCompositeParameters:
-                return self.sendRequest(target, request: request, callbackQueue: callbackQueue, progress: progress, completion: completion)
-            case .uploadFile(let file):
-                return self.sendUploadFile(target, request: request, callbackQueue: callbackQueue, file: file, progress: progress, completion: completion)
-            case .uploadMultipart(let multipartBody), .uploadCompositeMultipart(let multipartBody, _):
-                guard !multipartBody.isEmpty && endpoint.method.supportsMultipart else {
+    private func performRequest(_ target: Target,
```
jdisho: Since most of the functions does not have a new line after each parameter, I would suggest to have the same here and in the following cases. Just to be consistent with the style.

Sources/Moya/Task+Internal.swift
jdisho: Why are you importing Moya?

Sources/Moya/Task+Internal.swift
```diff
@@ -0,0 +1,86 @@
...
+
+extension Task {
+    typealias TaskParameters = (Encodable, ParameterEncoder)
+
+    func allParameters() throws -> [TaskParameters] {
+
+        var providers: [TaskParametersProvider?] = [bodyParams, urlParams]
```
jdisho: Why not? 

```swift 
var providers: [TaskParametersProvider?] = [bodyParams, urlParams, customParams]
```

You are filtering the nil values anyway.

Sources/Moya/Task+Internal.swift
```diff
@@ -0,0 +1,86 @@
+import Foundation
+import Moya
+
+extension Task {
+    typealias TaskParameters = (Encodable, ParameterEncoder)
```
jdisho: nit: Why not `Parameters` instead of `TaskParamters`?

Sources/Moya/Task.swift
```diff
@@ -3,39 +3,74 @@ import Foundation
...
-    case requestJSONEncodable(Encodable)
+        /// The given encodable will be url encoded in the request's body.
+        case urlEncoded(Encodable, URLEncodedFormEncoder = URLEncodedFormEncoder())
+    }
 
-    /// A request body set with `Encodable` type and custom encoder
-    case requestCustomJSONEncodable(Encodable, encoder: JSONEncoder)
+    /// The given encodable will be encoded in the url's query.
+    public struct URLParams {
+        public var encodable: Encodable
```
jdisho: `let` instead of` var`?

Sources/Moya/Task.swift
```diff
@@ -3,39 +3,74 @@ import Foundation
...
+    }
 
-    /// A requests body set with data, combined with url parameters.
-    case requestCompositeData(bodyData: Data, urlParameters: [String: Any])
+    /// The given encodable will be encoded according to the given custom parameter encoder
+    ///
+    /// The provided encoder must not be a `URLEncodedFormParameterEncoder` (use `BodyParams.urlEncoded` or `URLParams`instead)
+    /// or `JSONParameterEncoder`(use `BodyParams.json` instead). If this is the case, a `MoyaError.encodableMapping` will be raised.
+    public struct CustomParams {
+        public var encodable: Encodable
```
jdisho: `let` instead of `var`?

Sources/Moya/Task+Internal.swift
amaurydavid: customParams is `[TaskParametersProvider]` so I can't mix it with `TaskParametersProvider` values.

Sources/Moya/Task+Internal.swift
amaurydavid: More infos here: https://github.com/Moya/Moya/pull/1936#issuecomment-547849480

Sources/Moya/MoyaProvider+Internal.swift
amaurydavid: If functions have a lot of parameters like this I prefer breaking the style consistency to have it more readable.

Sources/Moya/Task+Internal.swift
amaurydavid: I think it was initially because Alamofire already has a `Parameters` type and I didn't want to mix them. Renaming it could indeed be a god idea.

Sources/Moya/Task+Internal.swift
jdisho: Missed that! How about this?

```swift 
var providers: [TaskParametersProvider?] = customParams.append(contentsOf: [bodyParams, urlParams])
```

Sources/Moya/Task.swift
amaurydavid: Those structs are just containers for those 2 properties. It could have been a tuple but I find it less clean with a tuple.
I think making it `var` allow a bit more flexibility when using it, even if I predict users should not have the need to mutate it.

Sources/Moya/MoyaProvider+Internal.swift
jdisho: I partially agree with you, but consistency matters too. 
In case you would like to break after each parameter, apply this style in every function that has more than **x** parameters and open a pull request.

Sources/Moya/Task.swift
jdisho: I think these shouldn't be mutable then.

Sources/Moya/Task+Internal.swift
amaurydavid: `append` mutates the receiver but doesn't return anything.

This would work but I find it less readable than the original:
```
var providers: [TaskParametersProvider?] = customParams ?? []
providers.append(contentsOf: ([bodyParams, urlParams] as [TaskParametersProvider?]))
```

Sources/Moya/Task+Internal.swift
jdisho: Right, but this `Parameters` typealias  **will not** be globally defined.

Sources/Moya/Task+Internal.swift
jdisho: It would have been nice if Swift supported something like `appending(contentsOf:)` 😄.
But, never mind then.

Sources/Moya/MoyaProvider+Internal.swift
amaurydavid: I guess its just indentation fixing from when I re-idented the file after working on it.

Sources/Moya/MoyaProvider+Internal.swift
amaurydavid: That's a fair point. I've updated it.

Sources/Moya/Task+Internal.swift
```diff
@@ -0,0 +1,91 @@
+import Foundation
+import Moya
+
+extension Task {
+    typealias Parameters = (Encodable, ParameterEncoder)
```
pedrovereza: This might be a personal preference, but I'd rather have a proper struct here instead of relying on a tuple 😬 

Thoughts on using something like this instead? 
```swift
public struct Parameters {
      public let encodable: Encodable
      public let encoder: ParameterEncoder

      public init(encodable: Encodable, encoder: ParameterEncoder) {
          self.encodable = encodable
          self.encoder = encoder
      }
  }
```

I feel my main problem with tuples is the non-obvious usage, like

```swift
return try task.allParameters().reduce(request) { initialRequest, parameters in
  let encodable = parameters.0
  let encoder = parameters.1
```

amaurydavid: That indeed could be a possibility, I find structs more elegant than tuples too.
But I've chosen to use tuples internally to avoid creating an other struct and because it's only used from `Task` to `URLRequest`. On the other side, I use structs for the public API (`URLParams` and `CustomParams`).
Using structs on both sides could lead to reusing some code, i'll see where it brings me.

amaurydavid: I'm quite pleased with this change: `URLParams` and `CustomParams` now inherit from `Task.Parameters`.

docs/Targets.md
```diff
@@ -65,47 +65,35 @@ we saw when calculating our `path` property.
...
-- `.downloadDestination(_:)` for a plain file download
-- `.downloadParameters(parameters:encoding:destination:)` for downloading with parameters sent alongside the request.
+
+With a `Task` we can provide some additional parameters that will be encoded into the request. In the `userRespositories` case, we are adding 1 parameter “sort” to the request's query string, with the value “pushed”.
+In the `branches` case, we also add 1 parameter to the query string, but with a twist: as the parameter's value is a `Bool`, we provide a custom encoder to make sure the `Bool` is converted into a literal (i.e “true” or “false”) instead of an int (0 or 1) by default.
+
+Alongside the `urlParams` associated value, you can provide a `bodyParams` value which works the same but encodes parameters in request's body instead of query string.
+You can also provide a `customParams` value if you want to encode your parameters in a way that is not possible with `bodyParams` or `urlParams`, if you want to encode some xml in the body for example.
+
+When using `Task.upload`, you have acces to 3 different sources:
```
LucianoPAlmeida:
```suggestion
When using `Task.upload`, you have access to 3 different sources: 
```

docs/MigrationGuides/migration_14_to_15.md
```diff
@@ -0,0 +1,43 @@
...
+### Determining associated values
+
+#### Body parameters
+If the old `Task` was `.requestData` or `.requestCompositeData`: the `bodyParams` associated value should be `.raw(data)`.
+
+If the old `Task` was `.requestJSONEncodable`, `.requestCompositeData`, or was using a `JSONEncoding` associated value: the new `bodyParams` associated value should be `.json(encodable)`, or `.json(encodable, encoder: <aCustomEncoder>)` if you need to provide a custom encoder.
+
+If the old `Task` was  `.requestCompositeParameters`, `.downloadParameters` or was using an `URLEncoding` associated value with a `.httpBody` destination: the new `bodyParams` associated value should be `.urlEncoded(encodable)` or `.urlEncoded(encodable, encoder: <aCustomEncoder>)` if you need to provide a custom encoder.
+
+If you are not concerned by any of the 3 points above, you don't need to provide the `bodyParams` associated value.
```
LucianoPAlmeida:
```suggestion
If you are not concerned by any of the points above, you don't need to provide the `bodyParams` associated value.
```

docs/MigrationGuides/migration_14_to_15.md
```diff
@@ -0,0 +1,43 @@
...
+
+#### URL parameters
```
LucianoPAlmeida: I'm thinking if it would be nice to add some small code examples here? 
What do you guys think?

Sources/Moya/Endpoint.swift
```diff
@@ -84,37 +83,11 @@ public extension Endpoint {
...
-            let parameterEncoding = URLEncoding(destination: .queryString)
-            return try request.encoded(parameters: urlParameters, parameterEncoding: parameterEncoding)
-        case let .requestCompositeParameters(bodyParameters: bodyParameters, bodyEncoding: bodyParameterEncoding, urlParameters: urlParameters):
-            if let bodyParameterEncoding = bodyParameterEncoding as? URLEncoding, bodyParameterEncoding.destination != .httpBody {
-                fatalError("Only URLEncoding that `bodyEncoding` accepts is URLEncoding.httpBody. Others like `default`, `queryString` or `methodDependent` are prohibited - if you want to use them, add your parameters to `urlParameters` instead.")
-            }
-            let bodyfulRequest = try request.encoded(parameters: bodyParameters, parameterEncoding: bodyParameterEncoding)
-            let urlEncoding = URLEncoding(destination: .queryString)
-            return try bodyfulRequest.encoded(parameters: urlParameters, parameterEncoding: urlEncoding)
+        //Encode params
```
LucianoPAlmeida: Is nice to see those spots on the code that shows the improvements :)

Sources/Moya/MoyaError.swift
```diff
@@ -16,7 +16,7 @@ public enum MoyaError: Swift.Error {
     case objectMapping(Swift.Error, Response)
 
     /// Indicates that Encodable couldn't be encoded into Data
-    case encodableMapping(Swift.Error)
+    case encodableMapping(String)
```
LucianoPAlmeida: Why does this changed to a `String`?

Sources/Moya/Task+Internal.swift
```diff
@@ -0,0 +1,77 @@
+import Foundation
+import Moya
+
+extension Task {
+
+    /// Returns all `TaskParametersType` objects to use when generating the `URLRequest`.
+    func allParameters() -> [TaskParametersType] {
+        var providers: [TaskParametersType?] = [bodyParams, urlParams]
```
LucianoPAlmeida: Nit: Could this be done like
```swift
var providers: [TaskParametersType?] = [bodyParams, urlParams, customParams]
return providers.compactMap { $0 }

// Or even
return  [bodyParams, urlParams, customParams].compactMap { $0 }
```
?

amaurydavid: @jdisho [had the same reaction](https://github.com/Moya/Moya/pull/1936#discussion_r345861114) ^^  `customParams` is an array while `bodyParams` and `urlParams` are not.

Sources/Moya/MoyaError.swift
amaurydavid: I changed it for a String because now it's only thrown when using a CustomParams with a wrong encoder.
But as you pointed it, i've read again the way parameters are encoded and it should definitively also still be thrown with the new way of encoding parameters (Endpoint.swift:L87). If I do so the associated value would be an Error again.
I would also need a new `MoyaError` with a `String` for missuses of `CustomParams`. Any suggestion for its name?

amaurydavid: Another option would be to have `MoyaError.encodableMapping(Swift.Error?, String?)`

LucianoPAlmeida: > Another option would be to have `MoyaError.encodableMapping(Swift.Error?, String?)`

I personally don't like this approach because it is not clear what this string means for the user. I mean it is probably better to have something like a custom moya error or even better, extend the MoyaError to have a localized description property e.g. NSError ... so that kind of message could go there :) What you think?

amaurydavid: I've pushed a new dedicated error so that `encodableMapping` is left untouched.

docs/MigrationGuides/migration_14_to_15.md
amaurydavid: I first was going for code exemples with a complete mapping of old vs new values, but there are lots of possibilities. So I came up with this. But I agree it's not "great".

amaurydavid:
I've studied some improvements I could make and here are 2 I've thought about

**1. Shrink all parameters associated values down to 1.**
It can be done by wrapping `BodyParams`, `URLParams` and `CustomParams` into an enum:
```
 public enum Task {
    enum ParametersKind {
        case body(BodyParams)
        case url(URLParams)
        case custom(CustomParams)
    }

    /// A task to request some data
    case request(parameters: [ParametersKind] = [])

    /// A task to upload some data
    case upload(source: UploadSource, parameters: [ParametersKind] = [])

    /// A task to download some data
    case download(destination: DownloadDestination, parameters: [ParametersKind] = [])
}
```

Pros:
- A more concise declaration.

Cons:
- The user would be able to provide `BodyParams` in `Task.upload`. A note in the documentation and a check at runtime can be added to notify the user via a warning log but it's better to have it at compile time. 




**2. Remove `CustomParams`**
`CustomParams` is only there to make sure it's only used as last resort when `BodyParams` and `URLParams` can't be. We could simply let the user use `Parameters` however he wants.

Pros: 
- The new `MoyaError.taskParametersUsage` is no longer needed and can be removed.
- `CustomParams` removed from the code base.

Cons:
- The user would be able to encode parameters in request's body when using `Task.upload`. Again, updating the docu and a check at runtime can be added.


What do you think about those code improvements ?

amaurydavid:
Hey guys :) Now that Moya 14 is out (thanks @sunshinejr for taking care of it), I'ld like to ping you out. I'ld be delighted to have your feedback on this PR, it's been opened for a while now.

sunshinejr:
@amaurydavid ah yes will do! let's merge the Swift 5.1 migration first and then try to get this one afterwards!

swizzlr:
# [Codecov](https://codecov.io/gh/Moya/Moya/pull/1936?src=pr&el=h1) Report
> Merging [#1936](https://codecov.io/gh/Moya/Moya/pull/1936?src=pr&el=desc) into [development-swift-5.1-updates](https://codecov.io/gh/Moya/Moya/commit/9fb97254376ae85cb24500a62c138d65027ae10b&el=desc) will **not change** coverage by `%`.
> The diff coverage is `n/a`.

[Impacted file tree graph](https://codecov.io/gh/Moya/Moya/pull/1936?src=pr&el=tree)

```diff
@@                      Coverage Diff                       @@
##           development-swift-5.1-updates    #1936   +/-   ##
==============================================================
  Coverage                          92.14%   92.14%           
==============================================================
  Files                                 29       29           
  Lines                               1018     1018           
==============================================================
  Hits                                 938      938           
  Misses                                80       80           
```



------

[Continue to review full report at Codecov](https://codecov.io/gh/Moya/Moya/pull/1936?src=pr&el=continue).
> **Legend** - [Click here to learn more](https://docs.codecov.io/docs/codecov-delta)
> `Δ = absolute <relative> (impact)`, `ø = not affected`, `? = missing data`
> Powered by [Codecov](https://codecov.io/gh/Moya/Moya/pull/1936?src=pr&el=footer). Last update [9fb9725...9fb9725](https://codecov.io/gh/Moya/Moya/pull/1936?src=pr&el=lastupdated). Read the [comment docs](https://docs.codecov.io/docs/pull-request-comments).

sunshinejr:
@amaurydavid can you rebase on top of current develop please? 🙏

amaurydavid:
I've rebased the PR on `development` and merged it in the branch. We are all set :)

Sources/Moya/Task.swift
```diff
@@ -3,39 +3,84 @@ import Foundation
...
+        ///   or `URLParams` instead) or `JSONParameterEncoder`(use `BodyParams.json` instead).
+        ///   If this is the case, a `MoyaError.encodableMapping` will be raised.
+        public required init(_ encodable: Encodable, encoder: ParameterEncoder) throws {
+            // CustomParams should only be used when `URLEncodedFormParameterEncoder` and `JSONParameterEncoder`
+            // are not enough. To enforce usage of BodyParams or URLParams, let's check the type of the given encoder
+            // to make sure CustomParams is correctly used.
+            if encoder is JSONParameterEncoder
+            || encoder is URLEncodedFormParameterEncoder {
+                throw MoyaError.taskParametersUsage(type(of: encoder))
+            }
```
sunshinejr: do we need to do this check? you could create your own encoding with different class name and still pass something like `JSONParameterEncoder` anyways.

amaurydavid: There is indeed the possibility to workaround the check is you really want to use params in a way not preferred by Moya. If so then the user explicitely says "I know what i'm doing" and should face possible strange behaviors.

If we remove this check, users would have the possibility to express their parameters either with `BodyParams` or `URLParams` and `CustomParams`, which I think is not a good API. We then would have to unify the 3 different params types into only one.
Please see https://github.com/Moya/Moya/pull/1936#issuecomment-569953087 i've studied impacts of such an update.

sunshinejr: ah very good explanation - thanks! it is indeed a hard one, was thinking about it for a while and not many choices left here... 

one thing that I had in mind was - what if we had a classic `Parameters` struct that would be suitable for `request`/`download` type and `UploadParameters` struct only for `upload` type? do you think it would eliminate most of the problems and not introduce new ones?

amaurydavid: I don't think we have problems with the actual code. I chose to help the user by providing a type safe API, at the lost of more code complexity. We can have a less complex code but the API would be less appealing to the user, and may allow some unexpected behaviors. I explored this second option in my comment above.

If your 2 structs idea simplifies the code while still being as appealing as the current solution I would gladly explore it. But right know I struggle to understand what properties would those 2 structs contain. Can you provide an example ?

sunshinejr: Yeah I didn't think about the actual code as problems, rather the check for types is something I would love to avoid if possible so that's why I'm thinking out loud. About the `Parameters`/`UploadParamaters` - I was thinking that the `Parameters` struct could allow for all properties (body/url/whatever) and `UploadParameters` would allow only for url ones?

amaurydavid: I don't see how such a distinction would suppress the current type check in CustomParams.

jdisho:
Hey folks! 👋
What is the plan about this PR, b/c it has been open ~ 7 months. Is it planned to be merged anytime soon?

Tests/MoyaTests/TaskSpec.swift
```diff
@@ -0,0 +1,91 @@
...
+@testable import Moya
+import Nimble
+import Foundation
+import Alamofire
+
+final class TaskSpec: QuickSpec {
+
+    override func spec() {
+
+        let encodable: [String: String] = ["Hello": "Moya"]
```
sunshinejr, line 11: can we update the test to cover for `[String: Any]`? e.g. `["a": "b", "c": 1]`

sunshinejr:
The PR looks mostly good, I just want to play with it in my projects to see how it looks/works - this is a second big rewrite to the `Task` type and so I really want to make sure we've got the most of it. I'd love if anyone that uses Moya in a bigger project could spin up the branch and play with it as well (cc @Moya/contributors)

Tests/MoyaTests/TaskSpec.swift
amaurydavid: For the dictionary to be `Encodable` the value must also be `Encodable`. `[String: Any]` won't be encodable.
Having a test with both String and Int we would have to use `[String: AnyEncodable]`, but I'm not sure that what you want.

sunshinejr: hmm, so currently if we want to pass `[String: Any]` as parameters we would need to convert each value to `AnyEncodable`? what I'm thinking is that we currently allow the parameters to be `[String: Any]` and ideally we would like to keep it this way

sunshinejr: @amaurydavid did you have a chance to think about the possible `[String: Any]` workaround?

amaurydavid: Sorry I didn't see your reply.
If think you already highlighted the situation [in this comment](https://github.com/Moya/Moya/pull/1936#pullrequestreview-314164520).

Using `Endocable` is actually enforced by the new parameter encoding from Alamofire, it's not a decision on my part. There have been some questions in the Alamofire github [here](https://github.com/Alamofire/Alamofire/issues/2763#issuecomment-476837964), [there](https://github.com/Alamofire/Alamofire/issues/2991#issuecomment-549216841), and surely in some other places.

It doesn't seem to be any way to use the dict `["a": "b", "c": 1]` into an `Encodable`. There are 2 workarounds for this:

- to replace the dict with a specific type:

```
struct MyParameters: Encodable {
 let a = "b"
 let c = 1
}
```

 - to wrap all values into `AnyEncodable`

sunshinejr: yeah, so from my personal experience I rarely use `Encodable`/separate structs for a simple API call and there are many places in which I want to pass [String: Any] as parameters

do you think we could create a helper function, e.g. `toEncodable()`, that could wrap `[String: Any]` to `Encodable`? ideally we wouldn't have to do such things, but as I see it right now it's hard with current design

cc @Moya/contributors

amaurydavid: Instead of a helper func we could had new inits to handle [String: Any] and ignore all values not being Encodable, but it would introduce a behavior users won't expect where some parameters won't be the request while passed to Moya.
I would rather let users make their own `Encodable`.

Noobish1:
What's the status of this PR?

sunshinejr:
@Noobish1 this is a very big change and I don't want to rush things as we did previously with introducing the `Task` type. It's mostly good, though we need to verify if it has all the needs we want (e.g. I would really like to be able to specify parameters by a literal, without introducing unnecessary interim `Encodable` type). Also would like someone else to clone & play with the API to see what else we need to discuss.

PR #1950, opened 2019-11-08
Update URL+Moya Change

dariukas:
Currently, without I put target.path into targetPath, in most cases target.path is called twice (when I setting path using Moya in a specific project) which makes an additional call: difficulties to debug, and performance change even small.

### DIFF
--- a/Sources/Moya/URL+Moya.swift
+++ b/Sources/Moya/URL+Moya.swift
@@ -8,7 +8,6 @@ public extension URL {
         // See: https://github.com/Moya/Moya/pull/1053
         // And: https://github.com/Moya/Moya/issues/1049
         let targetPath = target.path
-        let targetBaseURL = target.baseURL
         if targetPath.isEmpty {
             self = target.baseURL
         } else {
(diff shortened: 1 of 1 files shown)


Sources/Moya/URL+Moya.swift
```diff
@@ -7,10 +7,12 @@ public extension URL {
         // When a TargetType's path is empty, URL.appendingPathComponent may introduce trailing /, which may not be wanted in some cases
         // See: https://github.com/Moya/Moya/pull/1053
         // And: https://github.com/Moya/Moya/issues/1049
-        if target.path.isEmpty {
+        let targetPath = target.path
+        let targetBaseURL = target.baseURL
```
MoyaBot: - :warning: initialization of immutable value 'targetBaseURL' was never used; consider replacing with assignment to '_' or removing it
- :warning: initialization of immutable value 'targetBaseURL' was never used; consider replacing with assignment to '_' or removing it
- :warning: initialization of immutable value 'targetBaseURL' was never used; consider replacing with assignment to '_' or removing it

sunshinejr:
hey @dariukas thank you for the PR! would you please add [Changelog](https://github.com/Moya/Moya/blob/master/Changelog.md) entry (see [Changelog guidelines](https://github.com/Moya/contributors/blob/master/Changelog%20Guidelines.md)) and change the base branch from `master` to `development`?

sunshinejr:
Ah I see that you already did a PR to `development` here: https://github.com/Moya/Moya/pull/1957
In that case I'm closing this one - we are gonna merge `development` into `master` once we release a stable version 14! Thanks for the patch 🙇‍♂️

PR #1952, opened 2019-11-10
Fixing Data race conditions caught by TSAN

LucianoPAlmeida:
* Enable TSAN on MoyaTests Target, so when we run the tests TSAN is always enabled.

* Fixing issue caught by TSAN on interceptor when running tests with it 
Screen Shot 2019-11-10 at 18 03 46
It was just a concurrent read/write access to the willSend `MoyaInterceptor` property. 

cc @sunshinejr

### DIFF
--- a/Moya.xcodeproj/xcshareddata/xcschemes/Moya.xcscheme
+++ b/Moya.xcodeproj/xcshareddata/xcschemes/Moya.xcscheme
@@ -26,7 +26,8 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
-      shouldUseLaunchSchemeArgsEnv = "YES">
+      shouldUseLaunchSchemeArgsEnv = "YES"
+      enableThreadSanitizer = "YES">
       <Testables>
          <TestableReference
             skipped = "NO">
@@ -39,8 +40,6 @@
             </BuildableReference>
          </TestableReference>
       </Testables>
-      <AdditionalOptions>
-      </AdditionalOptions>
    </TestAction>
    <LaunchAction
       buildConfiguration = "Debug"
@@ -61,8 +60,6 @@
             ReferencedContainer = "container:Moya.xcodeproj">
          </BuildableReference>
       </MacroExpansion>
-      <AdditionalOptions>
-      </AdditionalOptions>
    </LaunchAction>
    <ProfileAction
       buildConfiguration = "Release"
--- a/Moya.xcodeproj/xcshareddata/xcschemes/ReactiveMoya.xcscheme
+++ b/Moya.xcodeproj/xcshareddata/xcschemes/ReactiveMoya.xcscheme
@@ -26,7 +26,8 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
-      shouldUseLaunchSchemeArgsEnv = "YES">
+      shouldUseLaunchSchemeArgsEnv = "YES"
+      enableThreadSanitizer = "YES">
       <Testables>
          <TestableReference
             skipped = "NO">
@@ -39,8 +40,6 @@
             </BuildableReference>
          </TestableReference>
       </Testables>
-      <AdditionalOptions>
-      </AdditionalOptions>
    </TestAction>
    <LaunchAction
       buildConfiguration = "Debug"
@@ -61,8 +60,6 @@
             ReferencedContainer = "container:Moya.xcodeproj">
          </BuildableReference>
       </MacroExpansion>
-      <AdditionalOptions>
-      </AdditionalOptions>
    </LaunchAction>
    <ProfileAction
       buildConfiguration = "Release"
--- a/Sources/Moya/Moya+Alamofire.swift
+++ b/Sources/Moya/Moya+Alamofire.swift
@@ -106,7 +106,7 @@ extension DownloadRequest: Requestable {
     }
 }
 
-final class MoyaRequestInterceptor: RequestInterceptor {
+final class MoyaRequestInterceptor: NSRecursiveLock, RequestInterceptor {
 
     var prepare: ((URLRequest) -> URLRequest)?
     var willSend: ((URLRequest) -> Void)?
@@ -117,8 +117,10 @@ final class MoyaRequestInterceptor: RequestInterceptor {
     }
 
     func adapt(_ urlRequest: URLRequest, for session: Alamofire.Session, completion: @escaping (Result<URLRequest, Error>) -> Void) {
-        let request = prepare?(urlRequest) ?? urlRequest
-        willSend?(request)
+        lock(); defer { unlock() }
+
+        let request = self.prepare?(urlRequest) ?? urlRequest
+        self.willSend?(request)
         completion(.success(request))
     }
 }
--- a/Sources/Moya/Moya+Alamofire.swift
+++ b/Sources/Moya/Moya+Alamofire.swift
@@ -106,10 +106,23 @@ extension DownloadRequest: Requestable {
     }
 }
 
-final class MoyaRequestInterceptor: NSRecursiveLock, RequestInterceptor {
+final class MoyaRequestInterceptor: RequestInterceptor {
+    private let lock: NSRecursiveLock = NSRecursiveLock()
 
     var prepare: ((URLRequest) -> URLRequest)?
-    var willSend: ((URLRequest) -> Void)?
+    private var internalWillSend: ((URLRequest) -> Void)?
+
+    var willSend: ((URLRequest) -> Void)? {
+        get {
+            lock.lock(); defer { lock.unlock() }
+            return internalWillSend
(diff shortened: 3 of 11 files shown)


Moya.xcodeproj/xcshareddata/xcschemes/Moya.xcscheme
```diff
@@ -26,7 +26,8 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
-      shouldUseLaunchSchemeArgsEnv = "YES">
+      shouldUseLaunchSchemeArgsEnv = "YES"
+      enableThreadSanitizer = "YES">
```
sunshinejr: should we have it enabled in the Moya.xcscheme as well?

Moya.xcodeproj/xcshareddata/xcschemes/ReactiveMoya.xcscheme
```diff
@@ -26,7 +26,8 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
-      shouldUseLaunchSchemeArgsEnv = "YES">
+      shouldUseLaunchSchemeArgsEnv = "YES"
+      enableThreadSanitizer = "YES">
```
sunshinejr: similar question for the other non-test schemes

Sources/Moya/Moya+Alamofire.swift
```diff
@@ -106,7 +106,7 @@ extension DownloadRequest: Requestable {
     }
 }
 
-final class MoyaRequestInterceptor: RequestInterceptor {
+final class MoyaRequestInterceptor: NSRecursiveLock, RequestInterceptor {
```
sunshinejr: oh nice, first time seeing inheritance from `NSRecursiveLock` rather then using an inner lock inside the object - are there any pros/cons using it that way?

sunshinejr:
thank you so much for the find @LucianoPAlmeida! and this is interesting - so do we actually call the same interceptor multiple times? did you check if it's alamofire or moya that does the call multiple times creating the race?

Moya.xcodeproj/xcshareddata/xcschemes/Moya.xcscheme
LucianoPAlmeida: I enabled in all the targets: Moya, ReactiveMoya, and RxMoya but for the Test action only (not enabled for build or run), so it is enabled for all the targets when we run cmd+u or `xcodebuild test` in the CI, also in MoyaTests too :)

Sources/Moya/Moya+Alamofire.swift
LucianoPAlmeida: This was only for simplicity, the concept is making the interceptor lockable, so we could manage lock/unlock inside and outside the class. The other option was to have a _lock property ... but then we would have to create an encapsulation logic to the willSend getter and setter. 
```swift
private _willSend: ...

var willSend : ... {
    get {
       _lock.lock(); defer { _lock.unlock() }
       return _willSend
    }

     set {
        _lock.lock(); defer { _lock.unlock() }
        _willSend = newValue
     }
}

```
I think we it was a public API that we wanted to make sure the access is atomic it would be a better option,because we encapsulate everything and users will not have to worry about. But since is an internal thing I think make the interceptor lockable is a more clear and simple option.
In the end the pro is just simplicity :)

LucianoPAlmeida:
> thank you so much for the find @LucianoPAlmeida! and this is interesting - so do we actually call the same interceptor multiple times? did you check if it's alamofire or moya that does the call multiple times creating the race?

The race is being caused by a concurrent read/write access to will, where the read is done by alamofire hook in the `adapt` which I think is on the alamofire thread, and the write(set the willSend property) is done in our setup method by another thread. So that what's causing the race ... so the solution was simply lock the access. I don't check if there is multiple calls to set, but I think there's not. The `adapt` may call multiple times, but that's not the cause of the data race :)

LucianoPAlmeida:
Two more additions to this:
* Since we are here, and we are enabling TSAN on tests I just replace the `objc_sync_enter/exit` APIs in favor of lock. 
* Increase the timeout of integration tests, because sometimes in slow networks tests will fail very often with the current 5.0 timeout.

Moya.xcodeproj/xcshareddata/xcschemes/Moya.xcscheme
sunshinejr: ah gotcha 👍

sunshinejr:
@LucianoPAlmeida what's the benefit of `NSLock` vs `objc_sync_enter/exit`?

Sources/Moya/Moya+Alamofire.swift
sunshinejr: This is a very good explanation, thanks! 

I think that making the class self-contained could help us in the future (meaning we lock it all in the interceptor). When we end up needing to use the interceptor again, we won't have to think about locking from the outside because it will be already taken-care-of. What do you think?

LucianoPAlmeida:
> @LucianoPAlmeida what's the benefit of `NSLock` vs `objc_sync_enter/exit`?

There are some benchmarks and stuff e.g. [this one](https://www.cocoawithlove.com/blog/2016/06/02/threads-and-mutexes.html) 
Where objc_sync_enter falls behind other APIs.
But the reason for changing was that:
* Normalizing the locking mechanism, because there some places with NSLock and some places with objc_sync_enter/exit... so the point was just to use the same mechanism for everything and since lock is the most common API, I'd go for that.
* objc_sync_enter is not available on linux, so if one day we would try to make an effort to support Moya on Linux, it would a thing that change anyways ;) 

In the end I think is not much difference, in terms of performance because of the places it is being used, but just normalizing the sync mechanism is a good thing :))

Sources/Moya/Moya+Alamofire.swift
LucianoPAlmeida: I personally prefer the current approach, because not all access need to be locked and most of the times the caller is the one who knows if we need locking or not.
But in this case think there is no problem in doing that, and is actually nicer at the API level (for the reason you mention that we don't to think about locking from the outside because it will be already taken-care-of).
I'll change that as soon as possible then :) 
Thanks

LucianoPAlmeida: Done :)

Sources/Moya/Moya+Alamofire.swift
```diff
@@ -107,9 +107,22 @@ extension DownloadRequest: Requestable {
 }
 
 final class MoyaRequestInterceptor: RequestInterceptor {
-
+    private let lock: NSRecursiveLock = NSRecursiveLock()
+
```
MoyaBot: - :warning: Lines should not have trailing whitespace. (`trailing_whitespace`)

Sources/Moya/Moya+Alamofire.swift
```diff
@@ -107,9 +107,22 @@ extension DownloadRequest: Requestable {
 }
 
 final class MoyaRequestInterceptor: RequestInterceptor {
+    private let lock: NSRecursiveLock = NSRecursiveLock()
 
     var prepare: ((URLRequest) -> URLRequest)?
-    var willSend: ((URLRequest) -> Void)?
+    private var internalWillSend: ((URLRequest) -> Void)?
```
MoyaBot: - :warning: Lines should not have trailing whitespace. (`trailing_whitespace`)

sunshinejr:
Hey @LucianoPAlmeida, looks good! One thing is that we want to support Xcode 10 in this newest release so the property wrapper would need to go for now, but we can try to reintroduce it afterwards into the `development-5.1` branch that Bas is working on.

LucianoPAlmeida:
> Hey @LucianoPAlmeida, looks good! One thing is that we want to support Xcode 10 in this newest release so the property wrapper would need to go for now, but we can try to reintroduce it afterwards into the `development-5.1` branch that Bas is working on.

That's unfortunate :(( 
But ok I'll remove the wrapper then. Thanks @sunshinejr :)

Sources/Moya/Moya+Alamofire.swift
```diff
@@ -107,9 +107,24 @@ extension DownloadRequest: Requestable {
 }
 
 final class MoyaRequestInterceptor: RequestInterceptor {
+    private let lock: NSRecursiveLock = NSRecursiveLock()
 
     var prepare: ((URLRequest) -> URLRequest)?
-    var willSend: ((URLRequest) -> Void)?
+    
+    // TODO: When update development to Swift 5.1 we should abstract
```
MoyaBot: - :warning: TODOs should be resolved (When update development to Swi...). (`todo`)

LucianoPAlmeida:
All reverted @sunshinejr

sunshinejr:
alright thanks @LucianoPAlmeida! one last thing - could you add changelog entry, please?

LucianoPAlmeida:
> alright thanks @LucianoPAlmeida! one last thing - could you add changelog entry, please?

Sure ;)

sunshinejr:
[beta.6 was just released 👍 ](https://github.com/Moya/Moya/releases/tag/14.0.0-beta.6)

LucianoPAlmeida:
Awesome 🎉

PR #1965, opened 2019-11-25
Add `mapString`, `mapDecodable` funtion take multiple key paths parameter

Woollim:
Hello all! I wanted to add new `mapString`, `mapDecodable` function

Much JSON Response, usually wrapped in many key path.
How could we get a `name` in below JSON Response?
It was not possible in the current `Moya` (We just made `XXX_ResponseData`)

```swift
let json = """
{
  result: {
    data: {
        name: "lee"
        korean_name: "이병찬"
    },
    response_code: 200
  }
}
"""
```

However, the new 'mapString' and 'mapDecodable' function take multiple key paths
Therefore, even the above complex JSON Response we can easily bring a `name`.

```swift
let name = try response.mapString(atKeypaths: ["result", "data", "name"])
print(name)
// "lee"

provider.rx.response(.getName)
    .map(String.self, atKeypaths: ["result", "data", "name"])
    // Type is now Observable<String>
    .subscribe { print($0) }
    // "lee"
```

### DIFF
--- a/Sources/Moya/Response.swift
+++ b/Sources/Moya/Response.swift
@@ -112,14 +112,31 @@ public extension Response {
     /// - parameter atKeyPath: Optional key path at which to parse string.
     func mapString(atKeyPath keyPath: String? = nil) throws -> String {
         if let keyPath = keyPath {
-            // Key path was provided, try to parse string at key path
-            guard let jsonDictionary = try mapJSON() as? NSDictionary,
-                let string = jsonDictionary.value(forKeyPath: keyPath) as? String else {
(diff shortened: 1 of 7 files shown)


amaurydavid:
I don't quite understand why you would need to have an array of String as keyPath, isn't one String enough ? `let name = dict.value(forKeyPath:"result.data.name") // prints "lee"`

Woollim:
Wow...... I learn the true meaning of `keyPath`
I didn't know until now I could access JSON Response lije `data.name`
Thank you @amaurydavid

PR #1969, opened 2019-12-07
Improved access token plugin

amaurydavid:
Some APIs require clients to use multiple access token types. For example, when using the Reddit API we need to use a basic token to get a bearer token, and then use the bearer token in every request.

With the current state of the AccessTokenPlugin, it's impossible to know in the token closure which token we are supposed to return. With that in mind, I added the AuthorizationType as parameter of the token closure so that we can now return the correct token.

Additionally, I've removed the `AuthorizationType.none` value because when implementing the token closure with the `AuthorizationType` as parameter there could have a possibility for it to be `AuthorizationType.none`, which doesn't make sense.
Also, it simplifies a bit the code when using an optional `AuthorizationType` instead of `AuthorizationType.none`.

### DIFF
--- a/Sources/Moya/Plugins/AccessTokenPlugin.swift
+++ b/Sources/Moya/Plugins/AccessTokenPlugin.swift
@@ -46,16 +46,18 @@ public enum AuthorizationType {
  */
 public struct AccessTokenPlugin: PluginType {
 
+    public typealias TokenClosure = (AuthorizationType) -> String
+
     /// A closure returning the access token to be applied in the header.
-    public let tokenClosure: () -> String
+    public let tokenClosure: TokenClosure
 
     /**
      Initialize a new `AccessTokenPlugin`.
 
      - parameters:
-       - tokenClosure: A closure returning the token to be applied in the pattern `Authorization: <AuthorizationType> <token>`
-    */
-    public init(tokenClosure: @escaping () -> String) {
+     - tokenClosure: A closure returning the token to be applied in the pattern `Authorization: <AuthorizationType> <token>`
+     */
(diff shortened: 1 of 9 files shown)


Sources/Moya/Plugins/AccessTokenPlugin.swift
```diff
@@ -6,15 +6,13 @@ import Foundation
 public protocol AccessTokenAuthorizable {
 
     /// Represents the authorization header to use for requests.
-    var authorizationType: AuthorizationType { get }
+    var authorizationType: AuthorizationType? { get }
```
LucianoPAlmeida: Hey @amaurydavid 
What's the advantage in making this optional? I mean instead of using `.none`. I'm asking because at the end using AuthorizationType optional nil a.k.a Optional<AuthorizationType>.none will be almost the same that  `AuthorizationType.none`

amaurydavid: The reasons are in the PR description ;)

LucianoPAlmeida: Oops 😅

LucianoPAlmeida:
LGTM, just one thing ... could you add a changelog entry for that?

sunshinejr:
@amaurydavid I like it a lot! Can you update the docs (migration/changelog/any other), please?

amaurydavid:
Sure ;)

sunshinejr:
[beta.6 was just released 👍 ](https://github.com/Moya/Moya/releases/tag/14.0.0-beta.6)

PR #1970, opened 2019-12-08
Support resume download with data

bauermusic:
Added a resume download case.
This is used when canceling a download and resuming at a later time.

Usually, when using background download, all requests are terminated when a user kills the App. It is possible however, to resume download. When a request is canceled, a task resume data is returned in the error's user info for NSURLSessionDownloadTaskResumeData. Using this data, we can resume the download when the App restarts.

### DIFF
--- a/Sources/Moya/Task.swift
+++ b/Sources/Moya/Task.swift
@@ -35,6 +35,9 @@ public enum Task {
 
     /// A file download task to a destination.
     case downloadDestination(DownloadDestination)
+    
+    /// A file download task with resume data, to a destination
+    case downloadResume(data: Data, destination: DownloadDestination)
 
     /// A file download task to a destination with extra parameters using the given encoding.
     case downloadParameters(parameters: [String: Any], encoding: ParameterEncoding, destination: DownloadDestination)
--- a/Sources/Moya/Endpoint.swift
+++ b/Sources/Moya/Endpoint.swift
@@ -85,7 +85,7 @@ public extension Endpoint {
         request.allHTTPHeaderFields = httpHeaderFields
 
         switch task {
-        case .requestPlain, .uploadFile, .uploadMultipart, .downloadDestination:
+        case .requestPlain, .uploadFile, .uploadMultipart, .downloadDestination, .downloadResume:
             return request
         case .requestData(let data):
             request.httpBody = data
--- a/Sources/Moya/MoyaProvider+Internal.swift
+++ b/Sources/Moya/MoyaProvider+Internal.swift
@@ -99,6 +99,9 @@ public extension MoyaProvider {
             case .downloadDestination(let destination), .downloadParameters(_, _, let destination):
                 return self.sendDownloadRequest(target, request: request, callbackQueue: callbackQueue, destination: destination, progress: progress, completion: completion)
             }
+            case .downloadResume(let resumeData, let destination):
+                return self.sendDownloadRequest(target, request: request, resumeData: resumeData, callbackQueue: callbackQueue, destination: destination, progress: progress, completion: completion)
+            }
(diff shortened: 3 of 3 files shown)


Sources/Moya/Task.swift
```diff
@@ -35,6 +35,9 @@ public enum Task {
 
     /// A file download task to a destination.
     case downloadDestination(DownloadDestination)
+
```
MoyaBot: - :warning: Lines should not have trailing whitespace. (`trailing_whitespace`)

amaurydavid:
Thanks for taking time fo make a PR! 
This sounds interesting. However if Moya were to support resuming downloads I think it would be great if we could go further by:
- providing an easy way to retrieve the `task resume data`.
- having a mechanism that automatically resumes those downloads. Maybe with #1924.

On a side note, if this update of `Task` is accepted, it would mean to make some updates to #1936.

Also, can you provide a test scenario or, even better, implementing this it in the example app?

bauermusic:
I should add a test, sure. As for auto-resume, I would still leave some decisions for the user (when and where to store the resume data), but sure, it should come as a more complete package.

LucianoPAlmeida:
> On a side note, if this update of `Task` is accepted, it would mean to make some updates to #1936.
> 

@amaurydavid @bauermusic :)
In my opinion, we should wait on this one until #1936 gets in since there is a lot of work already there and adding this would require some changes.
But that's only an opinion, I think that's something for @sunshinejr or someone else from the core team.

bauermusic:
Since the changes in this PR are minor (in quantity) and there's some more work to be done, let's wait for #1936.

bauermusic:
Not sure how to write a proper test for it since it involves a download-cancel setup. The download stub should be big enough so that the request could be canceled before completion.
Basically, when canceling a download request, the data in the underlying error is a small file (not the actual download data), that the iOS system uses to resume the download.

sunshinejr:
Hey @bauermusic - first of all, thank you so much for taking the time to implement this one! This is something I was thinking about lately and I think this could be a valuable addition.

In terms of the implementation, I was thinking more about changing current `Cacellable` to something like `Cancellable & Resumable`. This way when you have a reference to the download (`Cancellable & Resumable`), you can resume it when needed. What do y'all think @bauermusic @LucianoPAlmeida @amaurydavid?

As for the test - yes it's not that simple. You can take a look how we check upload progress with [OHHTTPSubs](https://github.com/Moya/Moya/blob/master/Tests/MoyaProviderSpec.swift#L724-L766). Basically we use response time to make sure the task, even though it's stubbed, is still taking longer to check for progress callbacks.

bauermusic:
Hi @sunshinejr,
The `Cancellable & Resumable` sounds really good. One thing to note, is that the 'resume data' is embedded in the response _error_. That means that if a user holds a pointer to the request, the same request will need to be updated.

Also, in the testing environment, it seems that a canceled request is not embedded with 'resume data', this is data created by the iOS system, so I'm not sure if it could be mocked.

There's another issue, not directly related, that I've not been able to figure out. When and App moves into the background, all active requests are canceled and are called back with canceled error. In **background download** however, I was expecting the same in `willTerminate` (by the user).
I was not able to 'catch' an active 'background' download before the App terminates. I had some code catching a canceled request and write the 'resume data' to disk, but it didn't run on termination. Also, manually canceling (in `willTerminate`) did not produce an error with resume data (completion block never got called).

What I was hoping for was: User download using 'background' configuration, kills the App, we then catch the error and write the resume data to disk, allowing the download to resume if needed.
I saw a few StackOverflow posts where the completion would be called back either before the App terminated or on the next run. For me, the completion never got called on termination.

PR #1979, opened 2019-12-29
MultiTarget conformance to AccessTokenAuthorizable

amaurydavid:
Just a fix for #1969 so that the `AccessTokenPlugin` correctly asks for a token when using `MultiTarget`.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,9 @@
 # Next
 
+### Changed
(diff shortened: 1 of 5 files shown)


LucianoPAlmeida:
LGTM 👍 @amaurydavid

sunshinejr:
ah good one @amaurydavid 👍 this is a breaking change tho, due to conforming to the protocol by default, would you please add a changelog entry and a note to the migration guide (e.g. that people don't have to conform to `AccessTokenAuthorizable` in `MultiTarget` anymore)? I think it's still worth it to get it into the current 14 release since Alamofire is not out yet

PR #1989, opened 2020-02-10
Adding support for MultiTarget with AccessTokenAuthorizable

embassem:
we often use AccessTokenPlugin to 'Signing' the request but when using a MultiTarget  the request didn't get 'Signing' as Multitarget itself is a TargetType wrapper,
adding support for  MultiTarget with AccessTokenAuthorizable will prevent the behavior the user doesn't understand, and make it easy to work with MultiTarget

### DIFF
--- a/Sources/Moya/Plugins/AccessTokenPlugin.swift
+++ b/Sources/Moya/Plugins/AccessTokenPlugin.swift
@@ -91,3 +91,15 @@ public struct AccessTokenPlugin: PluginType {
         return request
     }
 }
+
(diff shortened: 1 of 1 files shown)


amaurydavid:
Hello @embassem, the `MultiTarget` conformance have already been added with #1979 in the `development` branch, it's just not released yet.

sunshinejr:
ah yes I was thinking about releasing it with new Alamofire version but this is taking so long I might actually just release new version today. 

@embassem thank you very much for the contribution, though! I have to close this PR but I invited you to the org for spending your time trying to improve the library 🙇

PR #1992, opened 2020-02-15
[Release] 14.0.0 🚀

sunshinejr:
Alamofire 5.0 was just released so this prepares for official release 14.0.0!

In this PR:
- [x] Readme updates
- [x] Updated managers to use Alamofire 5.0
- [x] Upgraded to OHHTTPStubs 9.0
- [x] Fixed one bug in the JSON formatter in the example app


LucianoPAlmeida:
Danger Swift is giving a lot of errors lately, we should probably check it out :) 
cc @f-meloni

f-meloni:
I had a quick look at the 3 open PR that are failing at the moment
- 1 is just not passing the tests
- 2 are probably just missing the change that is in development 

```
      - run:
          name: Install MDSpell
          command: sudo npm install -g orta/node-markdown-spellcheck
```

f-meloni:
This specific failure is my fault :D Before Danger was not supporting SPM and to test your current version you had to tag it (in order to find it with Marathon), and an old tag was pushed by mistake on  https://github.com/f-meloni/danger-swift-xcodesummary.git during the laster release, that can be seen from `latest major:1000`.
I've already removed it, and took action to avoid that problem again :)

PR #1998, opened 2020-02-22
Generate Moya website script with rocket

f-meloni:
Fixes https://github.com/Moya/Moya/issues/1864

I couldn't fully try it, because it would push on on the repo, the part I tested looks working.

### DIFF
--- a/scripts/update_docs_website.sh
+++ b/scripts/update_docs_website.sh
@@ -1,0 +1,15 @@
+git clone git@github.com:Moya/moya.github.io.git _moya.github.io
+
+jazzy -o _moya.github.io
+
+cd _moya.github.io
+
+git add .
+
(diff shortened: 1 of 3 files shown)


scripts/update_docs_website.sh
```diff
@@ -0,0 +1,15 @@
+git clone git@github.com:Moya/moya.github.io.git _moya.github.io
```
f-meloni: Didn't know if you wanted the SSH url or the HTML one, can change it in case

sunshinejr: ssh is great, thanks!

sunshinejr:
Seems like jazzy broke and it says "0% documentation coverage" but I doubt it's an issue with this PR, gonna see if I can find out a root cause.

PR #2001, opened 2020-02-29
Introducing Path.

jdisho:
When new API endpoints are introduced, the `TargetType` enum can easily grow larger and, in such a case, I find it not easy to check if the endpoints are associated with the right `HTTP` methods. 

For the given reason, I'm introducing `Path` which combines in one place the endpoint and the `HTTP` method.

Example: 

```swift 
extension MyService: TargetType {
    var baseURL: URL { 
        return URL(string: "https://api.myservice.com")! 
     }

    var path: Path {
        switch self {
        case .zen:
            return .get(endpoint: "/zen")
        case .showUser(let id):
            return .get(endpoint: "/users/\(id)")
        case .updateUser(let id, _, _):
            return .post(endpoint: "/users/\(id)")
        case .createUser(_, _):
            return .post(endpoint: "/users")
        case .showAccounts:
            return .get(endpoint: "/accounts")
        }
    }
    ...
}
```

Let me know your thoughts! 
Thanks 😊

### DIFF
--- a/Examples/_shared/GiphyAPI.swift
+++ b/Examples/_shared/GiphyAPI.swift
@@ -9,16 +9,10 @@ public enum Giphy {
 
 extension Giphy: TargetType {
     public var baseURL: URL { URL(string: "https://upload.giphy.com")! } // swiftlint:disable:this force_unwrapping
-    public var path: String {
+    public var path: Path {
         switch self {
         case .upload:
-            return "/v1/gifs"
-        }
-    }
-    public var method: Moya.Method {
-        switch self {
-        case .upload:
-            return .post
+            return .post(endpoint: "/v1/gifs")
         }
     }
     public var task: Task {
--- a/Examples/_shared/GitHubAPI.swift
+++ b/Examples/_shared/GitHubAPI.swift
@@ -34,17 +34,16 @@ public enum GitHub {
 
 extension GitHub: TargetType {
     public var baseURL: URL { URL(string: "https://api.github.com")! }
-    public var path: String {
+    public var path: Path {
         switch self {
         case .zen:
-            return "/zen"
+            return .get(endpoint: "/zen")
         case .userProfile(let name):
-            return "/users/\(name.urlEscaped)"
+            return .get(endpoint: "/users/\(name.urlEscaped)")
         case .userRepositories(let name):
-            return "/users/\(name.urlEscaped)/repos"
+            return .get(endpoint: "/users/\(name.urlEscaped)/repos")
         }
     }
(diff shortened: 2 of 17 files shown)


LucianoPAlmeida:
@jdisho Thank you for the proposal :))
My two cents here:
Would it be a better idea to bring this in a retro compatible way, where the old option still valid ... and the user could choose? This is really a question because personally, I think some people may prefer the old way and this could be inserted as an option (we could do either way) and not a breaking change :)

LucianoPAlmeida:
cc @sunshinejr @Moya/contributors

colinta:
I agree with @LucianoPAImeida that this should try to be a backwards compatible change - with that said, I really like the idea! (I would agree with most “decluttering” of the Target protocol - it’s kind of a mess sometimes to find the right code). Would it be going too far to include the request params? Question: when will we be building a request that uses the path& http method, but *doesn’t* need the parameters?

For backwards compatibility some renaming would be in order; if instead of `var path: Path` which *replaces* `var path: String`, this could be an optional-and-default implemented `var request: Request`. Then we could migrate *towards* this feature without requiring another major version release. I really dig the name Path, but we’re already at version 5 in what was always meant to be just a thin network organization tool.

colinta:
On second thought adding parameters would be too much clutter. Http method and path make sense to me.

devxoul:
👍 for what it wants to solve. I'm using [MoyaSugar](https://github.com/devxoul/MoyaSugar) to reduce bunch of boilerplates. Hope this discussion goes well :)

sunshinejr:
I like the idea! Seems like a win in the long run. Thought I'm having a hard time thinking about a nice way of keeping it backwards compatible without introducing other problems. Then also we are refactoring the `Task` type in #1936, so when you add these two it might be too much for a single release (? not sure).

jdisho:
Hi folks! 👋
Any direction regarding this PR?

colinta:
My desire to have this implemented in a backwards-compatible way still stands.  As it is now including this feature would require a new major version release due to the change to `path`

colinta:
Also, we might want to look to the past as we think about this: https://github.com/Moya/Moya/pull/427 and https://github.com/Moya/Moya/issues/408

https://github.com/Moya/Moya/blob/9cc10d0f6ba8179c46695bf0831fd22a83e7a6c5/Sources/Moya/MultiTarget.swift

amaurydavid:
Could we rename `Path` into Route as in MoyaSugar ? I think it makes more sense than having the http method and the path into the notion of `Path`.

Issue #1123, opened 2017-06-13
Improving Plugins.


sunshinejr:
Hey all! I'm here to present my idea of improving Plugins and continue (kinda deadly) discussion from #985. 

Right now plugins:
- Are limited to one base type that has many functions that are not used and most of the time empty (but still executed).
- Are hard to expand (as a result of point above).
- Have limited functionality (as a result of points above).
- Do many things that `endpointClosure` & `requestClosure` do as well - making it ambiguous for users to use & for us to recommend a good practice.

Below is my proposition on to how make all of the above go away. Main idea is to remove `closures`, `manager` etc. from `MoyaProvider` and replace it with plugins. The first version may just introduce everything we need to a migration and then expand it as we go. Maybe we can even extract something from `TargetType`? 🤔 

### Types

```swift
// Base type

public protocol Plugin {}
```

```swift
// Current plugin methods extracted into Plugin types

public protocol RequestDelegatePluginType: Plugin {
    func willSend(_ request: RequestType, target: TargetType)
    func didReceive(_ result: Result<Moya.Response, MoyaError>, target: TargetType)
}

public protocol ResponseProcessingPluginType: Plugin {
    func process(_ result: Result<Moya.Response, MoyaError>, target: TargetType) -> Result<Moya.Response, MoyaError>
}
```

```swift
// `endpointClosure` & `requestClosure` & `stubbingClosure` into Plugin types

public protocol EndpointPreparingPluginType: Plugin {
    func prepare<T: TargetType>(endpoint: Endpoint<T>, target: T) -> Endpoint<T>
}

public protocol RequestPreparingPluginType: Plugin {
    func prepare(_ request: URLRequest, target: TargetType) -> URLRequest
}

public protocol ResponseStubbingPluginType: Plugin { // we may also remove this one and try to change `ResponseProcessingPluginType` a little bit to allow stubbing behavior :)
    func stubBehavior<T: TargetType>(_ target: T) -> StubBehavior
}
``` 

```swift
// Authentication moved from Request to Plugin type too

public protocol AuthenticationPluginType: Plugin {
    func authenticate(usingCredential credential: URLCredential) -> Self
}
```

### Migration from current closures
We could persist current closures as a default plugins, e.g. this `DefaultEndpointClosure` translated:
```swift
class DefaultEndpointPreparingPlugin: EndpointPreparingPluginType {
    
    func prepare<T>(endpoint: Endpoint<T>, target: T) -> Endpoint<T> where T: TargetType {
[...]

PR #2002, opened 2020-03-01
Add the possibility to modify the upload request in plugin.

VincentSit:
When a form request contains both files and a parameters, the existing method `prepare(_:target:)`  cannot handle `MultipartFormData`. So I added a method to deal with this situation.

More specifically, suppose your request needs to be encrypted. You usually create a plug-in, implement the `prepare(_:target:)` method. But when the request content contains files, for example, you update both the user's profile picture and name at the same time, the method is not called. That's the problem.

May not be a very common use case, what do you think?


Tests/MoyaTests/TestPlugin.swift
```diff
@@ -12,6 +12,12 @@ final class TestingPlugin: PluginType {
         request.addValue("yes", forHTTPHeaderField: "prepared")
         return request
     }
+    
+    func prepare(_ request: URLRequest, multipartBody: [MultipartFormData], target: TargetType) -> (URLRequest, [MultipartFormData]) {
+        var request = request
+        request.addValue("yes", forHTTPHeaderField: "prepared-multipart-body")
+        return request
```
LucianoPAlmeida: This is failing to compile because the return type is a tuple and you are returning a request :)
```suggestion
        return (request, multipartBody)
```

VincentSit: Thank you.

Tests/MoyaTests/TestPlugin.swift
```diff
@@ -12,6 +12,12 @@ final class TestingPlugin: PluginType {
         request.addValue("yes", forHTTPHeaderField: "prepared")
         return request
     }
+
```
MoyaBot: - :warning: Lines should not have trailing whitespace. (`trailing_whitespace`)

Tests/MoyaTests/TestPlugin.swift
```diff
@@ -13,6 +13,12 @@ final class TestingPlugin: PluginType {
         return request
     }
 
+    func prepare(_ request: URLRequest, multipartBody: [MultipartFormData], target: TargetType) -> (URLRequest, [MultipartFormData]) {
+        var request = request
+        request.addValue("yes", forHTTPHeaderField: "prepared-multipart-body")
```
LucianoPAlmeida: Shouldn't this be asserted at some point of the test? :)

swizzlr:
# [Codecov](https://codecov.io/gh/Moya/Moya/pull/2002?src=pr&el=h1) Report
> Merging [#2002](https://codecov.io/gh/Moya/Moya/pull/2002?src=pr&el=desc) into [development](https://codecov.io/gh/Moya/Moya/commit/d3206d70642443279313de7308d790c58f765823&el=desc) will **increase** coverage by `1.09%`.
> The diff coverage is `n/a`.

[Impacted file tree graph](https://codecov.io/gh/Moya/Moya/pull/2002?src=pr&el=tree)

```diff
@@               Coverage Diff               @@
##           development    #2002      +/-   ##
===============================================
+ Coverage        92.91%   94.01%   +1.09%     
===============================================
  Files               27       27              
  Lines              931      935       +4     
===============================================
+ Hits               865      879      +14     
+ Misses              66       56      -10     
```


| [Impacted Files](https://codecov.io/gh/Moya/Moya/pull/2002?src=pr&el=tree) | Coverage Δ | |
|---|---|---|
| [Sources/Moya/Response.swift](https://codecov.io/gh/Moya/Moya/pull/2002/diff?src=pr&el=tree#diff-U291cmNlcy9Nb3lhL1Jlc3BvbnNlLnN3aWZ0) | `95.34% <0.00%> (+4.65%)` | :arrow_up: |
| [Sources/RxMoya/Observable+Response.swift](https://codecov.io/gh/Moya/Moya/pull/2002/diff?src=pr&el=tree#diff-U291cmNlcy9SeE1veWEvT2JzZXJ2YWJsZStSZXNwb25zZS5zd2lmdA==) | `61.53% <0.00%> (+7.69%)` | :arrow_up: |
| [Sources/ReactiveMoya/SignalProducer+Response.swift](https://codecov.io/gh/Moya/Moya/pull/2002/diff?src=pr&el=tree#diff-U291cmNlcy9SZWFjdGl2ZU1veWEvU2lnbmFsUHJvZHVjZXIrUmVzcG9uc2Uuc3dpZnQ=) | `93.93% <0.00%> (+9.09%)` | :arrow_up: |
| [Sources/RxMoya/Single+Response.swift](https://codecov.io/gh/Moya/Moya/pull/2002/diff?src=pr&el=tree#diff-U291cmNlcy9SeE1veWEvU2luZ2xlK1Jlc3BvbnNlLnN3aWZ0) | `100.00% <0.00%> (+12.50%)` | :arrow_up: |

------

[Continue to review full report at Codecov](https://codecov.io/gh/Moya/Moya/pull/2002?src=pr&el=continue).
> **Legend** - [Click here to learn more](https://docs.codecov.io/docs/codecov-delta)
> `Δ = absolute <relative> (impact)`, `ø = not affected`, `? = missing data`
> Powered by [Codecov](https://codecov.io/gh/Moya/Moya/pull/2002?src=pr&el=footer). Last update [86f2b1f...c798a5d](https://codecov.io/gh/Moya/Moya/pull/2002?src=pr&el=lastupdated). Read the [comment docs](https://docs.codecov.io/docs/pull-request-comments).

sunshinejr:
Thanks for the work @VincentSit, we really appreciate it!

> More specifically, suppose your request needs to be encrypted. You usually create a plug-in, implement the prepare(_:target:) method. But when the request content contains files, for example, you update both the user's profile picture and name at the same time, the method is not called. That's the problem.

Is the method really not called? We call the `prepare` func from plugins when Alamofire calls to "adapt" the request. In that case would that be a bug on Alamofire's part or is that intentional?

VincentSit:
@sunshinejr Thanks. It's been a few months since I made this change. I don't remember too well. Let me do some tests, will reply to you later.

VincentSit:
@sunshinejr Sorry for the incorrect description, that method will be called. But it cannot return the modified multipartBody.

Tests/MoyaTests/TestPlugin.swift
VincentSit: Do you mean I need to write a complete upload file test? :)

LucianoPAlmeida: I'm really not sure, I just think if you add this header in a test you should assert that it on the request at somepoint ...

sunshinejr:
@VincentSit no issue at all, thanks for checking on that!

So the only issue we have is that we cannot modify multipart data in a plugin, correct? In that case it should be available to do in `Endpoint` closure as it has actually a better types for that task - you can modify the task however you want based on a condition.

I'm not fond of adding this case to the `Plugin` purely because it seems like we can do it already and the use case is fairly unusual... But I'm open for discussion (cc @Moya/contributors)

Now, should the functionality be in endpoint closure or in plugin type? Not sure. [I have been trying to come up with a better solution for customizable code (aka plugins) for a while](https://github.com/Moya/Moya/issues/1123), but with no luck.

VincentSit:
@sunshinejr Thanks. Yes, you are right. The only problem for me is that the multipart data cannot be  modified in plugin, and adding a method is the fastest way to solve the problem. So I did that.

I read your previous discussions. I agree with you, it deepens my understanding of Moya, and I have never considered doing it in Endpoint before.

Anyway, hopefully there will be a way to do this in future releases. You can also close this PR as it does not look like the best solution. Thank you.

sunshinejr:
@VincentSit no problem at all, thank you so much for taking the time and working on this feature! I'll close it for now due to the discussion above, but I invited you to the Moya organization as a token of appreciation for your hard work 🙇‍♂️

VincentSit:
@sunshinejr Thanks a lot.

Tests/MoyaTests/TestPlugin.swift
VincentSit: Hi @LucianoPAlmeida Since this PR has been closed, no further updates are needed. Thanks for your comments.

Issue #2003, opened 2020-03-04
Support RxSwift 5.1.0


chiquan91:
Could you update your framework with Rx 5.1.0 dependency because of Apple's hard deprecation, starting April 2020: https://developer.apple.com/news/?id=12232019b

sunshinejr:
@chiquan91 this should work out of the box, you just need to update the dependency graph to cover for new RxSwift release (we have ~> 5.0 constraint which should cover the 5.1 version as well). Did you try to see if it works?

jdisho:
Ups, I got a bit confused. 
@sunshinejr I think you are right.

chiquan91:
Thank you, i had run update and it had updated to RxSwift 5.1.0 as expected.

MhmdRizk:
Hello everyone, 
I'm using RxMoya
I installed Moya using carthage, github "Moya/Moya" ~> 14.0
I tried uploading my app to the store anyway review team rejected the binary since it's still using UIWebView. I checked my code very well. 
then entered this comment in terminal in my project. and found out that even RxSwift 5.1.1 is using UIwebview 

`./Carthage/Checkouts/Moya/Carthage/Checkouts/RxSwift/RxCocoa/iOS/UIWebView+Rx.swift://  UIWebView+Rx.swift
./Carthage/Checkouts/Moya/Carthage/Checkouts/RxSwift/RxCocoa/iOS/UIWebView+Rx.swift:    extension Reactive where Base: UIWebView {`

did anyone face this issue? 
Thanks in advance.

sunshinejr:
@MhmdRizk are you sure you're using RxSwift 5.1.1? did you check your Cartfile.resolved & Carthage build folder?

MhmdRizk:
@sunshinejr , yup made sure.
`*** Checking out Moya at "14.0.0"
*** Checking out RxSwift at "5.1.1"
*** Checking out ReactiveSwift at "6.2.1"`
I'm using RxGesture, but I made sure they updated and using latest. 
`*** Checking out RxGesture at "3.0.3"`

sunshinejr:
@MhmdRizk please double check the Carthage build folder, the files you mentioned are not there on the RxSwift 5.1.1 version: https://github.com/ReactiveX/RxSwift/tree/5.1.1/RxCocoa/iOS

MhmdRizk:
@sunshinejr  I removed the project and downloaded it again. and then after updating carthage again everything worked. 
Thank you

PR #2004, opened 2020-03-04
Upgrade to RxSwift 5.1

jdisho:
Since we have an important update from [RxSwift](https://github.com/ReactiveX/RxSwift/releases/tag/5.1.0), I'm upgrading it to the latest version, [5.1](https://github.com/ReactiveX/RxSwift/releases/tag/5.1.0) and bumping `Moya.podspec` version to `14.1.0`

This upgrade does not introduce any breaking changes. 

Closing #2003


LucianoPAlmeida:
@jdisho Changes looks good 👍 But I think we need to add a changelog entry for that :)

sunshinejr:
@LucianoPAlmeida I will hold on this for a sec since this PR shouldn't really be needed as we have a bit loosely tightened the constraint for deps (~> 5.0 should be able to update to 5.1). More in the issue mentioned in the PR.

LucianoPAlmeida:
> @LucianoPAlmeida I will hold on this for a sec since this PR shouldn't really be needed as we have a bit loosely tightened the constraint for deps (~> 5.0 should be able to update to 5.1). More in the issue mentioned in the PR.

@sunshinejr Sure! :)

sunshinejr:
@jdisho seems like we don't need to bump the version for now, the problem in the original issue was resolved - in that case I'm gonna close the issue. Thank you though for a very fast PR! 🙏

PR #2007, opened 2020-03-10
Add task uploadCompositeFile

mikaelbartlett:
This pull request adds missing task for uploading file with URL parameters.

I've seen #1936 and it seems this will solve this issue. This is a small fix to add this task until #1936 is merged and released.

### DIFF
--- a/Sources/Moya/Endpoint.swift
+++ b/Sources/Moya/Endpoint.swift
@@ -96,7 +96,7 @@ public extension Endpoint {
             return try request.encoded(encodable: encodable, encoder: encoder)
         case let .requestParameters(parameters, parameterEncoding):
             return try request.encoded(parameters: parameters, parameterEncoding: parameterEncoding)
-        case let .uploadCompositeMultipart(_, urlParameters):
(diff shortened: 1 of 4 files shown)


Sources/Moya/Endpoint.swift
```diff
@@ -96,7 +96,7 @@ public extension Endpoint {
             return try request.encoded(encodable: encodable, encoder: encoder)
         case let .requestParameters(parameters, parameterEncoding):
             return try request.encoded(parameters: parameters, parameterEncoding: parameterEncoding)
-        case let .uploadCompositeMultipart(_, urlParameters):
+        case let .uploadCompositeFile(_ , urlParameters), let .uploadCompositeMultipart(_, urlParameters):
```
MoyaBot, line 99: - :warning: There should be no space before and one after any comma. (`comma`)

sunshinejr:
Thanks for the PR @mikaelbartlett, really appreciate it! We're currently trying to figure out if we can get the refactoring PR in the next release - if we do then this PR wouldn't be needed. But in case we won't, we will think about this and the other PR for improving the task type. I'll keep you posted!

PR #2015, opened 2020-03-22
`sampleData` as an optional property.

jdisho:
Since, `sampleData` is used only to stub data for use in testing, I find it a non-required property. In most of the cases we end up just returning empty data. 

For this reason, I'm making the `sampleData` an optional property, by adding it into the protocol `TargetType` extension: 

```swift 
public extension TargetType {
    ...

    var sampleData: Data { Data() }
}
```

This is not a breaking change, therefore I'm opening this PR towards the `master` branch.

✌️


sunshinejr:
Yes this was something I was thinking about for a while now, but I'm summoning @SD10 from the graveyard as his #1502 is more than 2 years old now and also tried to simplify stubbing data 😄

amaurydavid:
It looks OK to me, though I would prefer having it really optional. But I think it makes a lot more code updates. 
As you mention it @sunshinejr I've been working for a while on a  way to improve (or at least try) stubbing but still struggle at some point and regularly put it aside. I'll give it another try in the next days.

SD10:
👋 

I suggested this awhile back when I was new to Moya [(discussion starts here)](https://github.com/Moya/Moya/issues/1126#issuecomment-311798808). We ultimately decided that we should [favour explicitly-defined behaviour over default implementations](https://github.com/Moya/Moya/blob/master/Vision.md#project-goals) as described in the vision for this project. cc @pedrovereza 

In regards to #1502, I think this is good progress to be the solution for the larger concern of separating stubbing/sampleData from `TargetType` but it was held up for several reasons:
1. making sure you can conveniently stub Moya requests in the test target [(see this comment)](https://github.com/Moya/Moya/pull/1502#issuecomment-404673285)
2. not enough feedback from the community
3. limiting the amount of code churn in an established project

It's something I'd like to take a look at again this year, but I have little time for open source still as I'm adjusting to a new job in crazy times

jdisho:
Folks, is this PR okay to be merged?

LucianoPAlmeida:
Looks fine to me :) 
cc @sunshinejr

sunshinejr:
So because we don't have an actual plan for bigger refactor for the stubbing behaviors yet, I'm keen to merge this in. What do you think @SD10? 

Also cc @ashfurrow: I know you had opinions on this one (I did as well). But, after working with Moya for a few more years, I think the `sampleData` started to get to me. Since I usually add stubs after I record them, it became quite tedious to add empty `Data()` to targets when I don't really have the data yet.

@jdisho can you update changelog / look for any docs that we need to update as well? Also, I know it's non-breaking, but can you also point to develop? I might want to release a new preview version sooner and we could include this too.

LucianoPAlmeida:
> So because we don't have an actual plan for bigger refactor for the stubbing behaviors yet, I'm keen to merge this in. What do you think @SD10 @amaurydavid?
> 

@sunshinejr There is a PR from @amaurydavid #2020 with a proposal of a new stub mechanism there. There is a discussion already started there too :)

sunshinejr:
@LucianoPAlmeida yeah I saw that, though it's not really a plan but rather one of the directions we could take. What I meant is that we don't know if/how we will refactor what we have right now and this small update makes sense to me. Though I would really like @Moya/contributors to chime in on this one as well.

Ah and btw, in terms of the plugins/closures/stubbing situation, I will be creating a separate thread to see what we have and how we can improve it...

SD10:
I think this change is rather trivial -- the reason we've rejected this proposal historically is because it didn't fit the vision -- maybe it's time to make a change

I'm OK with this change if we end up ultimately changing Moya's stubbing mechanism. It would be a good bridge to phase out `sampleData` by making it have this default implementation in a release prior to doing that.

sunshinejr:
merging this one - gonna add changelog & migration entry after the merge (`gh cli` has issues with me pulling this one for some reason... 🤷)

PR #2016, opened 2020-03-22
Update Basic.md

lzzhoujielun:
failure to  call  absoluteString() in the document

### DIFF
--- a/docs_CN/Examples/Basic.md
+++ b/docs_CN/Examples/Basic.md
@@ -118,7 +118,7 @@ JSON, images, text, 或者任何您希望从endpoint得到的.
 
 ```swift
 let endpointClosure = { (target: MyService) -> Endpoint in
-    return Endpoint(url: URL(target: target), sampleResponseClosure: {.networkResponse(200, target.sampleData)}, method: target.method, task: target.task)
+    return Endpoint(url: URL(target: target).absoluteString, sampleResponseClosure: {.networkResponse(200, target.sampleData)}, method: target.method, task: target.task)
 }
 ```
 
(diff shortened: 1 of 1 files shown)


jdisho:
Hmm, nothing against this PR, but I find a bit misleading that the parameter name is `url` and its expected type is `String`. Makes sense to change the type to `URL`? 🤔

LucianoPAlmeida:
> Hmm, nothing against this PR, but I find a bit misleading that the parameter name is `url` and its expected type is `String`. Makes sense to change the type to `URL`? 🤔

@jdisho I think it does make sense, but I don't know if maybe there is an overload(to simplify calls) and there another that takes an URL? Not sure, but if it is the only one, change to URL makes sense but that would be a breaking change...

sunshinejr:
@jdisho @LucianoPAlmeida I was actually surprised myself that we pass the `String` type, even though we still need to convert it to `URL` when passing to `URLRequest` afterwards. Not sure if there is anything against changing it the type to `URL` (besides the fact that people would need to update & need to release major version), but maybe we can try with a PR and see how it goes?

LucianoPAlmeida:
> @jdisho @LucianoPAlmeida I was actually surprised myself that we pass the `String` type, even though we still need to convert it to `URL` when passing to `URLRequest` afterwards. Not sure if there is anything against changing it the type to `URL` (besides the fact that people would need to update & need to release major version), but maybe we can try with a PR and see how it goes?

@sunshinejr @jdisho Maybe to not be a breaking change, we can just add another overload with the same signature, just with URL type, without removing the string one, maybe just deprecate the string version... :)

PR #2020, opened 2020-03-27
Feature/stubbing from plugins

amaurydavid:
Following recent (#2015) and ancient (#1502) discussions about stubbing, here is my solution to simplify it.
All settings for stubbing have been regrouped in a single function to be implemented in a custom plugin.

It can be part of a 2 part refactoring aiming to remove closures from provider, in benefit of using plugins.

Let me know your thoughts.

### DIFF
--- a/Sources/Moya/MoyaProvider+Internal.swift
+++ b/Sources/Moya/MoyaProvider+Internal.swift
@@ -125,8 +125,9 @@ public extension MoyaProvider {
                     return .failure(error)
                 }
             }
-
-            let validatedResponse = stubbedResult.flatMap { validate($0) }
+            
+            let completeResult: MoyaResult = stubbedResult.map { Response(statusCode: $0.statusCode, data: $0.data, request: request, response: $0.response) }
+            let validatedResponse = completeResult.flatMap { validate($0) }
             plugins.forEach { $0.didReceive(validatedResponse, target: target) }
             completion(validatedResponse)
         }
--- a/Sources/Moya/MoyaProvider+Internal.swift
+++ b/Sources/Moya/MoyaProvider+Internal.swift
@@ -21,8 +21,11 @@ public extension MoyaProvider {
     /// Performs normal requests.
     func requestNormal(_ target: Target, callbackQueue: DispatchQueue?, progress: Moya.ProgressBlock?, completion: @escaping Moya.Completion) -> Cancellable {
         let endpoint = self.endpoint(target)
-        let stubBehavior = self.stubClosure(target)
         let cancellableToken = CancellableWrapper()
+        var stubBehavior: StubBehavior?
+        if let stubbedTarget = target as? StubbedTargetType {
+            stubBehavior = plugins.compactMap { $0.stubBehavior(for: stubbedTarget) }.first
+        }
 
         // Allow plugins to modify response
         let pluginsWithCompletion: Moya.Completion = { result in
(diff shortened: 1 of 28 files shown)


Sources/Moya/MoyaProvider+Internal.swift
```diff
@@ -123,22 +123,10 @@ public extension MoyaProvider {
...
-                plugins.forEach { $0.didReceive(result, target: target) }
-                completion(result)
-            case .networkError(let error):
-                let error = MoyaError.underlying(error, nil)
-                plugins.forEach { $0.didReceive(.failure(error), target: target) }
-                completion(.failure(error))
-            }
+            let completeResult: MoyaResult = stubbedResult.map { Response(statusCode: $0.statusCode, data: $0.data, request: request, response: $0.response) }
+            let validatedResponse = completeResult.flatMap { validate($0) }
+            plugins.forEach { $0.didReceive(validatedResponse, target: target) }
```
LucianoPAlmeida: I can't see a good use case where the other plugin methods e.g. `didReceive` would be used on any stub plugin ...
Any examples @amaurydavid?

Sources/Moya/Plugin.swift
```diff
@@ -13,18 +13,25 @@ public protocol PluginType {
     /// Called immediately before a request is sent over the network (or stubbed).
     func willSend(_ request: RequestType, target: TargetType)
 
+    /// Called when the request is ready to be sent. Return a StubBehavior if you want your request to be stubbed,
+    /// or `nil` to actually send the request over the network
+    ///
+    /// If a provider has more than 1 plugin returning a non nil value,only the value of the first of those plugins will be used.
+    func stubBehavior(for target: TargetType) -> StubBehavior?
```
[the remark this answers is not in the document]
amaurydavid: I like to think of plugins to be the centralized place to customize providers so that we don't have to search if we have to provide a closure, a plugin, or anything else. I agree the new stubBehavior(for:) function in PluginType doesn't perfectly fit with existing ones, mainly because values could be ignored.

Sources/Moya/MoyaProvider+Internal.swift
amaurydavid: I'm not sure I understand your point. `didReceived` is called whether the response is a real response or a stubbed response. Once we built the stubbed response and we are ready to use it, we first notify plugins about the incoming response.

LucianoPAlmeida: My point is that if a user implements a stub plugin, there any case where it would be useful to implement other method besides the stub one? e.g. did receive

amaurydavid: If we think in terms of "1 plugin = 1 role" I don't think so. But I guess we could also have a single plugin acting as a stubbing plugin + AccesTokenPlugin + NetworkActivityPlugin.
Do you ask because thinking about separating stubbing into a proper plugin type ? I've thought about this but did not find a way to use multiple plugin types in the `provider.plugins` property without allowing to use non pluginType objects in it.

LucianoPAlmeida: > Do you ask because thinking about separating stubbing into a proper plugin type ?

Exactly, that goes to the point that plugins are much more general and since stub is much specific, seems like a good idea. And defining a proper stub structure we can have just the specific things we need for stubbing.

LucianoPAlmeida: Also, I think this allows users to make mistakes that would be hard to find out ... example: If you define a stub together in a `Plugin1` and another member of your team in `Plugin2` for the same case of a target, you might get an error on your stub being different that you expect and it could be really hard to track down from where this is coming from ... if you don't know that maybe there is some defined in other place and there isn't a rule defined to which one would be used :)

amaurydavid:
I'm closing it as it's a not a satisfying solution and will need additionnal discussions on the future for stubbing.

PR #2024, opened 2020-04-01
Revert "Remove Combine extensions"

### LINKED ISSUE #1928: Combine support mega-thread.

MaxDesiatov:
Closes #1928.

This reverts commit 7e82d4b1e012999e8e478f9aeb17389133bf009f.

This doesn't include updates to `README.md` and the linking fix for Xcode 10 from the original commit as I'm assuming the next version of Moya would drop support for Xcode 10?


sunshinejr:
ah that was fast, thanks @MaxDesiatov!
>  I'm assuming the next version of Moya would drop support for Xcode 10?
Yes next version of Moya would require Xcode 11+
- though this doesn't prevent us from a bad crash on different package managers, the linking we did in the project helps for the manual framework & Carthage integration, but we also need to think about SPM & Cocoapods
- CocoaPods should be easy as we can add the same linking flag we did here, just in a Podfile
- SPM is different though because we cannot pass a build flag like we do with all other providers, so we would have to either skip SPM support for Combine or create a separate

MaxDesiatov:
> we would have to either skip SPM support

Wouldn't adding the platforms argument in `Package.swift` be enough?

```swift
platforms: [
  .macOS(.v10_15),
  .iOS(v13)
],
```

sunshinejr:
@MaxDesiatov yup, but we can only set these per whole Package correct? This way people with min target lower (e.g. iOS 12) wouldn't be able to install new updates for Moya through SPM and we'd like to avoid that. [There was an idea for per-target platform on Swift forums](https://forums.swift.org/t/platform-per-target/29253/10) but I'm not sure if something was done in that regard.

MaxDesiatov:
Would it make more sense then to have a separate repository for this under the Moya organisation?

sunshinejr:
@MaxDesiatov we usually try to not create additional repositories for the Moya submodules and so we just use separate packages (like `Moya/RxSwift`, `Moya/ReactiveSwift`) so I was thinking about just creating `Moya/Combine` for all package managers to be consistent (it would be weird to have it by default on Carthage and not on SPM for instance). The only difference between package managers would be that the SPM would need iOS 13+ support.

MaxDesiatov:
>creating Moya/Combine for all package managers to be consistent

Sorry, not sure I'm following, how exactly would that work for SwiftPM? As far as I understand, it always fetches `Package.swift` from the root of the repository, how exactly would a SwiftPM user specify a dependency on `Moya/Combine` in `Package.swift` of their own library or Xcode SwiftPM dependencies list for their app?

For example, here's a dependencies list in one my of apps depending on Moya, how would that list look like if `Moya/Combine` was a separate package without being a separate repository?

Screenshot 2020-04-01 at 15 20 03

sunshinejr:
@MaxDesiatov yeah, AFAIK current SPM support in Xcode builds all targets and you can just opt-in to move to the project whatever you like. So in our case we produce 3 targets right now (Moya, RxMoya, ReactiveMoya) and you can move what you want to your project as a dependency.

MaxDesiatov:
Ok, I think I see what you mean, I've added a separate `CombineMoya` target to `Package.swift` and fixed the tests.

MaxDesiatov:
I've added a corresponding CombineMoya.xcscheme and target in the Xcode project. The tests fail for me locally too, but they do fail on the `development` branch just as well with the same error, unable to find the Alamofire module, would you like that to be fixed here?

sunshinejr:
@MaxDesiatov if you could that would be awesome, thanks!

grioja:
Is there any word on this PR?

MaxDesiatov:
Sorry for the delay, I'll pick this up in the next few days.

MaxDesiatov:
@sunshinejr this is now ready for review.

sunshinejr:
Looking... 👀

amaurydavid:
Alamofire 5.2 added support for Combine. If we update the dependency we might not need to handle Combine by ourself ? Also, there should have some Alamofire new COmbine func we can use.

MaxDesiatov:
@sunshinejr I've updated `Readme.md` with notes and also removed the weak linking flag from `Moya.podspec`. Is there anything else you'd like me to add or update?

sunshinejr:
@MaxDesiatov no I don't think so - thank you for your work, really appreciate it! Will give it a day or so to see if someone has concerns regarding this one, will also play with it a bit, but it looks good!

PR #2039, opened 2020-05-07
Update Basic.md

monologue555:
Fixes doc error with 'url' which should be 'URL'.

### DIFF
--- a/docs/Examples/Basic.md
+++ b/docs/Examples/Basic.md
@@ -124,7 +124,7 @@ You can also set up custom endpoints to alter the default behavior to your needs
 
 ```swift
 let endpointClosure = { (target: MyService) -> Endpoint in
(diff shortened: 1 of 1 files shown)


docs/Examples/Basic.md
```diff
@@ -145,7 +145,7 @@ let failureEndpointClosure = { (target: MyService) -> Endpoint in
             return .networkResponse(200, target.sampleData)
         }
     }
-    return Endpoint(url: url(target), sampleResponseClosure: sampleResponseClosure, method: target.method, task: target.task, httpHeaderFields: target.headers)
+    return Endpoint(url: URL(target), sampleResponseClosure: sampleResponseClosure, method: target.method, task: target.task, httpHeaderFields: target.headers)
```
sunshinejr, line 148: it seems like the URL initializer we have needs `target` label as well:
```suggestion
    return Endpoint(url: URL(target: target), sampleResponseClosure: sampleResponseClosure, method: target.method, task: target.task, httpHeaderFields: target.headers)
```

sunshinejr:
thanks @monologue555! 🚀

PR #2042, opened 2020-05-11
Add Task.uploadData and update tests

### LINKED ISSUE #1655: Add uploadData(Data) case to Task enum

julienbodet:
Fixes #1655

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,7 @@
 # Next
 
 ### Changed
+- Add uploadData(Data) case to Task enum. [#2042](https://github.com/Moya/Moya/pull/2042) by [@julienbodet](https://github.com/julienbodet).
 - Moya's Swift version is now Swift 5.1. [#1931](https://github.com/Moya/Moya/pull/1931) by [@BasThomas](https://github.com/BasThomas).
 # [14.0.0] - 2020-02-15
 
--- a/Sources/Moya/MoyaProvider+Internal.swift
+++ b/Sources/Moya/MoyaProvider+Internal.swift
@@ -84,6 +84,8 @@ public extension MoyaProvider {
                 return self.sendRequest(target, request: request, callbackQueue: callbackQueue, progress: progress, completion: completion)
             case .uploadFile(let file):
                 return self.sendUploadFile(target, request: request, callbackQueue: callbackQueue, file: file, progress: progress, completion: completion)
+            case .uploadData(let data):
+                return self.sendUploadData(target, request: request, callbackQueue: callbackQueue, data: data, progress: progress, completion: completion)
             case .uploadMultipart(let multipartBody), .uploadCompositeMultipart(let multipartBody, _):
                 guard !multipartBody.isEmpty && endpoint.method.supportsMultipart else {
                     fatalError("\(target) is not a multipart upload target.")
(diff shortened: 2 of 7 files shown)


LucianoPAlmeida:
@sunshinejr Seems like there is an issue with the CI. Cache maybe?

Sources/Moya/MoyaProvider+Internal.swift
```diff
@@ -195,6 +197,16 @@ private extension MoyaProvider {
         return sendAlamofireRequest(alamoRequest, target: target, callbackQueue: callbackQueue, progress: progress, completion: completion)
     }
 
+    func sendUploadData(_ target: Target, request: URLRequest, callbackQueue: DispatchQueue?, data: Data, progress: ProgressBlock? = nil, completion: @escaping Completion) -> CancellableToken {
```
LucianoPAlmeida, line 200: If I see it correctly `sendUploadData` is exactly the same as `sendUploadFile` can we find some way to unify them in only one method? I think is ok to do it like this, but maybe would be good to have it unified :)

sunshinejr:
@LucianoPAlmeida yeah seems like cache issue - just updated the CI, let's see how it goes!

LucianoPAlmeida:
I think the implementation looks good, but I think it would be good to add changelog entry :) 
cc @julienbodet

LucianoPAlmeida:
Ahh thanks for fixing the CI @sunshinejr :)

julienbodet:
> I think the implementation looks good, but I think it would be good to add changelog entry :)
> cc @julienbodet

Good idea, it's done now!

Changelog.md
```diff
@@ -1,5 +1,8 @@
 # Next
 
+### Changed
+- Add uploadData(Data) case to Task enum. [#2042](https://github.com/Moya/Moya/pull/2042) by [@julienbodet](https://github.com/julienbodet).
+
 ### Changed
 - Moya's Swift version is now Swift 5.1. [#1931](https://github.com/Moya/Moya/pull/1931) by [@BasThomas](https://github.com/BasThomas).
```
LucianoPAlmeida, line 7: There is already a section for changed, can you move it to here?

julienbodet: Sure!

Sources/Moya/MoyaProvider+Internal.swift
julienbodet: Good idea, but I don't see any way to do this without more refactoring. I think it's ok like this for now.

sunshinejr:
Thanks for the PR @julienbodet, really appreciate it! We're currently trying to figure out if we can get the refactoring PR in the next release - if we do then this PR wouldn't be needed. But in case we won't, we will think about this and the other PR for improving the task type. I'll keep you posted!

julienbodet:
> Thanks for the PR @julienbodet, really appreciate it! We're currently trying to figure out if we can get the refactoring PR in the next release - if we do then this PR wouldn't be needed. But in case we won't, we will think about this and the other PR for improving the task type. I'll keep you posted!

Great, thanks for the update!

Issue #2018, opened 2020-03-24
Update RxSwift in .podspec to ~>5.1.0


dimpiax:
Hello, where do you plan to update Moya with updated RxSwift version to which uses deprecated `UIWebView`.

https://developer.apple.com/news/?id=12232019b
https://github.com/ReactiveX/RxSwift/releases/tag/5.1.0

sunshinejr:
Hey @dimpiax - please take a look at https://github.com/Moya/Moya/issues/2003, we should be already supporting RxSwift 5.1 and above. Running `pod update` should help (and update to 5.1.0 if you don't have any other version pinning set-up) but let me know if you need any more guidance.

dimpiax:
@sunshinejr all right then, thanks.

agoubi:
woooh woke up on a nightmare man 
one of my apps for a big client just got rejected because of that 
just updated to 5.1 and it works

PR #2044, opened 2020-05-16
Align Package manager version requirements

lordzsolt:
## PR Description
Fixes the inconsistent behavior between version requirements of 3rd party dependencies, based on which package manager you use.

Using RxSwift as an example:
- `Package.swift` uses `.upToNextMajor(from: "5.0.0")`, which will resolve it to any 5.y.z version
- `Cartfile` uses `~> 5.0` which, surprisingly, will also resolve it to any 5.y.z version
- However, `.podspec` has `~> 5.0`, which will resolve it to any 5.0.z version. 

## Extra context
Initially, I only wanted to raise RxSwift's version requirement from `5.0` to `5.1`, because:
RxSwift 5.1.0 removed the deprecated UIWebView extensions. However, `Moya/RxSwift` insists on using 5.0.
Currently, no new app using `Moya/RxSwift` can be submitted to the app store.

### DIFF
--- a/Moya.podspec
+++ b/Moya.podspec
@@ -24,19 +24,19 @@ Pod::Spec.new do |s|
 
   s.subspec "Core" do |ss|
     ss.source_files  = "Sources/Moya/", "Sources/Moya/Plugins/"
-    ss.dependency "Alamofire", "~> 5.0"
+    ss.dependency "Alamofire", "~> 5"
     ss.framework  = "Foundation"
   end
 
   s.subspec "ReactiveSwift" do |ss|
     ss.source_files = "Sources/ReactiveMoya/"
     ss.dependency "Moya/Core"
-    ss.dependency "ReactiveSwift", "~> 6.0"
+    ss.dependency "ReactiveSwift", "~> 6"
   end
 
   s.subspec "RxSwift" do |ss|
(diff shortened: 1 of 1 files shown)


amaurydavid:
Hello :)
Thanks for taking time to improve Moya!

The RxSwift 5.1 subject have already been discussed in some recent opened issues like #2003 and #2018 . The conclusion was that there was no issue.

>However, .podspec has ~> 5.0, which will resolve it to any 5.0.z version.
I'm afraid you are wrong, here is an example from [the official Podfile guide](https://guides.cocoapods.org/using/the-podfile.html): 
> '~> 0.1' Version 0.1 and the versions up to 1.0, not including 1.0 and higher

With this correction it looks to me that versioning rules as the same on all 3 dependency manager (at least for RxSwift) and there is no changes needed.

Morever, your changes would allow Cocoapods to use any superior major version of Alamofire, ReactiveSwift and RxSwift, which we certainly don't want. A new major version contains breaking changes which would make Moya to not compile.

lordzsolt:
Oh, I seem to have misread the error message.

You are right, this change would allow **any** version of Alamofire, ReactiveSwift and RxSwift. 

Apologies for your time.

PR #2045, opened 2020-05-20
Changed the parameter type AuthorizationType to TargetType in the tok…

Coder-ZJQ:
Adding support for `MultiTarget` with different token:
``` swift
let fooToken = "eyeAm.AJsoN.weBTOKen.foo"
let barToken = "eyeAm.AJsoN.weBTOKen.bar"
let authPlugin = AccessTokenPlugin { target in 
    if target is FooService {
        return fooToken    
    } else if target is BarService {
        return barToken
    }
    return ""
}
let provider = MoyaProvider<MultiTarget>(plugins: [authPlugin])
```


Coder-ZJQ:
The parameter target pass to the token closure wasn't the real target.So close.

PR #2046, opened 2020-05-21
Added support for MultiTarget with different token

Coder-ZJQ:
Changed the parameter type `AuthorizationType` to `TargetType` in the token closure to add support for `MultiTarget` with different tokens:
``` swift
let fooToken = "eyeAm.AJsoN.weBTOKen.foo"
let barToken = "eyeAm.AJsoN.weBTOKen.bar"
let authPlugin = AccessTokenPlugin { target in 
    if target is FooService {
        return fooToken    
    } else if target is BarService {
        return barToken
    }
    return ""
}
let provider = MoyaProvider<MultiTarget>(plugins: [authPlugin])
```


sunshinejr:
Thanks for taking your time and issueing a PR @Coder-ZJQ (also sorry for a delayed review) - this looks cool! I can see a use-case where you would have different tokens for different targets, e.g. using facebook/google sign-ins.

This is, however, another breaking change to the auth plugin so I would like a double-check with @Moya/contributors if there is anything we should consider before merging it.

Coder-ZJQ:
It's my pleasure : ) And I have added the [changelog entry](https://github.com/Moya/Moya/pull/2046/files#diff-9b8fab691c00b9e5380b19ce882f3271), the docs([docs_CN/Authentication.md](https://github.com/Moya/Moya/pull/2046/files#diff-60b4b40900ca70b12445864bab156f9a), [docs/Authentication.md](https://github.com/Moya/Moya/pull/2046/files#diff-c3ae69f3d76aad8177df8dd3a54c9476)) I have updated in the first commit. @sunshinejr

sunshinejr:
@Coder-ZJQ thank you so much for working on this, I've added few docs changes but this is great and will be in the next beta 🙏 (sorry for the delay on that as well...)

PR #2052, opened 2020-06-11
Use correct Changelog file name

f-meloni:
I noticed that Danger was giving the Changelog warning also when the changelog was updated, this fixes it


BasThomas:
😬 thanks!

PR #2055, opened 2020-06-17
Update relevant instructions of session

dirtmelon:


sunshinejr:
@dirtmelon thanks for catching that! would you able to change the [docs in chinese](https://github.com/Moya/Moya/blob/master/docs_CN/Providers.md) as well?

dirtmelon:
@sunshinejr OK, I will update related Chinese documents in this weekend.

dirtmelon:
@sunshinejr Done.

PR #2062, opened 2020-06-24
Install rocket only on OSX and linux

f-meloni:
This should fix the issue reported on https://github.com/Moya/Moya/issues/1928

Adds Rocket to the dependencies only if the platform is Linux or OSX
For the rest, given rocket still has the `// dev` flag it will be anyway commented out on releases

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -32,7 +32,7 @@ let package = Package(
         .package(url: "https://github.com/ReactiveX/RxSwift.git", .upToNextMajor(from: "5.0.0")),
         .package(url: "https://github.com/Quick/Quick.git", .upToNextMajor(from: "2.0.0")), // dev
         .package(url: "https://github.com/Quick/Nimble.git", .upToNextMajor(from: "8.0.0")), // dev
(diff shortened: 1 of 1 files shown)


Package.swift
```diff
@@ -23,8 +33,7 @@ let package = Package(
         .package(url: "https://github.com/Quick/Quick.git", .upToNextMajor(from: "2.0.0")), // dev
         .package(url: "https://github.com/Quick/Nimble.git", .upToNextMajor(from: "8.0.0")), // dev
         .package(url: "https://github.com/AliSoftware/OHHTTPStubs.git", .upToNextMajor(from: "9.0.0")), // dev
```
MoyaBot, line 35: - :warning: Collection literals should not have trailing commas. (`trailing_comma`)

LucianoPAlmeida, line 35:
```suggestion
        .package(url: "https://github.com/AliSoftware/OHHTTPStubs.git", .upToNextMajor(from: "9.0.0")) // dev
```

f-meloni: Thank you

sunshinejr:
thank u @f-meloni! ❤️

PR #2067, opened 2020-07-07
[CircleCI] Branch-based cache

sunshinejr:
Because we have master/development pretty much always out of sync between versions (in case of Carthage dependencies), it makes sense to split cache into branch-based strategy. This will make new branches build longer, but it won't be reporting failures just because cache is out of date.


sunshinejr:
ok it seems to be working, merging!

PR #2069, opened 2020-07-13
Remove mentions of the former `Result` dependency from the README

xavierLowmiller:
Just a small documentation update.

The README's instructions for setting up Moya using submodules still mention the `Result` dependency which is no longer required.

There are other minor issues with the submodule setup descriptions since Xcode renamed some things. I'll submit a second pull request once I find a working setup.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -157,15 +157,14 @@ If this is your first time using Carthage in the project, you'll need to go thro
 $ git init
 ```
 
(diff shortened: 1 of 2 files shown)


xavierLowmiller:
Certainly!
Can somebody with Mandarin skills check out the Chinese translation? I'm not sure I can do any good there...

sunshinejr (approved):
@xavierLowmiller thank u!

PR #2079, opened 2020-08-25
#2078 Issue about path

### LINKED ISSUE #2078: The URL in the `?` is turned into `%3F`, causes the return data exception
In a project I'm working on, we had a problem to navigate between the pathes,
because it is a back-end driven-ui solution and there we from the json the path we should navigate to.

So basically, this path tells the whole url itself, without separating the query params.
So we get into that problem of having %3f in the place of ?
Debugging, we noticed that changing

from self = target.baseURL.appendingPathComponent(targetPath)
to self = URL(string: target.baseURL.absoluteString + target.path)!

It would work very well.
So this is the solution I'm suggesting.

andrevinic:
Fixes #2078 

In a project I'm working on, we had a problem to navigate between the pathes,

because it is a back-end driven-ui solution and there we from the json the path we should navigate to.

So basically, this path tells the whole url itself, without separating the query params.
So we get into that problem of having %3f in the place of ?
Debugging, we noticed that changing

from             self = target.baseURL.appendingPathComponent(targetPath)
to             self = URL(string: target.baseURL.absoluteString + target.path)!

It would work very well.
So this is the solution I'm suggesting, it worked for us.

### DIFF
--- a/Sources/Moya/URL+Moya.swift
+++ b/Sources/Moya/URL+Moya.swift
@@ -11,7 +11,7 @@ public extension URL {
         if targetPath.isEmpty {
             self = target.baseURL
         } else {
-            self = target.baseURL.appendingPathComponent(targetPath)
+            self = URL(string: target.baseURL.absoluteString + target.path)!
         }
     }
 }


Sources/Moya/URL+Moya.swift
```diff
@@ -11,7 +11,7 @@ public extension URL {
         if targetPath.isEmpty {
             self = target.baseURL
         } else {
-            self = target.baseURL.appendingPathComponent(targetPath)
+            self = URL(string: target.baseURL.absoluteString + target.path)!
```
MoyaBot, line 14: - :warning: Force unwrapping should be avoided. (`force_unwrapping`)

zhenyiyi: It would work very well.

andrevinic:
@sunshinejr everything OK to merge? thanks

andrevinic:
@sunshinejr could we do the merge?

thanks

andrevinic:
@stephencelis 
@steam 

would you guys help me here to merge?

thanks

stephencelis:
@andrevinic Sorry, I'm not a maintainer of this project.

PR #2099, opened 2020-10-12
Delayed stubs with range

sunshinejr:
So this is something I've been using for a while: delayed stubs with specified range instead of always the same (+ teardown, of course) number of seconds. My use-case for this is when I use stubs for the app, instead of real-backend responses, and I just want to make sure I imitate the real world response time a little bit better.

I'd love to see if anyone else sees that as a viable option to add to the core itself, so pinging @Moya/contributors!

### DIFF
--- a/Tests/MoyaTests/MoyaProviderSpec.swift
+++ b/Tests/MoyaTests/MoyaProviderSpec.swift
@@ -178,11 +178,11 @@ final class MoyaProviderSpec: QuickSpec {
         describe("a provider with delayed stubs") {
             var provider: MoyaProvider<GitHub>!
             var plugin: TestingPlugin!
-            let delay: TimeInterval = 0.5
+            let delayRange = 0.15...0.30
 
             beforeEach {
                 plugin = TestingPlugin()
-                provider = MoyaProvider<GitHub>(stubClosure: MoyaProvider.delayedStub(delay), plugins: [plugin])
+                provider = MoyaProvider<GitHub>(stubClosure: MoyaProvider.delayedStub(delayRange), plugins: [plugin])
             }
 
             it("delays execution") {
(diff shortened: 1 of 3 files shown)


Tests/MoyaTests/MoyaProviderSpec.swift
```diff
@@ -178,11 +178,11 @@ final class MoyaProviderSpec: QuickSpec {
         describe("a provider with delayed stubs") {
             var provider: MoyaProvider<GitHub>!
             var plugin: TestingPlugin!
-            let delay: TimeInterval = 0.5
+            let delayRange = 0.15...0.30
 
             beforeEach {
                 plugin = TestingPlugin()
-                provider = MoyaProvider<GitHub>(stubClosure: MoyaProvider.delayedStub(delay), plugins: [plugin])
+                provider = MoyaProvider<GitHub>(stubClosure: MoyaProvider.delayedStub(delayRange), plugins: [plugin])
```
colinta, line 185: These should be new specs, so we test both paths.

colinta (commented):
Do we need to expand the API to support this? It's no different than calling `delayedStub(TimeInterval.random(in: 0.15...0.30))`

sunshinejr:
@colinta is it really the same? My though was that by adding this we will generate random delay each time a request is made, versus with `delayedStub(TimeInterval.random(in: 0.15...0.30))` we would have always the same delay, even though it would be generated.

Tests/MoyaTests/MoyaProviderSpec.swift
sunshinejr: Yeah agree, though I didn't want to spend too much time in case it's not something we'd like to introduce so updated one to make sure it works.

colinta:
Ahh true, I didn't think about the fact that it's returning a closure there. It *is* still possible with the current API, by bringing that entire closure out:
```swift
provider = MoyaProvider<GitHub>(stubClosure: { _ in .delayed(seconds: TimeInterval.random(in: 0.15...0.3)) })
```
which we could certainly drop in the README/docs.  If others think this is a worthy addition I'm not here to block it.  Just on principle I think "less is more" when it comes to API.

sunshinejr:
@colinta yes, though you could say the same about other stubbing options - these are there and available to use without our extensions. This extension is something I use quite a bit and because of that I figured other people might not only be able to use that as well, but maybe discover a better way of testing stubbing code by exploring the API (_"oh, I didn't think of randomizing the delayed response time!!!"_). But again, even though this is something that I would like to be provided by default, it doesn't mean it has to be there if there is not enough positive feedback on it 👍

PR #2105, opened 2020-11-13
Combine: send progress response in completion

### LINKED ISSUE #2104: Combine: response is published before it is processed by plugins ✨
**Moya Combine - development branch**

When using the Combine extension and performing a `requestWithProgressPublisher` the response is published before it was processed by plugins.

**Expected**: the response is published after it was processed by plugins. As result, the response might not be published at all, in case it was mapped to an error by a plugin which would lead to the completion of the stream.

<details>
<summary>
<b>Details</b>
</summary>
<br/>
Here's an example to point out the issue:

- we are downloading an image from an URL that does not exist and the backend returns a 404 HTTP status error.
- the backend response data contains some error description
- we use a plugin to filter status codes and return an error if it is not a 2xx status code.

Currently, a subscription to the progress publisher will receive the response before it is was processed by the plugin. This leads to unexpected behavior. 

here is some made up code to explain it more:
```swift
moyaProvider.requestWithProgressPublisher(.image(name: "notFound"))
   .sink(receiveCompletion: { completion in
        guard case let .failure(error) = completion else { return }
        self.showErrorPopup(error)
    }, receiveValue: { progressResponse in
        guard let response = progressResponse.response else { return }
        do {
             self.image = try response.mapImage()
        } catch {
             self.image = .placehodler
        }
    }
```

Since the 404 response is published before it is processed by the plugins we end up in `receiveValue` and try to decode the error description in the response data as image. The decoding will fail and throw an error, in which case a placeholder image is used.

However, it's expected that the plugin processes the response before it is forwarded, and hence, maps it to an error. The error will be published as a completion event so we end up in `receiveCompletion` and show a popup with the error.
</details>

bensLine:
Fixes #2104 

The response is only sent from the completion after it was processed by plugins. If a plugin returns an error the response is not sent at all, instead, an error is sent as completion.

The change did not break existing test cases but I had not time to get in detail into Moya's test suit to add an appropriate test. Would be great if one can point me to the right place or help out.

### DIFF
--- a/Sources/CombineMoya/MoyaProvider+Combine.swift
+++ b/Sources/CombineMoya/MoyaProvider+Combine.swift
@@ -32,6 +32,7 @@ public extension MoyaProvider {
     func requestWithProgressPublisher(_ target: Target, callbackQueue: DispatchQueue? = nil) -> AnyPublisher<ProgressResponse, MoyaError> {
         let progressBlock: (AnySubscriber<ProgressResponse, MoyaError>) -> (ProgressResponse) -> Void = { subscriber in
             return { progress in
+                guard !progress.completed else { return }
                 _ = subscriber.receive(progress)
             }
         }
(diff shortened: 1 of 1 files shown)


bensLine:
@sunshinejr @MaxDesiatov can you take a look at this? :)

MaxDesiatov (commented):
This seems legit, but would be great to add a test that covers this change.

bensLine:
Great, I'll look into it. But it'll take some time until I get to it.

killmen136 (approved):

PR #2115, opened 2020-12-16
moved inflightsRequest getter inside thread lock to resolve data race…

### LINKED ISSUE #2085: Crash in iOS 13.0 in moya provider
I am facing a crash in moya provider as described in screenshot 
I think the issue caused by multithreading because I am using background thread for callback queue
Screen Shot 2020-09-13 at 10 24 07 AM

```swift
internal let provider = MoyaProvider<MultiTarget>(endpointClosure: MoyaProvider<MultiTarget>.defaultEndpointMapping, requestClosure: MoyaProvider<MultiTarget>.defaultRequestMapping(for:closure:), stubClosure: MoyaProvider<MultiTarget>.neverStub, callbackQueue: .global(qos: .userInteractive), session: JBNetworkManager.defaultAlamofireSession(), plugins: [NetworkLoggerPlugin(configuration: .init(logOptions: .default))], trackInflights: true)
```

Jeroenbb94:
Fixes #2085.

The `trackInflights` getter caused a data race issue when performing the same API request multiple times in a row. When removing or adding into the `trackInflights` dictionary, a `NSRecursiveLock` is being used. But not when getting the dictionary and looping through, which can result in a data race bad access crash.

To resolve this, the forEach loop should also be in the `NSRecursiveLock` to ensure no mutations are happening when getting the `trackInflights` for the specific endpoint.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,10 +40,8 @@ jobs:
       - run: mkdir -p $CIRCLE_ARTIFACTS $CIRCLE_TEST_REPORTS
       - restore_cache:
           keys:
-            - v{{ .Environment.CACHE_VERSION }}-dep-{{ checksum "Cartfile.resolved" }}
-      - restore_cache:
-          keys:
-            - v{{ .Environment.CACHE_VERSION }}-dep-{{ checksum "Dangerfile.swift" }}
+            - v{{ .Environment.CACHE_VERSION }}-{{ .Branch }}-dep-{{ checksum "Cartfile.resolved" }}
+            - v{{ .Environment.CACHE_VERSION }}-{{ .Branch }}-dep-{{ checksum "Dangerfile.swift" }}
       - run:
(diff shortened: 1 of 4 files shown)


Jeroenbb94:
For some reason the CI fails on installing SwiftLint, I have the same issue locally :|

LucianoPAlmeida (approved):
LGTM!

LucianoPAlmeida:
> For some reason the CI fails on installing SwiftLint, I have the same issue locally :|

The problem seems with something on CI since its failing for a lot of PRs

LucianoPAlmeida:
@Jeroenbb94 can you try re-target this to `development` instead of `master` branch ?

Jeroenbb94:
I notice that development now has a @Atmoic property wrapper around the inflight request. This PR can be closed cause the implementation of @Atmoic will resolve this issue.

PR #2111, opened 2020-12-07
Race condition fix for interceptor setup

anton-plebanovich:
Although rare it's possible that `interceptor` setup will happen after request `adapt` is called and so `willSend` call will be missed. It's easy to reproduce by blocking execution thread with `sleep` call:
Screen Shot 2020-12-07 at 16 12 45

Screen Shot 2020-12-07 at 16 13 29

Screen Shot 2020-12-07 at 16 15 10

After fix:
Screen Shot 2020-12-07 at 16 18 43

PR #2192, opened 2021-09-01
Race condition fix for interceptor setup

BasThomas:
Supersedes #2111 

This targets `development`, and fixes a naming issue.

cc @anton-plebanovich


BasThomas:
I wonder if and how we can write a test that covers this.

AndrewSB (approved):

OhKanghoon (approved):

Issue #2152, opened 2021-06-05
Use isEmpty to check whether the data is empty instead of count


jolyot:
First, thank you for the awesome product！
Moya enables me to handle the network layer more easily than before.

Now my proposal is simple. 
As the title says, you should use `isEmpty` instead of `count` to check whether the data is empty.

Here is my code:
```swift
if data.isEmpty && !failsOnEmptyData
```

The following links are the corresponding codes.
https://github.com/Moya/Moya/blob/e4b331474bdb6e9637f3e574fcde473170070d17/Sources/Moya/Response.swift#L103

https://github.com/Moya/Moya/blob/e4b331474bdb6e9637f3e574fcde473170070d17/Sources/Moya/Response.swift#L176

BasThomas:
Do you want to open a pull request for this? :)

PR #2194, opened 2021-09-02
Use isEmpty to check whether the data is empty

jolyot:
This PR is for the issue #2152 

This is not a breaking API change, therefore I'm opening this PR towards the master branch.


BasThomas (changes requested):
We should point this to `development` instead of the main branch; we're planning to release a new version imminently, and to make sure everything can be merged as expected would be great.

Thanks! 😊

BasThomas (approved):
Awesome, thanks!

PR #2196, opened 2021-09-02
Prepare Moya 15

BasThomas:
This sets up the last bits to release Moya 15; both Carthage and SwiftPM base the library version on git tags; this will be created as part of the release when this has been merged into `development`, and `development` subsequently gets merged into the main branch.

### DIFF
--- a/Moya.podspec
+++ b/Moya.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name         = "Moya"
-  s.version      = "15.0.0-alpha.1"
+  s.version      = "15.0.0"
   s.summary      = "Network abstraction layer written in Swift"
   s.description  = <<-EOS
   Moya abstracts network commands using Swift Generics to provide developers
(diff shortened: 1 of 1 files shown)


BasThomas:
Hm actually, probably want to update ReactiveSwift to latest, too.

OhKanghoon:
I think the major version of Reactive Swift is 6.0 right, isn't it?

OhKanghoon (approved):

Moya.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name         = "Moya"
-  s.version      = "15.0.0-alpha.1"
```
sunshinejr, line 3: afaik this will be set once you run the rake release command

sunshinejr:
@BasThomas one thing I wanted to also do is to bring back the original URL for ReactiveSwift (in SPM only, I think?) - we made a fork that's not needed anymore.

BasThomas:
> @BasThomas one thing I wanted to also do is to bring back the original URL for ReactiveSwift (in SPM only, I think?) - we made a fork that's not needed anymore.

https://github.com/Moya/Moya/pull/2197 😀

Moya.podspec
BasThomas: Oooh could be. Would it be an issue to have manually changed it?

sunshinejr: not sure, it might (?) since if there is nothing to commit it might be stuck on some command in the script? not really sure 😅

BasThomas: I'll see and fix if needed. :)

PR #2197, opened 2021-09-02
Update all non-dev dependencies to latest

BasThomas:
This also moves away from our ReactiveSwift fork, as it no longer seems to be necessary,


Package.resolved
```diff
@@ -75,20 +75,20 @@
...
       },
       {
         "package": "Rocket",
         "repositoryURL": "https://github.com/shibapm/Rocket",
         "state": {
           "branch": null,
-          "revision": "61166b38e229e09b53becef2e495eb99f437438a",
-          "version": "1.0.1"
+          "revision": "51a77ce5fa66c42715c14dcc542c01cd7a60fb27",
+          "version": "1.2.0"
```
BasThomas, line 91: And also one dev dependency I guess, oops

OhKanghoon (approved):
👍

PR #2200, opened 2021-09-04
Fix Release.md Rocket command

f-meloni:


BasThomas (approved):
Thanks @f-meloni, and thanks for the help getting this shipped 😄

PR #2203, opened 2021-09-05
[gardening] remove unecessary headers and plists causing SPM warnings

AndrewSB:
@BasThomas i was seeing a warning after i switched from my `fixes-ci` branch to your fixed `development`

i think its safe to delete these, since they aren't needed (at least for SPM) to generate a library. not very confident about this breaking Carthage/Cocoapods though, so lets merge this after CI can validate that?


BasThomas (approved):
Awesome, thanks for the cleanup!

Package.swift
```diff
@@ -27,19 +27,19 @@ let package = Package(
...
     dependencies: [
-        .package(url: "https://github.com/Alamofire/Alamofire.git", .upToNextMajor(from: "5.4.3")),
-        .package(url: "https://github.com/ReactiveCocoa/ReactiveSwift.git", .upToNextMajor(from: "6.6.1")),
-        .package(url: "https://github.com/ReactiveX/RxSwift.git", .upToNextMajor(from: "6.2.0")),
-        .package(url: "https://github.com/Quick/Quick.git", .upToNextMajor(from: "4.0.0")), // dev
-        .package(url: "https://github.com/Quick/Nimble.git", .upToNextMajor(from: "9.2.0")), // dev
-        .package(url: "https://github.com/AliSoftware/OHHTTPStubs.git", .upToNextMajor(from: "9.0.0")) // dev
+        .package(url: "https://github.com/Alamofire/Alamofire.git", from: "5.4.3"),
+        .package(url: "https://github.com/ReactiveCocoa/ReactiveSwift.git", from: "6.6.1"),
+        .package(url: "https://github.com/ReactiveX/RxSwift.git", from: "6.2.0"),
```
OhKanghoon, line 32: I think it's better to specify the same major version as podspec than to use the fixed version.
Is there a reason why you chose a fixed version?

BasThomas:
```

❌  error: Build input file cannot be found: '/Users/distiller/Moya/Moya/Sources/Supporting Files/Info.plist' (in target 'MoyaTests' from project 'Moya')
```

:(

Package.swift
AndrewSB: the `from: "semvar"` is an alias for up to the next major version. so selecting `Alamofire.git", from: "5.4.3"` allows any version from 5.x.x

we should definitely match what the Podspec & Cartfile have, I wish there was one place we could define all of those minimum versions to stay consistent. 

are the PodSpec/Cartfile significantly different?

AndrewSB:
let me see if i can fix that build error

Package.swift
OhKanghoon: Is `from: "5.4.3"` correct to include all 5.x.x?
It seems confusing to me 
because the document saying that it starting with the minimum version going up to the next major version.

In the podspec file it is written as `~> 5.0`
how about 5.0.0 in the Package.swift file? (`from: "5.0.0"` or `.upToNextMajor(from: "5.0.0")`)

AndrewSB: ahh you're right, its technically 5.4.3 until 6.0.0, not including 6.0.0.

so 5.4.2 or below aren't acceptable

yeah, it is a little confusing, they're updating the SPM syntax very often

OhKanghoon: Then, I think it would be good to change like podspec. (Alamofire, RxSwift etc..)
podspec : ~> 5.0 (`from: "5.0.0"` or .`upToNextMajor(from: "5.0.0")`)
What do you think about it?

OhKanghoon: #2207

TTOzzi:
That build warning removed at #2208

PR #2205, opened 2021-09-08
Prevent `Image` type conflict

### LINKED ISSUE #2204: Type conflict
swiftUI has Image type
But the Moya image.swift file also defines
public typealias Image = ImageType

BasThomas:
Resolves #2204

### DIFF
--- a/Sources/Moya/Image.swift
+++ b/Sources/Moya/Image.swift
@@ -7,4 +7,8 @@
 #endif
 
 /// An alias for the SDK's image type.
+@available(*, deprecated, renamed: "ImageType")
+public typealias Image = ImageType
+
+/// An alias for the SDK's image type.
 public typealias PlatformImage = ImageType
--- a/Sources/Moya/Image.swift
+++ b/Sources/Moya/Image.swift
@@ -7,4 +7,4 @@
 #endif
 
 /// An alias for the SDK's image type.
-public typealias Image = ImageType
+public typealias PlatformImage = ImageType
--- a/Tests/MoyaTests/TestHelpers.swift
+++ b/Tests/MoyaTests/TestHelpers.swift
@@ -230,13 +230,13 @@ extension URL {
     }
 }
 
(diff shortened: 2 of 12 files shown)


Tests/MoyaTests/TestHelpers.swift
```diff
@@ -230,13 +230,13 @@ extension URL {
...
 }
 
-// MARK: - Image Test Helpers
-// Necessary since Image(named:) doesn't work correctly in the test bundle
+// MARK: - PlatformImage Test Helpers
+// Necessary since PlatformImage(named:) doesn't work correctly in the test bundle
 extension ImageType {
     class TestClass { }
 
     static var testImage: ImageType {
```
BasThomas, line 238: Was a little surprised this returns the underlying `ImageType`, but thought I'd leave it as-is.

Sources/Moya/Image.swift
```diff
@@ -7,4 +7,4 @@
 #endif
 
 /// An alias for the SDK's image type.
-public typealias Image = ImageType
+public typealias PlatformImage = ImageType
```
BasThomas, line 10: We could also choose for something like `MoyaImage` but thought I'd go for a more generic name instead.

BasThomas (commented):
As this is breaking, this would have to wait to go in until 16.0.0 I'm afraid :(

Sources/Moya/Image.swift
LucianoPAlmeida: Hey Bas, just an idea we could do something like this 
```suggestion
@available(*, deprecated, message: "Use `PlatformImage` instead")
public typealias Image = ImageType

public typealias PlatformImage = ImageType
```
That way it would be not a breaking change, so users can disambiguate between SwiftUI image right away without need to wait a major version :)

BasThomas: Ooooh neat!

PR #2207, opened 2021-09-08
Pin dependencies to their major versions only

### LINKED ISSUE #2206: Moya 15 has too strict dependency on ReactiveSwift
Moya has too strict dependency on ReactiveSwift. I use ReactiveSwift 6.7.0, but I cannot update to Moya 15 since ReactiveSwift 6.6.1 is required.

BasThomas:
This resolves #2206.

This is an error on my part; I should not have pinned dependencies to specific versions. 😅 

This can go out as a 15.0.1.


OhKanghoon (approved):
👍

PR #2221, opened 2021-10-30
Fix danger

f-meloni:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,7 +41,6 @@ jobs:
       - restore_cache:
           keys:
             - v{{ .Environment.CACHE_VERSION }}-{{ .Branch }}-dep-{{ checksum "Cartfile.resolved" }}
-            - v{{ .Environment.CACHE_VERSION }}-{{ .Branch }}-dep-{{ checksum "Dangerfile.swift" }}
       - run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -41,7 +41,6 @@ jobs:
       - restore_cache:
           keys:
             - v{{ .Environment.CACHE_VERSION }}-{{ .Branch }}-dep-{{ checksum "Cartfile.resolved" }}
-            - v{{ .Environment.CACHE_VERSION }}-{{ .Branch }}-dep-{{ checksum "Dangerfile.swift" }}
```
f-meloni, line 44: Restore cache restores only the first matching key, is to offer some fallbacks for when you don't have a cache hit, so this logic was restoring almost all the times only the Carthage dependencies.

https://circleci.com/docs/2.0/caching/#restoring-cache

>CircleCI restores caches in the order of keys listed in the restore_cache step. Each cache key is namespaced to the project, and retrieval is prefix-matched. The cache will be restored from the first matching key. If there are multiple matches, the most recently generated cache will be used.

ashfurrow (approved):
Ah, interesting! Thank you for the fix and for the explanation 🙏

PR #2255, opened 2022-04-14
Fix tests not compilable when opening via Package.swift

antranapp:
When I open the project using Package.swift and run test, Xcode can't compile some test files because of missing CoreGraphics imports


AndrewSB (approved):
thanks @antranapp!

jbehrens94:
I can't reproduce this issue, @antranapp. It also doesn't make sense to me why CoreGraphics would be needed to run tests for a networking library, what error did you get?

antranapp:
@jbehrens94 We have this line `var size: CGSize?` in the files where I added `import CoreGraphics`. Not sure why you can't reproduce the issue, but when I open Package.swift directly, my Xcode reports error that CGSize is not found.

antranapp:
@AndrewSB what is blocking this MR to be merged? Is it because CI failing? I'm not sure why it's failing though. CI reports it failed to install swiftlint

Tests/MoyaTests/SignalProducer+MoyaSpec.swift
```diff
@@ -3,6 +3,7 @@ import Moya
 import ReactiveSwift
 import Nimble
 import Foundation
+import CoreGraphics
```
thejohnlima, line 6: It seems this PR is no longer required.
I just run the tests using `master` branch, latest version of Xcode, and all good.

Screenshot 2022-12-14 at 10 10 27 PM

Issue #1774, opened 2018-11-26
Don't follow redirects


xavierLowmiller:
## Description

We're looking for a way to not follow redirects on some targets.

Is there a way to handle redirects individually?

We're using the latest Moya version. From our Podfile.lock:
```
- Moya/Core (12.0.1):
  - Alamofire (~> 4.1)
  - Result (~> 4.0)
```

SD10:
@xavierLowmiller Hey, this might be configurable through Alamofire's manager.

 Can you check out [this stack overflow post on handling redirects in Alamofire](https://stackoverflow.com/questions/27759420/alamofire-follow-http-redirects-or-not) -- it links to a [related PR](https://github.com/Alamofire/Alamofire/pull/317) for support for this in Alamofire

> We're looking for a way to not follow redirects on some targets.

If your question is specifically asking about how to configure this on a target by target basis, I don't have any great ideas off the top of my head. The Alamofire `Manager` used by a `MoyaProvider` can only be configured per provider:

https://github.com/Moya/Moya/blob/master/Sources/Moya/MoyaProvider%2BDefaults.swift#L28-L37

xavierLowmiller:
Thanks for answering this!

We ended up setting the `taskWillPerformHTTPRedirection` closure on `MoyaProvider.manager.delegate` that conditionally follows redirects based on the URL's path.

What I was hoping for would be a `shouldFollowRedirect` or something as an extension on `TargetType`, similar to the optional `validationType`.

I'm willing to look into this. Would you be interested in a Pull Request?

SD10:
> What I was hoping for would be a shouldFollowRedirect or something as an extension on TargetType, similar to the optional validationType.

I'm willing to look into this. Would you be interested in a Pull Request?

@xavierLowmiller My concern is that `TargetType` will become bloated with all of these configurations. We have a [policy of not setting default values for protocol requirements in order to be explicit in the intent](https://github.com/Moya/Moya/blob/master/Vision.md).

I'm also concerned about how this would work with concurrent requests that all share the same `Manager` 🤔 

Will tag @Moya/contributors to see if anyone has input or more experience here

xavierLowmiller:
It might also fall under the „common things are easy, and uncommon things are possible“ category.
I’ve been using Moya for over a year in various projects, and this is the first time I’m running into this.

The thing I don’t like about our current implementation is that the decision to follow redirects is now completely independent of any `TargetType`, so I thought it might make sense to somehow combine this.

[...]

PR #2256, opened 2022-04-15
Add RedirectHandler to TargetType

antranapp:
Alamofire supports setting  a `RedirectHandler` per `Session` and per `Request`

When using Moya, I can already inject a custom `Session` to setup a global `RedirectHandler` for all requests. But to implement a `RedirectHandler` per request, it's unnecessarily complicated because Moya hides Alamofire's Request from in public API, there is no way Moya's users can customise Alamofire's Request easily.

I have tested by adding an optional RedirectHandler to TargetType protocol so that Moya's users can customise a redirect handler for each Target. By default, the value is nil, which means Alamofire will use the default handling mechanism for redirection.

I'm aware of [this discussion](https://github.com/Moya/Moya/issues/1774) but I don't agree with it. We basically hide an useful feature of Alamofire from Moya's users.

If this approach is ok, I'll add more tests to this PR to make it mergeable.

### DIFF
--- a/Sources/Moya/Moya+Alamofire.swift
+++ b/Sources/Moya/Moya+Alamofire.swift
@@ -1,6 +1,7 @@
 import Foundation
 import Alamofire
 
+public typealias RedirectHandler = Alamofire.RedirectHandler
 public typealias Session = Alamofire.Session
 internal typealias Request = Alamofire.Request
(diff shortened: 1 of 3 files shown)


gxs-jasper-nguyen:
This is exactly what we need in our project. Can we consider to review and merge this PR @xavierLowmiller please?

Thank you!

PR #2274, opened 2022-08-12
Add async Result modifiers

jbehrens94:
I was missing easy access to `Result` modifiers after `.request(...)`, so I took inspiration from the currently existing RxMoya and Moya wrappers.


jbehrens94:
@AndrewSB This is just a simple wrapper for some convenience modifiers, I see that there's some tests for Single/Observable/SignalProducer, do you think that's needed for the async wrapper as well? In general, this is just a simple extra compilation step, but if needed I can add tests. 😄

PR #2291, opened 2022-11-23
SPM Package iOS Target iOS 10 -> iOS 11 Update

HanweeeeLee:
It is an environment that uses Moya using SPM.
When I build a project, I can't build because the minimum target for iOS is iOS 10.

[xcode-14-release-notes](https://developer.apple.com/documentation/xcode-release-notes/xcode-14-release-notes)
The XCode 14 release notes show that the minimum build target is iOS 11.

I changed the iOS platforms version of the Package.swift file to iOS 11.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -16,7 +16,7 @@ let package = Package(
     name: "Moya",
     platforms: [
         .macOS(.v10_12),
-        .iOS(.v10),
+        .iOS(.v11),
         .tvOS(.v10),
         .watchOS(.v3)
     ],


naldal:
I have same issue too.
Moya should upgrade version of the minimum target for Xcode 14

JiHoonAHN:
I feel the same way.
Currently, the minimum version of Alamofire is iOS 11.0, causing problems.
To solve this problem, Moya also needs to upload to iOS 11.0.

thejohnlima (approved):

PR #2297, opened 2023-02-04
Fixing the CI for unlocking next PR 

frankois944:
Hello everyone,
I put my first stone at this project.

For now, I'm fixing the issue current issue inside the CI, Swiftlint is broken because it cant be compiled nor _brewed_ with the xcode version specified in the CI, the 12.5.1.

So I removed the swiftlint step which cant be used like this way anymore as we must use xcode 13 (then re-add swiftlint).

This is a temporary fix as I think we must upgrade the minimal version of xcode (> 13) to be aligned as Alamofire (so drop some macos/ios versions)

So, I tried to use use xcode 13.x in the pipeline and lot of issue with Carthage ... which need to be fixed(or to be removed/replace definitively by SPM?).

Also, actually, there is an another error inside the CI, Carthage takes to much time to build and CircleCI doesn't like that!

Edit: Finally, I removed Carthage test steps from the CI, it's really too much instable.

I really like this projet and use it for many years. I really want to be active.

Regards,
François

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,9 +45,6 @@ jobs:
           name: Set Ruby Version
           command: echo "ruby-2.4" > ~/.ruby-version
       - run:
-          name: Install Swiftlint
-          command: brew install swiftlint
-      - run:
           name: Add Python directory to PATH
           command: echo 'export PATH=~/Library/Python/2.7/bin:$PATH' >> $BASH_ENV        
       - run:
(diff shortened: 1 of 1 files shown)


frankois944:
If this fix is good and merged, I could make the Carthage migration with xcframework

frankois944:
Note: I disabled the build check carthage_without_swiftlint_integration  because of instability

PR #2302, opened 2023-02-22
boundary can now be passed to MultipartFormData

### LINKED ISSUE #1988: Custom Multipart Boundary
Currently it seems to be not possible to configure use a custom body part separator from Moya as it works in [Alamofire](https://alamofire.github.io/Alamofire/Classes/MultipartFormData.html#/s:9Alamofire17MultipartFormDataC11fileManager8boundaryACSo06NSFileF0C_SSSgtcfc). 

My proposal is to introduce an initializer argument to pass a custom multipart boundary into [MultipartFormData](https://moya.github.io/Structs/MultipartFormData.html#/s:4Moya17MultipartFormDataV8provider4name8fileName8mimeTypeA2C0cD8ProviderO_S2SSgAJtcfc). 

Currently I would have to manually implement the request using Alamofire if I want to provide a custom boundary, is that correct?

### LINKED ISSUE #2050: How to custom multipart boundary?
when i upload a file to server,i want to  custom multipart boundary,how can i do ? thanks!

### LINKED ISSUE #2051: Small refactor to MultipartFormData.
Related: #1988 #2050 and probably few more.

## Problem
So currently we have our own type for `MultipartFormData`, which is in fact a single part of that whole data (which is a bit different than the one from Alamofire):
```swift
public struct MultipartFormData {
    /// Method to provide the form data.
    public enum FormDataProvider {
        case data(Foundation.Data)
        case file(URL)
        case stream(InputStream, UInt64)
    }

    public init(provider: FormDataProvider, name: String, fileName: String? = nil, mimeType: String? = nil) {
        self.provider = provider
        self.name = name
        self.fileName = fileName
        self.mimeType = mimeType
    }

    /// The method being used for providing form data.
    public let provider: FormDataProvider

    /// The name.
    public let name: String

    /// The file name.
    public let fileName: String?

    /// The MIME type
    public let mimeType: String?
}
```

then we can pass an array of these in the `Task`:
```swift
enum Task {
    ...
    case uploadMultipart([MultipartFormData])
    ...
}
```

and then Moya will transform this array into the proper, Alamofire-provided `MultipartFormData`:
```swift
extension MoyaProvider {
    ...
    func sendUploadMultipart(_ target: Target, request: URLRequest, callbackQueue: DispatchQueue?, multipartBody: [MultipartFormData], progress: Moya.ProgressBlock? = nil, completion: @escaping Moya.Completion) -> CancellableToken {
        let formData = RequestMultipartFormData()
        formData.applyMoyaMultipartFormData(multipartBody)
    ...
    }
}
```

but this way we lack the ability to provide the `fileManager` and `boundary` properties that `Alamofire` allows:
```swift
open class MultipartFormData {
    ...
    public init(fileManager: FileManager = .default, boundary: String? = nil) {
        self.fileManager = fileManager
        self.boundary = boundary ?? BoundaryGenerator.randomBoundary()
        bodyParts = []
    }
    ...
}
```

## Solution

So two solutions comes to my mind. First, just use the multipart body provided by Alamofire. Though I really like the way we have it as an `enum` and the transition from enum to structs might be more time-consuming.

So I think we could refactor a little bit what we have. How?
- Rename the `MultipartFormData` to `MultipartBodyPart` (or something else, just an idea)
- Create new struct, called `MultipartFormData` which could look like the following:
```swift
struct MultipartFormData {
    init(fileManager: FileManager = .default, boundary: String? = nil, parts: [MultipartBodyPart]) {
        ...
    }
}
```

the callsite would change from:
```swift
enum MyTarget: Target {
    var task: Task {
        return .uploadMultipart([data, param])
    }
}
```

to:
```swift
enum MyTarget: Target {
    var task: Task {
        return .uploadMultipart(MultipartFormData(parts: [data, param]))
    }
}
```

Additionally, we could try and leverage `ExpressibleByArrayLiteral` so we could potentially not change the callsite at all for existing implementations and just add an option to provide these details.

thoughts? cc @Moya/contributors

ljubinkovicdj93:
✨

[MultipartFormData](https://github.com/Moya/Moya/blob/master/Sources/Moya/MultipartFormData.swift) can now take in `fileManager` and `boundary` inputs and pass it to [Alamofire's MultipartFormData](https://github.com/Alamofire/Alamofire/blob/master/Source/MultipartFormData.swift).

This doesn't introduce breaking API changes. Call site for `.uploadMultipart` will stay the same, ie:

```swift
enum MyTarget: TargetType {
   case test(Data, URL)

    var task: Task {
        switch self {
            case let .test(data, url):
                let bodyPart1 = MultipartFormBodyPart(provider: .file(url), name: "bp1")
                let bodyPart2 = MultipartFormBodyPart(provider: .data(data), name: "bp2")

                return .uploadMultipart([bodyPart1, bodyPart2]) // This is valid thanks to ExpressibleByArrayLiteral
        }
    }
}
```

Fixes #1988 
Fixes #2050 
Fixes #2051

### DIFF
--- a/Sources/Moya/Task.swift
+++ b/Sources/Moya/Task.swift
@@ -28,10 +28,10 @@ public enum Task {
     case uploadFile(URL)
 
     /// A "multipart/form-data" upload task.
-    case uploadMultipart([MultipartFormData])
+    case uploadMultipart(MultipartFormData)
 
     /// A "multipart/form-data" upload task  combined with url parameters.
-    case uploadCompositeMultipart([MultipartFormData], urlParameters: [String: Any])
+    case uploadCompositeMultipart(MultipartFormData, urlParameters: [String: Any])
 
     /// A file download task to a destination.
     case downloadDestination(DownloadDestination)
--- a/Sources/Moya/Task.swift
+++ b/Sources/Moya/Task.swift
@@ -28,10 +28,18 @@ public enum Task {
     case uploadFile(URL)
 
     /// A "multipart/form-data" upload task.
-    case uploadMultipart(MultipartFormData)
+    case uploadMultipartFormData(MultipartFormData)
+
+    /// A "multipart/form-data" upload task.
+    @available(*, deprecated, message: "use `uploadMultipartFormData(MultipartFormData)` instead")
+    case uploadMultipart([MultipartFormBodyPart])
 
     /// A "multipart/form-data" upload task  combined with url parameters.
-    case uploadCompositeMultipart(MultipartFormData, urlParameters: [String: Any])
+    case uploadCompositeMultipartFormData(MultipartFormData, urlParameters: [String: Any])
+
+    /// A "multipart/form-data" upload task  combined with url parameters.
+    @available(*, deprecated, message: "use `uploadCompositeMultipartFormData(MultipartFormData)` instead")
+    case uploadCompositeMultipart([MultipartFormBodyPart], urlParameters: [String: Any])
 
     /// A file download task to a destination.
     case downloadDestination(DownloadDestination)
--- a/Examples/_shared/GiphyAPI.swift
+++ b/Examples/_shared/GiphyAPI.swift
@@ -24,8 +24,9 @@ extension Giphy: TargetType {
     public var task: Task {
         switch self {
         case let .upload(data):
-            let multipartFormData = [MultipartFormData(provider: .data(data), name: "file", fileName: "gif.gif", mimeType: "image/gif")]
-            return .uploadCompositeMultipart(multipartFormData, urlParameters: ["api_key": "dc6zaTOxFJmzC", "username": "Moya"])
+            let multipartFormBodyParts = [MultipartFormBodyPart(provider: .data(data), name: "file", fileName: "gif.gif", mimeType: "image/gif")]
+            let multipartFormData = MultipartFormData(fileManager: .default, boundary: nil, parts: multipartFormBodyParts)
+            return .uploadCompositeMultipartFormData(multipartFormData, urlParameters: ["api_key": "dc6zaTOxFJmzC", "username": "Moya"])
         }
     }
     public var sampleData: Data {
--- a/Sources/Moya/Endpoint.swift
+++ b/Sources/Moya/Endpoint.swift
@@ -85,7 +85,7 @@ public extension Endpoint {
         request.allHTTPHeaderFields = httpHeaderFields
 
         switch task {
-        case .requestPlain, .uploadFile, .uploadMultipart, .downloadDestination:
+        case .requestPlain, .uploadFile, .uploadMultipart, .uploadMultipartFormData, .downloadDestination:
             return request
         case .requestData(let data):
             request.httpBody = data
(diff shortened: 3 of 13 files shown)


Sources/Moya/Task.swift
```diff
@@ -28,10 +28,10 @@ public enum Task {
     case uploadFile(URL)
 
     /// A "multipart/form-data" upload task.
-    case uploadMultipart([MultipartFormData])
+    case uploadMultipart(MultipartFormData)
```
AndrewSB, line 31: i'm not completely familiar, but are you sure this isn't breaking?

callsites like `.uploadMultipart([literal, array, works])` will continue to compile, but `.uploadMultipart(anArrayVariable)` wouldn't?

i might be totally wrong. correct me if i am

AndrewSB (commented):
nice work!

Sources/Moya/Task.swift
ljubinkovicdj93: You are completely right.

I missed the most obvious thing 😅

This means that the PR should target `development` branch instead then?

ljubinkovicdj93: Maybe this would be a better solution:

```swift
/// Represents an HTTP task.
public enum Task {
    // other cases...

    /// A "multipart/form-data" upload task.
    case uploadMultipartFormData(MultipartFormData)

    /// A "multipart/form-data" upload task.
    @available(*, deprecated, message: "use `uploadMultipartFormData(MultipartFormData)` instead")
    case uploadMultipart([MultipartFormBodyPart])

    /// A "multipart/form-data" upload task  combined with url parameters.
    case uploadCompositeMultipartFormData(MultipartFormData, urlParameters: [String: Any])

    /// A "multipart/form-data" upload task  combined with url parameters.
    @available(*, deprecated, message: "use `uploadCompositeMultipartFormData(MultipartFormData)` instead")
    case uploadCompositeMultipart([MultipartFormBodyPart], urlParameters: [String: Any])
}
```

This way it doesn't break the client's implementation.

I wanted to have the same case name with different associated values, but the Swift compiler [doesn't allow it (yet)](https://github.com/apple/swift/issues/52479)

What do you think?

AndrewSB: yeah i think that could work, mention me when you make the change and i'll check the diff again. it might also be useful to add the two examples as test cases, to make sure both continue to work

ljubinkovicdj93: @AndrewSB I've added the 2 new cases and deprecated the old ones. I've updated tests to make sure both old and new cases still pass, and the documentation and example code as well.

ljubinkovicdj93: @AndrewSB, @sunshinejr Can you please take a look at the latest changes when you get the chance. Thanks!

AndrewSB: @ljubinkovicdj93 the only thing i still see that `MultipartFormData.FormDataProvider` got renamed to `MultipartFormDataPart.FormDataProvider`. i think this is good to go if you just move that enum back to be nested under `MultipartFormData` instead of the new `MultipartFormDataPart`

ljubinkovicdj93: @AndrewSB [Done](https://github.com/Moya/Moya/pull/2302/commits/d4578c658ad5ebd92716777d3c53a92a7f2f21c3)

AndrewSB (approved):
appreciate you working through all of this!

JohnKuan:
Hi @AndrewSB, can I check when will this be in a release? I see the latest release is still 15.0.3 and that was back in 2022.

Issue #2174, opened 2021-07-20
NSURLSessionTask priority


ntnmrndn:
Hello, 

I was wondering wether Moya would consider adding support for prioritisation through `NSURLSessionTask` `.priority` property.

Do you have thoughts on the subject ?

ljluestc:
```
import Moya
import Foundation

class PrioritizedMoyaProvider<Target>: MoyaProvider<Target> where Target: TargetType {
    
    override func request(_ target: Target, callbackQueue: DispatchQueue? = .none, progress: ProgressBlock? = .none, completion: @escaping Completion) -> Cancellable {
        let requestCompletion: Completion = { result in
            // Call the original completion handler with the result
            completion(result)
        }
        
        // Create a URLRequest using the target
        let urlRequest = try! target.urlRequest()
        
        // Create a URLSessionTask from the URLRequest
        let task = session.dataTask(with: urlRequest) { data, response, error in
            // Process the URLSessionTask response
            let result = Result<Response, MoyaError>.init(catching: {
                // Handle the response and error to create a Moya Response or Moya Error
                // Note: This logic should be adapted to fit your specific needs
                // For demonstration purposes, we'll just create a simple response
                let response = Response(statusCode: 200, data: data ?? Data())
                return response
            })
            
            // Call the request completion handler with the result
            requestCompletion(result)
        }
        
        // Set the task priority based on your logic
        // For demonstration purposes, we'll set it to high priority for all requests
        task.priority = URLSessionTask.highPriority
        
        // Start the URLSessionTask
        task.resume()
        
        // Return a Cancellable object representing the URLSessionTask
        return AnyCancellable(task)
    }
}

```

PR #2330, opened 2023-11-08
Add priority customization to request

ntnmrndn:
✨

I would like to be able to set the NSURLSessionTask's priority. 
This PR adds a parameter to do that, with a default value to avoid breaking changes. 
Current proposal does not allow yet to change the priority dynamically after the request is created.

I targeted the main branch because it looks like develop is outdated and I tried to avoid making breaking changes.
Fixes https://github.com/Moya/Moya/issues/2174


Sources/Moya/MoyaProvider.swift
```diff
@@ -119,10 +119,15 @@ open class MoyaProvider<Target: TargetType>: MoyaProviderType {
...
                       progress: ProgressBlock? = .none,
+                      priority: Float = URLSessionTask.defaultPriority,
                       completion: @escaping Completion) -> Cancellable {
 
         let callbackQueue = callbackQueue ?? self.callbackQueue
-        return requestNormal(target, callbackQueue: callbackQueue, progress: progress, completion: completion)
+        return requestNormal(target, callbackQueue: callbackQueue, progress: progress, priority: priority, completion: completion)
+    }
+
+    public func request(_ target: Target, callbackQueue: DispatchQueue?, progress: ProgressBlock?, completion: @escaping Completion) -> Cancellable {
```
AndrewSB, line 129: why this additional function? 

does the function above with the new `priority` param with a default value not maintain API compatibility?

ntnmrndn: @AndrewSB Old C++ habit ! I fixed it.

PR #2376, opened 2026-06-25
Migrate CI to GitHub Actions and fix Xcode 16.4 / macOS 26 SDK compatibility

narlei:
## Summary

Replaces CircleCI with GitHub Actions and fixes a chain of build/test failures that surfaced when running under Xcode 16.4 + macOS 26 SDK on Apple Silicon (macos-15 runner).

### CI migration
- Add `.github/workflows/ci.yml` with two jobs: `build` and `carthage_without_swiftlint`
- Use `macos-15` runner with Xcode 16.4 (CircleCI's `m4pro.medium` resource class no longer supports `xcode:12.5.1`)
- Install SwiftLint via Homebrew (faster than building from source)
- Drop Ruby 2.4 and Python 2.7 references (incompatible with modern macOS images)
- Keep a minimal `.circleci/config.yml` as a pass-through to satisfy the existing required status checks

### Apple Silicon / Xcode 16.4 fixes

**Carthage — arm64-apple-ios swiftmodule overwriting arm64-apple-macos**

On Apple Silicon, iOS schemes with `SUPPORTS_MAC_DESIGNED_FOR_IPHONE_IPAD=YES` (the Xcode 16 default) claim macOS support. `carthage build --platform Mac` was picking those iOS schemes, producing `arm64-apple-ios` swiftmodules under `Carthage/Build/Mac/` and overwriting the native macOS ones. Fixed with two steps:
1. Patch every Carthage checkout's `project.pbxproj` to set `SUPPORTS_MAC_DESIGNED_FOR_IPHONE_IPAD = NO` and strip `IPHONEOS_DEPLOYMENT_TARGET` from all targets before the Mac build.
2. Delete `*iOS*.xcscheme` files from checkouts after the iOS/tvOS/watchOS bootstrap so `xcodebuild -list` can no longer discover iOS schemes when building for Mac.
3. Add a safety-net step that removes any `arm64-apple-macos.swiftmodule` still compiled against the iOS SDK (detected via `strings`).

**Nimble strict-prototypes warning in Xcode 16**

`DSL.m` uses K&R-style function declarations (`void f()` instead of `void f(void)`), which became an error in Xcode 16. Patched in-place with `sed` after checkout.

**Example project deployment target**

`Examples/Basic/ViewController.swift` uses Combine APIs (`AnyCancellable`, `requestPublisher`, `sink`) which require iOS 13+. The Rakefile was passing `IPHONEOS_DEPLOYMENT_TARGET=12.0`, causing a build failure. Bumped to `13.0`.

**CocoaPods version**

Alamofire's podspec now requires CocoaPods >= 1.13.0. The `Gemfile` was pinned to `~> 1.6`. Updated constraint to `>= 1.13.0` and added a `bundle update cocoapods` step before the lint.

**CocoaPods Spec Lint — ReactiveSwift macOS deployment target**

ReactiveSwift's podspec declares `osx.deployment_target = '10.9'`. With Xcode 16.4's macOS 26 SDK, Foundation types (`Date`, `Notification`, `URLRequest`, `Data`) now have `@available(macOS 10.10, *)` annotations, causing build errors in ReactiveSwift's own sources when linting the `Moya/ReactiveSwift` subspec for macOS. Since Moya's macOS compatibility is already validated by the "Test on macOS" step (which passes), the lint is run with `--platforms=ios,tvos,watchos` to skip the macOS platform.

---

## Action required (admin)

The CircleCI required status checks (`ci/circleci: build` and `ci/circleci: carthage_without_swiftlint_integration`) are still enforced on `master`. An admin needs to:

**1. Stop the project on CircleCI**
- Go to https://app.circleci.com/settings/project/github/Moya/Moya/setup
- Scroll down to **"Stop Building"** and confirm

**2. Remove the required checks on GitHub**
- Go to **Settings → Branches** in this repository
- Edit the branch protection rule for `master`
- Under **"Require status checks to pass before merging"**, remove:
  - `ci/circleci: build`
  - `ci/circleci: carthage_without_swiftlint_integration`
- Optionally add the new GitHub Actions checks:
  - `CI / build`
  - `CI / carthage_without_swiftlint`
- Save the changes

**3. Remove the minimal CircleCI config**
- Once the checks are removed, delete `.circleci/config.yml` from the repo

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,204 @@
+name: CI
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+
+jobs:
+  build:
+    runs-on: macos-15
+    env:
+      BUNDLE_PATH: vendor/bundle
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Select Xcode 16.4
+        run: sudo xcode-select -s /Applications/Xcode_16.4.app
+
+      - name: Install SwiftLint
+        run: brew install swiftlint
+
+      - name: Cache Bundler
+        uses: actions/cache@v4
+        with:
+          path: vendor/bundle
+          key: ${{ runner.os }}-gems-${{ hashFiles('Gemfile.lock') }}
+
+      - name: Install Ruby Dependencies
+        run: bundle install
+
+      - name: Cache Carthage
+        id: cache-carthage
+        uses: actions/cache@v4
+        with:
+          path: Carthage
+          key: ${{ runner.os }}-carthage-v15-${{ hashFiles('Cartfile.resolved') }}
+
+      - name: Install Carthage
+        run: brew install carthage
+
+      - name: Carthage Checkout
+        if: steps.cache-carthage.outputs.cache-hit != 'true'
+        run: carthage checkout
+
+      - name: Fix Nimble DSL.m for Xcode 16
+        if: steps.cache-carthage.outputs.cache-hit != 'true'
+        run: |
+          find Carthage/Checkouts/Nimble -name "DSL.m" -exec \
+            sed -i '' 's/\*NMB_\([a-zA-Z_]*\)()/\*NMB_\1(void)/g' {} \;
+
+      - name: Patch Carthage checkouts for native macOS builds on Apple Silicon
+        if: steps.cache-carthage.outputs.cache-hit != 'true'
+        # On Apple Silicon + Xcode 16, iOS schemes have SUPPORTS_MAC_DESIGNED_FOR_IPHONE_IPAD
+        # enabled by default, which makes them claim macOS support. Carthage --platform Mac
+        # then picks the iOS scheme (not the native macOS one), producing arm64-apple-ios
+        # swiftmodules instead of arm64-apple-macos. XCODE_XCCONFIG_FILE is the lowest-
+        # priority override mechanism and cannot fix this.
+        #
+        # Two-part fix at the highest file-based priority (target build settings in pbxproj):
+        # 1. Remove IPHONEOS_DEPLOYMENT_TARGET so native macOS targets don't inherit it.
+        # 2. Use the xcodeproj gem (already installed via CocoaPods) to inject
+        #    SUPPORTS_MAC_DESIGNED_FOR_IPHONE_IPAD = NO into each iOS target's build
+        #    settings (priority 6, above xcconfig at 5) so Carthage skips the iOS scheme
+        #    when building for --platform Mac and uses the native macOS scheme instead.
+        # scripts/carthage.sh re-adds IPHONEOS_DEPLOYMENT_TARGET for iphoneos/simulator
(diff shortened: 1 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,204 @@
...
+  build:
+    runs-on: macos-15
+    env:
+      BUNDLE_PATH: vendor/bundle
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Select Xcode 16.4
+        run: sudo xcode-select -s /Applications/Xcode_16.4.app
```
AndrewSB, line 18: why are we selecting 16.4 explicitly? can we not use whatever the latest Xcode on the macOS runner is?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,204 @@
+name: CI
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+
+jobs:
+  build:
+    runs-on: macos-15
```
AndrewSB, line 10: thoughts on macos-latest?

Gemfile
```diff
@@ -2,7 +2,7 @@ source 'https://rubygems.org'
 
 gem 'xcpretty'
 
-gem 'cocoapods', '~> 1.6'
+gem 'cocoapods', '>= 1.13.0'
```
AndrewSB, line 5: why?

AndrewSB (changes requested):
change overall looks good, left a couple questions. take a look, if you've thought those through, it makes sense to merge as is

.github/workflows/ci.yml
narlei: 16.4 is already the default on macos-15, so this line is really just a safety net in case that changes under us. The runner also has Xcode 26.x installed, but jumping there isn't as simple as switching the version — Nimble 9.2.1, Quick 4.0 and ReactiveSwift 6.7 (all pinned in Cartfile.resolved) already needed patching just to build on 16. A jump to 26 would break in new ways and deserves its own PR, ideally bumping those test deps too.

.github/workflows/ci.yml
narlei: Right now macos-latest == macos-15, so no real difference. But macos-26 doesn't ship Xcode 16.4 at all, so once GitHub flips the alias, xcode-select -s .../Xcode_16.4.app just breaks — on unrelated PRs, with no code change to blame. Pinning both together avoids that surprise. Can bump both in one go later.

Gemfile
narlei: Alamofire 5.10 requires CocoaPods >= 1.13 (privacy manifest stuff), and our lockfile had 1.11.3 — lint was failing on that. ~> 1.6 technically allowed 1.13+ already, the bundle update cocoapods step is what really fixes it, but figured it's worth raising the floor so this doesn't silently break again. Fine to make it ~> 1.13 instead if you'd rather keep a ceiling.

AndrewSB (approved):

